PR #303, opened 2014-04-18
Set default compiler to be the same that built dub

ibuclaw:
Raising this PR as a bug - there are likely other things that need to change too, such as the help message.
`"  dmd (default), gdc, ldc, gdmd, ldmd"`

It would be nice if the default compiler dub used was the same that built the application.  So for me, I don't need to specify --compiler=gdc explicitly every time.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -17,6 +17,7 @@ import dub.internal.vibecompat.inet.url;
 import dub.package_;
 import dub.packagemanager;
 import dub.packagesupplier;
+import dub.platform : determineCompiler;
 import dub.project;
 import dub.internal.utils : getDUBVersion;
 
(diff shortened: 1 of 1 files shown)


JohanEngelen:
@s-ludwig It looks like this functionality has been lost. `determineCompiler` is not called anywhere relevant. Thanks for looking into this.

s-ludwig:
I didn't remember the history for this anymore, but it was removed deliberately: #678 

But I'd say that #895 could be a good possibility to achieve the same practical outcome - assuming that that's the goal. It would probably make sense to implement that in conjunction with #875, for which I couldn't really make my mind up so far (any opinions on that one are very much appreciated).

Another, possibly cleaner in terms of usual OS behavior, alternative would be to support a configuration file in the same directory as the dub executable, which could have a pre-filled `"defaultCompiler"` field.

JohanEngelen:
Indeed, #895 is the goal. With #303 (this PR), it would still not solve that. Good pointer to #875 !

jacob-carlborg:
> I didn't remember the history for this anymore, but it was removed deliberately: #678

One reason I can think of why it was removed is to be able to use an optimizing compiler (ldc/gdc) when building binaries of Dub for a release but that shouldn't affect the default chosen compiler when running Dub.

PR #453, opened 2014-11-07
Added support for "dub init" to take a list of dependencies. It will search...

grogancolin:
... for the newest version numbers of each dependency on code.dlang.org (or whatever other PackageSupplier's are defined) and add the required string to dub.json.

The current <type> functionality (Where you can pass "vibe.d", "deimos" or "minimal" into dub init) has been moved to the --type=<type> option.
"minimal" is the default <type>, as is the case now.

Example:
dub init myProject vibe-d openssl logger --type=vibe.d
will produce dub.json:
{
    "name": "myproject",
    "description": "A simple vibe.d server application.",
    "copyright": "Copyright © 2014, Colin",
    "authors": ["Colin"],
    "dependencies": {
        "logger": "0.3.3",
        "vibe-d": "0.7.21-rc.4",
        "openssl": "1.1.3+1.0.1g"
    },
    "versions": ["VibeDefaultMain"]
}

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -342,7 +342,7 @@ class InitCommand : Command {
 				logInfo("Deprecated use of init type. Use --type=[vibe.d | deimos | minimal] in future.");
 			}
 		}
-		dub.createEmptyPackage(Path(dir), free_args, m_buildType);
+   		dub.createEmptyPackage(Path(dir), free_args, m_buildType);
 		return 0;
 	}
 }
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -622,12 +622,10 @@ class Dub {
 			}
 		}
         if(notFound.length > 1){
-            logError("Couldnt find packages: %-(%s, %). Exiting...", notFound);
-            return;
+            throw new Exception(format("Couldn't find packages: %-(%s, %).", notFound));
         }
         else if(notFound.length == 1){
-            logError("Couldnt find package: %-(%s, %). Exiting...", notFound);
-            return;
+            throw new Exception(format("Couldn't find package: %-(%s, %).", notFound));
         }
   		
         initPackage(path, depVers, type);


MartinNowak:
> It's a very interesting algorithm, so I'm doing this out of curiosity.

Still have the solution in my head, but didn't yet find enough time to implement it.
Meanwhile we have a first step for fuzzy searching running.
https://github.com/D-Programming-Language/dub-registry/issues/93
Interested in improving the init command @grogancolin?

grogancolin:
Absolutely. I've actually done some work on my local copy of dub and
dub-registry with the brute force levenstein distance method. It's hackish
and not suitable for public consumption yet, will clean it up this week.

grogancolin:
I'll pull down the referenced pull request and work off that actually. Makes more sense :)

s-ludwig:
I'm not sure if deprecating the old init behavior is a good idea, at least not without a suitable replacement. The idea was that init does not only add a few dependencies to dub.json, but also generates a proper program entry point, perhaps with hello-world-like example code. See #518 for a potential example.

Maybe we should just remove the deprecation message and extend the behavior to use a custom init routine for the first specified dependency if available and fall back to simple dependency mode otherwise. Later maybe we can do something like looking for a `xyz:app_skeleton` package and take that, so that dub init can be customized by package authors.

s-ludwig:
However, something like `dub init . deimos` would be hard to achieve with customization...

MartinNowak:
> at least not without a suitable replacement

There is a replacement `dub init myapp --type=vibe.d`.

MartinNowak:
> Maybe we should just remove the deprecation message and extend the behavior to use a custom init routine for the first specified dependency if available and fall back to simple dependency mode otherwise. Later maybe we can do something like looking for a xyz:app_skeleton package and take that, so that dub init can be customized by package authors.

That sounds like a good idea. We could simply name it `:skeleton`, because it also makes sense for libraries. I also like the convention to pick the first dependency.

PR #461, opened 2014-11-20
Added SublimeText Project Generator

londey:
- Added generator to create project files for SublimeText
- Adds import folders to the project for navigating etc.
-  Generated project inlcudes build systems to invoke DUB build test or run from inside SublimeText

Usage
dub generate sublimetext


MartinNowak:
Thanks looks greate and sorry for the late reply.

PR #471, opened 2014-12-08
Implement copyDirs parameter, similar to copyFiles

buggins:
Should recursively copy directory content to destination.

http://forum.rejectedsoftware.com/groups/rejectedsoftware.dub/thread/4875/

Working ok on dub build. Not sure about visuald/monod generators.

### DIFF
--- a/test/3-copyFiles/data/file_to_copy.txt
+++ b/test/3-copyFiles/data/file_to_copy.txt
@@ -1,0 +1,1 @@
+file_to_copy.txt content
--- a/test/3-copyFiles/data/file_to_copy_mask1.txt
+++ b/test/3-copyFiles/data/file_to_copy_mask1.txt
@@ -1,0 +1,1 @@
+file_to_copy_mask1.txt content
--- a/test/3-copyFiles/data/file_to_copy_mask2.txt
+++ b/test/3-copyFiles/data/file_to_copy_mask2.txt
@@ -1,0 +1,1 @@
+file_to_copy_mask2.txt content
--- a/test/3-copyFiles/data/res/.nocopy/file_inside_dot_prefixed_dir.txt
+++ b/test/3-copyFiles/data/res/.nocopy/file_inside_dot_prefixed_dir.txt
@@ -1,0 +1,1 @@
+some content
--- a/test/3-copyFiles/data/res/hdpi/file1.txt
+++ b/test/3-copyFiles/data/res/hdpi/file1.txt
@@ -1,0 +1,1 @@
+hdpi/file1 content
--- a/test/3-copyFiles/data/res/hdpi/file2.txt
+++ b/test/3-copyFiles/data/res/hdpi/file2.txt
@@ -1,0 +1,1 @@
+hdpi/file2 content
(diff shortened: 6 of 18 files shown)


xfront:
"copyFiles-windows": ["tools/win/usb_driver", "tools/win/*.dll", "tools/win/*.exe"]
will copy usb_driver/*.dll to bin,that is not my expect!

MartinNowak:
It should copy the folder usb_driver to bin, is this a bug?

s-ludwig:
Hm, `"copyDirs": ["tools/win/usb_driver"]` should copy the "usb_driver" folder, `"copyFiles"` just copies individual files. Any more complex operation currently needs to be done with a `"preGenerateCommands"` entry instead.

We'll have to think about a more sophisticated way to deal with runtime resource files at some point, but that's a relatively complex topic if the goal is to have that work well for all operating systems and applications. A [DEP](https://github.com/D-Programming-Language/dub/wiki/DUB-enhancement-proposals) would probably provide the best scope for discussing this.

BTW, I wish I had named `"copyFiles"` `"dllFiles"` instead, because it was really never intended to be good enough to handle anything else. Instead of a partial solution that `"copyFiles"`+`"copyDirs"` is, I would rather have liked to implement a more sophisticated solution that provides enough information to be useful for helping with automatically generating system packages (DEB etc.), for example.

buggins:
My initial implementation added copyDirs, has been replaced with fix of copyFiles behavior - to copy directory with content, retaining directory name.
To copy w/o directory name, "copyFiles-windows": ["tools/win/usb_driver/*"] may be used

s-ludwig:
I see... I'll have to go and go through the actual code. The new behavior should be documented in https://github.com/D-Programming-Language/dub-registry/blob/master/views/package_format.dt.

MartinNowak:
> a more sophisticated solution

The only sophistication still missing is matching across paths using an extended `**` glob.

MartinNowak:
> "copyFiles-windows": ["tools/win/usb_driver", "tools/win/*.dll", "tools/win/*.exe"]

You want `"copyFiles-windows": ["tools/win/usb_driver/*.dll", "tools/win/*.dll", "tools/win/*.exe"]` I think.

MartinNowak:
See https://github.com/D-Programming-Language/dub-registry/pull/101.

s-ludwig:
>> a more sophisticated solution

> The only sophistication still missing is matching across paths using an extended ** glob.

What I mean are things like copying to a different location than the target directory. The simplest example would be copying to a deeper directory, such as `bin/resources/some_folder` instead of just `bin/some_folder`. But it would also be a possible direction to enable copying to system locations, such as `/usr/share` or `C:\ProgramData`.

xfront:
the directory structure is:
 tools\win\usb_drivers\amd64\*.dll 
 tools\win\usb_drivers\i386\*.dll 
 tools\win\adb.exe
 tools\win\AdbWinApi.dll
image

I want to copy usb_drivers(directory) and adb.exe AdbWinApi.dll to bin.but "copyFiles-windows": ["tools/win/usb_driver", "tools/win/*.dll", "tools/win/*.exe"] will copy dll in amd64 and i386 too,
that is not my need.

PR #488, opened 2015-01-12
escaping local static array

sinkuu:
SHA1.finish returns a static array. In git head, dmd rejects escaping local static array.
PackageManager.hashPackage is used nowhere in dub though.

```
source/dub/packagemanager.d(608): Error: escaping reference to local variable hash
```


MartinNowak:
Thank, already fixed by 82525e23630a96c12376c2154953fd232858adbd.

PR #489, opened 2015-01-13
Add generator for CMake build scripts

Yoplitein:


MartinNowak:
How do I get a cmake with D support? Apparently I need some UseD.cmake script.
I tried https://github.com/dcarp/cmake-d to no avail.

Yoplitein:
There is a [fork of CMake](https://github.com/trentforkert/cmake) which, hopefully, will eventually be merged upstream.

MartinNowak:
```
-- To force a specific D compiler set the DC environment variable
--     ie - export DC="/usr/bin/dmd"
CMake Error at /home/dawg/Build/cmake-d/cmake-d/CMakeTestDCompiler.cmake:45 (message):
  The D compiler "/bin/dmd" is not able to compile a simple test program.

  It fails with the following output:

   

  

  CMake will not be able to correctly generate this project.
```
Yikes, it doesn't read the DC variable either.

source/dub/generators/cmake.d
```diff
@@ -0,0 +1,111 @@
...
+                script.put(
+                    "target_link_libraries(%s %s %s)\n".format(
+                        name,
+                        (info.dependencies ~ info.linkDependencies).dup.sort.uniq.map!sanitize.join(" "),
+                        info.buildSettings.libs.dup.join(" ")
+                    )
+                );
+            }
+            
+            string filename = "%s/%s.cmake".format(m_project.rootPackage.path.toString, name);
```
MartinNowak: Why not CMakeLists.txt?

Yoplitein: I figured CMakeLists.txt should be reserved for the user to set, e.g. project name, common compiler flags, non-dub generated targets, etc.

s-ludwig: How about creating a default CMakeLists.txt only when none exists, as a compromise?

source/dub/generators/cmake.d
```diff
@@ -0,0 +1,132 @@
...
+            script.put("include(UseD)\n");
+            script.put(
+                "add_d_conditions(VERSION %s DEBUG %s)\n".format(
+                    info.buildSettings.versions.dup.join(" "),
+                    info.buildSettings.debugVersions.dup.join(" "),
+                )
+            );
+            
+            foreach(directory; info.buildSettings.importPaths)
+                script.put("include_directories(%s)\n".format(directory));
```
MartinNowak: You're missing `-J` style string import paths.

source/dub/generators/cmake.d
```diff
@@ -0,0 +1,132 @@
...
+        super(project);
+    }
+    
+    override void generateTargets(GeneratorSettings settings, in TargetInfo[string] targets)
+    {
+        auto script = appender!(char[]);
+        auto scripts = appender!(string[]);
+        bool[string] visited;
+        string projectRoot = m_project.rootPackage.path.toString;
+        string cmakeListsPath = "%s/CMakeLists.txt".format(projectRoot);
```
MartinNowak: Would be better to use Path to concat paths, try `m_project.rootPackage.path ~ "CMakeLists.txt"` instead.
There is [`Path.toNativeString()`](http://vibed.org/api/vibe.inet.path/Path) to get the result.

source/dub/generators/cmake.d
```diff
@@ -0,0 +1,132 @@
...
+                script.put(
+                    "target_link_libraries(%s %s %s)\n".format(
+                        name,
+                        (info.dependencies ~ info.linkDependencies).dup.sort.uniq.map!sanitize.join(" "),
+                        info.buildSettings.libs.dup.join(" ")
+                    )
+                );
+            }
+            
+            string filename = "%s/%s.cmake".format(projectRoot, name);
```
MartinNowak: That produces 2 slashes `//vibe-d.make` for me, still works though.
Please use Path again to avoid this issue.

MartinNowak:
Rest looks good.

Yoplitein:
Should I squash my commits?

MartinNowak:
Not necessary, thanks.

PR #497, opened 2015-01-25
Use unique names for temporary files

Dicebot:
Partial implementation to https://github.com/D-Programming-Language/dub/issues/482 (does not add cache locks)

It currently includes only posix implementation - I need some hints to know how it should be done on Windows correctly.

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -386,15 +386,16 @@ Path generatePlatformProbeFile()
 	import dub.internal.vibecompat.data.json;
 	import dub.internal.utils;
 
-	auto path = getTempDir() ~ "dub_platform_probe.d";
+	auto path = getTempFile("dub_platform_probe", ".d");
 
 	auto fil = openFile(path, FileMode.CreateTrunc);
 	scope (failure) {
 		fil.close();
-		removeFile(path);
 	}
 
 	fil.write(q{
+		module dub_platform_probe;
+
 		template toString(int v) { enum toString = v.stringof; }
 
 		pragma(msg, `{`);
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -362,7 +362,7 @@ class Dub {
 			}
 
 			// generate main file
-			Path mainfile = getTempDir() ~ "dub_test_root.d";
+			Path mainfile = getTempFile("dub_test_root", ".d");
 			tcinfo.sourceFiles[""] ~= mainfile.toNativeString();
 			tcinfo.mainSourceFile = mainfile.toNativeString();
 			if (!m_dryRun) {
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -28,7 +28,70 @@ version(DubUseCurl) import std.net.curl;
 
 Path getTempDir()
 {
-	return std.file.tempDir;
+	return Path(std.file.tempDir());
+}
+
+private Path[] temporary_files;
+
+version (Posix)
+	private extern(C) int mkstemps(char *, int);
+else
+	private extern(Windows) uint GetTempFileName(
+		const(char)* lpPathName,
+		const(char)* lpPrefixString,
+		uint uUnique,
+		char* lpTempFileName
+	);
+
+Path getTempFile(string prefix, string extension = null)
+{
+	version (Posix)
+	{
+		auto name = (getTempDir() ~ prefix).toString();
+
+		// build 0-terminated template for mkstemp
+		auto suffix = "-XXXXXX" ~ extension;
+		auto pattern = new char[](name.length + suffix.length + 1);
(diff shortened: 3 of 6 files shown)


andralex:
Error: cannot read file test/1-staticLib-simple/aFcake-dub_test_root.d

What's that?

source/dub/internal/utils.d
```diff
@@ -28,13 +28,41 @@ version(DubUseCurl) import std.net.curl;
...
+Path getTempFile(string name)
+{
+	// build 0-terminated template for mktemps
+	enum prefix = "XXXXXX-";
+	auto pattern = new char[](prefix.length + name.length + 1);
+	pattern[0..prefix.length] = prefix[];
+	pattern[prefix.length .. $-1] = name[];
+	pattern[$-1] = '\0';
+
+	auto handle = mkstemps(pattern.ptr, to!int(name.length + 1));
```
MartinNowak: Does that work on Windows?

source/dub/dub.d
```diff
@@ -362,7 +362,7 @@ class Dub {
 			}
 
 			// generate main file
-			Path mainfile = getTempDir() ~ "dub_test_root.d";
+			Path mainfile = getTempFile("dub_test_root.d");
 			tcinfo.sourceFiles[""] ~= mainfile.toNativeString();
```
MartinNowak: Apparently there is some issue with relative vs. absolute paths here.

source/dub/internal/utils.d
Dicebot: No of course. I have never written a single program for Windows and will need to do some research on WinAPI to figure out the alternative.

source/dub/internal/utils.d
```diff
@@ -28,13 +28,69 @@ version(DubUseCurl) import std.net.curl;
...
+		temporary_files ~= path;
+		return path;
+	}
+	else version (Windows)
+	{
+		import core.sys.windows.windows : MAX_PATH;
+	
+		auto dir = getTempDir().toStringz();
+		auto name = prefix.toStringz();	
+		char[MAX_PATH] output;
```
MartinNowak: `= void`

source/dub/compilers/compiler.d
```diff
@@ -386,15 +386,16 @@ Path generatePlatformProbeFile()
...
+	auto path = getTempFile("dub_platform_probe", ".d");
 
 	auto fil = openFile(path, FileMode.CreateTrunc);
 	scope (failure) {
 		fil.close();
-		removeFile(path);
 	}
 
 	fil.write(q{
+		module dub_platform_probe;
```
MartinNowak: Why do you need a module declaration now?

Dicebot: Because uniquely generated file name may contain illegal symbols (illegal for a module name). `-` is one such symbol.

source/dub/internal/utils.d
Dicebot: Updated

MartinNowak:
```
source\dub\internal\utils.d(71): Error: None of the overloads of 'toStringz' are callable using argument types (Path), candidates are:
C:\D\dmd2\windows\bin\..\..\src\phobos\std\string.d(169):        std.string.toStringz(const(char)[] s)
C:\D\dmd2\windows\bin\..\..\src\phobos\std\string.d(212):        std.string.toStringz(string s)
source\dub\internal\utils.d(75): Error: no property 'ptr' for type 'immutable(char)*'
source\dub\internal\utils.d(78): Error: no property 'fromStrinz' for type 'char[260]'
```

MartinNowak:
Will this allow us to get your Jenkins running for the upcoming release?

MartinNowak:
After all the trouble with GetTempFileNameW and toUTF16z, I get something like this.
`C:/Users/dawg/AppData/Local/Temp/dubC3E2.tmp.rsp`
As the docs say, it uses __up to__ 3 chars of the prefix and apparently only 4 random digits.
It's also not necessarily unique as we have to append the file extension.

So I just generated a UUID based name instead.
This functionality is really missing from phobos.

source/dub/internal/utils.d
```diff
@@ -28,13 +28,70 @@ version(DubUseCurl) import std.net.curl;
...
+
+		pattern = pattern[0..$-1]; // get rid of \0
+		auto path = Path(assumeUnique(pattern));
+		temporary_files ~= path;
+		return path;
+	}
+	else version (Windows)
+	{
+		import core.sys.windows.windows : MAX_PATH;
+        import std.string : toStringz, fromStringz;
```
justanotheranonymoususer: Spaces!

MartinNowak:
> Ouch, I didn't realize this WinAPI function is that useless.

But it __does__ guarantee uniqueness, until we append the extension.

PR #498, opened 2015-01-31
Cached Parallel Separate Compilation of Project Modules

MartinNowak:
Is there a way to tell DUB to compile all its sources as separate objects which are then fed to ld similar to how most other build tools work.

This since my project has grown beyound 30k lines of code and takes 7.5 to build with DUB. I don't want DMD to recompile every line in my project if I change only of them. Since my cross-module interfaces are un-templatized I've already tested that this gives significant speeds-up in iterative build time (~2s for a typical iterative build in this 30k-project) when I edit one module at a time and use DMD together with ld.gold. 

The only way I managed to change behaviour of DUB was through the `--build-mode` flag as

```Shell
time dub build --build-mode=singleFile
```

which makes it compile each .d file separately:

```
Building knetquery ~master configuration "application", build 
type debug.
Compiling using dmd...
Compiling ../algorithm_ex.d...
Compiling ../arsd/characterencodings.d...
Compiling ../arsd/dom.d...
Compiling ../assert_ex.d...
Compiling ../backtrace/backtrace.d...
Compiling ../combinations.d...
Compiling ../dbg.d...
Compiling ../grammars.d...
Compiling ../ixes.d...
Compiling ../knet/base.d...
../sort_ex.d(143): Warning: use std.algorithm.sort instead of 
.sort property
../sort_ex.d(168): Warning: use std.algorithm.sort instead of 
.sort property
Compiling ../knet/cn5.d...
Compiling ../knet/decodings.d...
Compiling ../knet/folklex.d...
Compiling ../knet/languages.d...
Compiling ../knet/lectures/all.d...
Compiling ../knet/lemmas.d...
Compiling ../knet/moby.d...
Compiling ../knet/nell.d...
Compiling ../knet/origins.d...
Compiling ../knet/relations.d...
Compiling ../knet/roles.d...
Compiling ../knet/senses.d...
Compiling ../knet/swesaurus.d...
Compiling ../knet/synlex.d...
Compiling ../knet/thematics.d...
Compiling ../knet/wordnet.d...
Compiling ../mmfile_ex.d...
Compiling ../msgpack.d...
Compiling ../permutations.d...
Compiling ../predicates.d...
Compiling ../range_ex.d...
Compiling ../rational.d...
Compiling ../skip_ex.d...
Compiling ../sort_ex.d...
Compiling ../stemming.d...
Compiling ../traits_ex.d...
Compiling ../wordnet.d...
Compiling source/app.d...
Linking...
```

This takes about twice as long (~14s).

- It doesn't seem to support parallelism here. Where's the -j --jobs flag?!
- And I can't see any caching á lá SCons or ccache in action either.

Hasn't this been implemented in DUB yet or is something wrong?

If this is on the TODO list I would be glad to help out implementing support for parallell cached builds.

I'm using DUB master HEAD and ld.gold as my default linker on Ubuntu 14.04.

See also: http://forum.dlang.org/thread/wzkxfedfzsybwxpnmvmj@forum.dlang.org?page=1

https://github.com/gecko0307/Cook2 seems like something useful that can inspire this DUB in this regard.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -378,19 +378,22 @@ class BuildGenerator : ProjectGenerator {
 				removeFile(tpath);
 		}
 		if (settings.buildMode == BuildMode.singleFile && generate_binary) {
+			import std.parallelism, std.range : walkLength;
+
 			auto lbuildsettings = buildsettings;
-			auto objs = appender!(string[])();
+			auto srcs = buildsettings.sourceFiles.filter!(f => !isLinkerFile(f));
+			auto objs = new string[](srcs.walkLength);
 			logInfo("Compiling using %s...", settings.platform.compilerBinary);
-			foreach (file; buildsettings.sourceFiles.filter!(f=>!isLinkerFile(f))) {
-				logInfo("Compiling %s...", file);
-				objs.put(compileUnit(file, pathToObjName(file), buildsettings, settings));
+			foreach (i, src; srcs.parallel(1)) {
+				logInfo("Compiling %s...", src);
+				objs[i] = compileUnit(src, pathToObjName(src), buildsettings, settings);
 			}
 
 			logInfo("Linking...");
 			lbuildsettings.sourceFiles = is_static_library ? [] : lbuildsettings.sourceFiles.filter!(f=> f.isLinkerFile()).array;
 			settings.compiler.setTarget(lbuildsettings, settings.platform);
 			settings.compiler.prepareBuildSettings(lbuildsettings, BuildSetting.commandLineSeparate|BuildSetting.sourceFiles);
-			settings.compiler.invokeLinker(lbuildsettings, settings.platform, objs.data, settings.linkCallback);
+			settings.compiler.invokeLinker(lbuildsettings, settings.platform, objs, settings.linkCallback);
 
 		/*
 			NOTE: for DMD experimental separate compile/link is used, but this is not yet implemented


MartinNowak:
I added parallel builds, but even with that compiling dub-registry takes 1:34m vs. 0:20m using singleFile.

nordlow:
Nice. A significant improvement. However this doesn't help much in improving incremental build performance as the dependency logic in dub (git master HEAD at least) is faulty; It rebuilds all project modules every time a single source file is changed regardless of which modules that depend on it. This happens even for unused source files part of a project. Could you please fix this or at least point out where in the source I fix this?

nordlow:
Further, I'm missing a jobs flag to control `workUnitSize` argument to `std.parallelism.parallel` something like

`workUnitSize = totalCompilationUnits / numberOfJobs`

MartinNowak:
> Further, I'm missing a --jobs, -j flag in DUB to control workUnitSize argument to std.parallelism.parallel something like
> workUnitSize = totalCompilationUnits / numberOfJobs

That's not what workUnitSize is meant for, you want to create your own taskPool with less threads.

You want to add a 'jobs|j' flag to the global arguments ([here](https://github.com/D-Programming-Language/dub/blob/9c9c117d9a8d2f6edfed94fe4f3258dbf42d05cf/source/dub/commandline.d#L75)) and add a dub internal `taskPool` function that lazily initializes the pool on first use with the given number of jobs. If no jobs argument is given (jobs = 0) you can use the [default constructor](http://dlang.org/library/std/parallelism/task_pool.this.html) of TaskPool.

MartinNowak:
> However this doesn't help much in improving incremental build performance as the dependency logic in dub (git master HEAD at least) is faulty; It rebuilds all project modules every time a single source file (x.d) is changed regardless of which modules that depend on x.d

There is no dependency logic yet, you'd need to generate `-deps` files and compare the mod times.
I might restate that single file compilation is inherently slow.
Splitting large projects into smaller projects is much more efficient to reduce compile times.

nordlow:
Could you explain what `workUnitSize` is used for?

MartinNowak:
> Can you explain what workUnitSize is used for?

The number of items process for each worker step.
http://dlang.org/phobos/std_parallelism.html#.TaskPool.parallel

nordlow:
See https://github.com/D-Programming-Language/dub/pull/508

ThomasBrixLarsen:
I had to revert this change in my local dub. When compiling my Vibe.d project each module consumes about 40-60% of the available memory. With four hardware threads this obviously sends my system into swap hell.

s-ludwig:
Made this optional with a `--parallel` switch.

MartinNowak:
Single object compilation is simply not the right build mode for D, compiling small libraries at once is, because you can efficiently reuse parsed/analyzed module trees.

nordlow:
Has there been any progress on this?

Big projects, like one of mine (> 50 klines), simply cannot do without separate compilations, because there's not enough RAM on standard machines when DMD doesn't collect its garbage.

PR #504, opened 2015-02-04
Fix compilation with DMD git HEAD

Dicebot:
Recently merged cmake generator triggered one warning and one
error when compiled with DMD git HEAD.


SyntaxColoring:
:+1: I just ran into this problem as well.

andralex:
well lgtm :)

PR #505, opened 2015-02-09
Fixed singleFile build mode issues

machindertech:
+ Fixed error with long command line when invoking the linker on Windows
+ Fixed issue where singleFile build mode builds each file as a static library instead of as an object file
+ Fixed object file naming on Windows

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -189,14 +189,18 @@ class DmdCompiler : Compiler {
 	{
 		import std.string;
 		auto tpath = Path(settings.targetPath) ~ getTargetFileName(settings, platform);
-		auto args = [platform.compilerBinary, "-of"~tpath.toNativeString()];
+		auto args = ["-of"~tpath.toNativeString()];
 		args ~= objects;
 		args ~= settings.sourceFiles;
 		version(linux) args ~= "-L--no-as-needed"; // avoids linker errors due to libraries being speficied in the wrong order by DMD
 		args ~= settings.lflags.map!(l => "-L"~l)().array;
 		args ~= settings.dflags.filter!(f => isLinkerDFlag(f)).array;
-		logDiagnostic("%s", args.join(" "));
-		invokeTool(args, output_callback);
+
+		auto res_file = getTempFile("dub-build", ".lnk");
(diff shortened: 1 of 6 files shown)


source/dub/generators/build.d
```diff
@@ -350,7 +350,12 @@ class BuildGenerator : ProjectGenerator {
 	/// Output an unique name to represent the source file.
 	/// Calls with path that resolve to the same file on the filesystem will return the same,
 	/// unless they include different symbolic links (which are not resolved).
-	static string pathToObjName(string path) { return std.path.buildNormalizedPath(getcwd(), path~objSuffix)[1..$].replace("/", "."); }
+
+	static string pathToObjName(string path)
+	{
+		return std.path.stripDrive(std.path.buildNormalizedPath(getcwd(), path~objSuffix))[1..$].replace(std.path.dirSeparator, ".");
```
Geod24: Good catch.

source/dub/compilers/dmd.d
```diff
@@ -156,23 +156,23 @@ class DmdCompiler : Compiler {
...
-				settings.addDFlags("-lib");
-				break;
-			case TargetType.dynamicLibrary:
-				version (Windows) settings.addDFlags("-shared");
-				else settings.addDFlags("-shared", "-fPIC");
-				break;
-		}
-
-		if (tpath is null)
+		if (tpath is null) {
```
Geod24: I'm not sure I understand the logic here.
There are 2 very different codepath for this function, based on `tpath`. Looks like this function is being used for 2 differents purpose, isn't it ?

machindertech: The function is still being used to set the output file.  The only difference is that if tpath is set to something, then it means we're outputting an object file instead of the final target type.  In that case, we want to ignore the switch statement since that will also set the output type to something else (like a static library) and create a bunch of files that look like object files but aren't really.

MartinNowak: I think it would be better to fix the build settings here.
https://github.com/D-Programming-Language/dub/blob/76c8ec2166a4cfcb77193ffb5ff323087f4c5260/source/dub/generators/build.d#L355

machindertech:
Thank you for the feedback!  We've made changes to how object files are handled in the build settings based on the comments by MartinNowak.

PR #506, opened 2015-02-09
Fix: dub broken if compiled with absolute path DC

Dicebot:
More broken things found by my test suite ;)

Code that was supposed to find working compiler always tried to
build absolute path from compiler name - even it was absolute
path already.


MartinNowak:
tabs plz

MartinNowak:
Otherwise LGTM

MartinNowak:
I added a fixup

PR #508, opened 2015-02-10
Persistent Object Caching

nordlow:
This PR aims to implement robust caching of generated objects by adding logic that checks whether all sources files, compiler exectuables, etc an object depends upon are newer than the previously generated object file itself.

The logic will be activated only when the argument `--build-mode=singleFile` is fed to `dub`.

So far I've only added `-deps` flag to the compiler. Now I need guidance on

- where and how to parse the compiler output lines outputted by the `-deps` flag lines (lines beginning with `dep`).
- where to persistently save and load the associative array mapping source files to their previously scanned dependencies.

When this is done. I suggest to add logic in `BuildGenerator.compileUnit` that checks whether srcFile and all its dependencies are newer than the objPath (if it exists). 

A typical value for `objPath` is

`dub/build/application-debug-linux.posix-x86_64-dmd_2067-5C25CF5E0907389BD865D47A5E97D174/home.per.Work.justd.module_name.d.o`

I'm guessing this is a MD5-digest (32 hex digits, 16 bytes).
Should this digest be reused in a hash-chain used as keys in the persistent cache?

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -353,11 +353,16 @@ class BuildGenerator : ProjectGenerator {
 	static string pathToObjName(string path) { return std.path.buildNormalizedPath(getcwd(), path~objSuffix)[1..$].replace("/", "."); }
 	/// Compile a single source file (srcFile), and write the object to objName.
 	static string compileUnit(string srcFile, string objName, BuildSettings bs, GeneratorSettings gs) {
+            import std.stdio;
+            writeln("srcFile: ", srcFile);
+            writeln("objName: ", objName);
 		Path tempobj = Path(bs.targetPath)~objName;
 		string objPath = tempobj.toNativeString();
+                writeln("objPath: ", objPath);
 		bs.libs = null;
 		bs.lflags = null;
 		bs.addDFlags("-c");
+		bs.addDFlags("-deps");
 		bs.sourceFiles = [ srcFile ];
 		gs.compiler.prepareBuildSettings(bs, BuildSetting.commandLine);
 		gs.compiler.setTarget(bs, gs.platform, objPath);


MartinNowak:
We're using deps in phobos https://github.com/D-Programming-Language/phobos/blob/dc302d26a4168b2b21fb6cfc84edbcdbd32346a5/posix.mak#L336 and rdmd https://github.com/D-Programming-Language/tools/blob/ce191633a5777a25b5c2f0bfc084f72def1ccb06/rdmd.d#L581.

MartinNowak:
We already create unique target folders for each build config, hence the md5 hash in the target folder.
Just place any dependency metadata into this folder as well, it's also where the objects should be cached.

nordlow:
Ok, great, I'll reuse `getDependencies`. But where in the dub source can I get the output from the call to dmd?

Geod24:
Thanks for moving forward on this ! :)
The first problem I see ATM is that LDC does not implement `--deps`, only `--deps=file` (at least on my system).

> Ok, great, I'll reuse getDependencies. But where in the DUB source can I get the standard output from the call to dmd?
You can provide a delegate as a callback to Compiler.invoke. Arguments are return code and output. 
https://github.com/D-Programming-Language/dub/blob/ad9b43d6a150b53872db73c161e745c000cf8577/source/dub/compilers/dmd.d#L179

MartinNowak:
> The first problem I see ATM is that LDC does not implement --deps, only --deps=file (at least on my system).

No problem, simply write the deps file with the object, you want to cache it anyhow.

MartinNowak:
You've heard of [tup](http://gittup.org/tup/)? That tool tracks dependencies from bottom up thus achieves faster rebuilds because it doesn't have to scan all dependencies.

MartinNowak:
And if incremental rebuilds really become fast, I'd like to use them for a [dub watch command](https://github.com/D-Programming-Language/dub/issues/446).

MartinNowak:
Any updates on that?

nordlow:
How is this work related to https://github.com/D-Programming-Language/dub/issues/446?

MartinNowak:
> I'm asking because having a dynamic server-client solution for tracking file changes may obselete this solution.

Dub watch is intended for fast turnarounds when developing apps. There will be dub monitoring all dependencies and rebuilding/restarting the app.

PR #510, opened 2015-02-15
Stop linking standard library in build.sh

Dicebot:
Automatically linking proper static library for Phobos is duty
of proper compiler installation itself. Trying to repeat that
step in dub itself interferes with packaging if specific distro
uses different library names.

I may be missing the rationale for adding these line initially, if so -
please point them out and I will try to find an alternative solution.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -15,17 +15,6 @@ fi
 # link against libcurl
 LIBS=`pkg-config --libs libcurl 2>/dev/null || echo "-lcurl"`
 
-# fix for modern GCC versions with --as-needed by default
-if [ "$DC" = "dmd" ]; then
-	if [ `uname` = "Linux" ]; then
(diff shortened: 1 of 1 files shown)


s-ludwig:
These are workarounds for wrong linker command lines generated by DMD when linking with libcurl. I'm not sure if/when this was fixed, but it used to be the case that phobos came *before* curl on the command line, so that the curl symbols used by phobos were missing. The manual fix adds phobos a second time *after* curl to fix this.

MartinNowak:
You can probably replace it with the hack from the tools repo.
https://github.com/D-Programming-Language/tools/blob/546b9856427ed637fbffce11d5131a9c42d383b7/posix.mak#L105

I want to switch to dynamically loading curl but didn't yet got further than to add in it once.
https://issues.dlang.org/show_bug.cgi?id=13324

PR #511, opened 2015-02-16
copyFiles should not overwrite identical files on windows

MartinNowak:
This allows me to have bin and bin64 directories where first has 32 bit dll files and second has 64 bit dll files.
This makes program simpler because there is no difference in running 32 or 64bit executable since dlls are in the same location relative to executable location.

This can be archieved by supporting platform suffixes in targetPath and workingDirectory build settings.
Then I can do in `dub.json`
```
"targetPath" : "bin",
"workingDirectory" : "bin",
"targetPath" : "bin64",
"workingDirectory" : "bin64",
```

### DIFF
--- a/source/dub/internal/vibecompat/core/file.d
+++ b/source/dub/internal/vibecompat/core/file.d
@@ -137,6 +137,13 @@ void copyFile(string from, string to)
 
 version (Windows) extern(Windows) int CreateHardLinkW(in wchar* to, in wchar* from, void* attr=null);
 
+// guess whether 2 files are identical, ignores filename and content
+private bool sameFile(Path a, Path b)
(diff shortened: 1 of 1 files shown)


MrSmith33:
Added entry to cookbook https://github.com/D-Programming-Language/dub/wiki/Cookbook#creating-windows-package-that-has-32-and-64-bit-dlls

Geod24:
OT, but I don't know if you get notifications for edits to the wiki. I
added another entry to the cookbook. Very good idea btw :)

Mathias LANG
Epitech 2015

2015-02-16 23:25 GMT+01:00 Andrey Penechko <notifications@github.com>:

PR #513, opened 2015-02-16
add --version option to dub

MartinNowak:


Geod24:
It would probably be better to strip the CWD (e.g. `/home/travis/build/D-Programming-Language/dub/`), but other than that, LGTM.

source/dub/commandline.d
```diff
@@ -65,6 +65,7 @@ int runDubCommandLine(string[] args)
 	auto common_args = new CommandArgs(args);
 	try {
 		common_args.getopt("h|help", &help, ["Display general or command specific help"]);
+		common_args.getopt("version", &version_, ["Display dub version and exit"]);
```
Extrawurst: why no short version "v" ?

s-ludwig:
There is already a `--version` switch for certain commands (e.g. `dub fetch`), so those would clash. There are some switches à la `--print-platform`, so maybe `--print-version`?

MartinNowak:
> There are some switches à la --print-platform, so maybe --print-version?

Mmh, not really convinced. I made it so that `--version` is only recognized as first argument to dub itself.
That adds another "class" of options, but seems like the better choice here.
I have no idea about the nomenclature for those multi-command options, so I called them direct dub options.

MartinNowak:
> --version is only recognized as first argument to dub itself.

Git does the same.

s-ludwig:
Another possibility would be to make a separate `dub version` command, or maybe `dub info` with some more values. But the first parameter solution is probably OK, too, even if it lacks consistency. Any additional views on this? Otherwise I'd merge as is.

grogancolin:
command (--version | -v) seems to be pretty standard across CLI so I'd
guess it'd be good to stick to that as much as possible.
I find it annoying when a CLI doesnt have a -v switch ( dmd anyone? ).

Geod24:
2015-02-17 9:44 GMT+01:00 grogancolin <notifications@github.com>:

> command (--version | -v) seems to be pretty standard across CLI so I'd
> guess it'd be good to stick to that as much as possible.
> I find it annoying when a CLI doesnt have a -v switch ( dmd anyone? ).
>
>
`-v` is already used for verbose mode, which is widely used. `--version` is
rare enough, so it's not a problem to have only the verbose version.

s-ludwig:
Okay, then let's go with this.

MartinNowak:
> dmd anyone?

DMD has one ;).
https://github.com/D-Programming-Language/dmd/pull/4112

PR #515, opened 2015-02-19
Changed the init command to search for any packages it couldnt find on code.dlang.org

grogancolin:
Searches code.dlang.org for close package names if it couldnt find any during init.

Uses:
https://github.com/D-Programming-Language/dub-registry/pull/97

to achieve this.


grogancolin:
The build is failing because it cant download dmd v2.064.2. Did something change recently?

s-ludwig:
The path scheme for the downloads has changed recently (so that the year doesn't occur anymore). But it looks like only the downloads starting from 2.065 have been adjusted. @MartinNowak?

s-ludwig:
Looks good, except for spaces vs. tabs for indentation. BTW, if you can, I'd recommend to get a plugin for your editor that reads the .editorconfig file. Martin got one of these into many D repositories, so that it's not necessary to worry about such details most of the time.

I'll wait with the merge until we reach a conclusion in D-Programming-Language/dub-registry#86.

grogancolin:
I've reinstalled my OS recently. Still some teething issues with my .vimr.  Will fix these and recommit. Also noticed some erroneous debug messages I was using, will remove these too.

MartinNowak:
> The path scheme for the downloads has changed recently (so that the year doesn't occur anymore). But it looks like only the downloads starting from 2.065 have been adjusted. @MartinNowak?

Argh, that download URLs drive me crazy, how hard can it be.
Submitted a fix to the pending pull https://github.com/travis-ci/travis-build/pull/389.

MartinNowak:
> except for spaces vs. tabs for indentation

Might wanna try [EditorConfig](http://editorconfig.org/).

MartinNowak:
It could be done with a single round trip, if PackageSupplier.getVersions returned the suggestions as error body, but I think it's fine for now.

source/dub/packagesupplier.d
```diff
@@ -179,6 +186,11 @@ class RegistryPackageSupplier : PackageSupplier {
 		m_metadataCacheDirty = false;
 	}
 
+    Json searchForPackages(string[] names){
+        import std.array : join;
+        auto url = m_registryUrl ~ Path("api/search?q=" ~ names.join(","));
```
MartinNowak: Who does the URL encoding of the arguments?

grogancolin: When I went searching for what was encoding, I couldn't find any of the current uses of download() encoding the strings. To be safe however, I went and encoded it on the dub end using std.uri.encodeComponent (decoding on the dub-registry end with std.uri.decodeComponent).
Have tested and this works.

grogancolin:
Have updated the code with tabs/spaces. 
Also finally got editorConfig on my vim machine, about time.

I think I got them all, good to have a second pair of eyes over it though.

Also, what do we do about the FileSystemPackageSuppliers searchForPackage method? 
I haven't currently implemented anything there, and just threw in an assert(0) as it shouldnt ever be called really.

Does dub have a standard naming scheme for packages stored on its File System? I believe it's <package-name>-<version>.zip but I dont have any example on my local machine to check.
If it is so, it should be trivial to implement a function that will return a list of names similar to what RegistryPackageSupplier does now.

MartinNowak:
> Also, what do we do about the FileSystemPackageSuppliers searchForPackage method? 
> I haven't currently implemented anything there, and just threw in an assert(0) as it shouldnt ever be called really.

Sure, I thought it's in the normal `m_packageSuppliers` list.

> Does dub have a standard naming scheme for packages stored on its File System?

Try `dub list`, it's `${DUB_HOME}/packages/<package-name>-<version>/`.

grogancolin:
I will scrap this pull request and re-do it with the new changes @MartinNowak made to dub-registry in https://github.com/D-Programming-Language/dub-registry/pull/103

MartinNowak:
That would be very welcome.

PR #516, opened 2015-02-19
Use booleans instead of strings to represent version range limits.

s-ludwig:
Also disallows specifying invalid version ranges as strings. This makes the code less error prone and easier to reason about. Passes all tests, but could use a second pair of eyes to verify.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -33,9 +33,9 @@ struct Dependency {
 	private {
 		// Shortcut to create >=0.0.0
 		enum ANY_IDENT = "*";
-		string m_cmpA;
+		bool m_inclusiveA = true; // A comparison > (true) or >= (false)
 		Version m_versA;
-		string m_cmpB;
+		bool m_inclusiveB = true; // B comparison < (true) or <= (false)
 		Version m_versB;
 		Path m_path;
 		bool m_optional = false;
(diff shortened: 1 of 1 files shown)


source/dub/dependency.d
```diff
@@ -33,9 +33,9 @@ struct Dependency {
 	private {
 		// Shortcut to create >=0.0.0
 		enum ANY_IDENT = "*";
-		string m_cmpA;
+		bool m_inclusiveA = true; // A comparison > (true) or >= (false)
 		Version m_versA;
-		string m_cmpB;
+		bool m_inclusiveB = true; // B comparison < (true) or <= (false)
```
MartinNowak: I'd use mathematical boundaries to describe this, as in `"[)"`.
We use that in [uniform](http://dlang.org/phobos/std_random.html#.uniform) and independently from that I came up with it as well in my graphics library.
https://github.com/MartinNowak/graphics/blob/ea9c7fcea56993d8f48aef3b0f602f5a92c783bf/src/graphics/math/clamp.d#L19
You also only need `<`, `<=` and `==` by swapping the arguments to `>` or `>=`, as in the linked code above.

MartinNowak: > You also only need <, <= and == by swapping the arguments to > or >=, as in the linked code above.

You did that already with doCmp.

MartinNowak:
I find this very hard to follow, but this is mostly a refactoring and all the replacements looks correct.

source/dub/dependency.d
s-ludwig: "([" is definitely a convenient solution for template interfaces, but for the internal representation it seems to be sub-optimal. It means having magic constants in the code and using a type that has a greater value range than the problem domain.

Maybe a dedicated interval type, usable at runtime or at compile time, in Phobos would be nice: `Interval!int(3, 15, "(]")`
This would then define common operations (union, intersection, membership test and possibly arithmetic operations on the limits) and could then internally drive all places that work with such ranges.

BTW, I'll rename the `A` and `B` to `Min` and `Max` to make their semantics clearer.

s-ludwig:
Yeah, it was a bit unfortunate that the new representation caused some small semantic changes. It would have taken additional logic to simulate the old (useless) behavior in some places, so I left it out.

PR #518, opened 2015-02-23
Create empty DGui skeleton project

andre2007:
A minimal DGui application skeleton can be built

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -329,6 +329,7 @@ class InitCommand : Command {
 			"minimal - simple \"hello world\" project (default)",
 			"vibe.d  - minimal HTTP server based on vibe.d",
 			"deimos  - skeleton for C header bindings",
+			"dgui  - minimal DGui application",
 		]);
 	}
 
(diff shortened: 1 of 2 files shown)


Geod24:
I really think we should have a "plugin" system for that. It would mostly work by convention, so that for example, `dub init vibe.d` will look for `dub-init` folder (or subpackage) in vibe.d package, and copy it. That would also fix the issues of the packages not being in sync).

s-ludwig:
See also my reply in #453.

etcimon:
> I really think we should have a "plugin" system for that. It would mostly work by convention, so that for example, dub init vibe.d

A true plugin system would automatically search for and run shared libraries, and identify a set of callbacks, related or not to `init` files. This would be useful for eg. adding custom pre-processors, like a diet templates compiler that isn't CTFE but runs at compile-time.

s-ludwig:
There are many flavors of plugin systems, it doesn't always have to be shared library based. In this particular case I think that a template package is really a much better solution. But yeah, a "build" plugin system like you describe is also something that we need to think about. I've recently talked about such possibilities with Atila Neves via e-mail and could post those in a separate ticket or on the forums.

MartinNowak:
> See also my reply in #453.

https://github.com/D-Programming-Language/dub/pull/453#issuecomment-75568856

blm768:
Isn't this basically a subset of the features in #600?

s-ludwig:
@blm768: Yes, I think we should actually close this and go for a variant of #600 instead (using a sub package instead of a specific sub path).

PR #519, opened 2015-02-23
Test dub with DMD 2.067 beta

Geod24:


Geod24:
@MartinNowak : http://downloads.dlang.org/releases/2.x/2.064/ , not 2.064.2 (I don't understand it either...)

MartinNowak:
I made a fix already, supposed to be deployed to Travis today.

PR #523, opened 2015-03-07
Update CHANGELOG.md

grogancolin:
Removed the " - Non-existing packages will be ignored" line from the dub init section. 
This was removed after I wrote the changelog, now non-existant packages will cause dub to cancel the init.


MartinNowak:
> Pull Requests that have a failing status can’t be merged from a phone.

Well

s-ludwig:
> Pull Requests that have a failing status can’t be merged from a phone.

Sounds like they think they know best what's good for you...

PR #446, opened 2015-03-18
add dub watch command

MartinNowak:
Useful to automatically rebuild dub projects during development.
The command should do the same as dub run, but whenever any dependency changes it stops the process, rebuilds and restarts.
I'm willing to invest some time into this.
Maybe the command will only be available when dub was build with vibe.d or we roll a filewatcher by hand.


MartinNowak:
This doesn't yet work on Windows, but I wanted to get some early feedback.
Also the test seems to fail sometimes, maybe because the app is rebuild before the file is written.

MartinNowak:
> Also the test seems to fail sometimes, maybe because the app is rebuild before the file is written.

Fixed the test issue, on some filesystems (VirtualBox shared folder, XFS on docker) the modified notification arrived before the file was fully written. Adding a `1.msecs` sleep before killing the app seems to resolve the problem.

MartinNowak:
I could need some help with the Windows part of this, I disabled it for now so that we can already pull the existing change.
A followup improvement would be to track newly added files when recompiling a project, right now watches are only installed initially.

source/dub/generators/build.d
```diff
@@ -450,39 +457,103 @@ class BuildGenerator : ProjectGenerator {
...
+		auto cwd = Path(getcwd());
+		auto runcwd = switchWorkDir(cwd, buildsettings.workingDirectory);
+		scope(exit) chdir(cwd.toNativeString());
+
+		auto cmd = runCmd(exe_file_path, cwd, runcwd, run_args);
+		assert(!settings.runCallback);
+
+		while (true)
+		{
+			logInfo("Running '%-(%s, %)'", cmd);
```
s-ludwig: Interesting, never seen that format syntax before. Shouldn't it be `%-(%s %)` (without the comma), though?

s-ludwig:
I'm going to try this a bit on Linux and merge as a WIP. As soon as I get some time, I'll then add a Windows watcher.

source/dub/generators/build.d
MartinNowak: Indeed, there shouldn't be a comma.

MartinNowak:
> I'm going to try this a bit on Linux and merge as a WIP. As soon as I get some time, I'll then add a Windows watcher.

Please only merge this if you have time to follow-up to avoid release a half-finished feature.
The other important TODO is tracking dependency changes and updating the watches accordingly.
Should be simple to add a removeFile to the Watcher and diff a set of all watched files.

s-ludwig:
Can you push your branch to the DPL repository? We could finish it there before merging (will be after the 0.9.24 release anyway).

jcao219:
I've written up the Windows component of the Watcher, and I'l looking for feedback:  https://github.com/MartinNowak/dub/pull/1

Watcher does eventually need to track the json/sdl config file, and if that changes, dub could simply restart itself.

MartinNowak:
> Watcher does eventually need to track the json/sdl config file, and if that changes, dub could simply restart itself.

Good point, my implementation also doesn't keep track of a changing set of source files.
Maybe we need to move up the rebuild loop to before the config files are parsed. This would be less efficient (b/c the watches get setup everytime) but might still work well enough. We could later try to preserve and update the list of watches across rebuilds.

MartinNowak:
Any opinions on the design @s-ludwig? It already work nice enough to be highly useful, but in order to be accurate we'd have to regenerate the list of source files and reparse the config.
So maybe this should instead be a wrapper around dub run?

s-ludwig:
I'd say the watch logic should be in the `Dub` class. I can invoke `Dub.generateProject` whenever source files are added/removed/modified and can also reinitialize the project if the package recipe or the selections change (`Dub.upgrade` to add new dependencies, access `Project.selections` for reloading selections, and calling `Project.reinit` to load the changed dependency graph).

MartinNowak:
Sounds good.
Are we confident that a simply rebuilding the whole project will be fast enough, or that we could somehow leverage the list of changed files to speed it up if necessary?

s-ludwig:
File scanning tends not to be a big factor for typical projects, so leveraging the information from the directory watcher would be nice to have, but IMO it's not important enough to delay the introduction of a simpler watch command.

aliak00:
Is the missing piece for this pr windows related help?

aliak00:
Would a way to go about this to just use [Fswatch](https://code.dlang.org/packages/fswatch)? which seems to be cross platform proper.

jacob-carlborg:
> Would a way to go about this be to just use Fswatch? which seems to be cross platform proper.

Does not support macOS, at least not natively.

s-ludwig:
It should be noted that both, libasync and eventcore, support macOS for their FS watchers, and the sources of both can be reused without issues. External dependencies are currently not allowed for DUB, so the code should be copied instead of referenced.

jacob-carlborg:
Neither eventcore nor libasync are using the FSEvents API which is the preferred API on macOS. eventcore has a stubbed out implemented for FSEvents.

s-ludwig:
But kqueue should be fine, shouldn't it? The only drawback is that it requires manual recursive watch management.

jacob-carlborg:
> But kqueue should be fine, shouldn't it? The only drawback is that it requires manual recursive watch management.

That requires extra resources (file handlers).

etcimon:
At the time I studied solutions on Mac for the file watcher, I came across multiple problems with FSEvents. 1) It was very well documented for Objective C, could barely find any resources about using it in C. 2) I couldn't find out how I would use it in an event loop (kqueue). 3) It was fairly new it seems, not sure if it could have been compatible with older versions of Mac OS. 4) Writing the kqueue handler had the advantage of being also compatible with other BSD platforms.

Perhaps a few things have changed since then making it possible and easier

jacob-carlborg:
>  It was very well documented for Objective C, could barely find any resources about using it in C

First documentation link on Wikipedia for FSEvents links to examples using the C API [1].

> 3) It was fairly new it seems, not sure if it could have been compatible with older versions of Mac OS

It was introduced in macOS 10.5 and enhanced in version 10.7. Current version of macOS is 10.14 and D has a minimum requirement of 10.7 for the target system. So it's quite old now.

> 4) Writing the kqueue handler had the advantage of being also compatible with other BSD platforms.

Sure, it doesn't mean a kqueue handler cannot be implemented as well.

[1] https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/FSEvents_ProgGuide/UsingtheFSEventsFramework/UsingtheFSEventsFramework.html

etcimon:
That page from apple doesnt go past the table of contents from android chrome or firefox. I mean, the whole thing I had to develop on a pirated copy of vmware osx that were in torrents, because otherwise you need to buy a whole machine just do try it. I'll check when i get home but having the c headers directly with a nice flat guide like linux does on its kernel would be ideal.

etcimon:
Seems like a pretty big job to translate the core foundation headers to D. I'm not even sure we have the right to do that, considering the lawsuits from oracle to google for using java headers. Ideas?

rikkimax:
@etcimon Don't worry about it. Bindings are not the same as reimplementing the whole damn thing with modifications.

etcimon:
Ok, so I guess the next problem is, if FSEvent is supposed to be so easy, how come we don't even have Core Framework in druntime? Or in the whole dlang community? Not even a couple pieces of it?

etcimon:
Seems like there is this binding here we can use as a starting point: https://github.com/garlic-flavor/WebKit2d/tree/master/src/CoreFoundation

jacob-carlborg:
> Ok, so I guess the next problem is, if FSEvent is supposed to be so easy, 

No one said FSEvents was easy.

> how come we don't even have Core Framework in druntime? Or in the whole dlang community? Not even a couple pieces of it?

druntime only includes bindings for C and Posix or what's needed for druntime. Some people would prefer that druntime didn't include any public bindings. You'll get quite far with Posix bindings, I guess there has been no need for bindings to the Core Framework. It doesn't help that macOS is probably at the third place of the D user base. Those that are using macOS and D seem to only use Posix layer and no Apple specific libraries (except me).

jacob-carlborg:
I have a set of bindings for Apple specific libraries [1] but they were generated 9 years ago for D1, no idea if they'll work.

[1] http://dsource.org/projects/dstep/browser/dstep

jacob-carlborg:
The official documentation is available here [1]. It's a C API although it says that the language is Objective-C.

[1] https://developer.apple.com/documentation/coreservices/file_system_events?language=objc

Beyarz:
I think @andre2007s idea is good, we should implement this feature first and then work on optimizations.

drannex42:
Do we have any updates on this PR? Or has this already been implemented in any other way that I haven't been able to find?

PR #530, opened 2015-03-21
Doesn't play well with Emacs lockfiles

MartinNowak:
When there is an Emacs lockfile in the project directory (that starts with a ".#"), "dub describe" fails:

    $ dub describe
    Error executing command describe: /h/30/ozan/experiment/source/.#experiment.d: No such file or directory

It seems like for locking files Emacs creates a symlink that is not followable. Is is possible to exclude suh files from dub commands?


MartinNowak:
At first I thought about adding a specific handling of this case for `dub describe`, but it might make more sense to ignore hidden source files by default.
This also allows to build and run a project with a modified emacs buffer, same as with any other editor.

ozanerdem:
That would work for me -- I don't think I ever needed a hidden source file for any of my builds.

s-ludwig:
I think ignoring hidden source files is generally a good idea, they form an invalid module name anyway (at least on `Posix` systems), which could be problematic in certain cases. And I don't see a reason to *include* them to the build (looking at the problem in an additive way) - why would you want to deliberately hide one of the source files from the user?

MartinNowak:
That was my though too, so I kind of liked this solution more than a custom hack for emacs.
I was a little worried about being too blunt here, someone might want to generate source files before building and hide them, but it's a really small use-case as well and a separate output directory could be used instead.
They are invalid module names on windows too, as an [identifier](http://dlang.org/lex.html#Identifier) might not contain a dot. This already makes it fairly impractical, so we should be good to go with this.

s-ludwig:
> They are invalid module names on windows too

Dot`files are of course, I was just thinking of files with the hidden attribute set. But ignoring those would basically just be yet another rule, so maybe we can just ignore all dot files on all systems.

MartinNowak:
This is what this pull does, so then go on and merge it.

PR #538, opened 2015-03-31
Support packages without "lib" prefix in pkg-config

### LINKED ISSUE #300: pkg-config support
It'd be nice to have pkg-config support for DUB.

tomykaira:
This fixes #300.

I met the same problem with opencv.
Packages without "lib" prefix seems common between larger projects.

This patch tries `"lib"~lib` and `lib`, then add existing one to `pkgconfig_libs` list.


MartinNowak:
thx

PR #539, opened 2015-04-06
replace for loops and additional filtering with partition

MartinNowak:


MartinNowak:
What I noticed though when this pull was done, this doesn't preserve the order of the libraries, neither does old code.

s-ludwig:
The order should definitely be fixed in the long term, I didn't notice this in my initial implementation. But making that work in general is a little more involved, because currently three build settings fields are involved in specifying libraries (`"dflags"`, `"lflags"` and `"libs"`). We'd have to lower everything to `"dflags"` or `"lflags"`, possibly depending on the compiler.

MartinNowak:
Can we still merge the code cleanup?

s-ludwig:
Definitely! Forgot about that.

PR #546, opened 2015-04-13
exit with non-zero status code if run is skipped

MartinNowak:


s-ludwig:
Looks like the library test directories need to get a `.no_run` file in their path. I'm also thinking about changing the test runner to expect an individual `test.sh` in each directory, because building multi-package tests requires more specific invocations. We could instead extend the configurability of tests, but to me it seems like the added complexity may be worse than having a little redundancy.

MartinNowak:
Fixed, you can already use test scripts, see #551.

MartinNowak:
Motivation for this was that `dub dustmite` didn't complain about skipping execution.

s-ludwig:
Ah OK good to know...

PR #547, opened 2015-04-13
use thisExePath in recursive dub invocations

MartinNowak:


s-ludwig:
Hm.. that LDC failure is strange, maybe an OOM error?

MartinNowak:
Updated to also pass down config options.

PR #552, opened 2015-04-18
Add a command to DUB to list all of the import paths for a project.

w0rp:
Related to a recent discussion, this patch will add a command to dub to list all of the import paths used
for a project, line-by-line.

There might be some details I missed here, so I would appreciate some input. I have next to no experience with the DUB codebase.


MartinNowak:
I think it would be nicer to add this as `dub describe --import-paths` and `dub describe --string-import-paths` rather than adding lots of new commands.

MartinNowak:
A test for this and `dub describe` would be nice as well, https://coveralls.io/builds/2369240/source?filename=source%2Fdub%2Fcommandline.d#L761.
See [0-init-simple.sh](https://github.com/D-Programming-Language/dub/blob/c04d45d1882cb47b821c9d3aa1af858a845332b8/test/0-init-simple.sh) or [test-version-opt.sh](https://github.com/D-Programming-Language/dub/blob/c04d45d1882cb47b821c9d3aa1af858a845332b8/test/test-version-opt.sh).

w0rp:
Okay then. I'll see about adding the command as switches on `dub describe` instead. I'll see if I can add something for `--string-import-paths`. I believe it's not so hard for me to grab those either. I saw the message before about decreased code coverage, and I was wondering how to add tests for the project. It's slightly more complicated than my usual process of adding a `unittest` block.

I might have to come back to this during the week, but I'll get it done eventually.

MartinNowak:
> I was wondering how to add tests for the project

Just add a shell script, `dub init` a project, run `dub describe --import-paths` and grep for the source paths or so, see [test-version-opt.sh](https://github.com/D-Programming-Language/dub/blob/c04d45d1882cb47b821c9d3aa1af858a845332b8/test/test-version-opt.sh).

MartinNowak:
You found some time to implement it?

w0rp:
I just updated this again with the options as you requested. I added some test scripts for checking the output of each option, and a smoke test for dub describe. Maybe something could be done to test the JSON output for `dub describe`.

I modified the script which runs the other test scripts so the `die` and `log` functions can be used in them, as well as the `CURR_DIR` variable.

I'll see if I can figure out why the Travis CI script works on my machine, but fails on Travis CI.

w0rp:
There we go. I was dumb and committed my absolute paths, so now the scripts generate them in the environment. I also needed to add some dummy files so the right directories will be created, for the paths dub uses automatically.

source/dub/dub.d
```diff
@@ -422,6 +422,29 @@ class Dub {
 		write(dst.toPrettyString());
 	}
 
+	private void listPaths(string methodName)(BuildPlatform platform, string config)
+	{
+		string[] importPaths;
+
+		__traits(getMember, m_project, methodName)(importPaths, platform, config);
```
MartinNowak: You can return the paths array insyead of making it an out variable.

source/dub/dub.d
```diff
@@ -422,6 +422,29 @@ class Dub {
...
+
+		import std.stdio;
+
+		foreach(path; importPaths) {
+			writeln(path);
```
MartinNowak: This is the only code you share, doesn't justify the template function IMO.

source/dub/dub.d
```diff
@@ -422,6 +422,29 @@ class Dub {
...
+		}
+	}
+
+	void listImportPaths(BuildPlatform platform, string config)
+	{
+		listPaths!"listImportPaths"(platform, config);
```
MartinNowak:
```d
foreach (path; m_project.listImportPaths(platform, config))
    wrintln(path);
```

test/4-describe-import-paths.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+cd "$CURR_DIR"/describe-project
+
+temp_file=`mktemp`
+
+function cleanup {
+    rm $temp_file
+}
```
MartinNowak:
```sh
trap cleanup EXIT
```
http://redsymbol.net/articles/bash-exit-traps/

test/4-describe-import-paths.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
MartinNowak: Better use `set -e -o pipefail` instead of explicitly checking return codes.

MartinNowak:
Looks good, thanks.

source/dub/dub.d
w0rp: Fair point. I'll just essentially write the two functions out instead. They are quite small anyway.

source/dub/dub.d
w0rp: Can do. I was just following the way the other methods in the class were using output parameters.

test/4-describe-import-paths.sh
w0rp: Good suggestion. I've used `trap` before. I'll use it here.

test/4-describe-import-paths.sh
w0rp: Good idea. I'll do that.

w0rp:
I updated the D source files and the Bash scripts now. I created some very basic tests with `jq`, but I couldn't seem to get Travis to use it. I did see a whitelist for apt packages on Travis, but `jq` doesn't appear to be in it. I've pulled that out for now.

MartinNowak:
It should be preinstalled (http://docs.travis-ci.com/user/build-environment-updates/2014-12-09/#Additions), but not too important right now.

Abscissa:
W000t! This is awesome! Thanks a ton for this, @MartinNowak!!

source/dub/commandline.d
```diff
@@ -746,13 +746,28 @@ class TestCommand : PackageBuildCommand {
...
+	}
+
 	this()
 	{
 		this.name = "describe";
 		this.argumentsPattern = "[<package>]";
 		this.description = "Prints a JSON description of the project and its dependencies";
 		this.helpText = [
-			"Prints a JSON build description for the root package an all of their dependencies in a format similar to a JSON package description file. This is useful mostly for IDEs.",
+			"Prints a JSON build description for the root package an all of "
```
grogancolin: an = and?

PR #556, opened 2015-04-24
`enforce()` must take a condition, not just as message

schuetzm:
Caught by the new "array in bool context" warning in DMD.


s-ludwig:
This is fixed in a [separate branch](https://github.com/D-Programming-Language/dub/blob/refactor1/source/dub/internal/vibecompat/data/json.d#L1998) now by upgrading to the latest vibe.d code. I'd try to directly get that into master rather than fixing it in the old code, so that there will be no avoidable merge conflicts.

andralex:
I'll close this then.

Issue #562, opened 2015-05-02
Expanding on --import-paths for other BuildSettings fields


Abscissa:
I'd like to expand on the recent "dub describe --[string-]import-paths" to support some of the other fields in BuildSettings (such as libs and versions), and also optionally allow these to be output pre-formatted for a particular compiler (such as DMD-style: "-Ipath1 -Ipath2").

But before getting started, I wanted to get feedback regarding the CLI interface for it. I'm wondering whether following the example set by "--[string-]import-paths" by adding "--versions", "--debug-versions", "--libs", "--pre-generate-commands", etc might be a bit cluttered. Should it be changed to something like "--build-setting=import-paths"? Or is current approach fine?

w0rp:
I think I'd be happy with just about any approach which works and lets you get all of the values out in the right formats. When I added the path commands, I just wanted something so I could get the paths out without having to rely on some other program being installed, to make it easy to add the import paths to any syntax checker which uses dub already.

MartinNowak:
Mmh, what's wrong with processing the json output? After all you don't want 10 invocations of dub anyhow. What might make sense is something akin to `pkg-config --cflags --libs`, e.g. `dub describe --dflags --libs`, but I'm not sure how well that works with dependencies.

Abscissa:
> Mmh, what's wrong with processing the json output?

It's a fairly fundamental package manager feature that (realistically) is necessary for using dub's package management functionality while opting out of its build system. Because of that, an external tool (while probably technically feasible) is really a bit too unwieldy and inconvenient.

> After all you don't want 10 invocations of dub anyhow. What might make sense is something akin to ```pkg-config --cflags --libs```, e.g. ```dub describe --dflags --libs```,

I'm not all that familiar with pkg-config (I really should be though...), but I assume you're talking about  outputting multiple types of flags (compiler flags and library filepaths) with the same invocation. If so, yes, I was thinking of allowing the same thing, like you say, to avoid extra invocations.

For outputting the values in a compiler-formatted style that shouldn't be too much of an issue. I'm not entirely certain what to do about the current formatting of "list each value, compiler-independent format, on a separate line". Either just dump everything and let the user fend for themself at sorting them out. Or separate each section with a blank line. Or just ONLY allow ```dub describe --dflags --libs``` combinations of data when outputting in a compiler-specific command line format.

> but I'm not sure how well that works with dependencies.

I'm not sure what you mean by this. What potential issues do you see?

[...]

PR #566, opened 2015-05-05
#562: Expand on --import-paths to support other build settings and optio...

Abscissa:
...nal recursion into dependencies.

I'll add optional alternate output formatting (ie, the ```--format=(dmd|list)``` described in #562) in a separate pull request.


Abscissa:
May want to hold off on actually merging until I adjust the tests to handle these features. Also, a few of the settings don't seem to have any value, I'll need to look closer at those and possibly comment out any that may require further effort to get working.

Abscissa:
I still have some more work to do on this, so I'm going to temporarily close this until it's ready.

PR #598, opened 2015-06-21
Fix compilation failure on DMD/Phobos master

Dicebot:
Because of range-ification of some Phobos functions
implicit convertion of DirEntry to string does not
work anymore.

Looks like this won't be considered regression, thus submitting the fixup.


s-ludwig:
Interesting. I wonder if `alias this` shouldn't cause the range predicate of `globMatch` to match anyway...

s-ludwig:
Just hat a look and `isForwardRange` is the offender: It checks if `is(typeof(T.save) == T)`. To make `alias this` work, it would have to do something weaker, such as `is(T : typeof(T.save))`.

PR #600, opened 2015-06-22
add custom type for dub init

### LINKED ISSUE #557: Global user configuration
Is it possible to set global user configuration?
It would be handy to set default author's name, license, copyright and homepage to fill the newly initialized dub.json with them without the need for copy-pasting.

deviator:
- rework init types (now enum)
- add custom type (issue #557)

Example usage:
`dub init MyApp gtk-d deslog -tcustom` copies files from `.dub/custom_init_package/` and replace in them some vars:
- $name -> myapp
- $Name -> MyApp
- $deps_json -> "gtk-d": "3.1.3", "deslog": "0.2.0"

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -20,6 +20,7 @@ import dub.packagesupplier;
 import dub.platform : determineCompiler;
 import dub.project;
 import dub.internal.utils : getDUBVersion, getClosestMatch;
+import dub.init;
 
 import std.algorithm;
 import std.array;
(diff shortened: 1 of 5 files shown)


MartinNowak:
I guess your use-case is similar to git's [template directory](http://www.git-scm.com/docs/git-init#_template_directory), right? What exactly are you using this for?

It seems that copying default files in a dub directory is orthogonal to initializing a particular kind of dub folder. It's also fairly easy to write your own dub_init wrapper script.

For the purpose of having special init types for other libraries, I think it'd be best to do this as executable targets in the libraries, e.g. `dub init -t mylib args` would fetch mylib and run `dub run -c init mylib -- args`.

s-ludwig:
This looks like (form a really quick glance) it's simply the generalization of the `-t` switch that we've discussed in similar form some time ago. Unfortunately I can't find the discussion now, but the idea was different in that `dub init somelib` would automatically look for a sub package `somelib:init` and use that as the package template. For multiple libraries specified, I think we came to the conclusion to only look at the first one.

I think choosing the default like that makes more sense. It would usually be what people expect, and help pages are rarely looked at, so that an opt-in switch would probably end up being rarely used.

andre2007:
See pr https://github.com/dlang/dub/pull/1658

andre2007:
Closing as https://github.com/dlang/dub/pull/1658 is merged.

PR #593, opened 2015-06-18
Send additional env vars to build commands.

Abscissa:
Some of these (especially ```DC```, ```DUB_CONFIG```, ```DUB_BUILD_TYPE```, ```DUB_PACKAGE_DIR```, and ```DUB_RUN```) can be quite useful.

PR #607, opened 2015-06-26
Improve "dub describe" and use BitFlags

s-ludwig:
The commits by @Abscissa have already been reviewed as part of #572.


s-ludwig:
Main changes to "dub describe", apart from #572, is the addition of a "targets" field that lists the binary build targets, and making the API statically typed (vs. the old `Json` based API).

s-ludwig:
Pulling to prepare a beta-2 release with this included.

Abscissa:
Awesome :)

Is there any chance of #593 also getting into 0.9.24? That would also be really nice to have.

MartinNowak:
No we have all those fancy describe functions with lots of description and plenty of tests and I still can't get a plain list of dependencies. Are you sure this goes in the right direction?
This detailed level of introspection is better handled by using dub as library IMO. Making the latter more accessible/stable will improve the quality of dub while adding many more knobs and switches creates a maintenance nightmare.

Abscissa:
@MartinNowak: These "knobs and switches" are necessary to decouple package management from build system, which is necessary for other reasons. Whether or not there are other things *also* needed, like getting a plain list of deps (a good idea, btw), is a separate matter and has no bearing on whether other features are worthwhile. I've been trying to hold back on this, but honestly, I'm getting tired of your constant pushback and naysaying at nearly every attempt to improve anything in D. Progress doesn't come from blocking improvements.

MartinNowak:
Well have you even considered to use dub as library to get the data?
We have limited resources so it's important to think twice before increasing maintenance cost.
Also I don't clearly see the goal (or the outline of a plan) for this rather pedantic build tool vs. package manager distinction.

Abscissa:
On 10/31/2015 10:34 AM, Martin Nowak wrote:
> Well have you even considered to use dub as library to get the data?

You've brought that up several times before, and I've already addressed 
it. I'm done bikeshedding.

> We have limited resources so it's important to think twice before increasing maintenance cost.
> Also I don't clearly see the goal (or the outline of a plan) for this rather pedantic build tool vs. package manager distinction.
>

Where's the goal/plan for conflating to two? Again, this has already 
been discussed before, ad nauseum.

Issue #608, opened 2015-06-28
Not all DUB output is interesting to read


p0nce:
**Output today**

```
Building logger 0.3.3 configuration "library", build type debug.
Running dmd...
Building colorize 1.0.5 configuration "library", build type debug.
Running dmd...
Building gfm:logger 2.3.15 configuration "library", build type debug.
Running dmd...
Building derelict-util 2.0.0 configuration "library", build type debug.
Running dmd...
Building derelict-sdl2 1.9.2 configuration "library", build type debug.
Running dmd...
Building gfm:sdl2 2.3.15 configuration "library", build type debug.
Running dmd...
Building gfm:math 2.3.15 configuration "library", build type debug.
Running dmd...
Building derelict-gl3 1.0.12 configuration "library", build type debug.
Running dmd...
Building gfm:opengl 2.3.15 configuration "library", build type debug.
Running dmd...
Building simpleshader2 ~master configuration "application", build type debug.
Compiling using dmd...
simpleshader.d(193,5): Error: found '}' when expecting ';' following statement
simpleshader.d(196,5): Error: found '{' when expecting ';' following statement
FAIL .dub\build\application-debug-windows-x86-dmd_2068-FC44C28BB4F2F54AFF252D359
5300845\ simpleshader2 executable
Error executing command build:
dmd failed with exit code 1.
```

**What I'd rather like**

```
Building logger 0.3.3 configuration "library", debug build.
Building colorize 1.0.5 configuration "library", debug build.
Building gfm:logger 2.3.15 configuration "library", debug build.
Building derelict-util 2.0.0 configuration "library", debug build.
Building derelict-sdl2 1.9.2 configuration "library", debug build.
Building gfm:sdl2 2.3.15 configuration "library",debug build.
Building gfm:math 2.3.15 configuration "library", debug build.
Building derelict-gl3 1.0.12 configuration "library", debug build.
Building gfm:opengl 2.3.15 configuration "library", debug build.
Building simpleshader2 ~master configuration "application", build type debug.
simpleshader.d(193,5): Error: found '}' when expecting ';' following statement
simpleshader.d(196,5): Error: found '{' when expecting ';' following statement
dmd failed with exit code 1.
```

- I've removed `Running dmd...` because it's redundant, I know I build with DMD because I type the DUB command. It would be allright if said only once and ideally none. If I want to know what DUB really does, I will be interested in the whole command line and `-v` is there.
[...]

PR #621, opened 2015-07-10
Logging changes

p0nce:
PR for #608

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -223,7 +223,6 @@ int runDubCommandLine(string[] args)
 		return 1;
 	}
 	catch (Throwable e) {
-		logError("Error executing command %s:", cmd.name);
 		logError("%s", e.msg);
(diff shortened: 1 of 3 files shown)


p0nce:
Bump!

s-ludwig:
Looks good. The only thing I wonder that might still be worth to have is the "Linking..." message, so that failures can be immediately associated to either the compiler or the linker. Thoughts?

p0nce:
Yes. It's the one I was most worried of removing.
Also having a display of the used compiler + arch one time would be great if done one time before building.

s-ludwig:
Thinking about the latter, since the build type is also constant across all targets, that could also be moved to a single message at the beginning.

p0nce:
Reintroduced the Linking... messages.
About the single message at beginning, this need someone more knowledgeable in the codebase.

p0nce:
Rebased.

s-ludwig:
Okay, thanks! I'll have a look at that message at the beginning right now.

Issue #601, opened 2015-06-23
GNU/Linux [GDC/DMD] RangeError in path.d


bittwiddler1:
Similar to issue #137, my dub binaries throw RangeErrors when they are built with GDC or DMD. When built with LDC2, everything works perfectly.

Even when building dub with dub, this error manifests...
```
~/.../git/dlang/dub $  dub --compiler=dmd
....
iterating dir /home/bittwiddler/.dub/packages/ entry inifiled-0.0.3
Collecting dependencies for dub
Missing dependency vibe-d 0.7.23 of dub
Checking for missing dependencies.
Downloading metadata for memutils
Getting from http://code.dlang.org/packages/memutils.json
Getting http://code.dlang.org/packages/memutils.json...
Downloading metadata for libevent
Getting from http://code.dlang.org/packages/libevent.json
Getting http://code.dlang.org/packages/libevent.json...
Downloading metadata for vibe-d
Getting from http://code.dlang.org/packages/vibe-d.json
Getting http://code.dlang.org/packages/vibe-d.json...
Downloading metadata for openssl
Getting from http://code.dlang.org/packages/openssl.json
Getting http://code.dlang.org/packages/openssl.json...
Downloading metadata for libev
Getting from http://code.dlang.org/packages/libev.json
Getting http://code.dlang.org/packages/libev.json...
Downloading metadata for libasync
Getting from http://code.dlang.org/packages/libasync.json
Getting http://code.dlang.org/packages/libasync.json...
Configurations used for dependency resolution:
Interation (ci=-1)
Refreshing local packages (refresh existing: false)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/bittwiddler/.dub/packages/local-packages.json
iterating dir /home/bittwiddler/.dub/packages/
iterating dir /home/bittwiddler/.dub/packages/ entry libdparse-0.2.0
iterating dir /home/bittwiddler/.dub/packages/ entry inifiled-0.0.3
Collecting dependencies for dub
Missing dependency vibe-d 0.7.23 of dub
Checking for upgrades.
Using cached upgrade results...
Add config dub application
Add config dub library
Add config dub library-nonet
NON-PRIMARY:
Eliminating config library for dub
NON-PRIMARY:
Eliminating config library-nonet for dub
Using configuration 'application' for dub
Generating using build
Creating build generator.
Add config dub application
Add config dub library
Add config dub library-nonet
NON-PRIMARY:
[...]

PR #625, opened 2015-07-13
Bugfix for #601. Inplace assignment of 'p = p.relativeTo(cwd)' was

bittwiddler1:
causing memory corruption, probably due to reallocations. 
Using a second variable to hold the result of the operation as a fix.

Was a little hard to catch because of the single-line function.  I spread the operations out to help with the debugging.

### DIFF
--- a/source/dub/internal/vibecompat/inet/path.d
+++ b/source/dub/internal/vibecompat/inet/path.d
@@ -34,7 +34,7 @@ struct Path {
 	/// Constructs a Path object by parsing a path string.
 	this(string pathstr)
 	{
-		m_nodes = cast(immutable)splitPath(pathstr);
+		m_nodes = (splitPath(pathstr).idup);
 		m_absolute = (pathstr.startsWith("/") || m_nodes.length > 0 && (m_nodes[0].toString().countUntil(':')>0 || m_nodes[0] == "\\"));
 		m_endsWithSlash = pathstr.endsWith("/");
 	}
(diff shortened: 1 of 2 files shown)


source/dub/internal/vibecompat/inet/path.d
```diff
@@ -34,7 +34,7 @@ struct Path {
 	/// Constructs a Path object by parsing a path string.
 	this(string pathstr)
 	{
-		m_nodes = cast(immutable)splitPath(pathstr);
+		m_nodes = (splitPath(pathstr).idup);
```
s-ludwig: Is this a necessary part of the fix? Casting should be completely legal here and can in fact be dropped on recent compiler versions. I'd commit a change here to remove the cast for DMD >= 2.067.

source/dub/internal/vibecompat/inet/path.d
```diff
@@ -283,7 +283,7 @@ struct PathEntry {
 
 	string toString() const { return m_name; }
 
-	Path opBinary(string OP)(PathEntry rhs) const if( OP == "~" ) { return Path(cast(immutable)[this, rhs], false); }
+	Path opBinary(string OP)(PathEntry rhs) const if( OP == "~" ) { return Path(([this, rhs]).idup, false); }
```
s-ludwig: Same here, the cast can simply be be dropped for all DMD versions down to 2.064 as it seems.

bittwiddler1: The casting changes aren't part of the fix, but the possibility of undefined behavior convinced me to keep it in my patches.
That being said, I dont think it will be an issue unless another part of the code tries to change the memory. So it should be fine. 
The main issue was the inplace reassignment (and only in certain versions of glibc, I think.)
If you pull,  feel free to change the immutable casts as you want. 
I'd do it myself but I'm indisposed with real life stuff today (moving to a new home) :p

-------- Original message --------

s-ludwig: Okay, thanks! I'll just apply to workaround at the top then. Good luck with your move!

(The immutable stuff is safe, because all input is `immutable`, too. I'll update master to just avoid the casts.)

bittwiddler1:
Thanks for the good wishes. 
The move went very well!

Closing this out for now.

Issue #609, opened 2015-06-28
What if .visualdproj files were generated in the .dub directory?


p0nce:
When typing `dub generate visuald` without `--combined` the next action I usually do is click on the generated file, name.sln.
I find it great that it is generated next to dub.json

But it also comes with 9 other mylib.visualdproj files that I shouldn't click on and they take quite a lot of visual space. It clutters the root directory.

Why not put them in the `.dub\`  directory? Could be handy even with `--combined`.

s-ludwig:
Sounds like a good idea!

p0nce:
Same problems with the .lib target which are build in the working directory, and the .suo files that VS generates on top.

PR #627, opened 2015-07-17
.visualdproj files get generated in the .dub/ directory.

p0nce:
See #609.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 
 matrix:
   include:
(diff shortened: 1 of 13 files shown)


s-ludwig:
LGTM. I'll wait until the final 0.9.24 is tagged, though, so that a separate release branch isn't necessary.

p0nce:
It creates a workflow problem, "current working directory" of generated project should be set to ".." instead of left empty.

p0nce:
Closed because this PR should have been done in a branch.

PR #633, opened 2015-07-26
Fix #616: Infinite new-process recursion using dub decribe within…

### LINKED ISSUE #616: 0.9.24-beta.2: Infinite new-process recursion using dub decribe within preGenerateCommands
```bash
$ git clone https://github.com/Abscissa/safeArg.git
$ cd safeArg
$ dub describe
```

Works on DUB 0.9.23. Hangs on DUB 0.9.24-beta.2. Same thing for ```dub run```.

Abscissa:
…preGenerateCommands.

I used your ```DUB_DESCRIBE=package1,package2,...``` suggestion (although I changed the envar name from ```DUB_DESCRIBE``` to ```DUB_PACKAGES_USED```, which I thought was more appropriate internally as this isn't *directly* tied to ```dub describe``` even though that's where the problem manifests).


Abscissa:
Just wanted to ping this one. If we could get a DUB beta/RC with this in it, it would be a big help for me.

MartinNowak:
Why are you trying to block individual projects?
FWIW dub build each dependency and during that invokes their pre-build step, so this shouldn't be a recursive problem and a simple `environment["DUB_IN_PREBUILD"] = "1"` should suffice.

source/dub/generators/generator.d
```diff
@@ -485,5 +487,12 @@ void runBuildCommands(in string[] commands, in Package pack, in GeneratorSetting
 	env["DUB_PARALLEL_BUILD"]    = settings.parallelBuild? "TRUE" : "";
 
 	env["DUB_RUN_ARGS"] = (cast(string[])settings.runArgs).map!(escapeShellFileName).join(" ");
+	
+	PackagesUsed.add(proj.rootPackage.name);
+	foreach (dep; proj.dependencies)
+		PackagesUsed.add(dep.name);
+	PackagesUsed.store(env);
+	PackagesUsed.load(); // Refresh internal list from actual current environment
```
MartinNowak:
```d
environment["DUB_PACKAGES_USED"] = proj.dependencies.map!(d => chain(d.name, "=", d.path.toNativeString())).join(",");
```

source/dub/generators/generator.d
```diff
@@ -332,23 +332,24 @@ ProjectGenerator createProjectGenerator(string generator_type, Project project)
...
 /**
 	Runs post-build commands and copies required files to the binary directory.
 */
-private void finalizeGeneration(in Package pack, in GeneratorSettings settings, in BuildSettings buildsettings,
-	Path target_path, bool generate_binary)
+private void finalizeGeneration(in Package pack, in Project proj, in GeneratorSettings settings,
+	in BuildSettings buildsettings, Path target_path, bool generate_binary)
 {
-	if (buildsettings.postGenerateCommands.length) {
+	if (buildsettings.postGenerateCommands.length && !PackagesUsed.has(pack.name)) {
```
MartinNowak:
```d
environment.get("DUB_PACKAGES_USED", "").splitter(",").canFind!startsWith(chain(pack.name, "="));
```

Abscissa:
> Why are you trying to block individual projects?

From @s-ludwig in https://github.com/D-Programming-Language/dub/issues/616#issuecomment-120237193
> PS: Another idea: add a DUB_DESCRIBE=true environment variable and use that instead. That solves the read-only issue and is easier to implement, but wouldn't work if "dub describe" is used across multiple packages - DUB_DESCRIBE=package1,package2,... should work though.

source/dub/generators/generator.d
Abscissa: No, I'm deliberately *not* changing the current process's environment. If I did (which I actually did by accident at one point while developing this PR), then dependencies would occasionally get skipped when they shouldn't get skipped.

source/dub/generators/generator.d
Abscissa: The lack of abstraction seems very dirty there. Basic encapsulation principles apply: Any future changes that might be needed to the way the info is stored/passed would be messy and error-prone, plus it's mixing high-level intent with low-level implementation in the same line of code.

s-ludwig:
What I don't like about the `PackagesUsed` class is that it has global state, but is also used to store temporary state when building up the environment of the child process. What about replacing it by two free functions:
```
bool isRecursiveInvocation(string pack)
{
    return environment
        .get("DUB_PACKAGES_USED", "")
        .splitter(",")
        .canFind(pack.name);
}

void storeRecursiveInvokations(Environment env, string[] packs)
{
    env["DUB_PACKAGES_USED"] = environment
        .get("DUB_PACKAGES_USED", "")
        .splitter(",")
        .chain(packs)
        .join(",");
}
```

Those would stay together and provide the same level of encapsulation.

Abscissa:
> What I don't like about the PackagesUsed class is that it has global state, but is also used to store temporary state when building up the environment of the child process.

Yea, you guys are right, that is pretty ugly. Started going one way with it, then realized temporary state was also needed, and couldn't quite shake the original design from my mind.

The ```isRecursiveInvocation/storeRecursiveInvokations``` sounds good, I'll adjust.

Abscissa:
Ok, I've changed it to use @s-ludwig's suggestion.

s-ludwig:
Looks good. Pulling.

PR #644, opened 2015-08-07
Fix #624: --data=libs should output in compiler-, not linker-format.

### LINKED ISSUE #624: describe --data= needs way to get libs in compiler-format, not just linker-format
```dub describe --data=...``` needs a way to retrieve libs in compiler-format instead of linker-format. Currently it works like this:

```bash
$ dub describe --compiler=dmd --data=libs
-levent_pthreads -levent -lssl -lcrypto
```

That's good if you're passing them directly to the linker, but there needs to be a way to obtain it in compiler format:

```
-L-levent_pthreads -L-levent -L-lssl -L-lcrypto
```

I can do a pull request for this, but I wanted to 1. give a heads-up about it, and 2. ask about interface:

IMO, ```--data=libs``` should result in compiler-format (ie, ```-L-letc```) in order to be consistent with all other ```--data=...``` options. And a new option ```--data=linker-libs``` should be added for getting them in linker-format (ie, ```-letc```). Sound good?

Abscissa:
I'm going to leave ```--data=linker-libs``` and ```--data=linker-lflags``` I suggested in #624 up to a separate PR later on.

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -252,6 +252,9 @@ interface Compiler {
 	/// Invokes the underlying linker directly
 	void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback);
 
+	/// Convert linker flags to compiler format
+	string[] lflagsToDFlags(in string[] lflags) const;
+	
 	protected final void invokeTool(string[] args, void delegate(int, string) output_callback)
 	{
 		import std.string;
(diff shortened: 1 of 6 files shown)


s-ludwig:
Can you re-target this against the 0.9.24 branch? There are already some changes on master that shouldn't go into this release.

Hm, `--data=linker-lflags` sounds a bit strange and is inconsistent with the "lflags" field in the package recipe, we should probably turn this around and let  `--data=compiler-lflags` output in compiler format (or whatever the name).

PR #646, opened 2015-08-09
add buildType profile=gc

MartinNowak:
- allows to dub build -b profile=gc to trace memory allocations


s-ludwig:
The only thing I wonder if we should maybe name it "profile-gc" instead of using a an equals sign, which might produce a syntax conflict in certain situations (although it doesn't currently, AFAICS). The enum member name should be `profileGC` to comply with the general naming rules.

MartinNowak:
done

s-ludwig:
Thanks!

Issue #658, opened 2015-08-20
DUB_PACKAGE/DUB_PACKAGE_DIR is wrong for (pre|post)BuildCommands


Abscissa:
Within (pre|post)BuildCommands, the DUB_PACKAGE/DUB_PACKAGE_DIR envvars are set to the root package instead of the current package. But they *do* work correctly within (pre|post)GenerateCommands.

The cause appears to be ```BuildGenerator.buildTarget()```, ```BuildGenerator.performCachedBuild()``` and ```BuildGenerator.performDirectBuild()``` passing the wrong ```Package``` into ```runBuildCommands()```. Still looking into this further.

Abscissa:
Oh, I see what's going on. DUB combines all (pre|post)BuildCommands into one big (pre|post)BuildCommands list (well, one "pre" and one "post") as part of the BuildSettings for the root package. Then it makes *one* call to ```runBuildCommands()```, just for the root package, passing all the build commands for *all* dependencies at once.

I'm not sure what to do about that.

MartinNowak:
This makes sense b/c building many small libraries is just one strategy, another one would be to build everything at once.

Abscissa:
Regardless of why BuildCommands should or shouldn't be batched together, DUB_PACKAGE/DUB_PACKAGE_DIR still need to be correct. Currently, they are often incorrect.

nicolasjinchereau:
I'm trying to use $DUB_PACKAGE in both preGenerateCommands and preBuildCommands, but in both places, I get this error:

> Invalid variable: DUB_PACKAGE

I have this in a library project:

    "preBuildCommands":` [ "(cd src/cpp && make -f build.mk)" ],

When I build this library directly, the above command runs fine. If I use the same project as a dependency, get an error because dub tries to run that command in the directory of the dependant project.

> Running (cd src/cpp && make -f build.mk)
> /bin/sh: line 0: cd: src/cpp: No such file or directory

Paths in dub.json should not change meaning depending on whether the project is compiled directly or as a dependency.

Issue #618, opened 2015-07-04
Incorrect library extension for win64-msvc version of ldc2


mdparker:
dub assumes only dmd[1] uses the lib file extensions on Windows. Trying to compile a project that uses any libraries with the win64-msvc version of ldc2 results in:

> Error: unrecognized file extension a

[1] https://github.com/D-Programming-Language/dub/blob/master/source%2Fdub%2Fcompilers%2Fcompiler.d#L365

s-ludwig:
Is LDC on Windows generally outputting .lib? In that case I'd turn the condition around and only use libX.a for GDC or on non-Windows systems.

aldacron:
It depends on the backend. The release using MinGW puts out .a. I think the only two options to handle this are to either do it automatically by checking the output of ldc2 -version for 'windows-msvc' in the 'Default Target:' line, or to set up a separate configuration for ldc-msvc.

s-ludwig:
I see, so `getTargetFileName` would have to be moved to the `Compiler` class and the `LdcCompiler` class could then parse `ldc2 -version`.

cup:
https://github.com/dlang/dub/issues/835

PR #659, opened 2015-08-20
Fix #384: Switch the working directory when executing build commands.

### LINKED ISSUE #384: Switch the working directory when executing build commands
Currently, all pre/post build/generate commands are executed within the root package directory. The more useful default seems to be the directory of each package, though.

See also http://forum.rejectedsoftware.com/groups/rejectedsoftware.dub/thread/1953/

Abscissa:
Note that, due to #658, this only works for (pre|post)GenerateCommands, not (pre|post)BuildCommands. When #658 is fixed, this PR should automatically work for all four (pre|post)(Build|Generate)Commands.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -443,6 +443,7 @@ private void finalizeGeneration(in Package pack, in GeneratorSettings settings, 
 void runBuildCommands(in string[] commands, in Package pack, in GeneratorSettings settings, in BuildSettings build_settings)
 {
 	import std.conv;
+	import std.file;
 	import std.process;
 	import dub.internal.utils;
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
The only solution to #618 that doesn't require larger restructurings looks like it is to insert updates to the environment vars as commands into the command array. In that case it would make sense to also insert "cd" commands there to change directories. I'm not sure if this can be considered particularly pretty, but it would mean that the command lists stay naturally composable, reducing the risk that such issues occur in other parts of the code (VisualD project generator?).

Geod24:
Superseeded by #1820 (I don't have write access so someone will need to close).

PR #671, opened 2015-09-08
drop 2.064.2

MartinNowak:


s-ludwig:
Should we do this already? DUB generally has pretty low language requirements, So I'd usually have waited with this for a few additional months. After all, there are still some ancient compiler front-ends in use on recent Linux distributions (not sure about the actual versions, though).

MartinNowak:
OK, I'll fix #673 to not use vibe-d with 2.064.2 then.

PR #678, opened 2015-09-12
The default compiler should not depend on DUB's own build environment

schuetzm:
When building DUB with `DMD=ldmd2 ./build.sh`, the resulting `dub` binary uses `ldmd2` as default compiler at runtime. But IMO the behaviour of `dub` at runtime should not depend on the environment it was built in.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -5,7 +5,6 @@
 @for /f %%i in ('git describe') do @set GITVER=%%i
 @echo module dub.version_; > source\dub\version_.d
 @echo enum dubVersion = "%GITVER%"; >> source\dub\version_.d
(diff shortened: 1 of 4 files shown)


s-ludwig:
Since we now have the generic `findCompiler` functionality, I'd agree that this isn't necessary anymore and indeed not particularly desirable from a distribution POV. However, I'd not want to remove this without supporting a "defaultCompiler" field in `m_userConfig`/`m_systemConfig` as a run-time replacement (I think there have been requests for that anyway).

s-ludwig:
BTW, I'll have a look at the user config now.

PR #680, opened 2015-09-14
Get .visualdproj files to be created in the .dub directory

p0nce:
Not ready yet.

This creates a problem since when debugging an executable through VS the default working directory will be .dub/ where it was ./ previously. This can cause eg. files not being found.

This "debugging working directory" setting is stored in the .suo file generated by the IDE, so maybe a VisualD option is needed to get out of this.

### DIFF
--- a/source/dub/generators/visuald.d
+++ b/source/dub/generators/visuald.d
@@ -283,7 +283,7 @@ class VisualDGenerator : ProjectGenerator {
 					output_type = DynamicLib;
 					output_ext = "dll";
 				}
-				auto bin_path = pack == m_project.rootPackage.name ? Path(buildsettings.targetPath) : Path(".dub/lib/");
+				auto bin_path = pack == m_project.rootPackage.name ? Path(buildsettings.targetPath) : Path("lib/");
 				bin_path.endsWithSlash = true;
 				ret.formattedWrite("    <lib>%s</lib>\n", output_type);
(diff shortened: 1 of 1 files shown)


p0nce:
https://issues.dlang.org/show_bug.cgi?id=15058
This would need VisualD support

p0nce:
There is no way to do this and have the right current working directory for debugging in VisualD. A solution would be to generate the main package along with the .sln and hide others in .dub/

s-ludwig:
Thanks for opening a VisualD issue, I should have done that long ago. If the `<debugworkingdir>` solution should turn out to be insufficient, generating only the root package in the root directiory would be a decent workaround for now.

p0nce:
Yeah I would prefer that since it's quite common to generate VisualD project and keep them a while, while modifying some variable or even adding files. `<debugworkingdir>` would be more confusing.

s-ludwig:
Hm, well if the value of `<debugworkingdir>` ends up in the .suo file, it shouldn't matter that VisualD removes it from the project file, should it?

s-ludwig:
Also note #567.

p0nce:
> Hm, well if the value of <debugworkingdir> ends up in the .suo file, it shouldn't matter that VisualD removes it from the project file, should it?

Yes, adn't thought of this. I'll test this soon.

p0nce:
It worked. Now I don't think there is any problem left.

s-ludwig:
Very nice! Will pull and then fix #567.

Issue #647, opened 2015-08-09
-defaultlib=phobos2 flag has no purpose


p0nce:
**DUB version:** my version with the ldc2 fix in #638
**LDC version:** 0.15.2-beta2

**Problem statement**
I have a shared library which only build in OSX with `-b debug`.
Surprisingly it doesn't build with `-b release`or  `-b release-nobounds`.
(yet again https://github.com/p0nce/dplug/blob/master/examples/distort/distort.d with command-line `dub build --build=release --arch=x86 --compiler=ldc2`)

I get the error

```
ld: library not found for -lphobos2
```

**Explanation**
A shared library will be built with LDC with the flags: `-shared -defaultlib=phobos2`

https://github.com/D-Programming-Language/dub/blob/master/source/dub/compilers/ldc.d#L162
Hereby I state that the `-defaultlib=phobos2` should not be appended.
- **Reason 1** The configuration file of LDC already specifies this:

```
default:
{
    // 'switches' holds array of string that are appends to the command line
    // arguments before they are parsed.
    switches = [
        "-I%%ldcbinarypath%%/../import/ldc",
        "-I%%ldcbinarypath%%/../import",
        "-L-L%%ldcbinarypath%%/../lib",
        "-defaultlib=phobos2-ldc,curl,druntime-ldc",
        "-debuglib=phobos2-ldc-debug,curl,druntime-ldc-debug"
    ];
};
```

**\- Reason 2:** the pathes used are wrong. They should be the ones in the LDC config files because `phobos2.a` doesn't exist but `phobos2-ldc.a` and `phobos2-ldc-debug.a` do exist.

**\- Reason 3:** in the current state the flag is only effective when there isn't debug info. It halts because of the incorrect path. And when there is debug info, it does nothing since `-debuglib` would be expected instead of `-defaultlib`.

**\- Reason4**: I've patched DUB to skip this flag and everything works smoothly, since this flag does nothing. Only tested with LDC 0.15.2-b2 but LDC 0.15.1 has the same structure.

tl;dr: this flag looks obsolete and I think DUB shouldn't emit it.

MartinNowak:
Statically linking against phobos when you build a shared library is almost always wrong (and doesn't work b/c phobos isn't compiled with PIC).

p0nce:
Does this mean I have to distribute phobos separately?

Dicebot:
On distros that provide LDC you will almost always find shared LDC phobos as a separate package that can be specified as a dependency.

[...]

PR #681, opened 2015-09-14
Issue 635 bis

p0nce:
I have not merged the commits, what this does is allow the creation of dynlibs in OS X with DMD and LDC, also use -noboundscheck with LDC.

The name of the branch is misleading, this is actually about Issue #647 where there is a discussion about what should ideally be done.

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -175,7 +175,10 @@ class DmdCompiler : Compiler {
 				settings.addDFlags("-lib");
 				break;
 			case TargetType.dynamicLibrary:
-				settings.addDFlags("-shared");
+				version(linux)
+					settings.addDFlags("-shared", "-defaultlib=libphobos2.so");
+				else
+					settings.addDFlags("-shared");
 				break;
 			case TargetType.object:
 				settings.addDFlags("-c");
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -175,8 +175,7 @@ class DmdCompiler : Compiler {
 				settings.addDFlags("-lib");
 				break;
 			case TargetType.dynamicLibrary:
-				version (Windows) settings.addDFlags("-shared");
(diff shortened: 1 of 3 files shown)


source/dub/compilers/dmd.d
```diff
@@ -172,8 +172,10 @@ class DmdCompiler : Compiler {
 				settings.addDFlags("-lib");
 				break;
 			case TargetType.dynamicLibrary:
-				version (Windows) settings.addDFlags("-shared");
-				else settings.addDFlags("-shared", "-defaultlib=libphobos2.so");
+				version(linux)
+					settings.addDFlags("-shared", "-defaultlib=libphobos2.so");
+				else
+					settings.addDFlags("-shared");
```
MartinNowak: You're conflating 2 different things here, building an isolated shared D library that is statically linked against phobos, and building a shared D library that is linked against a shared phobos.
I think we need something along the line of https://github.com/MartinNowak/dub/commit/bfcc6ab3b2cb3bacb1bd090d4cdc5459b865b71b instead.

p0nce:
@MartinNowak Sure you are right. I was more reacting to the current situation where you can't have shared libs on Mac.
I'm more than OK for a better solution, I can use the branch meanwhile.

source/dub/compilers/dmd.d
John-Colvin: @MartinNowak How close is that branch of yours to a production solution?

s-ludwig:
I think the current diff is fine as a quick fix (it just specifies `-defaultlib` for Linux instead of for all Posix systems). Just the `addDFlagsNoDuplicates` changes should be removed and the rest squashed together into a single commit. The duplicate flags issue should have already been solved by #759.

p0nce:
Yes, I think about it regularly, the only thing I miss from Dthe current UB is that `-defaultlib` flag thing
This is somehow important, I will be able to make a tutorial on plugin-making without advertising an outdated private branch whenever OS X or LDC is used.
I'll work on it immediately.

p0nce:
This PR is now completely obsolete.

PR #684, opened 2015-09-15
enable merging of coverage reports to account for tests

MartinNowak:


s-ludwig:
BTW, not sure what is going on, but the coverage generation process seems to have gotten stuck at the same day as this PR (although it indicates some builds after the merge on the same day): https://coveralls.io/github/D-Programming-Language/dub?page=4

MartinNowak:
Yep, coveralls couldn't yet handle different coverage dirs.
Should work now.
https://github.com/ColdenCullen/doveralls/pull/11#issuecomment-142825349

PR #685, opened 2015-09-15
Add new 'registry' command that just open http://code.dlang.org

p0nce:
A way to solve #663

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -67,6 +67,7 @@ CommandGroup[] getCommands()
 			new RemoveOverrideCommand,
 			new ListOverridesCommand,
 			new CleanCachesCommand,
+			new RegistryCommand
 		)
 	];
 }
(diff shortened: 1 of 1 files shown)


MartinNowak:
I think the command line tool should provide the expected `dub search` functionality.
Adding a dedicated command just to browse an URL isn't justified IMO.

p0nce:
How would a dub search command be better than opening an URL?

s-ludwig:
Probably mostly for command-line fanatics, but since the core functionality is already there, it's an easy target. An advantage apart from working without a GUI is that the search output can be processed by a shell script to do additional things with it. Of course using the website has a lot advantages, too, and something like `dub search -web` could be an alternative if there is demand.

p0nce:
Well, I use dub day in day out. Opening `http://code.dlang.org` is a common task.
`dub search -web` would be ok of course.

MartinNowak:
I added the `dub search` command, so it would be easy now to make `dub search -web foobar` open http://code.dlang.org/search?q=foobar.

PR #688, opened 2015-09-18
Fix library extensions for LDC when using the MS linker. Fixes #618.

### LINKED ISSUE #618: Incorrect library extension for win64-msvc version of ldc2

s-ludwig:
Fixes the output binary extensions for the MSVC version of LDC. This requires runtime detection of the type of compiler invoked and thus requires moving the target filename handling code to the compiler class.

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -104,6 +104,9 @@ interface Compiler {
 	/// Removes any dflags that match one of the BuildOptions values and populates the BuildSettings.options field.
 	void extractBuildOptions(ref BuildSettings settings) const;
 
+	/// Computes the full file name of the generated binary.
+	string getTargetFileName(in BuildSettings settings, in BuildPlatform platform) const;
+
 	/// Adds the appropriate flag to set a target path
 	void setTarget(ref BuildSettings settings, in BuildPlatform platform, string targetPath = null) const;
 
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -166,6 +166,33 @@ class DMDCompiler : Compiler {
 		settings.dflags = newflags.data;
 	}
 
+	string getTargetFileName(in BuildSettings settings, in BuildPlatform platform)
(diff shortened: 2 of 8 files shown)


s-ludwig:
This affects the public API and thus should be merged prior to 1.0.0.

s-ludwig:
Should work now, but should be rebased on top of #785 before merge, because that makes some of the API changes in this PR obsolete.

s-ludwig:
Writing a specific test case is currently not possible, because this only affects the Windows versions of LDC.

p0nce:
You can assume the MS linker is always used nowadays.

s-ludwig:
So it's not possible to build LDC with cygwin anymore?

p0nce:
You probably meant Cygwin? The last LDC release which built programs with the MinGW linker was 0.16.0 IIRC. This version has been bested by every LDC release since then on Windows, to the point I doubt anyone uses it anymore for this platform.

s-ludwig:
("meant Cygwin?" Did you mean MinGW? :P - but you are right, I didn't even know which one of the two was used) But independently of LDC on Windows, I think it probably is a good idea to move the logic to the compiler class anyway. It will become important again once cross compilation comes into play.

p0nce:
> "meant Cygwin?" Did you mean MinGW?

Yes :-)

s-ludwig:
Anyone else with opinions on this change set?

@p0nce: Do you see any issues with pulling it like this? I'd like to keep the generally added flexibility and keeping support for a MinGW based LDC, even if rarely used, seems to be not the worst idea.

p0nce:
I see no issues in this. Was just signalling that MinGW is a minority case.

PR #692, opened 2015-09-21
Implement a commit hash based cache of local package versions.

s-ludwig:
To speed up the initialization process, this PR adds a file `.dub/version.json` that contains the last known pair of commit hash and version of a particular local GIT working copy. The current commit hash is then determined manually by reading the appropriate files in .git .

Open issues:
 - Can yield wrong results if tags to existing commits get added/changed/removed after the version has been cached
 - Currently doesn't read the "packed-refs" file, so mail fail on some repositories (which just means that it won't be able to hit the fast path)

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -93,6 +93,22 @@ void writeJsonFile(Path path, Json json)
 	f.writePrettyJsonString(json);
 }
 
+/// Performs a write->delete->rename sequence to atomically "overwrite" the destination file
+void atomicWriteJsonFile(Path path, Json json)
+{
+	import std.random : uniform;
+	auto tmppath = path[0 .. $-1] ~ format("%s.%s.tmp", path.head, uniform(0, int.max));
+	auto f = openFile(tmppath, FileMode.createTrunc);
+	scope (failure) {
+		f.close();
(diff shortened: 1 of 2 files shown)


MartinNowak:
Please always make sure to synchronize disk caches.
https://github.com/D-Programming-Language/dub/pull/683

MartinNowak:
The git command usually runs in 1ms for me, sure we need a cache?

s-ludwig:
On Windows it can take considerable time due to pure process startup. Might make sense to restrict the optimization to Windows. I'll change the storage procedure to "atomically" perform save->delete->rename.

s-ludwig:
A best-of-10 benchmark of `dub list` with 35 local working copies registered yields 275 ms for the cached version and 1344 ms for the uncached one (Windows). I'll do a test on Linux later, but for Windows it's definitely worth it.

s-ludwig:
Linux, for 17 working copies, yields 39 ms for the caching version and 89 ms for the uncached version (again best of 10). So it might be an improvement for people with large numbers of local working copies, but I'd be fine with restricting it to Windows-only.

MartinNowak:
I also used atomicWriteJsonFile for #683.

PR #693, opened 2015-09-21
LDC 0.16.0 removed the -noboundschecks flag, replace with -boundscheck=off

p0nce:
This flag is already existing in 0.15.x so it breaks nothing.

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -33,7 +33,7 @@ class LdcCompiler : Compiler {
 		//tuple(BuildOption.alwaysStackFrame, ["-?"]),
 		//tuple(BuildOption.stackStomping, ["-?"]),
 		tuple(BuildOption.inline, ["-enable-inlining"]),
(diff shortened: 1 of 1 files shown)


s-ludwig:
Did they remove it immediately, or has there been a transition phase? If not, we might need to make this dependent on the compiler version.

p0nce:
@s-ludwig I think 0.15.x was the transition phase, since this version has both flags.
I'm not aware for previous LDC versions (0.14.x).

s-ludwig:
Okay, I guess that's acceptable then.

SFrijters:
Is there any chance of getting a new dub release soon including this fix? Currently Travis CI doesn't work out of the box with 0.16.x LDC because it uses the latest release (0.9.24) which uses the old flag.

Geod24:
We might want to support it in DMD as well, as this switch appeared in 2.066: http://dlang.org/changelog/2.066.0.html

PR #695, opened 2015-09-25
use stable coveralls version

MartinNowak:


MartinNowak:
I reenabled coverage as PR test now that the merging works.

MartinNowak:
We don't get coverage for the gdc/ldc code b/c it's in a separate build btw.

PR #702, opened 2015-09-28
allow to specify ddox tool

MartinNowak:
- can be used to customize the ddox tool (and it's output)
- assumes similar commands/behavior


MartinNowak:
I'm open to other ways for configuring this as well.

MartinNowak:
Ping

s-ludwig:
I was wondering if it wouldn't make sense to use this occasion and switch from special handling of the two documentation build types to a separate "dub doc(s)" command. That could avoid additional special cases, such as those internal `BuildOption` members (or the hard-coded dflags of the previous implementation for that matter).

Anyway, the functionality makes sense, as well as a potential `"-ddoxGenerateArgs"` field.

MartinNowak:
> I was wondering if it wouldn't make sense to use this occasion and switch from special handling of the two documentation build types to a separate "dub doc(s)" command.

It does make sense, but can we separate that and use this PR as intermediate step?
Writing a doc command that handles ddoc, ddox, and other builds is quite some work.

MartinNowak:
> It does make sense, but can we separate that and use this PR as intermediate step?

What do you think? Go with the immediate improvement now, and do more later?

s-ludwig:
Should be alright. When I get some time I'll look into adding a really simple dedicated command (that just mirrors the current functionality) for the next release. We can extend it later then.

PR #704, opened 2015-09-28
Sets MACOSX_DEPLOYMENT_TARGET variable

p0nce:
to avoid trivial incompatibility with OSX version earlier than the one used to build DUB

It avoids the `Illegal instruction` error.

I've tested on OSX 10.6.8 and the next error is:

```
dyld: Library not loaded /usr/lib/libcurl.4.dylib
Referenced from: /path/to/dub
Reason: Incompatible library version: dub requires version 7.0.0 or later, but libcurl.4.dylib provides version 6.0.0
Trace/BPT trap
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -39,7 +39,7 @@ echo "module dub.version_;" > source/dub/version_.d
 echo "enum dubVersion = \"$GITVER\";" >> source/dub/version_.d
 
 # For OSX compatibility >= 10.6
-export MACOSX_DEPLOYMENT_TARGET=10.6
+MACOSX_DEPLOYMENT_TARGET=10.6
 
 echo Running $DMD...
 $DMD -ofbin/dub -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
--- a/build.sh
+++ b/build.sh
@@ -38,6 +38,8 @@ GITVER=$(git describe) || GITVER=unknown
 echo "module dub.version_;" > source/dub/version_.d
 echo "enum dubVersion = \"$GITVER\";" >> source/dub/version_.d
 
+# For OSX compatibility >= 10.6
+export MACOSX_DEPLOYMENT_TARGET=10.6
 
 echo Running $DMD...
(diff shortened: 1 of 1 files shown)


s-ludwig:
I think that library incompatibility might be fixed with DMD 2.069, because libcurl is loaded dynamically there. But since this is a Phobos dependency, there is nothing that can be done from DUB's side.

BTW, any reason why an explicit `export` is needed?

p0nce:
> BTW, any reason why an explicit export is needed? 

I've checked and indeed it isn't needed.

> But since this is a Phobos dependency, there is nothing that can be done from DUB's side.

Let's just wait DMD 2.069 then, nothing urgent. One can build on another machine.

s-ludwig:
Okay

jacob-carlborg:
Next version of DMD and the current version of LDC have a minimum requirement of OS X 10.7.

p0nce:
I target OS X 10.7+ for this reason, and don't need a DUB running on OS X 10.6 anymore. Do you want another PR?

s-ludwig:
Will/does setting it to 10.6 cause build issues with those compilers?

p0nce:
If phobos is built with 10.7 as minimum OS version then it could create problems if we don't do the same. Though I don't really know what could happen.

jacob-carlborg:
> Will/does setting it to 10.6 cause build issues with those compilers?

@s-ludwig yes:

```
$ cat main.d
int a = 3; // TLS
void main() {}

$ ./dmd main.d -L-macosx_version_min -L10.6 -L-lcrt1.10.6.o
ld: targeted OS version does not support use of thread local variables in _D4main1ai for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
--- errorlevel 1
```

Targeting 10.7 works perfectly fine: `./dmd main.d -L-macosx_version_min -L10.7 -L-lcrt1.o`

s-ludwig:
Okay, incremented to 10.7 now.

Issue #502, opened 2015-02-02
Protecting from import leak by fetching package one directory deeper


p0nce:
There were complaints in the NG that DUB does'nt support libraries whose source files are at the top-level directory.

Reference: http://forum.dlang.org/post/ysnfzpvrcvmjdekulfnq@forum.dlang.org
```
- When cloning repositories, dub does not preserve the 
repository's directory name (so e.g. fruit will be cloned to 
~/.dub/fruit-1.0.0/).

Somebody has created a Dub package for my library (excluding 
certain packages, due to point 1a above), and the other day 
someone complained that it doesn't work with Dscanner, because of 
the above issue - the module path "ae.dir.module" does not 
correspond to the filesystem path "ae-1.0.1/dir/module.d".
```

I think this is mostly solvable by fetching dependencies one directory deeper :

Reference: http://forum.dlang.org/post/ofznlktcmxwnpicjbfhv@forum.dlang.org
```
Currently, using ae checkouted in an ae/ directory means having 
an "import leak" meaning that you have to import a path outside 
the checkouted repositery. That's somehow unhygienic like, er, a 
global namespace. But yeah DUB could isolate checkouts one 
directory deeper (ae-1.0.0/ae/) to make this a non-problem, this 
would support libraries like ae/ and make DScanner work.
```

What do you think?

p0nce:
This prevent to use the ae library from DUB.
ae is a very useful library with the best command-line parser, best image library in the ecosystem, a HTTP server that supports DFeed, etc...

s-ludwig:
I think this is a viable workaround. The only thing that should ideally be made sure is that old directory hierarchies get upgraded to the new scheme when the new DUB version is first used.

PR #712, opened 2015-10-25
Fetch one folder deeper

p0nce:
This solves #502.

Packages now gets unzipped one directory deeper (eg: `~/.dub/packages/vibe-d-0.7.8/vibe-d/` instead of `/.dub/packages/vibe-d-0.7.8/`).

For scanning, here is how it works now:
- `vibe-d-0.7.8` directory is found by scanning
- (NEW) looking for a mandatory subfolder that is a prefix of `vibe-d-0.7.8`
- and which contains a package description

Once upgrading to the new directory structure, previously listed packages won't be found anymore.
New-style packages get installed over the older ones.
In my tests it works regardless.

What doesn't work is downgrading DUB from the new folder scheme to the older. If downgrading DUB, deleting packages manually is required.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -581,6 +581,13 @@ class Dub {
 		if (!placement.existsFile())
 			mkdirRecurse(placement.toNativeString());
 		Path dstpath = placement ~ (packageId ~ "-" ~ clean_package_version);
+		if (!dstpath.existsFile())
+			mkdirRecurse(dstpath.toNativeString());
+
+		// Support libraries typically used with git submodules like ae.
+		// Such libraries need to have ".." as import path but this can create
+		// import path leakage.
+		dstpath = dstpath ~ packageId;
 
 		auto lock = lockFile(dstpath.toNativeString() ~ ".lock", 30.seconds); // possibly wait for other dub instance
 		if (dstpath.existsFile())
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -533,7 +533,16 @@ class PackageManager {
 				try foreach( pdir; iterateDirectory(path) ){
 					logDebug("iterating dir %s entry %s", path.toNativeString(), pdir.name);
 					if( !pdir.isDirectory ) continue;
-					auto pack_path = path ~ (pdir.name ~ "/");
+
+					// Search for a single directory within this directory
+					FileInfo subdir;
+					foreach( pdir2; iterateDirectory(path ~ (pdir.name ~ "/")) )
+					{
+						subdir = pdir2;
+					}
+					if( !subdir.isDirectory ) continue;
+
+					auto pack_path = path ~ (pdir.name ~ "/") ~ (subdir.name ~ "/");
 					auto packageFile = Package.findPackageFile(pack_path);
 					if (packageFile.empty) continue;
 					Package p;
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -534,11 +534,13 @@ class PackageManager {
 					logDebug("iterating dir %s entry %s", path.toNativeString(), pdir.name);
 					if( !pdir.isDirectory ) continue;
 
-					// Search for a single directory within this directory
+					// Search for a single directory within this directory which happen to be a prefix of pdir
+					// This is to support new folder structure installed over the ancient one.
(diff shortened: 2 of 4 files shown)


MartinNowak:
Could you add a test for the upgrading?
And does this fully solve the git submodule vs. dub issue?

p0nce:
> Could you add a test for the upgrading?
Sorry not willing to do that. The upgrade is a one-time thing, and it worked in my local tests.

> And does this fully solve the git submodule vs. dub issue?
It solves that is a library like ae gets in a directory named ae/, which it expects.
Is there anything else to it? I don't think so.

MartinNowak:
> Sorry not willing to do that.

Fine, but we need a test case for the root import.
Should be straightforward to adapt one of the existing test cases.
https://github.com/D-Programming-Language/dub/tree/master/test

p0nce:
OK, let's try.
I've created https://github.com/p0nce/gitcompatibledubpackage because there was no package to fetch that were:
- used as git submodule
- and actually have a working `dub.json`

p0nce:
Added a test. It relies on a DUB registry package like the test using the `bloom` package.

source/dub/dub.d
```diff
@@ -581,6 +581,13 @@ class Dub {
 		if (!placement.existsFile())
 			mkdirRecurse(placement.toNativeString());
 		Path dstpath = placement ~ (packageId ~ "-" ~ clean_package_version);
+		if (!dstpath.existsFile())
+			mkdirRecurse(dstpath.toNativeString());
+
+		// Support libraries typically used with git submodules like ae.
+		// Such libraries need to have ".." as import path but this can create
```
MartinNowak: Well, after this fix they no longer need `..`, so this comment is misleading.

MartinNowak: Wait, they do require `..` bug there is no more leakage.

source/dub/packagemanager.d
```diff
@@ -533,7 +533,18 @@ class PackageManager {
...
+
+					// Search for a single directory within this directory which happen to be a prefix of pdir
+					// This is to support new folder structure installed over the ancient one.
+					FileInfo subdir;
+					foreach( pdir2; iterateDirectory(path ~ (pdir.name ~ "/")) )
+					{
+						if (pdir.name.startsWith(pdir2.name)) // eg: package vibe-d will be in "vibe-d-x.y.z/vibe-d"
+							subdir = pdir2;
+					}
+					if( !subdir.isDirectory ) continue;
```
MartinNowak: I would rather just upgrade the existing structure once and directly load subpaths here than supporting 2 different layouts.
You can place a `.new_package_layout` dummy in the dub folder to mark it as converted.

source/dub/dub.d
p0nce: Yep. Actually the former behaviour would work in *most* cases.

source/dub/packagemanager.d
p0nce: Actually the former layout isn't supported at all anymore. 

It won't get list by the scan, so the new layout will be installed on top of it (which is messy, but works in almost all cases: the problem case being when there was already a subdir called package-name with dub.json itself. Which I think is incredibly rare).

p0nce:
Something blocking this pull?

MartinNowak:
> Something blocking this pull?

Yes, we're lacking time, it doesn't yet upgrade existing repos, and apparently breaks our testsuite.

CyberShadow:
BTW, it's possible that with Git's recently improved symlink support on Windows, this might be fixable from `ae`'s side. I'd need to research the details though.

s-ludwig:
> apparently breaks our testsuite.

These are the failing scripts:

```
[INFO] Running /home/travis/build/D-Programming-Language/dub/test/issue564-invalid-upgrade-dependency.sh...
Failed to find a package named 'main'.
Script failure.
[INFO] Running /home/travis/build/D-Programming-Language/dub/test/issue586-subpack-dep.sh...
Failed to find a package named 'main'.
Script failure.
```

p0nce:
I lack the time to finish this.

s-ludwig:
Okay, thanks for the efforts so far anyway! Maybe I can manage to find some more time today to take this over. I'll close this PR then for now.

PR #715, opened 2015-11-02
dub init could use user details to get the full name

MartinNowak:
Instead of using the username to initialize author and copyright, we could use the user details which are often the full name of a user.

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -121,10 +121,7 @@ private void writePackageJSON(Path root_path, string description, string[string]
 
 	assert(!root_path.empty);
 
-	string username;
-	version (Windows) username = environment.get("USERNAME", "Peter Parker");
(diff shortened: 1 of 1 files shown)


MartinNowak:
Implemented

s-ludwig:
Looks good. Need to remember to adjust #734 to use `getUserName`.

s-ludwig:
I'll also have a look at [`GetUserNameEx`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724435(v=vs.85).aspx) to make this work on Windows, too.

s-ludwig:
Apparently `GetUserNameEx` works only if the user is in a domain. `NetQueryDisplayInformation` or `NetGetUserInfo` should work, though. But due to DMD missing all required .libs and function declarations, this is getting too much work for me ATM.

PR #723, opened 2015-11-18
[Recipe] Added error messages for SDL malformed dependency name.

lablanu:
Before, 
`dependency version="~>1.2.3"`
caused a range violation.
Now, it says that the dependency name is missing.

The fact that dependencies cannot have multiple names is expressed, as well.

Should there be a correct usage notice here?

P.S. I get the `Failed to parse package description` message twice on either error, how to fix that?


s-ludwig:
> I get the Failed to parse package description message twice on either error, how to fix that?

That error message is in an inappropriate place. I'll turn it into a `logDiagnostic` message, as it looks like it doesn't add any value.

PR #726, opened 2015-11-26
fix dmd/ldmd2 detection in build.sh

aG0aep6G:
Supposed to work with dlang.org's [peculiar way of setting DMD/DC](https://github.com/D-Programming-Language/dlang.org/blob/35bd01726086d434db470fe916e3122e90a2f257/posix.mak#L488).


s-ludwig:
Ah, that's why I got an unexpected build failure some time ago (I didn't investigate further, unfortunately). Thanks!

PR #727, opened 2015-11-28
update vibe-d for 2.069.x compat and test latest compilers

MartinNowak:


travis-ci.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e -o pipefail
 
-if [ -z "$FRONTEND" -o "$FRONTEND" \> 2.064.2 ]; then
+if [ -z "$FRONTEND" -o "$FRONTEND" \> 2.065.z ]; then
     dub fetch vibe-d --version=0.7.24 # get optional dependency
```
s-ludwig: Should also be updated to `--version=0.7.26`.

MartinNowak: If only dub describe was useful enough to give me that value.
https://github.com/D-Programming-Language/dub/pull/740

PR #732, opened 2015-12-10
"dub convert" command to convert between dub.json and dub.sdl. Fixes #689.

### LINKED ISSUE #689: SDL documentation and auto conversion from json
Hi Sonke.

SDL is a big improvement.  I love the conciseness and ability to add comments.  The documentation could be a bit more complete, with some more examples.  And auto conversion from JSON would be nice too.

Laeeth

s-ludwig:


blm768:
It might not be a bad idea to explicitly warn the user that the old config file will be deleted. It's not a major issue if people are using source control properly, but it seems dicey to rely on that.

s-ludwig:
It has a warning on the `dub convert --help` screen: https://github.com/D-Programming-Language/dub/pull/732/files#diff-e325b56891c9001a2e2cb1e3bc0a0a36R1665
I'd say that should usually be sufficient, since no actual data gets lost in the conversion process. Thoughts?

blm768:
Probably. I'm just thinking of the few users who might get tripped up, but with bidirectional conversion, it shouldn't be an issue unless there's a bug in the converter. Does the converter at least ensure that the new file is written before the old one is deleted?

s-ludwig:
Yes, the new file is written before deleting the old one (any error will skip the deletion).

MartinNowak:
Coverage looks good https://coveralls.io/builds/4673014/source?filename=source%2Fdub%2Frecipe%2Fsdl.d.

source/dub/recipe/sdl.d
```diff
@@ -174,6 +197,75 @@ private void parseConfiguration(Tag t, ref ConfigurationInfo ret, string package
...
+	foreach (suffix, arr; bs.versions) adda("versions", suffix, arr);
+	foreach (suffix, arr; bs.debugVersions) adda("debugVersions", suffix, arr);
+	foreach (suffix, arr; bs.importPaths) adda("importPaths", suffix, arr);
+	foreach (suffix, arr; bs.stringImportPaths) adda("stringImportPaths", suffix, arr);
+	foreach (suffix, arr; bs.preGenerateCommands) adda("preGenerateCommands", suffix, arr);
+	foreach (suffix, arr; bs.postGenerateCommands) adda("postGenerateCommands", suffix, arr);
+	foreach (suffix, arr; bs.preBuildCommands) adda("preBuildCommands", suffix, arr);
+	foreach (suffix, arr; bs.postBuildCommands) adda("postBuildCommands", suffix, arr);
+	foreach (suffix, bits; bs.buildRequirements) adda("buildRequirements", suffix, toNameArray!BuildRequirement(bits));
+	foreach (suffix, bits; bs.buildOptions) adda("buildOptions", suffix, toNameArray!BuildOption(bits));
```
MartinNowak: Looks like a 1-to-1 mapping, so maybe use generic code and convert all members of bs? You could blacklist members that shouldn't be converted, but doing this makes it impossible to forget something and will be easier to maintain.

s-ludwig: I'd like to postpone that to a separate PR - and then also for de-serialization and for JSON.

s-ludwig:
Fixed and rebased.

PR #733, opened 2015-12-11
Implement new optional dependency semantics. Fixes #361.

### LINKED ISSUE #361: Change semantics of optional dependencies
Currently, an optional dependency is satisfied once it is available locally. This doesn't allow the user to decide when to use an optional dependency on a per project basis, but rather will always be a per machine decision. The proposed new semantics are to satisfy an optional dependency iff:

 - another package in the dependency graph has a non-optional dependency to the same package
 - the optional package is explicitly listed in the `dub.selections.json` file

For the second option, there will also need to be a way to achieve this using the CLI, maybe simply by doing an explicit `dub upgrade <name of optional dependency>` within the project directory.

s-ludwig:
Three kinds of dependencies now:
- "required" - this is the default
- "optional", use `optional=true` - These will only be used if explicitly added to "dub.selections.json"
- "optional, default", use `optional=true default=true` - If no dub.selections.json exists, they will be selected if possible, otherwise behaves like "optional"

The local availability of a package doesn't play a role anymore.

In a later PR, a CLI functionality to manage optional dependencies should be added.


source/dub/dependency.d
```diff
@@ -460,11 +486,28 @@ unittest {
...
 	assertThrown(a.version_);
+	assert(a.matches(Version.MASTER));
+	assert(a.matches(Version("1.0.0")));
+	assert(a.matches(Version("0.0.1-pre")));
+	b = Dependency(">=1.0.1");
+	assert(b == a.merge(b));
+	assert(b == b.merge(a));
+	b = Dependency(Version.MASTER);
+	assert(a.merge(m) == m);
+	assert(m.merge(a) == m);
```
skoppe: Huh `m`? Should not be `b`?

skoppe: Same few lines down...

s-ludwig: The idea is that `merge` is commutative. And since `a` is the special `ANY` ("*") constant, the result should always equal the other operand. Probably this should be documented somewhere...

skoppe: I get the commutative thing. But I saw `b = Dependency(Version.MASTER);` on line 495, so I suspected that `b` meant to be used in the asserts.

s-ludwig: Oohh, now I see it (misread `b` for `a` in your first comment). `b` by accident happens to be the same as `m` here, but indeed `b` was intended. Will fix.

Abscissa:
This pull is important and @s-ludwig knows what he's doing, he did write dub after all.

I say we just merge this already and let any potential issues work their way out in the beta process.

John-Colvin:
Yeah, to be honest I can't see any obvious bugs in it, but I just don't have the time to get the context necessary to work out how it *really* works.

Abscissa:
Yea, I strongly suspect the holdup with this getting reviewers is a lack of people already familiar with this part of DUB. FWIW, I didn't notice any obvious issues when I browsed through it either, but like you, I'm also not that familiar with how this part of DUB works (and don't have time...and I didn't understand it when I did try some months back).

Kinda drawing on a recent discussion on D's NG here, I say: Sonke, the DUB guy himself, went to the trouble of making this PR. I think it's more reasonable to assume it's merge-worthy that not. If there are any glaring issues, they'll surface during beta, and then they can get fixed. So we can either continue waiting indefinitely for a more confident reviewer who may or may not show up, or we can just break the deadlock and get things rolling.

s-ludwig:
Okay, I think it should be pretty safe as the "critical" changes are only concerning optional dependencies, which have been more or less unusable anyway. I've read through it again, too, and didn't see any issues either. There is also the necessary documentation addition for the new "default" attribute: D-Programming-Language/dub-registry#134

PR #734, opened 2015-12-11
Implement interactive "dub init"

s-ludwig:
This adds an opt-out interactive mode to "dub init". It allows to customize the standard fields and to add any number of additional dependencies. Depends on #732.


MartinNowak:
Looks pretty good, please add a test case though.
I think this should work.
```
echo 'entry1\nentry2\n\nentry4' | dub init
```

s-ludwig:
Fixed and rebased.

MartinNowak:
Great, now that we can write recipes, we could add a `dub add-dependency mydep` command, similar to `npm install mydep --save`.

PR #735, opened 2015-12-14
Fetch packages into a folder with their name. Fixes #502.

### LINKED ISSUE #502: Protecting from import leak by fetching package one directory deeper

s-ludwig:
This changes the behavior of #712 to be backwards compatible with the old single-layer package structure. It also fixes searching within non-managed paths ("dub add-path") to *not* expect the new layout.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -203,9 +203,19 @@ class PackageManager {
 	bool isManagedPackage(Package pack)
 	const {
 		auto ppath = pack.basePackage.path;
+		return isManagedPath(ppath);
+	}
+
+	/** Determines if a specifc path is within a DUB managed package folder.
+
+		By default, managed folders are "~/.dub/packages" and
+		"/var/lib/dub/packages".
+	*/
+	bool isManagedPath(Path path)
+	const {
 		foreach (rep; m_repositories) {
 			auto rpath = rep.packagePath;
-			if (ppath.startsWith(rpath))
+			if (path.startsWith(rpath))
 				return true;
 		}
 		return false;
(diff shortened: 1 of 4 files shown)


skoppe:
Looks good to me.

s-ludwig:
Thanks! Practical testing didn't show any issues so far either, will merge.

MartinNowak:
Great, that this is merged now. Would be nice if some of the git submodule people could test this and write a small how-to @Cybershadow, @Hackerpilot.

source/dub/packagemanager.d
```diff
@@ -203,9 +203,19 @@ class PackageManager {
 	bool isManagedPackage(Package pack)
 	const {
 		auto ppath = pack.basePackage.path;
+		return isManagedPath(ppath);
+	}
+
+	/** Determines if a specifc path is within a DUB managed package folder.
+
+		By default, managed folders are "~/.dub/packages" and
+		"/var/lib/dub/packages".
```
MartinNowak: Nice, didn't knew there was a system-wide installation path.
This is useful for caching common dependencies, why is it deprecated?

skoppe: Huh, what gives you the idea it's deprecated?

PR #737, opened 2015-12-25
CMake generator: add a means to force inclusion of the main source file

Yoplitein:
Useful for libraries which specify such an attribute, e.g. vibe-d.

### DIFF
--- a/source/dub/generators/cmake.d
+++ b/source/dub/generators/cmake.d
@@ -86,12 +86,22 @@ class CMakeGenerator: ProjectGenerator
             
             if(addTarget)
             {
-                script.put("add_%s(%s %s\n".format(targetType, name, libType));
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'm not the most qualified reviewer for CMake stuff (I've used it, but only for C++ projects), but the changes themselves look good AFAICT. What I don't understand yet is in which situation this is useful or necessary. Usually, the main source file should be added automatically to the right target, like it is for all other generator types (https://github.com/D-Programming-Language/dub/blob/ab27aa2c246dd35bf57bdcb0d69e85ac902e3b63/source/dub/recipe/packagerecipe.d#L171).

Yoplitein:
After review, I realized that the main source file is added to the executable target. D'oh!

So, this is not necessary at all.

PR #738, opened 2015-12-25
CMake generator: always use forward slashes in paths

Yoplitein:
Fixes CMake choking on certain paths on Windows. Depends on #737.


s-ludwig:
It would be good if this change would be restricted to Windows, as currently, I think, backslashes as part of  file/folder names are supported on Posix systems. Backslashes as path separators on the other hand should only ever occur on Windows anyway.

Yoplitein:
Pruned #737 and added `version(Windows)` to sanitizeSlashes.

s-ludwig:
Thanks! Merging.

Issue #686, opened 2015-09-16
Still getting multiple -march with LDC


p0nce:
**OS** Mac OS X

**Project to reproduce**  https://github.com/p0nce/dplug/blob/master/examples/distort/dub.json

**DUB version** master @ 548dc1b0caf2be412b2d66a0117060c8a4d970df

**Problem**
`$ dub build --compiler=ldc2` => OK
```
Performing "debug" build using ldc2 for x86_64.
gfm:core 3.0.4: target for configuration "library" is up to date.
dplug:core 1.1.28: target for configuration "library" is up to date.
gfm:math 3.0.4: target for configuration "library" is up to date.
dplug:dsp 1.1.28: target for configuration "library" is up to date.
ae-graphics 0.0.7: target for configuration "library" is up to date.
dplug:plugin 1.1.28: target for configuration "library" is up to date.
derelict-util 2.0.3: target for configuration "library" is up to date.
derelict-cocoa 0.0.25: target for configuration "library" is up to date.
x11 1.0.8: target for configuration "library" is up to date.
dplug:window 1.1.28: target for configuration "library" is up to date.
imageformats 4.2.2: target for configuration "library" is up to date.
gfm:image 3.0.4: target for configuration "library" is up to date.
dplug:gui 1.1.28: target for configuration "library" is up to date.
dplug:vst 1.1.28: target for configuration "library" is up to date.
distort ~master: target for configuration "library" is up to date.
To force a rebuild of up-to-date targets, run again with --force.
```

`$ dub build --compiler=ldc2` -a x86 => FAIL
```
Performing "debug" build using ldc2 for x86.
gfm:core 3.0.4: target for configuration "library" is up to date.
dplug:core 1.1.28: target for configuration "library" is up to date.
gfm:math 3.0.4: target for configuration "library" is up to date.
dplug:dsp 1.1.28: target for configuration "library" is up to date.
ae-graphics 0.0.7: target for configuration "library" is up to date.
dplug:plugin 1.1.28: building configuration "library"...
ldc2: for the -march option: may only occur zero or one times!
ldc2 failed with exit code 1.
```

`$ dub build --compiler=ldc2` -a x86_64 => FAIL
```
Performing "debug" build using ldc2 for x86_64.
gfm:core 3.0.4: target for configuration "library" is up to date.
dplug:core 1.1.28: target for configuration "library" is up to date.
gfm:math 3.0.4: target for configuration "library" is up to date.
[...]

PR #752, opened 2016-01-20
Fix Issue #686

tsbockman:
This fixes [Issue #686](https://github.com/D-Programming-Language/dub/issues/686), "*multiple -march with LDC*".

Note that I have not run this through the full DUB test suite, as it wasn't obvious to me how to use it.

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -80,9 +80,9 @@ struct BuildSettings {
 		addPostBuildCommands(bs.postBuildCommands);
 	}
 
-	void addDFlags(in string[] value...) { dflags ~= value; }
+	void addDFlags(in string[] value...) { add(dflags, value, true); }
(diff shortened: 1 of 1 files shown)


tsbockman:
Hmm. I see this got broken for a reason: https://github.com/D-Programming-Language/dub/commit/51f6a89201b320af296b5fa0f215b5c2043825e7

How should I proceed here? I could special case `-march=`, but that seems like too much of a hack.

tsbockman:
Perhaps unique flags and those which may be duplicated should be tracked separately?

s-ludwig:
I need to dig into the code again, I forgot where the root of the issue was. But I think there was a structural issue in the code, so the fix will require some refactoring.

s-ludwig:
The refactoring fortunately turned out to be rather minor: #759

tsbockman:
@s-ludwig Thanks. I'll close this now.

PR #755, opened 2016-01-27
fix Issue #528 - fetch doesn't work for recently updated package

MartinNowak:
Revert "store metadata cache on disk"

This reverts commit 161aad05884ad52fb99b1fb0291037b562fcac7e.

We only ask the package suppliers for metadata when upgrading or
searching a specific package version. In particular building an already
fetched package does not ask the package suppliers (only the package
manager responsible for locally installed packages).

In all cases where we get data from a package supplier we want to get
the latest available information. Therefor it doesn't make sense to
cache any metadata from the package suppliers on disk.

Not sure whether we still need the in memory metadata cache (e.g. for
multiple queries during dependency resolution).

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -207,10 +207,7 @@ int runDubCommandLine(string[] args)
 	}
 
 	// execute the command
-	int rc;
-	try {
-		rc = cmd.execute(dub, remaining_args, app_args);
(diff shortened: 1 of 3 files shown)


MartinNowak:
I just attached the PR on the original issue #528 instead.

PR #528, opened 2016-01-27
Fetch doesn't work for recently updated package

MartinNowak:
I have updated a [package](http://code.dlang.org/packages/backtrace-d) of mine on the registry around 30 minutes ago. When I try using it as a dependency, I get:
`Root package psql-native contains reference to invalid package backtrace-d`

If I try to `dub fetch backtrace-d --version=0.1.0`, I get:
`No package backtrace-d was found matching the dependency 0.1.0`

If I try the `~master` version, it works.
If I try a different package, it works.

If I am doing a mistake, a better error message would be nice. Otherwise, there is a bug somewhere.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -207,10 +207,7 @@ int runDubCommandLine(string[] args)
 	}
 
 	// execute the command
-	int rc;
-	try {
-		rc = cmd.execute(dub, remaining_args, app_args);
-	}
+	try return cmd.execute(dub, remaining_args, app_args);
 	catch (UsageException e) {
 		logError("%s", e.msg);
 		logDebug("Full exception: %s", e.toString().sanitize);
@@ -222,10 +219,6 @@ int runDubCommandLine(string[] args)
 		logDebug("Full exception: %s", e.toString().sanitize);
 		return 2;
 	}
-
-	if (!cmd.skipDubInitialization)
-		dub.shutdown();
-	return rc;
 }
 
 struct CommonOptions {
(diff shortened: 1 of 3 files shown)


MartinNowak:
We only ask the package suppliers for metadata when upgrading or
searching a specific package version. In particular building an already
fetched package does not ask the package suppliers (only the package
manager responsible for locally installed packages).

In all cases where we get data from a package supplier we want to get
the latest available information. Therefor it doesn't make sense to
cache any metadata from the package suppliers on disk.

Not sure whether we still need the in memory metadata cache (e.g. for
multiple queries during dependency resolution).

The other need to query package suppliers (regukar upgrade check) is already covered by the project local upgrade cache in `./dub/dub.json`.

s-ludwig:
Sounds reasonable. Wonder how the current situation came to be.

MartinNowak:
> Sounds reasonable. Wonder how the current situation came to be.

I added it to not hit the website on a rebuild. Not sure why it did back then, maybe the upgrade cache wasn't yet working?
Pretty sure I tested it like described [here](https://code.dawg.eu/reducing-vibed-turnaround-time-part-1-faster-linking.html).
When I added the cache I didn't know much about the distinction between PackageManager and PackageSupplier and why the metadata was actually queried.

s-ludwig:
Yeah, I was mostly wondering, because there were hard numbers in #388. Anyway, not really important now either - I didn't notice any degrade in performance.

MartinNowak:
> s-ludwig commented 3 days ago
> Yeah, I was mostly wondering, because there were hard numbers in #388. Anyway, not really important now either - I didn't notice any degrade in performance.

Maybe sth. w/ dub.selections.json or the upgrade cache wasn't working back then?
I remember that the metadata cache got rid of a 150ms round trip for every build.
I did test __this__ PR with disabled network, just to be sure.

PR #762, opened 2016-02-01
Fix "dub dustmite" for path based dependencies. Fixes #240.

### LINKED ISSUE #240: Dustmite command fails for dependencies that are referenced by path
Trying to run dustmite on the vibe.d examples will result in a strange recursively repeating path and then fails with "command line too long".

s-ludwig:
This adds handling of path based dependencies to sub packages, as well as path based dependencies of dependencies. Previously, only path based dependencies of the root package god adjusted.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1592,6 +1592,31 @@ class DustmiteCommand : PackageBuildCommand {
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


MartinNowak:
Looks reasonable, but this code has zero test coverage.
https://coveralls.io/builds/5034828/source?filename=source%2Fdub%2Fcommandline.d#L1532

PR #763, opened 2016-02-01
gnu grep extensions aren't portable

John-Colvin:
fix suggested by @dallbee https://github.com/D-Programming-Language/dub/issues/749


s-ludwig:
Thanks! I had actually tried this locally, but didn't have the time to prepare a PR. Will pull once it's green.

PR #766, opened 2016-02-02
Let dub.selections.json always have precedence. Fixes #754.

### LINKED ISSUE #754: Sub-package path dependencies don't work
The following dependencies:
```
        "dependencies": {
      	          "vibe-d": {"path": "../vibe.d"}
        }
```

With `vibe-d` master in `../vibe.d` leads to the following error message:
`Unknown dependency: vibe-d:core`

s-ludwig:
Path based dependencies in dub.sdl/json used to have precedence over dub.selections.json, leading to the error message in #754.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -290,7 +290,8 @@ class Dub {
 			return;
 		}
 
-		foreach (p, ver; versions) {
+		foreach (p; versions.byKey) {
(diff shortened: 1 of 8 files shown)


s-ludwig:
Going to merge this after a self-review. Not sure if the peer-review development process is realistic for DUB at this point. I can't meaningfully continue to work on important things, because that would end up with merge conflicts all over the place due to PRs concerning the same code portions piling up.

Issue #782, opened 2016-03-10
dub build fails when using Debian GtkD packages from http://d-apt.sourceforge.net/


jeanbaptistelab:
Hi,

I've switched to using the GtkD packages from the above mentioned repository as they provide a shared-library version of GtkD which (as far as I know) is not (yet?) possible using dub itself.
My config uses `"libs" : ["gtkd3"]` which causes dub to use `pkg-config --libs gtkd3` for library and linker flags resolution.

The issue here is that the pkg-config `.pc` files for the Debian packages from that repository are written for a build context using the D compiler (dmd, ldc and friends) **_directly**_, either from the command line or from a Makefile. More generally, they are written so that the output of `pkg-config --libs <pure-d-library>` is passed unmodified to the D compiler. DUB however massages this output by calling the `BuildSettings.addLFlags` method for each argument returned by `pkg-config` which prefixes them with an extra `-L` flags.

An sample output might be easier to understand the above description:

Output of `pkg-config --libs gtkd3`:

```
-L-L/usr/lib/x86_64-linux-gnu -L-L/usr/lib/i386-linux-gnu -L-l:libgtkd3.so -L-l:libdl.so.2 -L--no-warn-search-mismatch -defaultlib=libphobos2.so
```

Ouput of `dub build`:

```
Using pkg-config to resolve library flags for gtkd3, sqlite3.
dmd -c -of.dub/build/posix-debug-linux.posix-x86_64-dmd_2070-51125FF0CA3A1BF02DDAD201D7FF67B9/ttracker.o -debug -g -w -version=USE_SQLITE -version=Have_ttracker -version=Have_docopt -version=Have_ttracker_ddbc -Isrc/ -I../../../../usr/include/dmd/gtkd3 -I../../.dub/packages/docopt-0.6.1-b.5/source -Iddbc/source/ -Jres/ src/ttracker/activitylist.d src/ttracker/app.d src/ttracker/db.d src/ttracker/models.d src/ttracker/package.d src/ttracker/utils.d -vcolumns
Linking...
[...]

Issue #444, opened 2014-10-20
Option to disable pkg-config


pszturmaj:
I had problem with manually built ffmpeg libs and pkg-config on Ubuntu 14.04. The problem with ffmpeg's pkg-config was that it appended "-pthread" to libs arguments and dmd/gdc was always returning with /usr/bin/ld error "unrecognized option `-pthread`".

This option may be passed to gcc when compiling sources. Unfortunately, when gcc is used as a linker (without specifying any sources, as dub/dmd does in linking stage) it passes all params to gnu LD. LD linker does not support "-phtread" option (instead one must use "-lpthread"). I also tried with gold linker, but the error was exactly the same.

Please consider adding an option to disable pkg-config's library resolution. Personally, I'm fine with setting "libs" manually. Another option may be skipping "-pthread" parameter when passing pkg-config --libs to the linker.

ximion:
I can confirm this issue, it's quite annoying...

s-ludwig:
For the time being, you could use the "dflags" or "lflags" fields to specify the necessary libraries manually. Not sure how to best approach this, since pkg-config generally outputs in GCC format, but DMD AFAIK doesn't allow to pass flags to the GCC linker command line verbatim. So the only real option seems to be to recognize the `-pthread` switch and remove it, since DMD already links against libpthread.

jeanbaptistelab:
@s-ludwig do you want me to update #783 to transform occurrences of `-pthread` in the output of `pkg-config --libs` to `-lpthread` as well ?

s-ludwig:
*If* you have the necessary time, then that would be great.

jeanbaptistelab:
@s-ludwig I've now added "-pthread" option handling in my fork, #783 should now include this change as well.

Geod24:
There is now an option in DMD to pass arguments directly to CC: https://github.com/dlang/dmd/pull/10441

Unfortunately it still requires handling directly from `dub` AFAICS, but it sounds a much better option than disabling `pkg-config`.

PR #783, opened 2016-03-10
Proposed fix for #782

jeanbaptistelab:
Hi, 

here is a proposed fix for #782 which should allow using dub  with pkg-config ".pc" files for pure D shared libraries...

Cheers,

JB

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -62,7 +62,7 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		string[size_t] package_names;
 		CONFIG[][] all_configs;
 		bool[] any_config;
-		bool[string] required_deps;
+		bool[string] maybe_optional_deps;
 		bool[TreeNode] visited;
 
 		void findConfigsRec(TreeNode parent, bool parent_unique)
@@ -74,7 +74,7 @@ class DependencyResolver(CONFIGS, CONFIG) {
 				auto basepack = rootPackage(ch.pack);
 				auto pidx = all_configs.length;
 
-				if (ch.depType == DependencyType.required) required_deps[ch.pack] = true;
+				if (ch.depType != DependencyType.required) maybe_optional_deps[ch.pack] = true;
 
 				CONFIG[] configs;
 				if (auto pi = basepack in package_indices) {
@@ -110,11 +110,11 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		// getChildren() returns no configurations for an optional dependency,
 		// but getAllConfigs() has already provided an existing list of configs)
 		foreach (i, ref cfgs; all_configs)
-			if (cfgs.length == 0 || package_names[i] !in required_deps)
+			if (cfgs.length == 0 || package_names[i] in maybe_optional_deps)
 				cfgs = cfgs ~ CONFIG.invalid;
 
 		logDebug("Configurations used for dependency resolution:");
-		foreach (n, i; package_indices) logDebug("  %s (%s%s): %s", n, i, n in required_deps ? "" : ", optional", all_configs[i]);
+		foreach (n, i; package_indices) logDebug("  %s (%s%s): %s", n, i, n in maybe_optional_deps ? ", maybe optional" : ", required", all_configs[i]);
 
 		auto config_indices = new size_t[all_configs.length];
 		config_indices[] = 0;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Just one question about the `-defaultlib` switch. Isn't that supposed to go to the D compiler instead of the linker? It doesn't seem to be handled in the test case.

jeanbaptistelab:
I don't know, I need to look at this. Later today probably...

jeanbaptistelab:
I've added handling of the "-pthread" option, this should also help #444 ...

jeanbaptistelab:
Re the `-defaultlib` handling, you're right, this is a dmd option which means its "-L" (if present) needs to be stripped off so as to apply to the dmd invokation. Note (and "this may be a bad ting"™) that dmd expects the *actual* file to be given (for instance, `libphobos2.so`) while ldc seems to expect the "Unix linker" conventional name to be used (i.e. `phobos2`). I've updated my fork to pass `-defaultlib` parameters to the compiler (using the addDFlags() method), but this might not work with ldc (which would then require massaging of the actual `-defaultlib` argument...

jeanbaptistelab:
The tests are now all green :+1:

jeanbaptistelab:
Is there anything I should do to help this being merged ?

s-ludwig:
Sorry, I seem to have forgotten to submit my reply earlier. So we just need to get the tests pass on all compilers. For now I think it's okay to just restrict the new test case to DMD >= 2.068. This can be done in the test script for restricting to DMD:
https://github.com/D-Programming-Language/dub/blob/e865ff2f5d5336d72ff71b4ec1ac394d3ebb9c15/test/issue613-dynlib-pic.sh#L5

And the frontend version can be restricted with a xxx.sh.min_frontend file:
https://github.com/D-Programming-Language/dub/blob/e865ff2f5d5336d72ff71b4ec1ac394d3ebb9c15/test/ddox.sh.min_frontend

jeanbaptistelab:
I've implemented scope restrictions for #782 to DMD frontend, version >= 2.068.

jeanbaptistelab:
Sorry about the serial commits, but it looks like this is good now...

s-ludwig:
Okay, thanks! The diff looks good. Do you have experience with interactive rebase? It would be nice to get the commits cleaned up by squashing related changes and rebasing on top of master instead of merging. If you are not and it takes too much time, I'm not going to be pedantic about this, though ;-)

jeanbaptistelab:
I'm not familiar with it, but I can't think of a better opportunity to learn! Let me have a crack at Google...

jeanbaptistelab:
Ok, I think I got the gist of the rebase business, here is my proposed commit message:

    Fix and test case for issue #782. Analyze the output of pkg-config --libs <lib> and massage them if they appear to already be ready for direct use with the compiler.

    This change allows using pkg-config within dub with .pc files that were written for use with the compiler directly, which can prove troublesome in some cases (for instance, when using the GtkD debian packages as mentioned in issue #782).
    Basically, we analyze the output of pkg-config and fix up arguments so that they are suitable for the D compiler invokations through dub.
    Arguments handled include:
    - double "-L" flags (from gtkd3.pc for example) -> passed as single "-L" flags
    - "-defaultlib" flags (from phobos2.pc for example) -> passed as-is (previous dub code used to pass them as "-Ldefaultlib")
    - "-pthread" flags (from libavcode.pc for example) -> passed as "-L-l-pthread" (previous dub code used to pass them as-is)

    The test case tests for these arguments explicitly, is currently only applicable to DMD frontend versions >= 2.068 as other frontends/versions seems to have some issues with shared libraries.

s-ludwig:
Sounds good! The test case could stay as a separate commit, but the diff is so small that it doesn't matter either way. BTW, I really like the rebase dialog of TortoiseGIT, if you are on Windows.

jeanbaptistelab:
I'm not a fan of Windows, though when I use it, I use Emacs. Actually, I use Emacs everywhere...

s-ludwig:
I gotta admit that I don't really use Windows either anymore, but I really do miss TortoiseGIT.

jeanbaptistelab:
Anything else I need to do ?

s-ludwig:
Oh sorry, I didn't realize that you already did the rebase, because all the old commits were still there.  The problem is that you probably used `git pull` after the rebased to bring your local branch up to date (which added the merge commit at the end and in turn all of the earlier commits as ancestors). Instead you need to `git push --force-with-lease` right after the rebase to overwrite the remote branch with the rebased one.

PR #785, opened 2016-03-11
Clean up API and documentation for the 1.0.0 release

s-ludwig:
This mostly just adds documentation comments. Some code got deprecated/removed and some code got moved. Review should be pretty straight forward, despite the size of the diff.


s-ludwig:
Okay, the API should be good to go for 1.0.0 now (apart from a change in `PackageRecipe.getPlatformSettings`). A lot could still be done to modernize certain aspects (using ranges mostly), but IMO it doesn't make much sense to make that a 1.0.0 priority at this point. I'll rebase and squash related commits, so that reviewing commit-by-commit makes sense. Apart from the first commit, there are no functional changes, so concentrating mostly on the API/doc aspects should be sufficient for the review.

s-ludwig:
Rebase finished. Review status page: https://github.com/D-Programming-Language/dub/wiki/API-review-for-1.0.0

JackStouffer:
There's something to be said for Walter's policy of breaking up PR's by commits :/

source/dub/internal/utils.d
```diff
@@ -359,10 +359,25 @@ auto bitFieldNames(T)(T value) if(is(T==enum) && isIntegral!T)
...
 
 
+bool isIdentChar(dchar ch)
+{
+	return ch >= 'A' && ch <= 'Z' || ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' || ch == '_';
+}
+
+string stripDlangSpecialChars(string s)
+{
+	import std.array;
```
JackStouffer: `import std.array : appender;`

source/dub/internal/utils.d
```diff
@@ -359,10 +359,25 @@ auto bitFieldNames(T)(T value) if(is(T==enum) && isIntegral!T)
...
+	import std.uni;
```
JackStouffer: This seems to be unused.

source/dub/internal/utils.d
```diff
@@ -359,10 +359,25 @@ auto bitFieldNames(T)(T value) if(is(T==enum) && isIntegral!T)
 }
 
 
+bool isIdentChar(dchar ch)
+{
+	return ch >= 'A' && ch <= 'Z' || ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' || ch == '_';
```
JackStouffer:
```
import std.ascii : isAlphaNum;
return ch.isAlphaNum || ch == '_';
```

s-ludwig: I'll fix these. But it should be noted that this code is just moved from another file, so I wouldn't make all shortcomings of the existing code part of this PR, as that would maybe take months to process.

s-ludwig:
> There's something to be said for Walter's policy of breaking up PR's by commits

What difference would that make in this case, though, apart from greatly increasing the overhead for PR creation/merging? I've tried to make one commit for each step, so that reviewing commit-by-commit reduces the pain a bit.

source/dub/dub.d
```diff
@@ -190,9 +224,12 @@ class Dub {
 	deprecated void shutdown() {}
 	deprecated void cleanCaches() {}
 
-	/// Loads the package from the current working directory as the main
-	/// project package.
-	void loadPackageFromCwd()
+	deprecated("Use loadPackage instead.")
+	alias loadPackageFromCwd = loadPackage;
```
wilzbach: maybe add a removal date?

source/dub/dub.d
```diff
@@ -455,7 +507,9 @@ class Dub {
 		writeln(desc.serializeToPrettyJson());
 	}
 
-	void listImportPaths(BuildPlatform platform, string config, string buildType, bool nullDelim)
+	/** Prints a list of all import paths necessary for building the root package.
+	*/
+	deprecated void listImportPaths(BuildPlatform platform, string config, string buildType, bool nullDelim)
```
wilzbach: needs removal date and annotation - e.g. `@@@DEPRECATED_2016-07@@@`

source/dub/packagesupplier.d
```diff
@@ -23,26 +23,74 @@ import std.file;
 import std.string : format;
 import std.zip;
 
-// TODO: drop the "best package" behavior and let retrievePackage/getPackageDescription take a Version instead of Dependency
+// TODO: Could drop the "best package" behavior and let retrievePackage/
```
wilzbach: probably TODO should be put as open issues ...

wilzbach:
I found a bit of time and had a look over this.
@s-ludwig yep this really just moves around a lot of stuff and adds documentation.
Imho this is sitting here far too long - let's go with!
We won't spot bugs without an alpha release anyhow ;-)

Maybe you want to set a removal date for the deprecation and annotate them so it's easier to find?
(Btw SemVer allows breaking changes for major releases)

source/dub/dub.d
s-ludwig: My plan is to remove all deprecated symbols for the 1.0.0 release, so that we can start with a clean slate. So instead of a date, I'd mention the version, but I'll add that for all deprecations throughout the API.

source/dub/packagesupplier.d
s-ludwig: I was half way into putting that TODO into practice, before I noticed that this would produce some redundant HTTP requests and reverted the changes. I left the comment with a more careful wording, but it really should probably just be removed instead.

s-ludwig:
Okay, deprecations are augmented with removal notes now and everything is rebased onto master.

ZombineDev:
Hi Sönke,

Here's my review:

General suggestions:

+ Group `dub.semver`, `dub.dependency` and `dub.dependencyresolver` into a
  package for dependency resolution.
+ Group `dub.description`, `dub.init`, `dub.package_`, `dub.packagemanager`,
  `dub.packagesupplier`, `dub.project` and `dub.recipe.packagerecipe` into a package for DUB package management.
+ Group `dub.recipe.io`, `dub.recipe.sdl`, `dub.recipe.json` and `dub.internal.utils` and some parts for `dub.commandline`
  into a package for IO related functionality
+ Leave only `dub.dub`, `dub.commandline` and `dub.version_` as top level `dub.*` modules
+ Use selective imports everywhere
+ Use local imports if something is used in < 5-10% (or in only 2-3 places) of the whole module
+ Avoid global variables
+ Don't use exceptions for return values (e.g. determineVersionFromSCM
  should return either a valid `Version` or Version.init (== Version.unknown)).
+ Don't catch bare `Exception`s  
+ Remove constructor overloads that do IO
+ 80% of the code should be pure and separate from the code that does the IO
+ Use ranges instead of arrays and strings where possible.
  Prefer lazily algorithms. Avoid GC usage. Remove things like
  `.map!(f => cast(string)f.filename).array;`. Change iterator delegates to ranges.
+ Avoid trivial `@property`s - instead make the particular fields public and const (where applicable). Use properties only when more compicated logic is needed than plain get/set.
+ [very low priority]: Use phobos code formatting - braces, indentation, etc. - (although I personally prefer snake_case for local variables).
+ I see no point in deprecations as DUB is still not 1.0.0, and if I were you, I would just remove the old stuff, but I leave this decision to you.

Commit-by-commit review:

* bfe27fb430e080ef09d270b46a7964666c034b82 (Remove pointless extra abstraction for package recipes)
  + the changes look good

* a11a28047bf6ff2048b4e1bbf986ad1e85cedd98 (Remove PathAndFormat struct)
  + all good

* 083947453f32246719d5c116629d614e565e0b17 (Code cleanup)
  + the change is OK, but:
  + constructors that do IO are a bad idea. Instead remove this overload and add a unittest showing the user how he can accomplish the same by manually reading the file.

* 357f5f231991aa19d7bf7b51b7f4ef13c24f49bc (Clean up and document the dub.package_ public API)
  + the change is OK, but:
  + Maybe Package.dependencies should be named Package.allDependencies
  + Change Package.allDependencies documentation to: "Returns a union of the dependencies of all configurations" (it sounds more clear this way).
  + Explain which version will Package.allDependencies when there is conflict - the newest, the oldest, or a union?
  + Merge Package.getDependencies into Package.allDependencies

* b58ad63f8b11843e87a99f83a77616a8cf9c26ae (Review dub.semver and improve documentation)
  + the change is OK, but:
  + it seams there is some overlap with dub.dependency. Move more stuff from there to this module?

* e48ec4935731950803b6b75e312d0835d95e7e66 (Review dub.init and improve documentation)
  + all good
  + Maybe initPackage should support custom user-defined init templates in the
    future (other than minimal, vibe.d and deimos).

* a0eb519765b99384fe2ea0ca8ba60a7b0a274241 (Review dub.description and improve documentation)
  + all good

* 53d838f3d3dc64be3eeb2e8ee2408f0ef52987d7 (Move BuildPlatform to dub.platform and improve docs)
  + the change is OK, but:
  + I'm not sure determineBuildPlatform and friends are useful. They just provide information about the DUB binary. They can only be useful when the package target platform is the same as the platform on which DUB was compiled.
  + There is duplication with `dub.compilers.utils.generatePlatformProbeFile`. Refactor with mixins?

* 5a6cd75d1cb4ac30b95037bcda73fd5b5f9ca61e (Review and improve documentation for dub.commandline)
  + all good

* 77c4d181e08133a71af80e5d6b0a343b180f8279 (Review dub.recipe.io and improve documentation)
  + Rename `readPackageRecipe(Path file, ...)` to `Path filename` so that it shows up in documentation next to `string filename`.
  + otherwise all good

* df8fb07227123cf3baa8bd742f66e6d1f87cac81 (Make internal function in dub.recipe.json private.)
  + all good

* 59bb9aa803d7aab78436430b2287d79662a6dbef (Review dub.dependency and improve documentation)
  + the changes look good, but:
  + `Dependency` should be backed by an `Algebraic` (or something conceptually similar), because it can't be both a specific version, version range and path at the same time.
  + Static properties like `any`, `invalid`, etc should have `Dependency` return type for clarity, instead of relying on return type inference
  + Magic constants like `ANY_IDENT`, `"~master"`, etc should be all enums declared next to each other in the code
  + No need for trivial get and set properties - expose fields directly. Use properties only when more compicated logic is needed.
  + Add more comments on the internals of `Dependency.versionSpec`
  + There are `Dependency.fromJson` and `Dependency.toJson`. What about SDL?
  + There is some overlap between `Version` and `Dependency`. I think you simplify the code if you separate it more cleanly.

* 6d616ee1ffc49cccd77f70244297107ab3bbf4b3 (Fix naming convention for the compiler handler classes and add documentation)
  + the changes look good, but:
  + dub.compilers.compiler and dub.compilers.[dmd|gdc|ldc] import each other.
  Break the import cycle by moving all but the `dub.compilers.compiler.Compiler` 
  interface to `dub.compilers.package`. Maybe rename `dub.compilers.compiler`
  to `dub.compilers.interfaces`
  + Now I saw you did some of that in 5b2f303a510593f13a1acadc86b9215deef3374c, but there are still a couple of things left

* 706d03b6b845327136ebc23560495aba4618f3b7 (Review dub.generators.generator and improve documentation)
  + all good

* dd1a0653dd3b2a5920d888cf77902e3b22dd89b2 (Make all internal generator functionality private)
  + all good

* bec55078a9f64f0a777cb3d4a17ca62b838a12d6 (Avoid any of the newly deprecated symbols)
  + all good

* 95da1741f48dd93e6a1f2a3f5e42d22ac43ab923 (Review dub.packagemanager and improve documentation)
  + the changes look good

* 5b2f303a510593f13a1acadc86b9215deef3374c (Move utility functions from dub.compilers.compiler to dub.compilers.utils)
  + the changes look good, but:
  + there are still some misc stuff left in dub.compilers.compiler that shouldn't be there
  + Move the `isLinkerFile` unittest from `dub.generators.build` to `dub.compilers.utils`

* 85d1540897d8731a954e3e230e00bd1852b9821a (Add documentation for dub.compilers.compiler/utils)
  + all good

* 63f8612ffbfc8ff98414b44c07293a333d7f0dd1 (Review and document dub.packagesupplier)
  + all good

* 2e8fa59403038c00ef5d877729c1b6dc79e2e7da (Review and document module dub.dub)
  + the changes look good

* 088633e381762a58ebc95279c39c3255f0b6aa90 (Review and document dub.project)
  + the changes look good

* a49678f3e4b564691bad1b131a481341f8c52f83 (Make defaultPlacementLocation a member of Dub instead of a global)
  + the changes look good

* 62122b8d4df97bb015a869ad0313ce3a4e53af23 (Fix behavior of "--data-list --data-0" and improve the Dub/Project.listBuildSettings API)
  + the changes look good

* 660c2b5a45a964b2868c425d2c0225d1b105e1cd (Review fixes)
  + all good

* bc3c1c522ce2160e49d23ffaedd1c5c240bdf96b (Mention version of removal for all deprecated public symbols)
  + all good

Overall this PR LGTM.
Most of my complaints are about better organization of the code, making most of the functions pure (by moving out the IO), local imports and less arrays and more ranges, and DRY.

ZombineDev:
@JackStouffer I personally found it easy to do a commit-by-commit review. Most commits add documentation, move stuff around, or change names / add deprecations. But you still need look at each module as a whole to get a sense of the bigger picture.

s-ludwig:
@ZombineDev: Thanks a lot for the detailed review! I agree with almost all of the points, but I think that the time frame isn't sufficient to implement most of them. The goal is to get DUB to 1.0.0 ASAP and just avoid the most prominent kinks in the public API. That's why I restrained my changes to the most obvious and least intrusive ones for now.

But 2.0.0 should definitely aim for a thorough design revision - the codebase is dated, D and Phobos changed a lot over the years, and the first months of development were rather rushed, to get something that works fast (as a package manager only for vibe.d and not intended as a general purpose tool). When this PR is merged, I'll put your remaining points into an issue or on the wiki so that they can eventually get processed after the 1.0.0 release (can happen additively, flagging the old APIs as "scheduled for deprecation").

As the first step, I'll go through the specific points for each commit (might take some time, but hopefully I can allocate a good chunk of time tomorrow).

BTW, regarding deprecations, I don't think that pre- or post-1.0.0 really makes a practical difference. Some projects already use DUB as a library and not having a grace period for breaking changes is always a pain point (requiring `static if` workarounds or adjusting foreign code bases), so I like to do that anyway.

s-ludwig:
> * 083947453f32246719d5c116629d614e565e0b17 (Code cleanup)
>   + the change is OK, but:
>   + constructors that do IO are a bad idea. Instead remove this overload and add a unittest showing the user how he can accomplish the same by manually reading the file.

Replaced the constructor by a static `load()` method.

> * 357f5f231991aa19d7bf7b51b7f4ef13c24f49bc (Clean up and document the dub.package_ public API)
>   + the change is OK, but:
>   + Maybe Package.dependencies should be named Package.allDependencies
>   + Change Package.allDependencies documentation to: "Returns a union of the dependencies of all configurations" (it sounds more clear this way).
>   + Explain which version will Package.allDependencies when there is conflict - the newest, the oldest, or a union?
>   + Merge Package.getDependencies into Package.allDependencies

Replaced `.dependencies` by `.getAllDependencies()`, which returns a list instead of an AA.

> * b58ad63f8b11843e87a99f83a77616a8cf9c26ae (Review dub.semver and improve documentation)
>   + the change is OK, but:
>   + it seams there is some overlap with dub.dependency. Move more stuff from there to this module?

You mean the `Version` class, right? The only issue is that `Version` also supports branch identifiers and maybe commit references in the future, so it wouldn't fully fit in a `semver` module.

> * e48ec4935731950803b6b75e312d0835d95e7e66 (Review dub.init and improve documentation)
>   + all good
>   + Maybe initPackage should support custom user-defined init templates in the
    future (other than minimal, vibe.d and deimos).

Definitely on the radar.

> * 53d838f3d3dc64be3eeb2e8ee2408f0ef52987d7 (Move BuildPlatform to dub.platform and improve docs)
>   + the change is OK, but:
>   + I'm not sure determineBuildPlatform and friends are useful. They just provide information about the DUB binary. They can only be useful when the package target platform is the same as the platform on which DUB was compiled.

They are useful when DUB is used as a library. I'm for example using it for a CI slave, so that it can report the actual platform for a particular set of compiler switches.

>   + There is duplication with `dub.compilers.utils.generatePlatformProbeFile`. Refactor with mixins?

Should definitely be done. I'll get to this later, since it's not part of the API.

> * 77c4d181e08133a71af80e5d6b0a343b180f8279 (Review dub.recipe.io and improve documentation)
>   + Rename `readPackageRecipe(Path file, ...)` to `Path filename` so that it shows up in documentation next to `string filename`.

Done.

> * 59bb9aa803d7aab78436430b2287d79662a6dbef (Review dub.dependency and improve documentation)
>   + the changes look good, but:
>   + `Dependency` should be backed by an `Algebraic` (or something conceptually similar), because it can't be both a specific version, version range and path at the same time.

Good point. I think this needs to be postponed to after 1.0.0, though.

>   + Static properties like `any`, `invalid`, etc should have `Dependency` return type for clarity, instead of relying on return type inference

Done.

>   + Magic constants like `ANY_IDENT`, `"~master"`, etc should be all enums declared next to each other in the code

Do you mean globally in the module? They are now next to each other within their struct context. Naming convention still needs to be fixed, but they are not public.

>   + No need for trivial get and set properties - expose fields directly. Use properties only when more compicated logic is needed.

At least the `optional`/`default` ones could still come in handy, because it would make sense to switch over to using an enum (`dub.dependencyresolver.DependencyType`) to avoid the redundant state `optional == false && default == true`.

>   + Add more comments on the internals of `Dependency.versionSpec`

Do you have something specific in mind? Maybe the behavior of the "fuzzy" version match could be explained there, or a link to the relevant section in the package format specification could be added.

>   + There are `Dependency.fromJson` and `Dependency.toJson`. What about SDL?

The SDLang serialization code is in `dub.recipe.sdl`, but is `private`. The `toJson`/`fromJson` methods are there for generic serialization and because they are used in other places (dub.selections.json), but I'd really rather like to remove them, too.

>   + There is some overlap between `Version` and `Dependency`. I think you simplify the code if you separate it more cleanly.

What do you mean here specifically?

> * 6d616ee1ffc49cccd77f70244297107ab3bbf4b3 (Fix naming convention for the compiler handler classes and add documentation)
>   + the changes look good, but:
>   + dub.compilers.compiler and dub.compilers.[dmd|gdc|ldc] import each other.
>   Break the import cycle by moving all but the `dub.compilers.compiler.Compiler` 
>   interface to `dub.compilers.package`. Maybe rename `dub.compilers.compiler`
>   to `dub.compilers.interfaces`
>   + Now I saw you did some of that in 5b2f303a510593f13a1acadc86b9215deef3374c, but there are still a couple of things left

The only reason `dub.compilers.compiler` imports the implementations is to register them at startup. That code could for example also be moved to some higher level module (e.g. `dub.dub`).

> * 5b2f303a510593f13a1acadc86b9215deef3374c (Move utility functions from dub.compilers.compiler to dub.compilers.utils)
>   + the changes look good, but:
>   + there are still some misc stuff left in dub.compilers.compiler that shouldn't be there

There is `getCompiler`, `registerCompiler` and `defaultCompiler`. The latter the only one that doesn't fully fit, but the `utils` module is intended to contain functionality that is useful to compiler class implementations, not for external code. A possibility would be to move this function to the `Dub` class (which is the only user so far and already exposes a `defaultCompiler` property).

>   + Move the `isLinkerFile` unittest from `dub.generators.build` to `dub.compilers.utils`

Done.

s-ludwig:
Hm that test failure for 2d8f991 is a mystery...

s-ludwig:
Cyclic compiler imports and `defaultCompiler` are also done now.

s-ludwig:
Okay, since the remaining points are of general nature and note specific to the change set here, I'll merge this as soon as Travis is green to unblock further development and will create an issue for the rest.

PR #800, opened 2016-03-31
Fix static lib extension for ldc on windows.

MrSmith33:


s-ludwig:
The correct fix is more involved, as there are two versions of LDC on Windows. See #618 and #688 for a partial fix.

MrSmith33:
Yep, looks difficult! Thanks for info.

s-ludwig:
#688 is now ready for merge after a review.

PR #818, opened 2016-04-16
update links to dlang

wilzbach:
Updated links to our new namespace.
Be a bit careful - travis needs a trigger to know about the move.


s-ludwig:
BTW, great that you took the initiative to make the switch happen. I thought about it many times, but never really did something about it so far.

wilzbach:
> BTW, great that you took the initiative to make the switch happen. I thought about it many times, but never really did something about it so far.

Seems like sometimes you just need to ping people enough and cool stuff is happening :)

PR #821, opened 2016-04-20
add dynamic library build target

linux-fan-dave:
I want to add d / dub support to qt-creator and want to use dub as a library. The problem with the static library is that it will not get compiled with -fPIC support which is needed to be linked against the c++ qtcreator plugin api. I figured that compiling dub as a shared library solves this problems, hence the additional configuration.


s-ludwig:
Should be good to go as a workaround. We should really start to add more support for dynamic libraries (e.g. adding a command line option and forcing a library dependency to build as static/dynamic).

PR #827, opened 2016-04-27
init with empty deps

wilzbach:
This would initialize `dependencies` for the `json` file - it leave `sdl` untouched.
The idea ist that it is a bit easier to copy & paste dependencies once the new code.dlang.org website with the clipboard buttons is released.

Btw quite a silly workaround, but I couldn't come up with something smarter.

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -90,6 +90,15 @@ void initPackage(Path root_path, string[string] deps, string type,
 private void initMinimalPackage(Path root_path, ref PackageRecipe p)
 {
 	p.description = "A minimal D application.";
+
+	// we want to create a empty, not-null dictionary,
+	// s. t. the generated recipe contains an empty "dependencies" section
+	// in case of PackageFormat.json
+	auto bar = Dependency("~>0.0.0");
+	p.buildSettings.dependencies["foo"] = bar;
+	p.buildSettings.dependencies.remove("foo");
+	assert(p.buildSettings.dependencies !is null);
+
 	createDirectory(root_path ~ "source");
 	write((root_path ~ "source/app.d").toNativeString(),
 q{import std.stdio;
--- a/test/0-init-simple-json.sh
+++ b/test/0-init-simple-json.sh
@@ -1,16 +1,32 @@
 #!/bin/bash
 
 packname="0-init-simple-pack"
+currentYear=$(date +"%Y")
 
 $DUB init -n $packname -f json
+
+cat > $packname/dub2.json <<- EOL
+{
+	"name": "$packname",
+	"authors": [
+		"$USER"
+	],
+	"dependencies": {},
+	"description": "A minimal D application.",
+	"copyright": "Copyright © $currentYear, $USER",
+	"license": "proprietary"
+}
+EOL
 
 function cleanup {
     rm -rf $packname
 }
 
+diff -w $packname/dub.json $packname/dub2.json
+
 if [ ! -e $packname/dub.json ]; then # it failed
-    cleanup
-    exit 1
+	cleanup
+	exit 1
 fi
 cleanup
 exit 0


ZombineDev:
1. Can you add a test in the `tests/` directory?
2. Nitpick: dub and vibe.d [use](https://github.com/dlang/dub/blob/master/.editorconfig) tabs for indentation. Maybe you forgot to setup [editorconfig](http://editorconfig.org) for your editor?

wilzbach:
> Can you add a test in the tests/ directory?

Sure. Done.

> Nitpick: dub and vibe.d use tabs for indentation. Maybe you forgot to setup editorconfig for your editor?

Done (also for the modified test).

source/dub/init.d
```diff
@@ -90,6 +90,13 @@ void initPackage(Path root_path, string[string] deps, string type,
 private void initMinimalPackage(Path root_path, ref PackageRecipe p)
 {
 	p.description = "A minimal D application.";
+
+	// we want to create a empty, not-null dictionary
```
s-ludwig: I'd also mention the motivation here - looks suspicious for something that will be unclear when looking at it in a year or so, e.g. "so that the generated recipe contains an empty "dependencies" section in case of PackageFormat.json". Otherwise, LGTM.

test/0-init-simple-json.sh
```diff
@@ -1,16 +1,32 @@
...
 
+cat > $packname/dub2.json <<- EOL
+{
+	"name": "$packname",
+	"authors": [
+		"$USER"
+	],
+	"dependencies": {},
+	"description": "A minimal D application.",
+	"copyright": "Copyright © $currentYear, $USER",
```
s-ludwig: I'm a bit surprised that this works, due to #715, I hope that's just the Travis setup. I don't know how difficult it is to get the users full name from the CLI, but we'd need either that or just searching for `"dependencies": {}` within the file would be a viable alternative.

wilzbach: why? For Posix #715 checks $USER

source/dub/init.d
wilzbach: makes sense. done.

test/0-init-simple-json.sh
s-ludwig: But only if `getpwuid` fails: https://github.com/dlang/dub/pull/715/files#diff-2f485d9efbc154e37bbdbdb8ee846aaeR176

MartinNowak: Yes, better just grep for dependencies.

source/dub/init.d
```diff
@@ -90,6 +90,15 @@ void initPackage(Path root_path, string[string] deps, string type,
 private void initMinimalPackage(Path root_path, ref PackageRecipe p)
 {
 	p.description = "A minimal D application.";
+
+	// we want to create a empty, not-null dictionary,
+	// s. t. the generated recipe contains an empty "dependencies" section
+	// in case of PackageFormat.json
+	auto bar = Dependency("~>0.0.0");
+	p.buildSettings.dependencies["foo"] = bar;
+	p.buildSettings.dependencies.remove("foo");
```
MartinNowak: Would it make more sense to change the json writer to always add a dependencies section?

s-ludwig: On first sight it feels like a too leaky design to me. The approach here is of course also rather hackish, but it has a very specific target. Always outputting a dependencies field on the other hand would affect 90% of cases where this isn't actually needed/desired.

andre2007:
@wilzbach is this pr still relevant as there is now a dub command to add dependencies?

Geod24:
> @wilzbach is this pr still relevant as there is now a dub command to add dependencies?

I would say no.

Issue #831, opened 2016-04-29
dub stalls, searching for unused optional dependencies every recompilation


cyisfor:
Every compilation is taking an extra 10 or 15 seconds, because `dub -v` pauses on things like:
```
Search for versions of botan (1 package suppliers)
...
Search for versions of botan-math (1 package suppliers)
```

According to #361 the fact that botan is not in dub.selections.json should imply that it's not being used, but it should also imply not to search for versions of it online. Since it's not used, dub never downloads a local copy of botan (or any optional dependency AFAIK), and since it never has a local copy, dub always checks online to find what versions _not_ to include, _every recompilation._

Fixing that would make my edit-compile-run loop go faster, and reduce the amount that servers are getting hammered for possibly complex database queries. Unless there's something I'm not understanding here? Why does it pause on botan, if it's an optional dependency, and not selected?

PR #832, opened 2016-04-29
No check optional deps

cyisfor:
Avoid stalling on polling online servers for versions of dependencies that we aren't going to use.

Not sure if I did this right... works for stopping optional deps, but I don't have an optional dependency to actually select, to test if that sets `depType` to `required` early enough.

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -74,13 +74,13 @@ class DependencyResolver(CONFIGS, CONFIG) {
 				auto basepack = basePackage(ch.pack);
 				auto pidx = all_configs.length;
 
-				if (ch.depType != DependencyType.required) maybe_optional_deps[ch.pack] = true;
+				if (optional) maybe_optional_deps[ch.pack] = true;
 
 				CONFIG[] configs;
 				if (auto pi = basepack in package_indices) {
 					pidx = *pi;
 					configs = all_configs[*pi];
-				} else {
+				} else if(ch.depType == DependencyType.required) {
 					if (basepack == root_base_pack) configs = [root.config];
 					else configs = getAllConfigs(basepack);
 					all_configs ~= configs;
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -74,19 +74,23 @@ class DependencyResolver(CONFIGS, CONFIG) {
 				auto basepack = basePackage(ch.pack);
 				auto pidx = all_configs.length;
 
-				if (optional) maybe_optional_deps[ch.pack] = true;
+				if (ch.depType != DependencyType.required) maybe_optional_deps[ch.pack] = true;
 
 				CONFIG[] configs;
 				if (auto pi = basepack in package_indices) {
 					pidx = *pi;
 					configs = all_configs[*pi];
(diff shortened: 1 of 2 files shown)


source/dub/dependencyresolver.d
```diff
@@ -81,13 +81,17 @@ class DependencyResolver(CONFIGS, CONFIG) {
 					pidx = *pi;
 					configs = all_configs[*pi];
 				} else {
-					if (basepack == root_base_pack) configs = [root.config];
-					else configs = getAllConfigs(basepack);
-					all_configs ~= configs;
-					package_indices[basepack] = pidx;
-					package_names[pidx] = basepack;
+				  if (basepack == root_base_pack) configs = [root.config];
+				  else if(ch.depType == DependencyType.required)
```
MartinNowak: Looking at https://github.com/dlang/dub/blob/9a37109faa3a956d625a95a5946314a71da0ffd7/source/dub/dub.d#L1218-L1227, this should be `ch.depType != DependencyType.optional`.

MartinNowak: Also there should be a space after the `if`.

source/dub/version_.d
```diff
@@ -1,2 +0,0 @@
-module dub.version_; 
-enum dubVersion = "v0.9.24";
```
MartinNowak: Please don't delete this file.

s-ludwig:
Looks like we overlooked a valid case for the meta data cache. I'm not fully sure about it in general, yet, but one issue with this change is that it doesn't work correctly in all cases (in particular the case when there is a second non-optional dependency to the same package must be handled).

MartinNowak:
Can we get the configs lazily once an optional package is required as dependency?

MartinNowak:
Should we defer resolution of this bug until we rework the dependency resolver?

s-ludwig:
Yes, let's defer this to 1.0.0. It's annoying with no/slow internet connection, but can be worked around with the `--nodeps` switch. 1.0.0 is scheduled for June 15, so 0.9.25 needs to get out ASAP. Do you think that you'll have time to implement the dependency resolver until then (minus beta phase)? But my hope is that this won't change the public API, so that it could also wait for 1.1.0.

s-ludwig:
I've fixed the general #831 issue now with a different approach: #861 - However, this one may still be good to have, but I'll have to think this through in-depth to be sure about all side-effects.

cyisfor:
@s-ludwig Well, thanks for fixing it. I don't care what approach as long as it works. I just wrote this attempt to avoid being a totally useless lump over in the issue section.

s-ludwig:
Definitely appreciated! My fix doesn't catch all the cases that this one does, so if it doesn't accidentally break something, it makes sense to integrate it.

s-ludwig:
I just thought of an issue with this approach: If the same package is referenced twice, once optional and once non-optional, if the optional dependency is visited first, it will end up with no version candidates, even though there is the other non-optional dependency. This could probably be fixed with a `purely_optional_packages` set and some twisting of the program logic, but it raises the question of the additional complexity still being worth it.

MartinNowak:
One could try to overwrite the configs of earlier optional packages of they are required later on.
Using a hash table for all_configs might be enough to do that.

PR #846, opened 2016-05-20
warn about .dub/packages layout change

MartinNowak:
Had some issues with still using 0.9.24 in some places, e.g. when building dlang.org.
Even preferable if we can find a way to keep both layouts, e.g. using symlinks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,12 @@ v0.9.25 - 2016-05-
 ### Features and improvements ###
 
  - Builds on DMD 2.064.2 up to 2.071.0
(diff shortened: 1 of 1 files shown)


s-ludwig:
If I'm not totally off, it doesn't actually upgrade the layout at once, but just places newly downloaded packages into the new structure, while still recognizing old packages. There has been an obscure issue with this, though, and I need to look up if I actually committed a fix, yet.

s-ludwig:
Okay, the fix is in beta-3.

MartinNowak:
Sounds fine then.

PR #847, opened 2016-05-22
Use std.stdio instead of the deprecated std.stream.

s-ludwig:
Since opening non-ASCII file names on Windows is fixed for the latest DMD version, it now only has to be made sure to use a recent compiler to build DUB.

### DIFF
--- a/source/dub/internal/vibecompat/core/file.d
+++ b/source/dub/internal/vibecompat/core/file.d
@@ -17,7 +17,7 @@ import std.datetime;
 import std.exception;
 import std.file;
 import std.path;
-static import std.stream;
+import std.stdio;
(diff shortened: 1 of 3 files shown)


wilzbach:
Looks reasonably to me (especially getting rid of std.stream), but I don't use Windows.
Maybe we can setup [AppVeyor](https://github.com/libmir/mir/blob/master/appveyor.yml) to ensure that we don't accidentally break something?

s-ludwig:
AppVeyor would be a good idea, I personally already broke things on Windows one or two times in the past since I started working mostly on Linux. For now, I've verified that it works locally and I also just added a test that will be run once we have a Windows CI solution.

PR #848, opened 2016-05-23
Sped up -> Speed up

burner:


s-ludwig:
This is actually intended to be past tense, could also use "speeded", but isn't "sped" correct, too?

burner:
my bad http://english.stackexchange.com/questions/114620/speeded-vs-sped

sorry for the noise

PR #851, opened 2016-05-24
Implement single-file packages. Fixes #103.

### LINKED ISSUE #103: Single-file packages
(Note: I'd post it to the newsgroup, but [issue #8](https://github.com/rejectedsoftware/vibenews/issues/8) prevents me)

### Problem
When dealing with large libraries including multiple files, sub-packages, and so on, dub is well-designed and easy to use. There is one target audience, however, where the user interface could be simpler: packages with only a single file.

The current choice offered is either to make one repository for every single-file package (e.g. my own `embd`), or to lump together several often unrelated packages into one repo (e.g. the great `arsd`). The former option is clunky, and the latter requires users to download and compile several modules they won't need. Another option is needed.

### Solution
An SFP is a folder named `example` with a single file inside named `example.d` (not including `.git`, etc). Somewhere in the file will be a comment like so:

```
/+
dub:
 {
     ...what would be in package.json...
 }
+/
```

This file could be automatically generated by:
```
$ dub init --sfp example
    create folder example
    create file example/example.d
```

### Questions
Here's some additional ideas:

  * Requiring that the packages be `sfp/[name]` in the repository?
  * Adding support for [gist](https://gist.github.com/) and [dpaste](http://dpaste.dzfl.pl/)?
  * An option to generate a symlink in the parent folder: `example.d -> example/example.d`

School's about to start, so I'll be slow, but I'm willing to work on it.

Thoughts?
NMS

s-ludwig:
See #103.


s-ludwig:
Inferring the package name requires some additional changes.

test/issue103-single-file-package.d
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env dub
+/+ SDL
+name "hello_world"
++/
```
wilzbach: Is it possible to have a tested  example with dependencies?

s-ludwig: Good point. While normal dependencies work fine, while extending the test case, I noticed that path based dependencies are broken. Fixing this cleanly requires some more thought.

s-ludwig:
Okay, should be ready now. I propose to defer inferring the package name, because that requires some package parser API changes that may or may not be worth it.

ZombineDev:
Overall LGTM. BTW is dub.json supported as a comment?

ZombineDev:
The pretty informative comment on `loadSingleFilePackage` answered my question. Though a test case would be nice.

s-ludwig:
Thanks for looking through it! I'll add a dub.json test and rebase + pull.

source/dub/dub.d
```diff
@@ -247,6 +247,75 @@ class Dub {
...
+	{
+		import dub.recipe.io : parsePackageRecipe;
+		import std.file : mkdirRecurse, readText;
+
+		path = makeAbsolute(path);
+
+		string file_content = readText(path.toNativeString());
+		auto idx = file_content.indexOf("/+");
+		enforce(idx >= 0, "Missing /+ ... +/ recipe comment.");
+		file_content = file_content[idx+2 .. $].strip();
```
ZombineDev: I think calling `strip` here is unnecessary, as the call below should cover its purpose.

s-ludwig: Indeed, that was left-over from an earlier code version.

source/dub/dub.d
```diff
@@ -247,6 +247,75 @@ class Dub {
 		m_project = new Project(m_packageManager, pack);
 	}
 
+	/** Loads a single file package.
+
+		Single-file packages are D files that contain a package receipe comment
```
ZombineDev: Indentation is a bit off.

source/dub/dub.d
```diff
@@ -247,6 +247,75 @@ class Dub {
...
+		auto recipe_filename = recipe_content[0 .. idx].strip();
+		recipe_content = recipe_content[idx+1 .. $];
+
+		auto recipe = parsePackageRecipe(recipe_content, recipe_filename);
+		recipe.buildSettings.sourceFiles[""] = [path.toNativeString()];
+		recipe.buildSettings.mainSourceFile = path.toNativeString();
+		if (recipe.buildSettings.targetType == TargetType.autodetect)
+			recipe.buildSettings.targetType = TargetType.executable;
+
+		auto root_path = m_tempPath ~ (".dub/single/"~recipe.name);
```
ZombineDev: It's unclear where root_path is used. I would have expected it to be passed to the Package constructor below. Or maybe it should be handled in a different place?

source/dub/dub.d
s-ludwig: The doc comments are currently all done that way (using a single tab to indent the intermediate lines). With a 4-space tab stop they happen to line up, but if we want to aim for *proper* alignment, we'd have to switch to prefixing with "*" to avoid .editorconfig getting in the way.

source/dub/dub.d
s-ludwig: Another left-over from the initial version. Removed now.

source/dub/dub.d
ZombineDev: Hmm, I would have expected github to recognize .editorconfig and to use 4 space spaces for tabs. Anyway, not important.

ZombineDev:
Thanks for the changes (especially for removing the path creation from `loadSingleFilePackage`). Now everything LGTM :P

s-ludwig:
Travis is green again. Merging.

PR #854, opened 2016-05-27
Escaping LDC arguments given in a @cmdfile (Issue 834)

p0nce:
This works on Windows when issue 834 was spotted in the first place, but need to be tested for POSIX first. (Presumably just adding " " wouldn't work there, and there was no problem in the first place in POSIX AFAIK)

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -217,8 +217,15 @@ class LDCCompiler : Compiler {
 	void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback)
 	{
 		auto res_file = getTempFile("dub-build", ".rsp");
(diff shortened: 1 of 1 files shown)


p0nce:
May I ask why we need to give LDC arguments in a @cmdfile in the first place?

s-ludwig:
> May I ask why we need to give LDC arguments in a @cmdfile in the first place?

That's just because the command line length can easily exceed the OS limit and the command file doesn't have any limitation.

It would of course be good to have a fully correct solution (e.g. when an argument with `"` is encountered), but I guess this is good enough a step forward, and DMD for example doesn't currently do anything else either.

p0nce:
OK. Let me rework this to do like for DMD, this `escapeArgs` function should be reused.

PR #861, opened 2016-06-02
Only resolve dependencies if necessary. Fixes #831.

### LINKED ISSUE #831: dub stalls, searching for unused optional dependencies every recompilation

s-ludwig:
This avoids the default dependency resolution pass to search for missing dependencies if all dependencies are satisfied already.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -615,11 +615,14 @@ abstract class PackageBuildCommand : Command {
 			// TODO: only upgrade(select) if necessary, only upgrade(upgrade) every now and then
 
 			// retrieve missing packages
-			logDiagnostic("Checking for missing dependencies.");
-			if (m_single) dub.upgrade(UpgradeOptions.select | UpgradeOptions.noSaveSelections);
-			else {
-				dub.upgrade(UpgradeOptions.select);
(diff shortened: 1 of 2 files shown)


andralex:
going on a limb here :)

MartinNowak:
Another approach which would also allow partial updates of dependencies would be to sort the selected versions to the front of the available version, thus starting with resolving the current selection.
But for a normal build, not resolving anything makes sense.

PR #862, opened 2016-06-03
Use release flags by default in build.sh

tmccombs:


s-ludwig:
`-O -inline` makes sense. I'd weigh debugability as important than the last bit of performance, though, so I'd replace `-release` with `-g`, so that assertions stay and that ideally there are meaningful stack traces.

tmccombs:
According to this https://forum.dlang.org/post/pantqnvhhaozetsnhjjq@forum.dlang.org asserts take a considerable amount of time.

Geod24:
`build.sh` is gone now. Users of `build.d` can pass the flag they want as argument.
Given DMD track record, I doubt it's a good idea to do it by default, but packagers tend to use GDC/LDC with optimization flags.

PR #866, opened 2016-06-07
Pass all remaining args to the script for "dub [file].d" invocations.

### LINKED ISSUE #856: Pass all arguments to dub scripts
As probably the first tester of the new dub script feature I want to make a request:
Dub should send all arguments after the script to it, because

```
dub my-script.d --flag // doesn't work
./my-script --flag // doesn't work
```

of course I can use `dub my-script.d -- --flag`, but this feels more than unintuive.

@s-ludwig would you accept a PR that fixes this?

Related to #805.

s-ludwig:
Fixes #856.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -148,7 +148,9 @@ int runDubCommandLine(string[] args)
 	if (args.length >= 1 && !args[0].startsWith("-")) {
 		if (args[0].endsWith(".d")) {
 			cmdname = "run";
-			args = "--single" ~ args;
+			if (app_args.length) app_args = args[1 .. $] ~ "--" ~ app_args;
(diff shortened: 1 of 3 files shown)


s-ludwig:
@Geod24: Do you agree with this, or do you still have doubts or maybe an alternative?

Geod24:
> However, this syntax is already there primarily to support shebang scripts, and AFAIK this is the only portable way to be able to pass parameters to the application.

I think that's a compelling argument, so the idea LGTM.
Could this be documented anywhere though ?

s-ludwig:
Yes, this definitely needs to be documented. I'll put a section at http://code.dlang.org/getting_started for now.

PR #869, opened 2016-06-08
allow inline dub single file package declaration anywhere in file

rjmcguire:
In case you agree with my argument about the position of the /+
BTW I made the ':' optional in the regex, not sure you'd want that, I consider the \n to be the end filename marker.

To support all comment styles we'd have to implement all the comment parsing part of D.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -281,24 +281,17 @@ class Dub {
 	{
 		import dub.recipe.io : parsePackageRecipe;
 		import std.file : mkdirRecurse, readText;
+		import std.regex : ctRegex, matchFirst;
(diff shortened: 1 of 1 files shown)


s-ludwig:
```
[INFO] Running /home/travis/build/dlang/dub/test/issue103-single-file-package.sh...
hello_world ~master: building configuration "application"...
Linking...
Running ./hello_world foo -- bar
Hello, World!
Missing /+ ... +/ recipe comment.
Script failure.
```

I wonder if it makes sense to combine this with the added support for `/* */`/`//`. But that could also just be added to the regex (with missing support for parsing nested `/+ +/`).

I'd opt to use a runtime regex, though, as `ctRegex` incurred a huge compile time hit last time I checked.

PR #870, opened 2016-06-08
Improve single file packages

s-ludwig:
See https://forum.dlang.org/post/mailman.3032.1465413646.26339.digitalmars-d-announce@puremagic.com


s-ludwig:
I'll add a test, but let me slightly defer that, so that the release schedule isn't affected.

PR #871, opened 2016-06-08
allow any style comments for single file packages

rjmcguire:
Not extensively tested but it works for a few tests I tried. either + or * comments and anywhere in file.
I'm still not convinced it worth having so much more code in here but the feature will make at least 3 people happier.


rjmcguire:
where is cancel pull request.

s-ludwig:
You can usually just let the PR stay open and simply reset/rebase/squash the existing branch until it's ready.

PR #872, opened 2016-06-08
only consider first level comments * or + comments.

rjmcguire:


source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
 
 		The script above can be invoked with "dub --single test.d".
 	*/
+		struct Parser {
+			static typeof(this) opCall(string sc) {
```
s-ludwig: Interesting that `typeof(this)` works in contexts where there is no `this`...

source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
 
 		The script above can be invoked with "dub --single test.d".
 	*/
+		struct Parser {
```
s-ludwig: The Parser struct should be put into `dub.internal.utils` or similar.

s-ludwig: `DCommentParser` would then also make sense as a name to make it more clear what it does.

source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
...
+			string front;
+			bool empty;
+			char inComment = '\0';
+			bool hadFileName;
+			bool hadData;
+			int depth;
+			void popFront() {
+				import std.uni : isWhite;
+				if (empty) return;
+				for (;sourcecode.length > 0;) {
```
s-ludwig: I think `while()` would be more clear here.

source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
...
+					if (sourcecode[0].isWhite) {
+						sourcecode = sourcecode[1..$];
```
s-ludwig: It maybe doesn't matter in practice, but since this compares a single byte, `std.ascii : isWhite` would be the right choice here. A value greater than `0x7F` could be erroneously matched as a whitespace character (a non-breaking space for example).

source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
...
+					empty = true;
+					return;
+				}
+
+				int i,j;
+				for (; i<sourcecode.length; i++) {
+					if (sourcecode[i] == '/') {
+						j=i;
+						if (inComment=='\0') {
+							j = i;
```
s-ludwig: This line is redundant with the one on l. 312.

source/dub/dub.d
s-ludwig: Oh sorry, I see now that this also checks the comment contents. But I guess that we could just separate the two using an additional `.filter!(...)` call and keep this struct to do only the generic comment parsing.

source/dub/dub.d
```diff
@@ -277,28 +277,104 @@ class Dub {
...
+							i++;
+							if (sourcecode[i] == '+' || sourcecode[i]=='*') {
+								depth++;
```
s-ludwig: Only `/+ +/` supports nesting. `/* /* */ */` would be parsed as `/* /* */`, `*`, `/` by the compiler.

s-ludwig:
The parser definitely needs some unit tests. Some that come to mind after going through the code:

    /* foo: */

-> nothing

    /+ foo: +/

-> nothing

    /* /* */ /* dub.json: {} */

-> valid

    /+ /+ +/ /+ dub.json: {} +/

-> invalid

    /+ /+ +/ +/ /+ dub.json: {} +/

-> valid

    // dub.json:
    // {}

-> valid

    /** dub.json: */

-> invalid, should probably emit a warning that doc comment syntax is not allowed for recipe comments

    /++ dub.json: */

-> invalid

    /// dub.json:

-> invalid

source/dub/dub.d
rjmcguire: On Thu, Jun 9, 2016 at 8:15 AM, Sönke Ludwig <notifications@github.com>
wrote:

> In source/dub/dub.d
> <https://github.com/dlang/dub/pull/872#discussion_r66387841>:
>
> > +					}
> > +				}
> > +				if (sourcecode.length <= 0) {
> > +					empty = true;
> > +					return;
> > +				}
> > +
> > +				int i,j;
> > +				for (; i<sourcecode.length; i++) {
> > +					if (sourcecode[i] == '/') {
> > +						j=i;
> > +						if (inComment=='\0') {
> > +							j = i;
> > +							i++;
> > +							if (sourcecode[i] == '+' || sourcecode[i]=='*') {
> > +								depth++;
>
> Only /+ +/ supports nesting. /* /* */ */ would be parsed as /* /* */, *, /
> by the compiler.
>
>
Its not really supposed to be a full tokenizer or comment parser. It uses
quick hacks to identify the correct location of the package spec.
Perhaps I should make that more obvious?

s-ludwig: That's one of the issues of allowing the comment to be anywhere in the file. We now basically need a proper D lexer (strings really would have to be parsed, too). If it's required to be the first comment, it's unlikely to hit something such as `"/*"` before the recipe comment.

rjmcguire: On Thu, Jun 9, 2016 at 8:15 AM, Sönke Ludwig <notifications@github.com>
wrote:

> In source/dub/dub.d
> <https://github.com/dlang/dub/pull/872#discussion_r66387841>:
>
> > +					}
> > +				}
> > +				if (sourcecode.length <= 0) {
> > +					empty = true;
> > +					return;
> > +				}
> > +
> > +				int i,j;
> > +				for (; i<sourcecode.length; i++) {
> > +					if (sourcecode[i] == '/') {
> > +						j=i;
> > +						if (inComment=='\0') {
> > +							j = i;
> > +							i++;
> > +							if (sourcecode[i] == '+' || sourcecode[i]=='*') {
> > +								depth++;
>
> Only /+ +/ supports nesting. /* /* */ */ would be parsed as /* /* */, *, /
> by the compiler.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/dlang/dub/pull/872/files/6d6ee0e79d1f8a8ab1bc2beff4ea760963d618d4#r66387841>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABU8Ca5R1Xh308M9nUskZ6LL2lVMDbrqks5qJ69xgaJpZM4IxbiW>
> .
>
At the moment this will just let the compiler show the error to the user.

rjmcguire:
I have updated the parser to show it is not really a D comment parser.
Added some unit tests based on those you provided.

`/+ /+ +/ /+ dub.json: {} +/` still shows as valid even though it shouldn't:
This means user cannot comment out a single package description using nested comments.
I don't handle single line comments _at all_ so they won't be valid, and that is also how one can comment out a package description.

Would be good to have a warning about documentation style comments, not done yet.

PR #875, opened 2016-06-11
Choose default compiler using path based priority.

KevinBrogan:
Old Code: Do a compiler first scan, stopping when one is found, which results in dmd always being the default compiler.

New Code: Do a path scan first, stopping when a compiler is found, and use that compiler instead.

Result: The default compiler selected by dub will depend on how your path is setup, so that the first one found is selected, as oposed to always choosing compilers in a pre determined order.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1052,10 +1052,11 @@ class Dub {
 		version (Posix) enum sep = ":", exe = "";
 
 		auto compilers = ["dmd", "gdc", "gdmd", "ldc2", "ldmd2"];
+		typeof(compilers[0]) found;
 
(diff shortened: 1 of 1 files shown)


JohanEngelen:
@KevinBrogan Could you have a look at #895 ? Does it make sense to search dub's own directory first (before using the system path)?

s-ludwig:
With #895 being resolved by searching for a configuration file beneath the executable, I'm wondering if this is a good idea. Is selecting programs based on PATH order common practice? My feeling is that this is rather obscure from the outside and using the "defaultCompiler" configuration field as an explicit alternative is already possible.

PR #878, opened 2016-06-15
Enforce that recipe comments are the first thing in the source file.

s-ludwig:
Apart from the optional shebang line...

This is a temporary replacement for #872.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -285,17 +285,27 @@ class Dub {
 		path = makeAbsolute(path);
 
 		string file_content = readText(path.toNativeString());
-		auto idx = file_content.indexOf("/+");
-		enforce(idx >= 0, "Missing /+ ... +/ recipe comment.");
-		file_content = file_content[idx+2 .. $];
-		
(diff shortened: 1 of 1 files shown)


s-ludwig:
@rjmcguire: I've made this alternative to keep the changes minimal for now and so that we can work on #872 without having to fit into the 1.0.0 release schedule (which is already overdue).

rjmcguire:
Thanks, I had sorted out the problems with both my implementations but I can't figure out how to update my pull-requests, think I may have messed up my local repo by not branching before doing my changes.

s-ludwig:
While working on it, you can just add additional commits to the branch that is used for the PR and those will appear automatically. Once everything is finished, you can use rebase/squash to merge the commits together. Alternatively, if the PR is supposed to be a single commit anyway, `git commit --amend` will overwrite the latest commit instead of adding a new one, so that no rebase is necessary at the end. In both cases, you'll have to push the changes with `--force`, to force overwriting the commit history of the branch on github.

s-ludwig:
Merging for rc.1.

PR #879, opened 2016-06-17
interactive package removal

MartinNowak:
- ask user which version of a package to remove

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1213,6 +1213,10 @@ class InstallCommand : FetchCommand {
 }
 
 class RemoveCommand : FetchRemoveCommand {
+	private {
+		bool m_nonInteractive;
+	}
+
 	this()
 	{
 		this.name = "remove";
@@ -1226,6 +1230,7 @@ class RemoveCommand : FetchRemoveCommand {
 	override void prepare(scope CommandArgs args)
 	{
 		super.prepare(args);
+		args.getopt("n|non-interactive", &m_nonInteractive, ["Don't enter interactive mode."]);
 	}
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
@@ -1236,7 +1241,29 @@ class RemoveCommand : FetchRemoveCommand {
 		auto package_id = free_args[0];
 		auto location = dub.defaultPlacementLocation;
 
-		dub.remove(package_id, m_version, location, m_forceRemove);
+		size_t resolveVersion(in Package[] packages) {
+			writeln("Select version of '", package_id, "' to remove from location '", location, "':");
+			foreach(i, pack; packages)
+				writeln(i, ". ", pack.version_);
+			writeln(packages.length, ". ", "all versions");
+			while (true) {
+				writef("> ");
+				auto inp = readln();
+				if (!inp.length) // Ctrl+D
+					return size_t.max;
+				if (inp.length > 1) {
+					try {
+						immutable selection = inp[0 .. $ - 1].to!size_t;
+						if (selection <= packages.length)
+							return selection;
+					} catch (ConvException e) {
+					}
+					logError("Please enter a number between 0 and %s.", packages.length);
+				}
+			}
+		}
+
+		dub.remove(package_id, m_version, location, m_forceRemove, m_nonInteractive ? null : &resolveVersion);
 		return 0;
 	}
 }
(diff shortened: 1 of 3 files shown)


source/dub/commandline.d
```diff
@@ -1226,6 +1230,7 @@ class RemoveCommand : FetchRemoveCommand {
 	override void prepare(scope CommandArgs args)
 	{
 		super.prepare(args);
+		args.getopt("n|non-interactive", &m_nonInteractive, ["Don't enter interactive mode."]);
```
MartinNowak: Should maybe make a common option out of this.

MartinNowak:
```sh
Select version of 'dub' to remove from location 'user':
0. 0.9.20
1. 0.9.21
2. all versions
> 
```

MartinNowak:
Ping

source/dub/commandline.d
```diff
@@ -1236,7 +1241,29 @@ class RemoveCommand : FetchRemoveCommand {
 		auto package_id = free_args[0];
 		auto location = dub.defaultPlacementLocation;
 
-		dub.remove(package_id, m_version, location, m_forceRemove);
+		size_t resolveVersion(in Package[] packages) {
+			writeln("Select version of '", package_id, "' to remove from location '", location, "':");
+			foreach(i, pack; packages)
+				writeln(i, ". ", pack.version_);
```
ZombineDev: Nitpick - I like the following format better:

```
1) 0.9.20
2) 0.9.21
3) all versions
```
I.e. `)` instead of `.` and 1-based index.

So how about this:

```D
foreach (idx, option; packages.map!(p => p.version_.toString).chain(only("all versions")))
    writefln("%s) %s, idx + 1, option);
```

source/dub/commandline.d
```diff
@@ -1236,7 +1241,29 @@ class RemoveCommand : FetchRemoveCommand {
...
+			writeln(packages.length, ". ", "all versions");
+			while (true) {
+				writef("> ");
+				auto inp = readln();
+				if (!inp.length) // Ctrl+D
+					return size_t.max;
+				if (inp.length > 1) {
```
ZombineDev: No need for the if - the condition is always true.

ZombineDev:
```
if (inp.length > 1)
    continue;
```

Makes the flow more clear.

ZombineDev: BTW, would that work on Windows where newline is `\r\n`? Would `inp.length` equal 2 then?

s-ludwig: On Windows it also uses a single `\n`, but it's good I double checked now, because that assumption is also used for the interactive init code (a Windows CI test instance would of course be even better than double-checking, need to look into AppVeyor).

Another possibility to express this in a clear way:
```
inp = inp.stripRight('\n');
if (inp.length == 0) continue;
```

source/dub/dub.d
```diff
@@ -738,8 +738,10 @@ class Dub {
 				name/version.
 			force_remove = Forces removal of the package, even if untracked
 				files are found in its folder.
+			resolveVersion = Callback to select package version.
 	*/
-	void remove(string package_id, string version_, PlacementLocation location, bool force_remove)
+	void remove(string package_id, string version_, PlacementLocation location, bool force_remove,
+				scope size_t delegate(in Package[] packages) resolveVersion=null)
```
s-ludwig: What about adding a second overload instead, which removes the `version_` argument? The original one could then be implemented in terms of the resolver callback. Another change that might be worth it would be to change the callback to `bool delegate(in Package)`, so that it can opt to remove multiple packages at once, too.

Nitpick: "resolve_version" to match the naming convention of other parameters.

source/dub/commandline.d
ZombineDev: Thanks for clarifying. I would have checked that myself, but I was on the phone.

+1 for stripRight

source/dub/commandline.d
MartinNowak: I'm a huge fan or PRs for other PRs, in particular for such nitpicks.

MartinNowak:
All done

MartinNowak:
> Another change that might be worth it would be to change the callback to bool delegate(in Package), so that it can opt to remove multiple packages at once, too.

That is tricky b/c the command behaves differently when only a single version is present.
And both the interactive and non-interactive version would need to collect all packages for the choice/error message anyhow, so I don't think it's worth doing that.

source/dub/commandline.d
ZombineDev: Me too, but 95% of the time I do dlang PR reviews on my phone, which highly unusable for dev work :/

s-ludwig:
LGTM, thanks!

Issue #883, opened 2016-06-23
Interactive init always generates source/app.d file


s-ludwig:
Quoting from https://github.com/rejectedsoftware/vibe.d/issues/1520:

> interactive initialization generates source/app.d file. If in the middle of answering the questions, I decide to abort (i.e. ctrl-c), the source/app.d file is still there, and dub init now complains "The target directory already contains a 'source/' file. Aborting." The directory is not in a usable state.
>
> The source directory + app.d file should not be generated until interactive initialization is complete.

> Clarification: NO files should be written until interactive initialization is complete. Not just source/app.d.

PR #884, opened 2016-06-23
Call the recipe callback for initPackage before actually writing files to disk.

s-ludwig:
See #883.


ZombineDev:
Looks, good, but can you add a test?

s-ludwig:
Okay, test added.

Issue #887, opened 2016-06-29
single file package. A way to generate exe on project folder ?


Zardoz89:
I'm trying to use ```dub build --single my_program.d``` with a "script" program that I wrote. Previously, I was using rdmd, and now I'm trying to use dub 1.0.0, so I can access dub packages. 

I use "-v" option to see that dub puts my executable file inside of a mangled directory on the temporal files folder. I need to have the executable file on the same folder where the source file is, or at least on a predictable path, so I can copy it to /usr/local/bin/ . 

I try to set **targetPath** and **wordkingDirectory** to "/usr/local/bin" and to "$PACKAGE_DIR" without any success.

s-ludwig:
Currently single-file are packages always forced to be built with `--temp-build`. The main idea of this was to support D script files with shebang lines, but for pure "dub build" it indeed doesn't really make sense. The question is if it should only be enabled for the `dub somescript.d` syntax, or also for `dub run --single somescript.d` (and never for `dub build --single somescript.d`). I'd say it should only be enabled for the special shebang syntax, to keep the other two consistent with normal builds.

Zardoz89:
Perhaps  --temp-build not should be forced  if targetPath is defined.

PR #888, opened 2016-06-29
Only use temporary builds shebang invocations.

s-ludwig:
See issue #887 - now builds single-file packages normally in the package's parent directory unless the `dub file.d` syntax is used.

### DIFF
--- a/test/issue103-single-file-package.sh
+++ b/test/issue103-single-file-package.sh
@@ -1,6 +1,20 @@
 #!/bin/sh
 set -e
 cd ${CURR_DIR}
-./issue103-single-file-package.d foo -- bar
+rm -f single-file-test
+
 ${DUB} run --single issue103-single-file-package-json.d --compiler=${DC}
-${DUB} issue103-single-file-package-w-dep.d --compiler=${DC}
+if [ ! -f single-file-test ]; then
+	echo "Normal invocation did not produce a binary in the current directory"
+	exit 1
+fi
+rm single-file-test
+
+./issue103-single-file-package.d foo -- bar
+
+${DUB} issue103-single-file-package-w-dep.d
+
+if [ -f single-file-test ]; then
+	echo "Shebang invocation produced binary in current directory"
+	exit 1
+fi
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -100,7 +100,7 @@ int runDubCommandLine(string[] args)
 
 	// special single-file package shebang syntax
 	if (args.length >= 2 && args[1].endsWith(".d")) {
-		args = args[0] ~ ["run", "-q", "--single", args[1], "--"] ~ args[2 ..$];
+		args = args[0] ~ ["run", "-q", "--temp-build", "--single", args[1], "--"] ~ args[2 ..$];
 	}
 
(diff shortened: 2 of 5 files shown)


test/issue103-single-file-package.sh
```diff
@@ -1,6 +1,18 @@
...
 ./issue103-single-file-package.d foo -- bar
+if [ -f single-file-test ]; then
+	echo "Shebang invocation produced binary in current directory"
+	exit 1
+fi
+
 ${DUB} run --single issue103-single-file-package-json.d --compiler=${DC}
-${DUB} issue103-single-file-package-w-dep.d --compiler=${DC}
+if [ ! -f single-file-test ]; then
+	echo "Normal invocation did not produce a binard in the current directory"
```
ZombineDev: Small typo: binard -> binary

test/issue103-single-file-package.sh
```diff
@@ -1,6 +1,18 @@
...
+	exit 1
+fi
+rm single-file-test
+
+${DUB} issue103-single-file-package-w-dep.d
```
ZombineDev: Add a test for this one too?

test/issue103-single-file-package.sh
```diff
@@ -1,6 +1,18 @@
 #!/bin/sh
 set -e
 cd ${CURR_DIR}
+rm -f single-file-test
 ./issue103-single-file-package.d foo -- bar
+if [ -f single-file-test ]; then
+	echo "Shebang invocation produced binary in current directory"
+	exit 1
+fi
```
ZombineDev: IIUC, the binary will placed in the parent directory. Can you verify that with a test?

s-ludwig: It should be placed within the system's temp folder. But since the sub folder contains a hash of all build parameters, it's hard to do a positive check.

test/issue103-single-file-package.sh
s-ludwig: Fixed.

test/issue103-single-file-package.sh
s-ludwig: Since this is using the "shebang" short syntax, the test would basically be redundant with the first one. Do you see value in it anyway?

ZombineDev: If you move it just after the "shebang" one, the test will cover both.

ZombineDev:
BTW, could we allow the user to override the location by specifying `targetPath`, as mentioned in the issue? I also think that would be useful.

test/issue103-single-file-package.sh
ZombineDev: Sorry, I misunderstood. It's ok as it is.

test/issue103-single-file-package.sh
s-ludwig: Okay, I'll shuffle the tests around.

s-ludwig:
Setting the target path should work just like for normal packages. It just didn't work, because the "temporary" build mode was always active.

ZombineDev:
Thanks for the clarification and addressing my nitpicks. All LGTM

s-ludwig:
Okay, thanks for doing the review!

PR #897, opened 2016-07-12
Search for dub configuration in ../etc/dub/. Fixes #895.

### LINKED ISSUE #895: Search dub's own path first to find compiler
Would it be possible to have the dub executable search its own path first for the compiler? That way, dub will use the compiler it is packaged with. (note: this assumes dub is not on the path, but called e.g. like `~/ldc/ldc2-1.1.0-alpha1-osx-x86_64/bin/dub`)

s-ludwig:
The search happens relative to the executable file. This also contains two small refactorings to prepare for a separation of configuration files and cache files, which currently reside in the same folder.

#895

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -94,7 +94,7 @@ class Dub {
 		Path m_rootPath;
 		Path m_tempPath;
 		Path m_userDubPath, m_systemDubPath;
-		Json m_systemConfig, m_userConfig;
+		DubConfig m_config;
 		Path m_projectPath;
 		Project m_project;
 		Path m_overrideSearchPath;
@@ -133,19 +133,10 @@ class Dub {
 
 		PackageSupplier[] ps = additional_package_suppliers;
 
-		if (skip_registry < SkipPackageSuppliers.all) {
-			if (auto pp = "registryUrls" in m_userConfig)
-				ps ~= deserializeJson!(string[])(*pp)
-					.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url)))
-					.array;
-		}
-
-		if (skip_registry < SkipPackageSuppliers.all) {
-			if (auto pp = "registryUrls" in m_systemConfig)
-				ps ~= deserializeJson!(string[])(*pp)
-					.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url)))
-					.array;
-		}
+		if (skip_registry < SkipPackageSuppliers.all)
+			ps ~= m_config.registryURLs
+				.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url)))
+				.array;
 
 		if (skip_registry < SkipPackageSuppliers.standard)
 			ps ~= defaultPackageSuppliers();
@@ -183,8 +174,9 @@ class Dub {
 		}
 
 		m_tempPath = Path(tempDir);
-		m_userConfig = jsonFromFile(m_userDubPath ~ "settings.json", true);
-		m_systemConfig = jsonFromFile(m_systemDubPath ~ "settings.json", true);
+
+		m_config = new DubConfig(jsonFromFile(m_userDubPath ~ "settings.json", true), null);
+		m_config = new DubConfig(jsonFromFile(m_systemDubPath ~ "settings.json", true), m_config);
 
 		determineDefaultCompiler();
 	}
@@ -402,7 +394,7 @@ class Dub {
 				if (basename == rootbasename) continue;
 
 				if (!m_project.selections.hasSelectedVersion(basename)) {
-					logInfo("Package %s can be installed with version %s.",
+					logInfo("Non-selected package %s is available with version %s.",
 						basename, ver);
 					any = true;
 					continue;
@@ -1057,10 +1049,7 @@ class Dub {
 	{
 		import std.process : environment;
 
-		m_defaultCompiler = m_userConfig["defaultCompiler"].opt!string();
-		if (m_defaultCompiler.length) return;
-
-		m_defaultCompiler = m_systemConfig["defaultCompiler"].opt!string();
+		m_defaultCompiler = m_config.defaultCompiler;
 		if (m_defaultCompiler.length) return;
 
 		version (Windows) enum sep = ";", exe = ".exe";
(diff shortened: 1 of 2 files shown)


source/dub/dub.d
```diff
@@ -172,19 +162,21 @@ class Dub {
...
 		}
 
-		m_tempPath = Path(tempDir);
-		m_userConfig = jsonFromFile(m_userDubPath ~ "settings.json", true);
-		m_systemConfig = jsonFromFile(m_systemDubPath ~ "settings.json", true);
+		m_dirs.temp = Path(tempDir);
+
+		m_config = new DubConfig(jsonFromFile(m_dirs.userSettings ~ "settings.json", true), null);
+		m_config = new DubConfig(jsonFromFile(m_dirs.systemSettings ~ "settings.json", true), m_config);
+		m_config = new DubConfig(jsonFromFile(Path(thisExePath) ~ "../etc/dub/settings.json", true), m_config);
```
JohanEngelen: If I understand correctly, this ordering means that first `../etc/dub/settings.json` is used, then `/etc/`, and then `~/.dub/`?

s-ludwig: Hm, yes indeed, I missed that. User and system should be in reverse.

source/dub/dub.d
```diff
@@ -172,19 +162,21 @@ class Dub {
...
 		}
 
-		m_tempPath = Path(tempDir);
-		m_userConfig = jsonFromFile(m_userDubPath ~ "settings.json", true);
-		m_systemConfig = jsonFromFile(m_systemDubPath ~ "settings.json", true);
+		m_dirs.temp = Path(tempDir);
+
+		m_config = new DubConfig(jsonFromFile(m_dirs.systemSettings ~ "settings.json", true), m_config);
+		m_config = new DubConfig(jsonFromFile(m_dirs.userSettings ~ "settings.json", true), null);
+		m_config = new DubConfig(jsonFromFile(Path(thisExePath) ~ "../etc/dub/settings.json", true), m_config);
```
JohanEngelen: `thisExePath` is a file, not a directory. This needs to be (something like):
```d
m_config = new DubConfig(jsonFromFile(Path(thisExePath).parentPath() ~ "../etc/dub/settings.json", true), m_config);
```
With this change, it works for me :-)

JohanEngelen:
With this https://github.com/dlang/dub/pull/897#discussion_r70489186 it works for me.

Perhaps an additional change could be made to dub's output message. With a settings.json: `{  "defaultCompiler" : "../bin/ldc2" }`
it outputs `Performing "debug" build using ../bin/ldc2 for x86_64.` Without `../bin` and perhaps with compiler version would be nicer imho.

JohanEngelen:
I think `~/.dub/` should have priority over `../etc/dub/settings.json`, such that the user can override any setting. When dub is installed in `/opt/bin/` then the config in `/opt/etc` cannot be overridden.

source/dub/dub.d
s-ludwig: I hoped to be able to avoid adding a test, but I'll do one at least for the `../etc/` case.

s-ludwig:
Hm, good point, it should probably have the lowest precedence of all. Setting a system-wide configuration property and not having an effect would be equally bad.

JohanEngelen:
> it should probably have the lowest precedence of all

Well, in that case it is not possible to download a new compiler package while a system compiler is installed too: even if the system compiler is not in the path, Dub will find the system settings.conf and use the system compiler...

s-ludwig:
Should the system configuration file even have a "defaultCompiler" set by default? The other question is if, in this situation, users would realistically even try to use `/path/to/custom/compiler/bin/dub` to invoke the custom compiler, or rather just `dub --compiler=...`.

JohanEngelen:
I don't know how these packaging/installation things usually work, sorry.
What I was thinking is that if we ship an LDC package with Dub in it, it'd be nice if that dub selected the LDC it is packaged with. So that would mean packaging `../etc/dub/settings.conf` with it (similarly there is an `../etc/ldc2.conf` in the package). You can unpack the package anywhere, and it just works. LDC searches for its ldc2.conf in this order (`.` and `..` relative to the executable): current working dir, `.`, `~/.ldc`, `../etc`, install-prefix/etc, install-prefix/etc/ldc, `/etc`, `/etc/ldc`.
If installation of LDC means "copy the package to system install path", then the "etc/dub/settings.conf" with "defaultCompiler" set will end up in the system settings.

I think it makes (some) sense for a local dub settings file to be able to override the system setting, and that user setting overrides all. I guess there is no perfect solution here.

JohanEngelen:
I think it is unavoidable to have system settings file with a "defaultCompiler" setting. Otherwise installing LDC will still default to DMD as a compiler (note that `../etc/dub/` will become the same as the system setting file upon installation).
With the order after your change just now (user file, system file, `../etc/dub` file), it will be impossible to have multiple packages on a system with a different default compiler setting per package (because of the, I think, unavoidable system defaultCompiler setting).

jacob-carlborg:
> Should the system configuration file even have a "defaultCompiler" set by default? The other question is if, in this situation, users would realistically even try to use /path/to/custom/compiler/bin/dub to invoke the custom compiler, or rather just dub --compiler=...

DVM will install the compiler in `~/.dvm/compilers/<vendor>-<version>`. The binaries will be placed in a sub directory `<platform>/bin/`. DVM then adds the binaries to the PATH. When switching compiler DVM will add the compiler to the beginning of `PATH` so the correct compiler is always found by Dub.

DVM doesn't currently support LDC, but if it did (when it does 😃), it would need to be able to have a configuration per compiler. Even though the correct version of LDC would be in the beginning of `PATH` if Dub still prefers DMD it will not use the correct compiler.

JohanEngelen:
@s-ludwig Did I convince you that `../etc/dub/` should have preference over `m_dirs.systemSettings` ? :)

s-ludwig:
Still not sure about the order, but I agree that there probably isn't a "correct" or "perfect" here. AFAICS, both, "User->System->Local" and "User->Local->System", will solve the issue for the downloaded package. For installed packages, I'd expect `dub` to be in a separate package in a non-compiler-specific form, so it wouldn't really matter there.

I'm really not really sure how to weigh the pros and cons personally, so I'll just follow the democratic :wink: way and change it to "User->Local->System".

JohanEngelen:
Thanks!

s-ludwig:
Does anyone see any issues left?

JohanEngelen:
@redstar It'd be nice if we could package Dub with LDC1.1.0 with this PR included. How do we go about that?

redstar:
@JohanEngelen I'll update the build script for release.

PR #898, opened 2016-07-12
trivial: Fix typos

ximion:
Found by automatic QA in Debian.


s-ludwig:
Thanks! Didn't know that it could handle D code.

ximion:
It just examines the binaries we ship with Debian, and the strings they contain :)
This fortunately is relatively language-agnostic.

Issue #820, opened 2016-04-19
"dub convert" inserts redundant fields


s-ludwig:
Since the source information for the conversion process is taken from the `Package` instance, automatically inferred information gets inserted into the output. Instead, the raw package recipe should be loaded.

PR #901, opened 2016-07-19
Fix "dub convert" adding extra fields - issue #820

s-ludwig:
This changes "dub convert" to use the original package recipe instead of the one that gets augmented with inferred configurations and settings.


andralex:
I have no idea what I'm doing :)

PR #905, opened 2016-07-27
Support create project for hunt framework

zoujiaqing:
Can use dub create hunt framework based project, example:

``` shell
dub init myproject -t hunt
```


andralex:
cc @s-ludwig

s-ludwig:
Yeah, we should prioritize making this user controlled instead. I'll have to search for the ticket, but the idea is to search for something like a special sub package of the first dependency (e..g. "hunt:init" or "hunt:example" in this case) and copy+modify that accordingly. Then all special knowledge of certain packages can be faded out.

andre2007:
Implemented with pr https://github.com/dlang/dub/pull/1658

PR #922, opened 2016-09-06
Link against the intermediate binary files of dependencies. Fixes #921.

### LINKED ISSUE #921: DUB scripts, need to call DUB two times to build dependencies
I've started to use this feature because it's perfect for libraries examples. However I've noticed that the first time one types `dub example.d`, the compilation fails because of a missing dependency (e.g with a message saying `folder/libstuff.a: no file or folder of this type`. The second time the dependency is build and it works. I'd expect the script to works _directly_ the first time.

s-ludwig:
This change causes the build generator to pass the path to the intermediate build directory (e.g. ".dub/build/*/libfoo.a") instead of the final target path. The major consequence is that temporary builds with dependencies now work (they won't copy the intermediate binary to the final target path).

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -52,6 +52,8 @@ class BuildGenerator : ProjectGenerator {
 
 		bool any_cached = false;
 
+		Path[string] target_paths;
+
 		bool[string] visited;
 		void buildTargetRec(string target)
 		{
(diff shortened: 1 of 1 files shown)


BBasile:
Ok , thanks, I confirm that your pull fix the issue.

PR #924, opened 2016-09-08
Fix purging of optional dependencies. Fixes #923.

### LINKED ISSUE #923: Fails to download dependencies in a sub package scenario
A scenario such as the following leads to non-optional packages staying unresolved and thus not getting downloaded:

```
package "main" depends on "a:bar"
package "a:bar" depends on "a"
package "a" depends on "a:foo"
package "a:foo" depends on "b"
```

-> Package "b" does not get resolved.

s-ludwig:


s-ludwig:
This simply makes sure that the whole package graph is visited while marking non-optional dependencies, in contrast to searching only a part of the whole package graph, by stopping as soon as the same *base* package is visited twice (i.e. two sub packages of the same base package). If nobody objects, I'll merge this, so that it can get into 1.1.0.

Issue #932, opened 2016-09-16
dub convert - need converted data in stdout


BBasile:
Currently in Coedit I use the result of `dub desribe` to get the JSON equivalent of a SDL. I could use the `convert` command however it replaces the existing description. What I would need is an option that would have for effect to just print the result without writing a file, e.g

`dub convert --format=json`

would be the same as

`dub convert --format=json --target=file`

and to get the result in the standard output:

`dub convert --format=json --target=stdout`

PR #933, opened 2016-09-21
Issue 932 - convert to stdout

s-ludwig:
See #932.


s-ludwig:
Yeah, "describe" has to do a bunch of things (dependency and configuration resolution, as well as determining the build environment properties probably being major ones) so for just getting general information about a package, this definitely makes sense.

BBasile:
Thank you.

PR #941, opened 2016-09-28
Expand '~' in paths.

### LINKED ISSUE #771: Fails to expand ~ in path
If I have a dependency in `dub.sdl` like this:

```
dependency "vibe-d" path="~/development/d/vibe.d"
```

Dub fails to expand `~` to the home directory. The error message I get is something like:

```
No package file found in /Users/jacob/development/d/foo/~/development/d/vibe.d/, expected one of dub.json/dub.sdl/package.json
```

rcorre:
Resolves #771: Fails to expand ~ in path.
For example, a ~ in a path dependency path in dub.json will now be
expanded.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -28,6 +28,7 @@ import std.conv;
 import std.datetime;
 import std.exception;
 import std.file;
+import std.path;
 import std.process;
 import std.string;
 import std.typecons;
@@ -1346,7 +1347,7 @@ final class SelectedVersions {
 		if (j.type == Json.Type.string)
 			return Dependency(Version(j.get!string));
 		else if (j.type == Json.Type.object)
-			return Dependency(Path(j["path"].get!string));
+			return Dependency(Path(j["path"].get!string.expandTilde));
 		else throw new Exception(format("Unexpected type for dependency: %s", j.type));
 	}
 


s-ludwig:
I'm not sure where is the best place\*, but putting it in `Path` like this has the drawback of breaking down once the optional vibe-d dependency is used (and thus the `Path` struct from vibe.d). So this either has to be moved down to DUB specific code, or something has to be added to vibe.d at the same time (I'd probably make this explicit and add something like a `Path.expand` method).

\* `dub.project.processVars` seems to be the most promising candidate at first sight.

rcorre:
`processVars` isn't hit during `dub upgrade`, I'm still digging around for another reasonable place to put the path expansion. I just realized `$HOME` works as dub expands environment variables so this isn't as annoying as I thought.

s-ludwig:
Wouldn't expanding tilde during "dub upgrade" only be useful for expanding path based dependencies that use it? But at least sing absolute paths in the package recipe is something that should be discouraged. Putting such paths in "dub.selections.json" is fine, though. A dedicated expansion call is definitely necessary there, though.

rcorre:
My use case was testing out a dependency that I had cloned and modified locally, so I think modifying dub.selections was the right thing to do. Still it night seem inconsistent to expand there but not dub.json

rcorre:
@s-ludwig: the new solution avoids modifying `Path` and expands `~` when reading `dub.selections.json` (but not `dub.json`.

s-ludwig:
This definitely makes sense. Should we merge this now as a first step, or do you want to look into it for the dependency part and for other paths of the recipe, too? My best bet would still be  `processVars` + somewhere for dependencies. In the case of dependencies, I'd add a warning message at the same time that it's discouraged to hard-code absolute paths.

rcorre:
Still looking into handling the recipe. I was pretty sure it failed before hitting `processVars`, but that doesn't make sense if `$HOME` worked ... I'll have to look again.

rcorre:
Never mind. Variables actually aren't expanded in the recipe, just `dub.selections.json`. I think that would require some more substantial reorganization so I'd just take this as-is for now.

Geod24:
@rcorre : Do you know if this is still an issue ?

rcorre:
@Geod24 I don't, I haven't used `dub` in a while and I was cleaning up old PRs. I'm still happy to merge it, but I've forgotten all the context around it.

PR #945, opened 2016-09-29
Use std.encoding in SDLang lexer

JollieRoger:
Replace deprecated and removed std.stream with std.encoding for BOM code.

### DIFF
--- a/source/dub/internal/sdlang/lexer.d
+++ b/source/dub/internal/sdlang/lexer.d
@@ -13,7 +13,7 @@ import std.bigint;
 import std.conv;
 import std.datetime;
 import std.file;
-import std.stream : ByteOrderMarks, BOM;
+import std.encoding : getBOM, BOM;
 import std.traits;
 import std.typecons;
 import std.uni;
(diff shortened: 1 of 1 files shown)


s-ludwig:
Unfortunately, this change will have to wait for a a relatively long time before it can actually replace the old code, because older frontend/Phobos versions still have to be supported. Wrapping the new code in `static if (__VERSION__ >= 2072)` should be fine, though.

It would be good to also have this change in [sdlang-d](https://github.com/Abscissa/SDLang-D/blob/master/src/sdlang/lexer.d) - it  appears to also still use the old symbols (without a `std.stream` import, though). I periodically pull changes from that repository, so the risk would otherwise be to accidentally lose this fix during such a merge.

s-ludwig:
Just saw that sdlang-d has its [own version](https://github.com/Abscissa/SDLang-D/blob/ea90f434541d32fef9dfc5c05e2d7a3c41111a87/src/sdlang/util.d#L184-L200) of the std.stream symbols now. I'll just merge the latest version - then there is also no need for `static if` workarounds. Thank you for taking the time to open a PR anyway!

PR #949, opened 2016-10-04
Upgrade sdlang-d.

s-ludwig:
This supersedes #945.

### DIFF
--- a/build-files.txt
+++ b/build-files.txt
@@ -32,6 +32,7 @@ source/dub/internal/sdlang/parser.d
 source/dub/internal/sdlang/symbol.d
 source/dub/internal/sdlang/token.d
 source/dub/internal/sdlang/util.d
+source/dub/internal/taggedalgebraic.d
 source/dub/internal/utils.d
 source/dub/internal/vibecompat/core/file.d
 source/dub/internal/vibecompat/core/log.d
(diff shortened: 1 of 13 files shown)


s-ludwig:
Looks like this isn't easily portable to anything below DMD 2.067.0. Since this would currently rule out GDC (The available releases still appear to be 2.066 front end based), I guess this should be put on hold for a while.

wilzbach:
> Looks like this isn't easily portable to anything below DMD 2.067.0. Since this would currently rule out GDC (The available releases still appear to be 2.066 front end based), I guess this should be put on hold for a while.

That sucks :/
May I stupidly ask whether maintaining compatibility with GDC is really worth the pain? 
LDC has done a great job at catching up lately and they now also support building completely statically linked binaries (one of the few reasons for GDC AFAIK).

s-ludwig:
Well, at least I think we don't lose anything right now by not upgrading. All bug fixes or improvements that I noticed are not relevant for DUB.

It's sad that GDC's development cannot really keep up since a while, but after all, GDC is still in the Debian repositories, so may not be the worst idea to at least enable building DUB, even if most packages will probably fail.

wilzbach:
> Looks like this isn't easily portable to anything below DMD 2.067.0. 

FYI 2.068.0 is now the oldest supported platform, so it might be possible to revive this?

Geod24:
I think we have drifted far enough that, should you want to revive this, it could use a new PR.

PR #956, opened 2016-10-06
Remove some duplicated code, implement platform probing for LDC.

epi:
The repeated code was moved from gdc and dmd implementations to a protected final method of their base `interface Compiler`. `LDCCompiler.determinePlatform` was implemented using the same method.
I've tested it with dmd and a few different builds of gdc and ldc, but only on linux x86_64 build system (x86, x86_64 and various arm targets).


s-ludwig:
Looks good to merge! Thanks!

PR #961, opened 2016-10-09
Changes for operation with enviroment variables

MGWL:
I very need a possibility of installation of a envoroment variable for application  directly from dub. I suggest to consider a question of introduction of the following change for a possibility of setting enviroment variables from dub.

```
   . . .
    "configurations": [
        {
            "platforms": [  "linux" ],
            "workingDirectory": "exampleLinux",
            "name": "exampleLinux",
            "targetType": "executable",
            "targetName": "exampleLinux",
            "enviromentCommands": [
                "LD_LIBRARY_PATH=$PACKAGE_DIR/exampleLinux",
                "MY_VARIABLE=String",
            ],
            "targetPath": "exampleLinux",
            "mainSourceFile": "example/example.d"
        }
    ],
   . . .
```

I ask to consider this change. Without it there is no possibility of full application launch from dub.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -102,6 +102,7 @@ int runDubCommandLine(string[] args)
 	if (args.length >= 2 && args[1].endsWith(".d")) {
 		args = args[0] ~ ["run", "-q", "--temp-build", "--single", args[1], "--"] ~ args[2 ..$];
 	}
+
 	// split application arguments from DUB arguments
 	string[] app_args;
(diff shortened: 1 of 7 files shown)


wilzbach:
> Without it there is no possibility of full application launch from dub.

Could you maybe explain this in more details?
Why can't your application use default values for environment variables or detect them?

Geod24:
Closing due to lack of response from the author.

PR #963, opened 2016-10-11
Always use -singleobj when invoking LDC + drop -od/-oq switches.

s-ludwig:
See #938. This is also a workaround for ldc-developers/ldc#1819, but currently testing LDC 1.1.0-beta3 doesn't work through TravisCI, so this yet needs to get verified.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ sudo: false
 matrix:
   include:
     - d: dmd-beta
-      env: [FRONTEND=2.071]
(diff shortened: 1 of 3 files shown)


kinke:
Quoting from https://github.com/ldc-developers/ldc/issues/1819#issuecomment-253262297:

> However, it seems like building a static library from a single .o would always cause all the modules in that library to be linked in (so all the static constructors to be run, etc.), which is undesirable.

PR #962, opened 2016-10-10
Fix collecting path based sub package + base package dependencies. Fixes #959.

s-ludwig:
Having a sub package dependency followed by a dependency to the base package of the sub package incorrectly tried to obtain the base package as a sub package with an empty name.

PR #969, opened 2016-10-24
Fix resolution of path based dependencies when no selections exist yet. Fixes #934.

### LINKED ISSUE #934: wrong dependency solving of local packages
If I create 3 projects A, B, C, and use A in B, A and B in C I get strange error: Sub package a: doesn't exist. But if I don't write dependency A in dub.sdl in C project all builds successful.

packA/dub.sdl:
```
name "packa"
targetType "library"
```
packA/source/packa.d:
```d
module packa;
float superFunc(float a, float b) { return a + b; }
```
packB/dub.sdl:
```
name "packb"
dependency "packa" path="../packA"
targetType "library"
```
packB/source/packb.d:
```d
module packb;
import packa;
float someCode(float a, float b) { return superFunc(a,b) * a; }
```
packC/dub.sdl:
```
name "packc"
dependency "packa" path="../packA" // if comment it out all builds and works fine
dependency "packb" path="../packB"
targetType "executable"
```
packC/source/app.d:
```d
import std.stdio;

import packa;
import packb;

void main()
{
    writeln(superFunc(3,4));
    writeln(someCode(3,4));
}
```

dub from master
```
dub --version
DUB version 1.0.0, built on Sep 23 2016
```

s-ludwig:


MrSmith33:
Checked. Issue is resolved.

s-ludwig:
Just noticed that this has in fact already been fixed by #962. I tested against master instead of stable, so I missed that at first. I'll still merge this, so that the additional regression test is added.

PR #971, opened 2016-10-25
[trivial] Fix bug with ldmd2

9il:
-L= works with with ldc2 only. -L works both with ldc2 and ldmd2

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -221,7 +221,7 @@ class LDCCompiler : Compiler {
 
 	string[] lflagsToDFlags(in string[] lflags) const
 	{
-		return  lflags.map!(s => "-L="~s)().array();
+		return  lflags.map!(s => "-L"~s)().array();
 	}
 
 	private auto escapeArgs(in string[] args)


s-ludwig:
Do you have a test case for this? Invoking "ldmd2" *should* use the `DMDCompiler` class instead of the `LDCCompiler` one and thus should already pass `-L...`.

9il:
This fails https://travis-ci.org/libmir/mir-glas/builds/170611531#L2573 . Maybe this is linker bug (libmir-glas depends on libmir-cpuid). It is hard to test for me because i have not linux

s-ludwig:
The `-L.. -L-lname` approach doesn't seem to work, I get the same errors when building with DMD. What does work, interestingly, is to repeat the libraries twice: `"lflags": ["-L../", "-lmir-cpuid", "-lmir-glas", "-lmir-cpuid", "-lmir-glas"],`. A guess would be that this is an issue with the order in which the files are passed to the linker (i.e. correct order or `--no-as-needed` is required). Repeating the library might cause the linker to assume that all symbols that are referenced within the library are actually needed, but that's just a shot in the blue.

PR #974, opened 2016-10-27
update travis compilers

John-Colvin:
seems sensible to make it all specific seeing as you're specifying the frontend version anyway.

Also, moved coverage to latest stable dmd


John-Colvin:
huh, seems that moving the coverage caused a problem

John-Colvin:
once https://github.com/dlang/dub/pull/975 is merged, this should go green.

PR #976, opened 2016-10-27
Fix build

John-Colvin:
combines https://github.com/dlang/dub/pull/975 and https://github.com/dlang/dub/pull/974 because both are needed for travis to pass.


s-ludwig:
Thanks, looks good now. I missed that the old compiler versions stopped working when libdparse was fixed to work on the latest DMD beta.

PR #979, opened 2016-10-28
Fix defaultlib

John-Colvin:
replaces https://github.com/dlang/dub/pull/977 and https://github.com/dlang/dub/pull/978

The reason why defaultlib is necessary for dmd is that dmd always chooses static phobos by default, unless you specify defaultlib. Because ldc doesn't do this, the defaultlib flag does nothing useful there and causes problems if the library is renamed on a particular distro.

if we add defaultlib, we should prepend it, because otherwise the user can't override it (the last one is used, all others are ignored)


s-ludwig:
Makes sense!

Issue #980, opened 2016-10-28
dub --cache should be able to take a path


John-Colvin:
sometimes neither of the 3 options currently available are good enough.

~~Workaround: cd to the directory you want to use as a cache, use `--cache=local` and `--root=xxxx` to point dub to the right place.~~ That doesn't work, as the cache also seems to be moved by `root`

WebFreak001:
you can rely on the regular cache and use the new DUB_HOME or DPATH environment variable to control where the cache is put now.

PR #982, opened 2016-10-28
add cachePath option

John-Colvin:
to fix https://github.com/dlang/dub/issues/980

There are cleaner ways to implement this, but I wanted to minimize impact on the library API.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1120,6 +1120,9 @@ class Dub {
 			version(Windows) enum pathsep = ";";
 			else enum pathsep = ":";
 			if (p.length) paths ~= p.split(pathsep).map!(p => Path(p))().array();
+			if (!defaultPlacementPath_.empty) // e.g. cachePath was set
+				paths ~= defaultPlacementPath; // look there for packages as well
+
 			m_packageManager.disableDefaultSearchPaths = false;
 			m_packageManager.searchPath = paths;
 		}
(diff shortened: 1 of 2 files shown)


source/dub/dub.d
```diff
@@ -655,7 +678,7 @@ class Dub {
 		}
 
 		if (existing) {
-			if (!ver.startsWith("~") || !(options & FetchOptions.forceBranchUpgrade) || location == PlacementLocation.local) {
+			if (!ver.startsWith("~") || !(options & FetchOptions.forceBranchUpgrade)/* || location == PlacementLocation.local*/) {
```
John-Colvin: I commented this out because I didn't want to have to deal with it for now, will fix it, but first perhaps someone could explain why this is special-cased for local?

John-Colvin: @s-ludwig This was introduced in https://github.com/dlang/dub/commit/2be278a19b74a5791886bfe8963551d865f27e0a, perhaps you could talk me through it?

John-Colvin:
have also made it so that `--cachePath` is used for searching as well as storage, which looks like a special case but actually gives you the same behaviour as the default user/system paths enjoy when they are specified with `--cache`

John-Colvin:
seems like I've been going about this the wrong way. I think what is actually needed is some way to create a new `Repository` and have a `PackageType.custom` or similar.

MartinNowak:
I find cache path a bit misleading, it's the local packages that are stored in there.

John-Colvin:
closed in favour of https://github.com/dlang/dub/pull/985

PR #985, opened 2016-11-04
rethink of caching and user/system repos

John-Colvin:
I recognise this is an absolute monster diff, so I'll do my best to put in a lot of comments in here to explain motivations soon.

If you like to break things down by commit: https://github.com/dlang/dub/commit/06a40361386cfd746aa152687d535ec50b00ead6, https://github.com/dlang/dub/commit/13d6ec41849b7680076bd5c0f6cb1e10413b23f6 and https://github.com/dlang/dub/commit/82d25b62a151021a2916f4fc9bf36e72046de324 are the places to look, the test suite passes for each of them individually. Not much to be gained by looking at the other two in isolation.

The changes are also broadly contained within each file: `packagemanager.d` first, then `dub.d`, then `commandline.d`. Alternatively, the opposite order might make the "why" clearer.

## The point:
adding support for custom directories for packages and local registration files.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -31,7 +31,7 @@ import std.datetime;
 import std.exception;
 import std.file;
 import std.process;
-import std.range : only;
+import std.range : chain, only;
 import std.string;
 import std.typecons;
 import std.zip;
@@ -148,7 +148,7 @@ class Dub {
 	}
 	/// ditto
 	this(Path root_path = Path("."), PackageSupplier[] additional_package_suppliers = null,
-			SkipPackageSuppliers skip_registry = SkipPackageSuppliers.none)
+			SkipPackageSuppliers skip_registry = SkipPackageSuppliers.none, Path[] extraRepoPaths = null)
 	{
 		m_rootPath = makeAbsoluteRelativeToCwd(root_path);
 
@@ -165,7 +165,8 @@ class Dub {
 			ps ~= defaultPackageSuppliers();
 
 		m_packageSuppliers = ps;
-		m_packageManager = new PackageManager(only(userRepoPath, systemRepoPath));
+		m_packageManager = new PackageManager(
+				chain(extraRepoPaths, only(userRepoPath, systemRepoPath)));
 		updatePackageSearchPath();
 	}
 
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -31,6 +31,7 @@ import std.datetime;
 import std.exception;
 import std.file;
 import std.process;
+import std.range : only;
 import std.string;
 import std.typecons;
 import std.zip;
@@ -97,6 +98,7 @@ class Dub {
 		Path m_projectPath;
 		Project m_project;
 		Path m_overrideSearchPath;
+		Path m_defaultRepoPath;
 		string m_defaultCompiler;
 	}
 
@@ -106,8 +108,22 @@ class Dub {
 		of the normal upgrade process are stored in a certain location. This is
 		how the "--local" and "--system" command line switches operate.
 	*/
-	PlacementLocation defaultPlacementLocation = PlacementLocation.user;
-
+	Path defaultRepoPath() @property
+	{
+		return m_defaultRepoPath;
+	}
+	/// Ditto
+	void defaultRepoPath(Path path) @property
+	{
+		path = makeAbsolute(path);
+		assert(m_packageManager.isManagedPath(path, false));
+		m_defaultRepoPath = path;
+	}
+	/// Ditto
+	deprecated defaultPlacementLocation(PlacementLocation location) @property
+	{
+		defaultRepoPath = toPath(location);
+	}
 
 	/** Initializes the instance for use with a specific root package.
 
@@ -141,7 +157,7 @@ class Dub {
 			ps ~= defaultPackageSuppliers();
 
 		m_packageSuppliers = ps;
-		m_packageManager = new PackageManager(m_dirs.userSettings, m_dirs.systemSettings);
+		m_packageManager = new PackageManager(only(userRepoPath, systemRepoPath));
 		updatePackageSearchPath();
 	}
 
@@ -149,13 +165,13 @@ class Dub {
 		loading a package.
 
 		This constructor corresponds to the "--bare" option of the command line
-		interface. Use 
+		interface. Use
 	*/
 	this(Path override_path)
 	{
 		init();
 		m_overrideSearchPath = override_path;
-		m_packageManager = new PackageManager(Path(), Path(), false);
+		m_packageManager = new PackageManager(cast(Path[])[], false);
 		updatePackageSearchPath();
 	}
 
@@ -172,6 +188,11 @@ class Dub {
 				m_dirs.userSettings = Path(getcwd()) ~ m_dirs.userSettings;
 		}
 
+		m_dirs.systemRepo = m_dirs.systemSettings ~ "packages/";
+		m_dirs.userRepo = m_dirs.userSettings ~ "packages/";
+
+		m_defaultRepoPath = m_dirs.userRepo;
+
 		m_dirs.temp = Path(tempDir);
 
 		m_config = new DubConfig(jsonFromFile(m_dirs.systemSettings ~ "settings.json", true), m_config);
@@ -179,6 +200,15 @@ class Dub {
 		m_config = new DubConfig(jsonFromFile(m_dirs.userSettings ~ "settings.json", true), m_config);
 
 		determineDefaultCompiler();
+	}
+
+	Path toPath(PlacementLocation location) const
+	{
+		final switch (location) {
+			case PlacementLocation.local: return m_rootPath;
+			case PlacementLocation.user: return m_dirs.userRepo;
+			case PlacementLocation.system: return m_dirs.systemRepo;
+		}
 	}
 
 	@property void dryRun(bool v) { m_dryRun = v; }
@@ -192,6 +222,12 @@ class Dub {
 		m_rootPath = root_path;
 		if (!m_rootPath.absolute) m_rootPath = Path(getcwd()) ~ m_rootPath;
 	}
+
+	/// Returns the user repository path (e.g. ~/.dub/)
+	@property Path userRepoPath() const { return m_dirs.userRepo; }
+
+	/// Returns the system repository path
+	@property Path systemRepoPath() const { return m_dirs.systemRepo; }
 
 	/// Returns the name listed in the dub.json of the current
 	/// application.
@@ -243,7 +279,7 @@ class Dub {
 
 		Single-file packages are D files that contain a package receipe comment
 		at their top. A recipe comment must be a nested `/+ ... +/` style
-		comment, containing the virtual recipe file name and a colon, followed by the 
+		comment, containing the virtual recipe file name and a colon, followed by the
 		recipe contents (what would normally be in dub.sdl/dub.json).
 
 		Example:
@@ -435,7 +471,7 @@ class Dub {
 			FetchOptions fetchOpts;
 			fetchOpts |= (options & UpgradeOptions.preRelease) != 0 ? FetchOptions.usePrerelease : FetchOptions.none;
 			fetchOpts |= (options & UpgradeOptions.forceRemove) != 0 ? FetchOptions.forceRemove : FetchOptions.none;
-			if (!pack) fetch(p, ver, defaultPlacementLocation, fetchOpts, "getting selected version");
+			if (!pack) fetch(p, ver, defaultRepoPath, fetchOpts, "getting selected version");
 			if ((options & UpgradeOptions.select) && p != m_project.rootPackage.name) {
 				if (ver.path.empty) m_project.selections.selectVersion(p, ver.version_);
 				else {
@@ -614,8 +650,8 @@ class Dub {
 		if (existsFile(path ~ ".dub/obj")) rmdirRecurse((path ~ ".dub/obj").toNativeString());
 	}
 
-	/// Fetches the package matching the dependency and places it in the specified location.
-	Package fetch(string packageId, const Dependency dep, PlacementLocation location, FetchOptions options, string reason = "")
+	/// Fetches the package matching the dependency and places it in the specified repository
+	Package fetch(string packageId, const Dependency dep, Path repoPath, FetchOptions options, string reason = "")
 	{
 		Json pinfo;
 		PackageSupplier supplier;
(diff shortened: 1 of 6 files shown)


source/dub/packagemanager.d
```diff
@@ -780,15 +719,178 @@ private enum LocalOverridesFilename = "local-overrides.json";
...
 	Path[] searchPath;
 	Package[] localPackages;
 	PackageOverride[] overrides;
 
 	this(Path path)
 	{
-		this.path = path;
-		this.packagePath = path ~"packages/";
+		this.packagePath = path;
+	}
```
John-Colvin: The decision to use "packages/" as a subdir for holding downloaded packages is now localised in `Dub.init()` (which should definitely be renamed, something for another day...). With that done there doesn't seem to be any point in `Repository.path`.

source/dub/packagemanager.d
```diff
@@ -624,128 +661,30 @@ class PackageManager {
...
-		loadOverrides(LocalPackageType.user);
-		loadOverrides(LocalPackageType.system);
 	}
 
-	alias Hash = ubyte[];
+	deprecated alias Hash = ubyte[];
 	/// Generates a hash value for a given package.
 	/// Some files or folders are ignored during the generation (like .dub and
 	/// .svn folders)
-	Hash hashPackage(Package pack)
```
John-Colvin: seemed strange to have this here, seeing as it only applies to a single package, so it's now a member of `Package`. Kept a `deprecated` shim here for compatibility seeing as it was `public`.

source/dub/packagemanager.d
```diff
@@ -624,128 +661,30 @@ class PackageManager {
...
-				sha1.put(openFile(Path(file.name)).readAll());
-				logDebug("Hashed file contents from %s", Path(file.name).head);
-			}
-		}
-		auto hash = sha1.finish();
-		logDebug("Project hash: %s", hash);
-		return hash[].dup;
-	}
-
-	private void writeLocalPackageList(LocalPackageType type)
```
John-Colvin: now a member of `Repository`

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+		repoPath.normalize();
+		m_repos[repoPath].removeSearchPath(path);
 	}
 
 	void refresh(bool refresh_existing_packages)
 	{
 		logDiagnostic("Refreshing local packages (refresh existing: %s)...", refresh_existing_packages);
 
 		// load locally defined packages
-		void scanLocalPackages(LocalPackageType type)
```
John-Colvin: There's a big diff here because of adding a level of indent. See individual comments for actual changes.

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
-
-							Package pp;
-							if (!refresh_existing_packages) {
-								foreach (p; m_repositories[type].localPackages)
-									if (p.path == path) {
-										pp = p;
-										break;
-									}
-							}
+		if (!m_disableRepoSearchPaths)
```
John-Colvin: moved this check outside as you don't do anything at all if `m_disableRepoSearchPaths == true`

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+			over_repos: foreach (ref repo; m_repos)
```
John-Colvin: loop instead of local function.

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+			{
+				Path list_path = repo.packagePath;
+				Package[] packs;
+				Path[] paths;
+				try {
+					auto local_package_file = list_path ~ LocalPackagesFilename;
+					logDiagnostic("Looking for local package map at %s", local_package_file.toNativeString());
+					if( !existsFile(local_package_file) ) continue over_repos;
```
John-Colvin: making sure to continue the loop instead of returning! (A lot of time debugging my stupidity here...)

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+	void addSearchPath(Path path, Path repoPath)
+	in { assert(repoPath.absolute); }
+	body
+	{
+		repoPath.normalize();
+		m_repos[repoPath].addSearchPath(path);
 	}
 
 	/// Removes a search path from the given type.
-	void removeSearchPath(Path path, LocalPackageType type)
```
John-Colvin: now a member of `Repository`.

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+	void removeLocalPackage(Path path, Path repoPath)
+	in { assert(repoPath.absolute); }
+	body
+	{
+		repoPath.normalize();
+		m_repos[repoPath].removeLocalPackage(path);
 	}
 
 	/// For the given type add another path where packages will be looked up.
-	void addSearchPath(Path path, LocalPackageType type)
```
John-Colvin: The work is now done as member of Repository, but there's a Path-based function here to refer to the right member of m_repos.

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
...
+	}
+	Package addLocalPackage(Path path, string verName, Path repoPath)
+	in { assert(repoPath.absolute); }
+	body
+{
+		repoPath.normalize();
+		return m_repos[repoPath].addLocalPackage(path, verName);
 	}
 
-	void removeLocalPackage(Path path, LocalPackageType type)
```
John-Colvin: The work is now done as member of Repository, but there's a Path-based function here to refer to the right member of m_repos.

source/dub/packagemanager.d
```diff
@@ -445,138 +494,126 @@ class PackageManager {
 		logInfo("Removed package: '"~pack.name~"'");
 	}
 
-	Package addLocalPackage(Path path, string verName, LocalPackageType type)
```
John-Colvin: The work is now done as member of Repository, but there's a Path-based function here to refer to the right member of m_repos.

source/dub/packagemanager.d
```diff
@@ -297,41 +325,62 @@ class PackageManager {
...
+	}
+	/// ditto
+	deprecated void addOverride(LocalPackageType scope_, string package_, Dependency version_spec, Path target)
 	{
-		m_repositories[scope_].overrides ~= PackageOverride(package_, version_spec, target);
-		writeLocalPackageOverridesFile(scope_);
+		addOverride(m_defaultPaths[scope_], package_, version_spec, target);
 	}
 	/// ditto
-	void addOverride(LocalPackageType scope_, string package_, Dependency version_spec, Path target)
```
John-Colvin: The work is now done as member of Repository, but there's a Path-based function here to refer to the right member of m_repos.

source/dub/packagemanager.d
```diff
@@ -297,41 +325,62 @@ class PackageManager {
 		return &iterator;
 	}
 
-
 	/** Returns a list of all package overrides for the given scope.
 	*/
-	const(PackageOverride)[] getOverrides(LocalPackageType scope_)
```
John-Colvin: The work is now done as member of Repository, but there's a Path-based function here to refer to the right member of m_repos.

source/dub/packagemanager.d
```diff
@@ -243,15 +260,26 @@ class PackageManager {
...
+		Passing allowSubDirs = true will cause only the roots of the managed
+		folders to be matched
 	*/
-	bool isManagedPath(Path path)
-	const {
-		foreach (rep; m_repositories) {
-			auto rpath = rep.packagePath;
-			if (path.startsWith(rpath))
-				return true;
+	bool isManagedPath(Path path, bool allowSubDirs = true)
```
John-Colvin: added a quick path for checking if a path is actually a root of a repo.

source/dub/packagemanager.d
```diff
@@ -55,12 +74,14 @@ class PackageManager {
 	/// ditto
 	@property const(Path)[] searchPath() const { return m_searchPath; }
 
+	deprecated alias disableDefaultSearchPaths = disableRepoSearchPaths;
```
John-Colvin: `default` doesn't seem right now it applies to all repos including custom ones.

source/dub/packagemanager.d
```diff
@@ -30,23 +30,42 @@ import std.zip;
 /// packages.
 class PackageManager {
 	private {
-		Repository[LocalPackageType] m_repositories;
+		Repository[Path] m_repos;
```
John-Colvin: This is the fundamental change. There are now an unlimited number of repositories, so we index them by path instead of an `enum`. Indexing by path is a little scary, which is why the `public` interface of this class has contracts enforcing absolute paths and normalised paths before using them.

source/dub/dub.d
```diff
@@ -1106,8 +1217,9 @@ class Dub {
 		m_defaultCompiler = res.empty ? compilers[0] : res.front;
 	}
 
-	private Path makeAbsolute(Path p) const { return p.absolute ? p : m_rootPath ~ p; }
-	private Path makeAbsolute(string p) const { return makeAbsolute(Path(p)); }
+	private Path makeAbsoluteRelativeToRoot(Path p) const { return p.absolute ? p : m_rootPath ~ p; }
+	private Path makeAbsoluteRelativeToRoot(string p) const { return makeAbsoluteRelativeToRoot(Path(p)); }
+	private static Path makeAbsoluteRelativeToCwd(Path p) { return p.absolute ? p : getWorkingDirectory() ~ p; }
```
John-Colvin: seemed good to be really explicit what the reference point is for making absolute paths.

source/dub/dub.d
```diff
@@ -1075,17 +1187,16 @@ class Dub {
...
-			Path[] paths;
-
 			version(Windows) enum pathsep = ";";
 			else enum pathsep = ":";
-			if (p.length) paths ~= p.split(pathsep).map!(p => Path(p))().array();
-			m_packageManager.disableDefaultSearchPaths = false;
-			m_packageManager.searchPath = paths;
+			m_packageManager.searchPath = environment.get("DUBPATH")
+				.splitter(pathsep)
+				.map!(p => makeAbsoluteRelativeToCwd(Path(p)))
```
John-Colvin: previously these paths were free-floating, seems better to make the absolute relative to what the cwd is *now* as opposed to when the path is actually used.

s-ludwig:
Just two generic comments, because I don't have the time for doing a review, yet: There should be no breaking API changes (just in case), and I have the feeling that the diff might be larger than it needs to be in order to implement the functionality. I'll have to take a closer look to be certain, but I think that this should be possible by reusing existing functionality.

John-Colvin:
> There should be no breaking API changes

There are (almost) none. I think the only one is that `defaultPlacementLocation` is no longer readable, only writeable, because it would no longer make sense (the location could be something that isn't in `enum PlacementLocation { ... }`. Every other change is backwards compatible, but the old versions are marked `deprecated`. Users inheriting from the various classes shouldn't see breakage as I've been quite careful to keep methods virtual (i.e. no templates, even when they would have been nice).

> the diff might be larger than it needs to be in order to implement the functionality

It is larger than strictly necessary for the command-line flag. 1) repository locations are now controllable from the API in a straightforward manner. 2) There's a fair amount of refactoring and general improvement.

source/dub/dub.d
```diff
@@ -105,28 +107,50 @@ class Dub {
...
+	{
+		return m_defaultRepoPath;
+	}
+	/// Ditto
+	void defaultRepoPath(Path path) @property
+	{
+		path = makeAbsoluteRelativeToCwd(path);
+		//assert(m_packageManager.isManagedPath(path, false), path.toNativeString);
+		m_defaultRepoPath = path;
+	}
```
John-Colvin: `defaultRepoPath` fulfils exactly the same purpose as `defaultPlacementPath`, but is based on paths instead of `PlacementLocation`

source/dub/dub.d
```diff
@@ -727,17 +783,23 @@ class Dub {
...
 				scope size_t delegate(in Package[] packages) resolve_version)
 	{
 		enforce(!package_id.empty);
-		if (location == PlacementLocation.local) {
+
+		auto rootNormalized = rootPath;
+		rootNormalized.normalize();
+		repoPath = makeAbsoluteRelativeToCwd(repoPath);
+		repoPath.normalize();
+		if (repoPath == rootNormalized) {
```
John-Colvin: comparing `Path`s for logical equality (i.e. do they point to the same place) is a bit of a pain. Is there some easy reliable way that I'm missing?

source/dub/dub.d
```diff
@@ -692,12 +743,17 @@ class Dub {
 			return m_packageManager.getPackage(packageId, ver, dstpath);
 		}
 
-		auto path = getTempFile(packageId, ".zip");
-		supplier.fetchPackage(path, packageId, dep, (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
-		scope(exit) std.file.remove(path.toNativeString());
+		auto zipPath = getTempFile(packageId, ".zip");
+		supplier.fetchPackage(zipPath, packageId, dep, (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
+		scope(exit) std.file.remove(zipPath.toNativeString());
```
John-Colvin: just a renaming for clarity.

source/dub/dub.d
```diff
@@ -655,10 +703,13 @@ class Dub {
 		}
 
 		if (existing) {
-			if (!ver.startsWith("~") || !(options & FetchOptions.forceBranchUpgrade) || location == PlacementLocation.local) {
+			repoPath.normalize();
+			auto rootNormalized = rootPath;
+			rootNormalized.normalize();
+			if (!ver.startsWith("~") || !(options & FetchOptions.forceBranchUpgrade) || repoPath == rootNormalized) {
```
John-Colvin: What was the logic behind skipping upgrades for packages that are `PlacementLocation.local`?

source/dub/dub.d
```diff
@@ -181,18 +213,32 @@ class Dub {
...
 	}
 
+	Path toPath(PlacementLocation location) const
+	{
+		final switch (location) {
+			case PlacementLocation.local: return m_rootPath;
+			case PlacementLocation.user: return m_dirs.userRepo;
+			case PlacementLocation.system: return m_dirs.systemRepo;
+		}
+	}
```
John-Colvin: useful for transitioning to using paths when you have `PlacementLocation` all over your code, lets people translate from `PlacementLocation`. If all the `PlacementLocation` overloads were to be removed, I would imagine this would remain to allow stragglers to stay on board.

source/dub/dub.d
```diff
@@ -105,28 +107,52 @@ class Dub {
...
+	{
+		path = makeAbsoluteRelativeToCwd(path);
+		//assert(m_packageManager.isManagedPath(path, false), path.toNativeString);
+		m_defaultRepoPath = path;
+	}
+	/// Ditto
+	deprecated defaultPlacementLocation(PlacementLocation location) @property
+	{
+		defaultRepoPath = toPath(location);
+	}
```
John-Colvin: Currently this property is write-only, this should be the only breaking change in the whole PR. However, if breaking changes aren't possible at this point, an alternative would be to add this:
```D
deprecated PlacementLocation defaultPlacementLocation() @property
{
	if (defaultRepoPath == m_rootPath) return PlacementLocation.local;
    if (defaultRepoPath == m_dirs.userRepo) return PlacementLocation.user;
    if (defaultRepoPath == m_dirs.systemRepo) return PlacementLocation.system;
    throw new Exception("defaultPlacementLocation attempted to be read, "
			~ "but defaultRepoPath is set to a custom path");
}
```

Panke:
I think I need something similar for #1898. What's the state here?

Geod24:
@John-Colvin : I see a lot of overlap between what you were attempting to do here and the current state of the `PackageManager`. Namely, things were re-centered around `Location` (former `Repository`) and the `PackageManager` is now a hub to access `Location`s.

I see that the main change was to index things by `Path`, which, as you described it, is scary. The way one can support unlimited `Path` right now is by adding search paths to a specific location. Do you think the current state of master covers what you tried to do (albeit differently), or is there something still missing ?

PR #988, opened 2016-11-11
introduce x86-mscoff arch

John-Colvin:
`x86-mscoff` is more like an architecture than a normal flag, it needs propagating down to dependencies.


s-ludwig:
Can you rename it to `x86_mscoff`? We need this in one form or another, and while I'm not sure if this is the cleanest solution, using an underscore will ensure that this doesn't conflict with a different solution later on (e.g. the object/executable file format might become a separate entity, orthogonal to the architecture and then the "-" would be ambiguous for platform specifications).

John-Colvin:
ok... github has completely screwed up here, it can't see the commits any more and when i tried to retarget it (hoping that might get it to re-find them, it closed the pr!). I'll open a new one

Issue #992, opened 2016-11-13
Avoid special-casing POSIX in copyFile


WebDrake:
The `copyFile` function in `dub.internal.vibecompat.core.file` uses a `version (Posix)` block to special-case the preservation of file permissions:
https://github.com/dlang/dub/blob/v1.0.0/source/dub/internal/vibecompat/core/file.d#L118-L129

However, this ignores the `preserve` parameter of `std.file.copy`:
https://dlang.org/phobos/std_file.html#.copy

Is there a good reason for this special-casing and avoiding using the `preserve` parameter, or is it an oversight?  If the latter, I would suggest just setting the `preserve` parameter to always preserve permissions (regardless of platform).

WebDrake:
I should declare a vested interest here: I'm trying to create a snap package of dub, and the `chown` call on L126 conflicts with snap-package containerization.  So it would be great if that was avoidable.

`std.file.copy` uses a call to `fchmod` but no `chown` call.

s-ludwig:
Is see no places where preserving ownership would be critical, switching to use the preserve parameter (which I *think*  didn't exist back when this code was introduced in DUB) should be safe.

WebDrake:
Cool, thanks.  I'll submit a PR sometime this week; before I do, I would like to verify that it really does solve the containment issue I encountered when trying to create a snap package, and that it doesn't create any other issues (e.g. wrong user for the built binaries).

Assuming all goes OK, would it be possible to have a v1.0.1 release with the fix in?

s-ludwig:
v1.1.1 actually, but it should be small enough to put it in a patch release. I'll add it to the milestone.

WebDrake:
Great, thank you very much.  I'll follow up when I've confirmed the change solves my own issues.

s-ludwig:
Fixed by #994.

PR #994, opened 2016-11-13
Remove POSIX special-casing from copyFile

WebDrake:
Since `std.file.copy` already provides a means to preserve attributes of a file, it is preferable to use that functionality rather than having a custom-written solution.

Note that where the original dub code used a `chown` call followed by a `chmod` call, `std.file.copy` only uses a `chmod`.  This is not expected to cause any issues but should be borne in mind as a possible breaking change for some use-cases.

Fixes https://github.com/dlang/dub/issues/992.

### DIFF
--- a/source/dub/internal/vibecompat/core/file.d
+++ b/source/dub/internal/vibecompat/core/file.d
@@ -112,20 +112,26 @@ void copyFile(Path from, Path to, bool overwrite = false)
 		removeFile(to);
 	}
 
-	.copy(from.toNativeString(), to.toNativeString());
-
-	// try to preserve ownership/permissions in Posix
-	version (Posix) {
-		import core.sys.posix.sys.stat;
-		import core.sys.posix.unistd;
-		import std.utf;
-		auto cspath = toUTFz!(const(char)*)(from.toNativeString());
-		auto cdpath = toUTFz!(const(char)*)(to.toNativeString());
-		stat_t st;
-		enforce(stat(cspath, &st) == 0, "Failed to get attributes of source file.");
-		if (chown(cdpath, st.st_uid, st.st_gid) != 0)
-			st.st_mode &= ~(S_ISUID | S_ISGID);
-		chmod(cdpath, st.st_mode);
+	static if (is(PreserveAttributes))
+	{
+		.copy(from.toNativeString(), to.toNativeString(), PreserveAttributes.yes);
+	}
+	else
+	{
+		.copy(from.toNativeString(), to.toNativeString());
+		// try to preserve ownership/permissions in Posix
+		version (Posix) {
+			import core.sys.posix.sys.stat;
+			import core.sys.posix.unistd;
+			import std.utf;
+			auto cspath = toUTFz!(const(char)*)(from.toNativeString());
+			auto cdpath = toUTFz!(const(char)*)(to.toNativeString());
+			stat_t st;
+			enforce(stat(cspath, &st) == 0, "Failed to get attributes of source file.");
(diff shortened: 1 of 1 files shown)


WebDrake:
Note, this also fixes an issue with snap packaging of dub: the `chown` was blocked by snap-package containment, while the `chmod` is not.

WebDrake:
Note, this PR might have to be retargeted against a v1.1.x branch as it's intended for a v1.1.1 patch release.

WebDrake:
Looks like this is failing for frontends 2.067 and older (cf. remarks in https://github.com/dlang/dub/issues/992 about which frontend version the `PreserveAttributes` parameter was introduced).

I would presume the best option here is to introduce a `version` conditional based on the frontend version.

s-ludwig:
I've updated the version on "stable" and "master" now (there is a check already, but forgot to update it).

s-ludwig:
So to be more clear, this was an existing build error on master/stable. Can you rebase the PR against stable? It's green again now.

WebDrake:
Updated to handle the case where `PreserveAttributes` is not available.

```diff
diff --git a/source/dub/internal/vibecompat/core/file.d b/source/dub/internal/vibecompat/core/file.d
index 9490c11..444c5e1 100644
--- a/source/dub/internal/vibecompat/core/file.d
+++ b/source/dub/internal/vibecompat/core/file.d
@@ -112,7 +112,27 @@ void copyFile(Path from, Path to, bool overwrite = false)
                removeFile(to);
        }
 
-       .copy(from.toNativeString(), to.toNativeString(), PreserveAttributes.yes);
+       static if (is(typeof(PreserveAttributes)))
+       {
+               .copy(from.toNativeString(), to.toNativeString(), PreserveAttributes.yes);
+       }
+       else
+       {
+               .copy(from.toNativeString(), to.toNativeString());
+               // try to preserve ownership/permissions in Posix
+               version (Posix) {
+                       import core.sys.posix.sys.stat;
+                       import core.sys.posix.unistd;
+                       import std.utf;
+                       auto cspath = toUTFz!(const(char)*)(from.toNativeString());
+                       auto cdpath = toUTFz!(const(char)*)(to.toNativeString());
+                       stat_t st;
+                       enforce(stat(cspath, &st) == 0, "Failed to get attributes of source file.");
+                       if (chown(cdpath, st.st_uid, st.st_gid) != 0)
+                               st.st_mode &= ~(S_ISUID | S_ISGID);
+                       chmod(cdpath, st.st_mode);
+               }
+       }
 }
 /// ditto
```

WebDrake:
Please hold off on merging this for a while -- I'm not certain this is going to work with snap packaging.

WebDrake:
Not sure why the build is failing -- looks like it's not down to my changes ... ?

s-ludwig:
Looks like network issues w.r.t code.dlang.org or GitHub. I've restarted the builds.

WebDrake:
OK, this works for my snap-package use-case, so as far as I'm concerned it's ready for review.

s-ludwig:
Okay, looks ready to be merged then.

WebDrake:
Thanks!  Can you ping me when the v1.1.1 release is made?  (No urgency from my side.)

WebDrake:
On 13/11/16 14:59, Sönke Ludwig wrote:
> I've updated the version on "stable" and "master" now (there is a check already,
> but forgot to update it).

I was assuming there would be a tagged v1.1.1 ... ?

s-ludwig:
v1.1.1 will be tagged together with DMD 2.072.1 if everything goes as planned.

WebDrake:
OK, thanks!

PR #1002, opened 2016-11-22
Source globbing

John-Colvin:
the docs suggest that dub supports globbing on `sourceFiles`.
`sourceFiles	"<pattern1>" ["<pattern2>" [...]]"` from https://code.dlang.org/package-format?lang=sdl#build-settings. "pattern" is only elsewhere used for `excludedSourceFiles` and `copyFiles`, both of which support globbing.

This PR implements globbing for `sourceFiles`


source/dub/project.d
```diff
@@ -1161,12 +1161,41 @@ private string processVars(string var, in Project project, in Package pack, bool
...
+			case '/': sepIdx = i; goto default;
+			default: ++i; break switch_;
+			}
+		}
+		if (i == res.length) //no globbing found in the path
+			return [res];
+		import std.path : globMatch;
+		return dirEntries(res[0 .. sepIdx], SpanMode.depth)
+			.map!(de => de.name)
+			.filter!(name => globMatch(name, res))
```
wilzbach: This will iterate over all files and not stop if the folder is not included in the glob - in other words it's not globing, more filtering with a regex ;-)

John-Colvin: sadly, yes. Well, getting the non-globbed prefix makes it not *so* bad in common cases, but yes, it should be better.

John-Colvin: Ideally an efficient (correct? I guess doing it my way would fail if it visits folders it doesn't have permission to read, that the ideal implementation wouldn't look at) implementation should be put in the pattern overload of std.file.dirEntries

s-ludwig:
I think the filter semantics are okay for now, as the other fields are processed in a similar way, but instead of `processVars` it should happen in `BuildSettingsTemplate.getPlatformSettings.collectFiles`, so that `BuildSettings` doesn't carry around invalid file names half of the time.

John-Colvin:
my reasoning for doing it in `processVars` was because of environment variables: You can't assume usable paths until after they've been resolved by `processVars`.

John-Colvin:
@s-ludwig `BuildSettings` is carrying around invalid names until `processVars` anyway due to environment variables.

John-Colvin:
ping @s-ludwig, I explained my reasoning, am I misunderstanding?

John-Colvin:
ping?

MartinNowak:
Having to list all files is a bit of an anti-pattern for build systems, since they need to be checked every time for rebuilds. We do that with dub's search in sourcePaths anyhow, so it shouldn't matter much.

What's the use-case for globs in sourceFiles over sourcePaths?
Usually `sourceFiles` is intended to add a few additonal files (or libs).

I'd also say that the default globMatch is fairly limiting, a common extension is to allow `**` to match in subdirectories. Might be interesting for any glob matching in dub (and could be added to phobos).

MartinNowak:
> my reasoning for doing it in processVars was because of environment variables: You can't assume usable paths (for globbing or anything else) until after they've been resolved by processVars.

Do we allow environment variables in file lists anywhere? I mean packages should be self-contained. Am I misunderstanding anything? If so an example might help.

MartinNowak:
Ping @John-Colvin

Geod24:
One reason to add globbing in `sourceFiles` is to support passing object files to the compiler.
I have a project which interacts with a lot of C++ code. Said code uses a rather complex build system, but I don't need that complexity, as I seldom touch those files.

The approach I took was to write a quick D script to compile those files. As I didn't want to deal with dynamic libraries, and static libraries don't work on OSX, I just compile them to object files, and added all of them to `sourceFiles`.

My `dub.json` looks like this (without the comments :) ):
```json
    "excludedSourceFiles": [ "source/scpp/*.d" ],    // A bunch of scripts to update/build those .cpp/.h
    "preBuildCommands": [ "source/scpp/build.d" ],   // Build the .o out of the .cpp
    "sourceFiles": [
        "source/scpp/build/BallotProtocol.o",
        "source/scpp/build/ByteSliceHasher.o",
        "source/scpp/build/DUtils.o",
        "source/scpp/build/HashOfHash.o",
        "source/scpp/build/Hex.o",
        "source/scpp/build/KeyUtils.o",
        "source/scpp/build/LocalNode.o",
        "source/scpp/build/Math.o",
        "source/scpp/build/NominationProtocol.o",
        "source/scpp/build/QuorumSetUtils.o",
        "source/scpp/build/SCP.o",
        "source/scpp/build/SCPDriver.o",
        "source/scpp/build/SHA.o",
        "source/scpp/build/SecretKey.o",
        "source/scpp/build/Slot.o",
        "source/scpp/build/StrKey.o",
        "source/scpp/build/crc16.o",
        "source/scpp/build/jsoncpp.o",
        "source/scpp/build/marshal.o",
        "source/scpp/build/numeric.o",
        "source/scpp/build/uint128_t.o"
    ],
    "lflags": [ "-lsodium", "-lstdc++" ],
```

As you can imagine, having `source/scpp/build/*.o` would be much nicer.

thewilsonator:
(Hope I didn't bungle that too bad.)

thewilsonator:
```
0.12s$ wget https://raw.githubusercontent.com/wilzbach/installer-dub/master/script/install.sh -O ~/dlang/install.dub.sh
--2019-07-17 07:27:04--  https://raw.githubusercontent.com/wilzbach/installer-dub/master/script/install.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 151.101.0.133, 151.101.64.133, 151.101.128.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|151.101.0.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 23625 (23K) [text/plain]
Saving to: ‘/home/travis/dlang/install.dub.sh’
100%[======================================>] 23,625      --.-K/s   in 0.01s   
2019-07-17 07:27:04 (2.28 MB/s) - ‘/home/travis/dlang/install.dub.sh’ saved [23625/23625]
0.85s$ . $(bash ~/dlang/install.dub.sh -a dub)
curl: (22) The requested URL returned error: 502 Bad Gateway
Unknown compiler 'dub-v1.11.0-alpha.1'
/home/travis/.travis/functions: line 104: .: filename argument required
.: usage: . filename [arguments]
The command ". $(bash ~/dlang/install.dub.sh -a dub)" failed and exited with 2 during .
Your build has been stopped.
```
The rest of them passed so I thin its fine.

John-Colvin:
Thanks @thewilsonator for fixing this up, sorry for abandoning it for so long.

thewilsonator:
No problems. I think its probably green now. The globbing is still unchanged for your implementation, but it should work.

thewilsonator:
@Geod24 Cleaned up the spaghetti `static if`

Geod24:
Red though

Geod24:
Ah didn't see you pushed again. Going to try the feature out, thanks.
Note that you should have used `auto-merge-squash` here (and that's why I ask people to squash).

Issue #652, opened 2015-08-13
Optimize dependency resolution algorithm?


Abscissa:
Is it possible to adjust the dependency resolution algorithm to scale better? I've hit situations (not easy to reproduce AFAICT so far) where the dependency resolution step can take several minutes or even half-hour or more, and (unless --vverbose is used) with no output indicating what it's doing or any indication that no, it has not actually locked up.

Abscissa:
FWIW, I just hit a case where it took approx 12 hours. As dub grows and gets more heavily used, this problem will probably only get bigger and more common.

Geod24:
Sounds like a bug. Any idea what the bound is ? I take from your comment it's IO (registry) bound ?

Abscissa:
Appears to be CPU. I happen to have it running right now (without ```--vverbose```). KDE's System Activity program is reporting approx 5MB (each) in/out IO total, no current IO at all and 30MB memory (completely unchanging). But it's showing an unchanging 25% CPU (I have two hyper-threaded cores, so I'm guessing that might indicate it's a single-threaded program fully saturating one of the 4 available CPU threads).

I've done it before with ```--vverbose```, and I could see that it goes through tons of cycles of trying lots and lots of different combinations of version numbers for the different packages. The first time it happened, I suspected an infinite loop, but I couldn't find any duplicate combinations in the very long ```--vverbose``` output. So I let it run and it eventually completed. So I'm suspecting a poorly-scaling algorithm, like an O(2^n) or some such.

Abscissa:
FWIW, the program I'm hitting this issue with (it's not on github right now) involves 17 packages total (including both direct and indirect dependencies). 19 if you count sub-packages separately.

s-ludwig:
The issue is that the current ((hopefully) exhaustive) algorithm has a worst case exponential runtime. There are a number of quick paths that make the process linear in the usual case, but some dependency configurations contain pathological cases. There are basically two options how to solve this:
1. Make hitting such cases less likely by adding more quick paths.
2. Use a different algorithm that is not exhaustive, but has guaranteed linear or close to linear runtime.

The latter would only be possible by making some assumptions, such as that a newer version of a package will have dependency specifications that are newer or equal to the older versions. Since this will not always be true, some dependency trees may yield a sub-optimal set of version selections, or, in the worst case, will fail to fine one at all.

s-ludwig:
[...]

PR #1012, opened 2016-12-09
fix upgrade <package> not yet implemented error

MartinNowak:
- packages are passed as free_args


MartinNowak:
See https://github.com/dlang/dub/issues/652#issuecomment-219219378 for a proposal on how to implement partial upgrades via `dub upgrade specific-package`.

s-ludwig:
I'll look into implementing this. Should be pretty easy.

Issue #1014, opened 2016-12-11
Unnecessary junk files created when running dub with ldc


joakim-brannstrom:
The directory .dub/obj is created in every directory a shebang script is ran when using ldc2.

Example:
```sh
$ cat issueXXX.d
#!/usr/bin/env dub
/+ dub.sdl:
    name "myapp"
+/
void main(string[] args) {
}

$ ./issueXXX.d
$ find .dub
.dub/obj
.dub/obj/myapp.o
```

ZombineDev:
Which version of ldc are you using? Since I upgraded to `ldc-1.1.0-beta6` I don't recall to have encountered such problems.

joakim-brannstrom:
Same version.
Maybe it is how I installed it (guessing).
I unpacked it and added the "ldc-..../bin" to my PATH.
Which forced me to having to modify ldc-..../etc/dub/settings.json to "ldc2" for dub to find the binary.

joakim-noah:
Just ran into this myself, when using dub on Android with ldc.  Nothing to do with how you installed it, other than forcing dub to use ldc.  It comes from this commit d55d741, which needs to be fixed, as it sprays `.dub` directories all over your filesystem if you use dub to run commands.

PR #1015, opened 2016-12-11
Compiling with LDC generates object files in the current directory

joakim-brannstrom:
Fix to generate them in the target build path.


joakim-brannstrom:
Fixes issue #1014

s-ludwig:
I'm a little skeptical that this change will work in all cases, but obviously that was the case with the earlier hack, too. As a proper solution, I think `BuildSettings` will have to introduce an `intermediateFilesPath` field, so that this can be controlled from outside of the `Compiler` class.

But merging this as an intermediate fix sounds good anyway. There are currently some test errors left:
- It tried to build the package in `tests/issue1014`, as the folder is left there after the test script exits. To be consistent with the other tests, the .d file should be in `tests/issue1014` in the first place, with an empty `.no_build` file placed next to it. The `mkdir`/`cp` code in the script can then be removed
- There is an error in the `tests/ddox` test. I didn't look close enough yet to have an explanation.

PR #1024, opened 2016-12-18
Implement selective package upgrades.

s-ludwig:
See #1012. This makes it possible to pass a list of package names to "dub upgrade". The upgrade will be restricted to those packages.


MartinNowak:
Nice.

What the similar option of `bundler upgrade --source package` does, is updating a particular package __and__ all dependencies of that package. Not allowing recursive updates of dependencies will easily create invalid configurations.
I think the better trick is to simply put the currently selected version to the front after sorting _all_ available versions of a package.

Issue #1037, opened 2017-01-02
Root package X reference Y cannot be satisfied: need to show why


timotheecour:
When you get `Root package X reference Y cannot be satisfied`, I don't see anything in the logs (--vverbose) showing what caused this.

In my case it should show:
```
"gtk-d ~>3.4" violates "gtk-d:gtkd": "~>3.3.0" in gtk-d (and maybe show dependency chain)
```

```
"dependencies": {
"gtk-d" : "~>3.4",
"ggplotd": ">=1.1.2",
},
```

burner:
the same problem as in https://github.com/dlang/dub/issues/1001

The problem I found is that
package A depends on package C of version 1.0.0 and
package B depends on package C of version 1.0.1

The question is what version of package C to use?

SingingBush:
in this scenario could dub inform the user of the problem and let the user decide by putting package C as a dependency in the users project and using that version. 

As a developer making use of package A and package B, I have decided that I want to specifically use version 1.0.3 of package C and am fully aware the it's down to me to test and make sure that packages A & B are ok with that.

timotheecour:
Can you try with https://github.com/dlang/dub/pull/1038 ? this is meant as a simple hotfix for cases like that

PR #1038, opened 2017-01-02
[do not pull] make ~>x.y.z mean >=x.y.z <(x+1).0.0

timotheecour:
make ~>x.y.z mean >=x.y.z <(x+1).0.0 instead of >=x.y.z <x.(y+1).z
This solves a lot of problems I keep running into. Theoretically at least, X.Y.Z should be backward compatible with x.y.z so long X.Y.Z>=x.y.z and X=x (ie so long we keep the major version to x; we therefore should allow minor version increments.

eg:
```
"dependencies": {
"gtk-d" : "~>3.4",
"ggplotd": ">=1.1.2",
},
```
Currently fails because
```
"gtk-d ~>3.4" violates "gtk-d:gtkd": "~>3.3.0"
```
This is a common scenario responsible for a lot of breakages.

With this fix, it passes.

[if this idea is accepted, I could update unittests and docs]

### DIFF
--- a/source/dub/semver.d
+++ b/source/dub/semver.d
@@ -231,6 +231,7 @@ string bumpVersion(string ver) {
 	// Increment next to last version from a[.b[.c]].
 	auto splitted = split(ver, ".");
 	assert(splitted.length > 0 && splitted.length <= 3, "Version corrupt: " ~ ver);
+	if(splitted.length==3) splitted=splitted[0..2];
 	auto to_inc = splitted.length == 3? 1 : 0;
 	splitted = splitted[0 .. to_inc+1];
 	splitted[to_inc] = to!string(to!int(splitted[to_inc]) + 1);


timotheecour:
NOTE: this would fix the case that I was referring to in https://github.com/dlang/dub/issues/1037 among other cases; it's obviously a very common scenario

p0nce:
If you use `~>2.0` it means `>= 2.0.0 < 3.0.0`

rjframe:
In theory, you're right - it should never be necessary so specify more than "x.y", but in practice semver is not enforced, and if a package is 0.y.z there are no guarantees anyway, and I need to be able to prevent auto-upgrading on y (without resorting to "==" rather than "~>") - most people seem to use y for breaking changes when x is 0.

I'm against this change in functionality. As @p0nce says, we have the ability to specify the level of upgrades we want; this PR takes that away. At minimum, the PR should warn me that it's overriding my plan - not do so silently (or it should just warn anytime I specify a version by "x.y.z" that z is ignored).

In your example, I can resolve the issue by choosing to either downgrade my gtk-d dependency to 3.3, or update/fix ggplotd. Perhaps instead of this PR, modify the documentation to recommend the developer only specify "x.y" if (x > 1)? If ggplotd did this, your problem would be resolved.

s-ludwig:
It should also be noted that this is de-facto standard behavior used by other package managers as well and changing it would be very confusing for anyone who is familiar with those.

timotheecour:
would it be ok if it were controlled by an option (off by default), eg:
```
dub build #default behavior
dub build --compatibility=default #default behavior
dub build --compatibility=lax
```

timotheecour:
> most people seem to use y for breaking changes when x is 0.

how about special casing this 0 case:

```
dub build #default behavior
dub build --compatibility=default #default behavior
dub build --compatibility=lax
dub build --compatibility=0_is_special
```

p0nce:
There is no significant to depart from how other package managers do.
And even less so if configurable, that would break the porject if build with another option.

s-ludwig:
Closing this, as possible advantages over `~>x.y` are dwarfed by the breakage and/or inconsistency with other package managers. I felt the urge to specify the patch version of the base version as well, in cases where really only the major version was important. But to be realistic, if it's already *necessary* to specify the patch version of the base version, it's rather unlikely that the package in question will be able to keep backwards compatibility during the whole major version phase anyway and `~>x.y.z` is probably the more appropriate specification.

p0nce:
A quick poll on IRC showed that most consider small breaking change in minor version acceptable!

s-ludwig:
The SemVer spec doesn't, though, and we should really follow that as good as possible from the package manager POV. But I'd still suggest to make an exception for deprecations (which could be considered breaking changes for people who compile with `-de`), so that they don't require a separate major version.

p0nce:
Well deprecation are often for renaming/moving files and they are the easiest to do in a backwards-compatible way. I guess the reason why so much people use ~>x.y.z is because of the perception things could break with minor versions.
In no-less than DerelictUtil (most depended upon library?), we are using ~>x.y.z because its author thinks minor version allow breaking change, so it would be unwise.

s-ludwig:
Yeah, I guess that's the practical reason for choosing those semantics for `~>x.y.z` (maybe also for `0.x.y`, but there you'd usually need to match the exact version anyway to avoid breakage). Maybe we can one day realize some kind of partial automation to check for breaking API changes (e.g. generate dummy calls to all functions for the previous version and compile them against the current one, won't work very well for templates of course).

MartinNowak:
> A quick poll on IRC showed that most consider small breaking change in minor version acceptable!

That's a bad direction, which you know if you ever had to maintain/update a system with more than a handful of dependencies.
If changes are so frequent people are hopefully still in pre `1.0.0` where they are allowed.
Down the road, tagging sth. `1.0.0` without committing to API stability will be huge pain for users.

PR #1050, opened 2017-01-23
Fix issue 824: More comprehensive gitignore file.

rjframe:


BBasile:
You forgot to pull on your fork...you commit is 355 behind official master. You need to rebase and solve the conflicts.

rjframe:
Thanks. This is done.

source/dub/init.d
```diff
@@ -158,7 +158,20 @@ private void initDeimosPackage(Path root_path, ref PackageRecipe p, scope void d
...
-		".dub\ndocs.json\n__dummy.html\n*.o\n*.obj\n__test__*__\n");
+q"{.dub
+docs.json
+__dummy.html
+docs/
+*.so
+*.dylib
+*.dll
+*.a
+*.lib
```
s-ludwig: I have some repositories that have library files checked in, so I think this is too broad as a default. It would definitely make sense to ignore "(project name).{a,lib,dll,dylib,exe}".

source/dub/init.d
```diff
@@ -158,7 +158,20 @@ private void initDeimosPackage(Path root_path, ref PackageRecipe p, scope void d
...
+*.exe
+*.o
+*.obj
+__test__*__
```
s-ludwig: The test pattern has changed to "(project name)-test-(configuration)[.exe]" recently.

wilzbach:
FYI: There's also https://github.com/github/gitignore/pull/2300

source/dub/init.d
rjframe: @s-ludwig My apologies for the delay; this is fixed.

dram:
Seems there are several problems:

* `.so` line and `.dylib` line are joined together
* there is a superfluous `"` after `.so`
* `*.so` may also need to be changed to `{name}.so`, like `.dylib` and `.dll`

rjframe:
@s-ludwig @dram You are correct; does Github let me continue pushing changes after a merge or does that need to be a new PR?

Though I would personally keep *.so; I often place shared library dependencies in the source directory rather than my system paths but I want them ignored.

wilzbach:
> does Github let me continue pushing changes after a merge or does that need to be a new PR?

New PR

s-ludwig:
I've pushed a quick fix: #1209

PR #1051, opened 2017-01-24
VisualD : generate all build modes

Flamaros:
Hi,

This changes make the VisualD solution always generated with debug, release and unittest modes. 
Any Visual Studio should expect that as every c++ projects comes with debug and release modes by default. This improve Windows user productivity by reducing the usage of terminal.

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -108,6 +108,7 @@ struct BuildSettings {
 	void addOptions(in BuildOptions value) { this.options |= value; }
 	void removeOptions(in BuildOption[] value...) { foreach (v; value) this.options &= ~v; }
 	void removeOptions(in BuildOptions value) { this.options &= ~value; }
+	void resetOptions() { this.options = BuildOption.none; }
 
 	// Adds vals to arr without adding duplicates.
 	private void add(ref string[] arr, in string[] vals, bool no_duplicates = true)
--- a/source/dub/generators/visuald.d
+++ b/source/dub/generators/visuald.d
@@ -93,24 +93,36 @@ class VisualDGenerator : ProjectGenerator {
 
 			// Global section contains configurations
 			ret.put("Global\n");
-			ret.put("\tGlobalSection(SolutionConfigurationPlatforms) = preSolution\n");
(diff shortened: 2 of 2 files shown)


Flamaros:
I may found a little issue, all builds use the same output file path. So a manual clean is needed to generate the proper target.

Need I fix that? I think to fix it by adding the build mode as directory in the output path (eg. bin/debug/prog.exe)

Flamaros:
Need I improve test coverage to have a review?

s-ludwig:
Sorry for the review delay, we have a little shortage of regular reviewers currently. I think this is generally a viable solution for the time being, being able to specify different target names/paths is an orthogonal issue and can be solved later.

The only issue is that just changing the build options may be somewhat surprising if more than that is defined in the package recipe, such as a custom build mode. Instead of hard-coding the build options, `addBuildTypeSettings` would have to be used here. However, since `ProjectGenerator` already does that, we'd need another field in `TargetInfo`, let's call it `baseBuildSettings`, which includes everything except for the build type specific settings.

Then, `VisualDGenerator` would just have to duplicate that object and replicate https://github.com/dlang/dub/blob/28a1c02fa773f5b82f5fb447975ea31244461251/source/dub/generators/generator.d#L104-L105 for each build type.

In the long term, I'd like to implement something akin to `dub generate visuald --build=debug,release --arch=x86,x86_64` for defining the build matrix, but until then implementing it just within the VisualD generator is the most practical solution.

Flamaros:
No problem.

I'll try to do modifications that you suggest.

My first try was to manage command line exactly like you see it for the futur, but I am not enough familiar with dub code (and confortable with D) to manage changes with this kind of impact. It will be a major refactor that impacts all generators.

Pombal:
Any updates on this?

I'm trying to find a way to generate a visuald solution with x86,  x86-64, debug, and release configurations, and I can't find any way to do it.

warwick-davis:
Also looking for an update. Would be willing to take over if required.

rikkimax:
> Also looking for an update. Would be willing to take over if required.

Looks like they are doing their own thing and this isn't important to them. So yeah go ahead and take it over if you want to continue it!

Issue #533, opened 2015-03-25
[Feature request] .hgignore for dub init


dmi7ry:
`dub init projectname` will create a directory for the project with a `.gitignore` file. But some developers use Mercurial. Please add option for support Mercurial (or simple add `.hgignore` file - then developer can leave the desired file and delete other).

Geod24:
The more I think about it, the more I think `dub init` should be more configurable: a dub.json template (for Authors / dependencies), custom file / structure, etc...

s-ludwig:
But remember that `dub init` is mainly just a helper for newcomers or for having a quick sane default in case of experienced users. Making it too complex could easily raise the learning curve too much in relation to the benefit it provides.

A `--hg(ignore)` switch sounds fine, though.

ColdenCullen:
It might be nice to make `dub init` interactive, in a similar vein to npm. This would make creating packages even easier, and would result in more complete starting points:
```bash
$ dub init
Package name [DIRNAME]: mypackage
Description ["A simple dub package"]: A basic package
License [MIT]: BSD
Authors [USERNAME]: Colden Cullen
```
would generate the following:
```json
{
    "name": "mypackage",
    "description": "A basic package",
    "license": "BSD",
    "copyright": "Copyright 2015 Colden Cullen",
    "authors": [ "Colden Cullen" ],
    "dependencies": {
    }
}
```
And so on and so forth.

s-ludwig:
Could be an interesting default behavior indeed!

grogancolin:
I think making init more customisable would be good, but personally I
wouldn't like it being interactive.

I'd rather just add lots of switches for setting each parameter, and if
switch isn't given just give it a default value.

s-ludwig:
I'd just make the argument-less `dub init` use the interactive route. Adding any custom parameters would switch to standard mode.

The big advantage is that it drives people to use sane values for fields like description or author. There are currently a number of public projects where these have been left with default values.

grogancolin:
Yeah, this is a good point.

MartinNowak:
+1 for interactive
Another idea is to combine `dub init` with a `git init` call (or `hg init` for that matter).

gedaiu:
I would like to have dub interactive by default. Maybe with a -i flag?

MartinNowak:
https://github.com/D-Programming-Language/dub/pull/600#issuecomment-143574460
[...]

Issue #830, opened 2016-04-29
Skip the format selection step for "dub init" if the --format switch was used


s-ludwig:
See #826 - not necessary to ask the user again.

PR #1057, opened 2017-01-30
Fix issue #533: Support .hgignore file in dub init.

rjframe:
Note that this also includes my PR #1050 (add globs to the ignorefiles).

Git remains the default. I'm going to create a separate issue concerning #533's comments about running `{git|hg} init` to another ticket, since it is a separate issue.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -432,6 +432,7 @@ class InitCommand : Command {
 	private{
 		string m_templateType = "minimal";
 		PackageFormat m_format = PackageFormat.json;
+		string m_repoType = "git";
 		bool m_nonInteractive;
 	}
 	this()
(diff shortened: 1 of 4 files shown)


rjframe:
This is also affected by and expands the scope of issue #830.

rjframe:
The check failure is "Permission denied" in the "test/0-init-ignorefiles.sh" script; I cannot reproduce it on my machine and the script does nothing that isn't also done by the other init tests - how can I debug this?

rjframe:
I'm not sure what other package managers do (I'm only familiar with pip, and compose as a consumer). I agree it would be better not to list it during init (I think a few of the items asked for should be moved to the global configuration file, this among them); I added it since it was part of the relevant issue, but would be glad to remove it.

rjframe:
The `dub init` process no longer asks for the repository type, defaulting to git.

Geod24:
There doesn't seem to be a high demand for it, and in fact the trend is very clearly down, as can be seen from e.g. Bitbucket removing support. [This article](https://www.theregister.co.uk/2019/08/21/bitbucket_mercurial_repositories/) claims 4% of users.
It also doesn't make sense for `dub` to have a half-baked support, with "just" `.hgignore`, so I think we should close this.

PR #1059, opened 2017-01-30
Fix 'Failed to apply the selected architecture x86_mscoff. Got ["x86"]' when passing --arch=x86_mscoff

MrSmith33:


source/dub/compilers/utils.d
```diff
@@ -301,7 +301,7 @@ Path generatePlatformProbeFile()
 		string determineArchitecture()
 		{
 			string ret;
-			version(X86) ret ~= `"x86", `;
+			version(X86) ret ~= `"x86", "x86_mscoff", `;
```
MrSmith33: Another variant is:
``` D
version(Windows)
	version(X86) ret ~= `"x86", "x86_mscoff", `;
else
	version(X86) ret ~= `"x86", `;
```

s-ludwig:
I think that the fix should target the place where the warning is generated. Putting it here results in a platform description that is not necessarily accurate (i.e. when not compiling with `-m32mscoff`, it still outputs "x86_mscoff") and other users of this functionality might get confused.

What I would probably do, before proper generic support for selecting object file formats is implemented, would be to explicitly replace `canFind(arch_override)` with `canFind(arch_override == "x86_mscoff" ? "x86" : arch_override)` , together with a comment explaining the hack. Then, once that generic support is implemented, the "mscoff" wouldn't be part of "arch_override" anymore and the hack can be removed again.

MrSmith33:
Ok, I will change this to your suggestion

s-ludwig:
Okay, thanks, looks good to merge then!

BTW, just for future reference, pure bug fixes should be made on "stable" instead of "master" now since DUB was integrated into the DMD release process. I'll just handle that manually for this one.

s-ludwig:
I just realized that my suggestion also made it impossible to use a platform specifier specific to x86_mscoff, which is needed by libasync. I've adjusted the code to also add to the `.architecture` array, but dragged it out from the platform probe file to `probePlatform()`. Sorry for the confusion.

MrSmith33:
> BTW, just for future reference, pure bug fixes should be made on "stable" instead of "master" now since DUB was integrated into the DMD release process. I'll just handle that manually for this one.

Thanks, good to know!

PR #1068, opened 2017-02-12
Prevent hardLinkFile from unlinking a target just to then turn around…

daym:
… and link the same source to the same target again (on Posix).

### DIFF
--- a/source/dub/internal/vibecompat/core/file.d
+++ b/source/dub/internal/vibecompat/core/file.d
@@ -145,8 +145,14 @@ version (Windows) extern(Windows) int CreateHardLinkW(in wchar* to, in wchar* fr
 // guess whether 2 files are identical, ignores filename and content
 private bool sameFile(Path a, Path b)
 {
-	static assert(__traits(allMembers, FileInfo)[0] == "name");
-	return getFileInfo(a).tupleof[1 .. $] == getFileInfo(b).tupleof[1 .. $];
+	version (Posix) {
+		auto st_a = std.file.DirEntry(a.toNativeString).statBuf;
+		auto st_b = std.file.DirEntry(a.toNativeString).statBuf;
+		return st_a == st_b;
+	} else {
+		static assert(__traits(allMembers, FileInfo)[0] == "name");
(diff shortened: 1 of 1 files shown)


source/dub/internal/vibecompat/core/file.d
```diff
@@ -145,8 +145,14 @@ version (Windows) extern(Windows) int CreateHardLinkW(in wchar* to, in wchar* fr
 // guess whether 2 files are identical, ignores filename and content
 private bool sameFile(Path a, Path b)
 {
-	static assert(__traits(allMembers, FileInfo)[0] == "name");
-	return getFileInfo(a).tupleof[1 .. $] == getFileInfo(b).tupleof[1 .. $];
+	version (Posix) {
+		auto st_a = std.file.DirEntry(a.toNativeString).statBuf;
+		auto st_b = std.file.DirEntry(a.toNativeString).statBuf;
+		return st_a == st_b;
```
MartinNowak: This does an identity check, but an identical copy would be OK as well.
How about just using the same getFileInfo heuristic?

source/dub/internal/vibecompat/core/file.d
```diff
@@ -145,8 +145,14 @@ version (Windows) extern(Windows) int CreateHardLinkW(in wchar* to, in wchar* fr
 // guess whether 2 files are identical, ignores filename and content
 private bool sameFile(Path a, Path b)
 {
-	static assert(__traits(allMembers, FileInfo)[0] == "name");
-	return getFileInfo(a).tupleof[1 .. $] == getFileInfo(b).tupleof[1 .. $];
+	version (Posix) {
+		auto st_a = std.file.DirEntry(a.toNativeString).statBuf;
+		auto st_b = std.file.DirEntry(a.toNativeString).statBuf;
```
MartinNowak: Typo s/a/b/

MartinNowak:
Why did you close it?

daym:
Replaced by #1071 where I used a non-master branch so I can make more changes.

PR #1071, opened 2017-02-13
Prevent hardLinkFile from unlinking a target just to then turn around…

daym:
… and link the same source to the same target again (on Posix).


daym:
It's using statBuf because only that contains (st_ino, st_dev) which one needs to compare (at least) for hard links.

PR #1072, opened 2017-02-13
fix dub builds for FreeBSD

MartinNowak:
- curl workaround was linking against shared phobos library on FreeBSD (shipped since 2.069)
- fixed by moving pkg-config and LIBS stuff under <2.069.0 workaround

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,28 +17,29 @@ if [ "$DMD" = "" ]; then
 fi
 
 VERSION=$($DMD --version 2>/dev/null | sed -n 's|DMD.* v||p')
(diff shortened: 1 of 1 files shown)


MartinNowak:
Actually running `./dub --version` uncovered a sleeping bug in the FreeBSD distribution and now prevented me from building 2.073.1-b2. Because we're building tools with the recent stable compiler (2.072.2) and shared libraries from that aren't on LD_LIBRARY_PATH, dub couldn't run.

MartinNowak:
We should be able to remove that workaround soon.

MartinNowak:
Merging so I can build the beta.

PR #1078, opened 2017-02-22
Avoid excessively complex regex and improve test coverage.

s-ludwig:
This speeds up "dub test" by several seconds, which is the time required to build the original comments_pattern regex.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -427,7 +427,7 @@ string getModuleNameFromContent(string content) {
 	static Regex!char comments_pattern, module_pattern;
 
 	if (!regex_initialized) {
-		comments_pattern = regex(`(/\*([^*]|[\r\n]|(\*+([^*/]|[\r\n])))*\*+/)|(//.*)`, "g");
+		comments_pattern = regex(`//[^\r\n]*\r?\n?|/\*.*\*/|/\+.*\+/`, "g");
 		module_pattern = regex(`module\s+([\w\.]+)\s*;`, "g");
 		regex_initialized = true;
 	}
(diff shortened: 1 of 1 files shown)


source/dub/internal/utils.d
```diff
@@ -427,43 +427,32 @@ string getModuleNameFromContent(string content) {
...
 	auto result = matchFirst(content, module_pattern);
 
-	string moduleName;
-	if(!result.empty) moduleName = result.front;
+	if (!result.empty) return result[1];
 
-	if (moduleName.length >= 7) moduleName = moduleName[7..$-1];
-
-	return moduleName;
+	return null;
```
wilzbach: This looks like`Nullable` would be a clearer return type ;-)

wilzbach:
How about writing a small, specialized lexer for this?
This should by a lot faster than the regex. I just outlined the rough idea here as a proof-of-concept and share some simple perf stats:

```d
dstring findModuleName(string filename)
{
    import std.algorithm, std.file, std.range, std.utf;

    auto content = filename.readText;

    bool starCommentScope, plusCommentScope, inlineComment;

    // iterate with 2-mer (avoids need of most special handling
    foreach (i, e, next; content.lockstep(content.save.dropOne))
    {
        switch (e) {
            case '/':
                switch (next) {
                    case '*':  starCommentScope = true; break;
                    case '+': plusCommentScope = true; break;
                    case '/': inlineComment = true; break;
                    default:
                }
                break;
            case '*':
                if (next == '/') starCommentScope = false;
                break;
            case '+':
                if (next == '/') plusCommentScope = false;
                break;
            // might need a bit of OS tuning
            case '\n':
            case '\r':
                inlineComment = false;
                break;
            default:
                if (!(plusCommentScope || starCommentScope)) {
                    // check for match
                    if (content[i .. $].startsWith("module ")) {
                        return content[i .. $].until(';').drop(7).array;
                    }
                }
        }
    }
    return null;
}

void main(string[] args)
{
    import std.stdio;
    findModuleName(args[1]).writeln;
}
```

Another huge performance boost would come from only reading on a stream instead of reading the entire file.
A quick benchmark:

| method|1K|10K|
|--------|----|---------|
|parser | 0.6s | 6.2s|  
|regex | 3.8s | 38s |

source/dub/internal/utils.d
s-ludwig: I could have used "" here, because the user code checks for `.length == 0` anyway, but I'm kind of used to using `null`. But in general, I think that using an empty string to denote the absence of a `module` statement should be fine here, as there are no ambiguities and `Nullable` would mainly just complicate the usage.

s-ludwig:
Yes, a custom parser would definitely be a good idea. But from my side I'd propose to keep that as a follow-up PR, because I'm so overloaded with higher priority issues right now. But in any case, I'll add a couple of more tests that should also be useful for a custom parser later on.

wilzbach:
> Yes, a custom parser would definitely be a good idea. But from my side I'd propose to keep that as a follow-up PR, because I'm so overloaded with higher priority issues right now. But in any case, I'll add a couple of more tests that should also be useful for a custom parser later on.

Oh I understand :/
I pushed a cleaned version of my proof of my simple parser as #1080, so that it doesn't get forgotten ;-)

PR #1079, opened 2017-02-22
Fix cubic runtime complexity for collecting source/import files.

s-ludwig:
See rejectedsoftware/vibe.d#1690.

This is still not very efficient, doing several temporary allocations and copies, but replaces cubic runtime by almost linear (except for allocation and AA creation times).

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -149,14 +149,14 @@ struct BuildSettings {
 	// add string import files (avoids file name duplicates in addition to path duplicates)
 	private void addSI(ref string[] arr, in string[] vals)
 	{
-		outer:
+		bool[string] existing;
+		foreach (v; arr) existing[Path(v).head.toString()] = true;
 		foreach (v; vals) {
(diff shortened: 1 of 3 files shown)


source/dub/compilers/buildsettings.d
```diff
@@ -149,14 +149,14 @@ struct BuildSettings {
 	// add string import files (avoids file name duplicates in addition to path duplicates)
 	private void addSI(ref string[] arr, in string[] vals)
 	{
-		outer:
+		bool[string] existing;
```
wilzbach: A next iteration could cache this as well.

s-ludwig: True, I was a bit afraid to forget a place where it could get out of sync, because the arrays are public fields, so I wanted to leave that to a more in-depth optimization attempt.

s-ludwig:
Okay, since this is just a local change, I'm also pretty confident that it should be without mistakes. I'll rebase and merge after Travis passes, unless any issues come up.

PR #1080, opened 2017-02-23
[Perf] Use a custom parser to get module name from content

wilzbach:
Coming from #1078, here's a simple, custom parser to get the module name. It's about 10x faster than using the RegExp.

Another big gain would be to use a file stream instead of parsing the entire file, but that's an opportunity for another PR.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -416,27 +416,60 @@ string determineModuleName(BuildSettings settings, Path file, Path base_path)
 /**
  * Search for module keyword in D Code
  */
-string getModuleNameFromContent(string content) {
-	import std.regex;
-	import std.string;
+string getModuleNameFromContent(string _content)
+{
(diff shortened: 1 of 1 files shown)


s-ludwig:
Do you have objections to merge #1078? Then, if this passes the additional tests, I think we can just merge it as well.

wilzbach:
> Do you have objections to merge #1078?

Not at all ;-)
As said I just made the PR, s.t. this doesn't get forgotten.

> Then, if this passes the additional tests, I think we can just merge it as well.

Let's hope for the best.

s-ludwig:
Still errors out with
> core.exception.AssertError@source/dub/internal/utils.d(479): unittest failure

BBasile:
You should add tests for laughable (but valid) cases like
```d
module /**/hu./**/he/**/./**/ho;
```

Geod24:
Done in https://github.com/dlang/dub/pull/2897

Issue #909, opened 2016-08-12
Lift "Single-file packages are not allowed to specify source paths" restriction


wilzbach:
> Single-file packages are not allowed to specify source paths.

This seems to me a rather strict restriction into which I have bumped into now a couple of times already.

Use cases:

1) I have one helper file/library and multiple plotting script that are each runnable (currently the helper file needs to be in it's own folder and have it's own dub.json)
2) It really forces one to think in the "dub-way", e.g. I can't mess with `sourcePaths` to include globally installed, pre-compiled packages or include other temporary files (e.g. from git)

@s-ludwig what were your reasons to restrict the use of source paths for single-file packages?

PR #1081, opened 2017-02-23
Use a default package name for single file packages in case none is provided

wilzbach:
Usually the `name` is set to the `file_name` without extensions anyways, so imho this a very sane fallback for single file packages.

BTW
- you have a lot of trailing whitespace in your files. How about removing them and enabling a quick CI check to prevent regressions in the future?
- if we are already at this topic: would be quite cool if we could remove the [restriction not to allow source paths](https://github.com/dlang/dub/issues/909)

### DIFF
--- a/source/dub/recipe/io.d
+++ b/source/dub/recipe/io.d
@@ -53,11 +53,14 @@ PackageRecipe readPackageRecipe(Path filename, string parent_name = null)
 			to determine the file format from the file extension
 		parent_name = Optional name of the parent package (if this is a sub
 		package)
+		default_package_name = Optional default package name (if no package name
+		is found in the recipe this value will be used)
 
 	Returns: Returns the package recipe contents
 	Throws: Throws an exception if an I/O or syntax error occurs
 */
-PackageRecipe parsePackageRecipe(string contents, string filename, string parent_name = null)
+PackageRecipe parsePackageRecipe(string contents, string filename, string parent_name = null,
+								 string default_package_name = null)
 {
 	import std.algorithm : endsWith;
 	import dub.internal.vibecompat.data.json;
@@ -65,6 +68,8 @@ PackageRecipe parsePackageRecipe(string contents, string filename, string parent
 	import dub.recipe.sdl : parseSDL;
 
 	PackageRecipe ret;
+
+	ret.name = default_package_name;
 
 	if (filename.endsWith(".json")) parseJson(ret, parseJsonString(contents, filename), parent_name);
 	else if (filename.endsWith(".sdl")) parseSDL(ret, contents, parent_name, filename);
--- a/test/single-file-sdl-default-name.sh
+++ b/test/single-file-sdl-default-name.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+set -e
+cd ${CURR_DIR}
+rm -f single-file-sdl-default-name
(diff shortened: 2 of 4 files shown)


source/dub/recipe/io.d
```diff
@@ -66,6 +69,9 @@ PackageRecipe parsePackageRecipe(string contents, string filename, string parent
 
 	PackageRecipe ret;
 
+	if (default_package_name)
```
s-ludwig: Nitpick: the `if` is redundant here.

test/single-file-sdl-default-name.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -e
+cd ${CURR_DIR}
+rm -f single-file-sdl-default-name
+
+${DUB} run --single single-file-sdl-default-name.d --compiler=${DC}
+if [ ! -f single-file-test ]; then
```
s-ludwig: Shouldn't this be "single-file-sdl-default-name"?

s-ludwig:
Good idea and the implementation is in the right spot, AFAICS.

Regarding the white space issue, `.editorconfig` should usually avoid these mistakes and I wonder if they were introduced recently or before `.editorconfig` existed.

I'm not completely decided what to think about the source path topic. On one hand it would be cleaner to let single-file packages be what they are, self-contained in a single file. The effect of source paths can be achieved by making a (sourceLibrary) package out of the source folder and adding a dependency.

On the other hand, single-file packages are private things anyway (not listed on the registry etc.), so the need to be particularly pedantic about such things isn't really there either.

wilzbach:
> Regarding the white space issue, .editorconfig should usually avoid these mistakes and I wonder if they were introduced recently or before .editorconfig existed.

-> #1084

> I'm not completely decided what to think about the source path topic. On one hand it would be cleaner to let single-file packages be what they are, self-contained in a single file. The effect of source paths can be achieved by making a (sourceLibrary) package out of the source folder and adding a dependency.

Another possibility is to create the `dub.sdl` and use different configurations - though it just doesn't feel that "nice". For reference an example for:

- shared folder: would be [mir/examples/flex]([mir.random examples](https://github.com/libmir/mir/blob/a1db2f1931d19d284d56320725b1a812a316cdf6/examples/flex_plot/test_density_with_poles.d) (it does make sense here to use an own `folder`)
- `dub.sdl` + configuration: would be [dlang/tools/styles](https://github.com/dlang/tools/tree/master/styles)

> On the other hand, single-file packages are private things anyway (not listed on the registry etc.), so the need to be particularly pedantic about such things isn't really there either.

It just feels arbitrary to me and I bumped into this already a couple of times whenever I created a more complex script or two versions of one script and wanted to add some kind of another/shared file to the build. 
I agree that it's not a "best practice", but it would be pretty convenient from time to time ;-)

wilzbach:
@s-ludwig Do you already have hardened opinion on this?
[We plan](https://github.com/dlang-tour/dlang-tour/pull/530) to introduce DUB support for the dlang-tour, s.t.
- examples with DUB packages can be shown on the Tour
- DUB packages can implement a similar "run examples" feature like dlang.org for examples

For both use cases, it would be quite annoying to be required to always add a line like `name "a_runnable_example`"` to the DUB header ...

PR #1082, opened 2017-02-23
dub run <package> should search the registry on failure

wilzbach:
Adds a simple fallback to `dub run`, s.t. the following pattern simply works:

```
dub run dlang-tour
dub run dfmt
dub run dscanner
dub run avgtime
...
```

I am not sure how we can best test this _without_ querying the registry. Also I don't know any package that is compatible with all the compilers that DUB supports.
What would be the best strategy here?

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -821,6 +821,45 @@ class BuildCommand : GenerateCommand {
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		if (free_args.length >= 1) {
+			auto package_name = free_args[0];
+			auto pack = dub.packageManager.getFirstPackage(package_name);
+
+			if (!pack) {
+				bool input(string caption, bool default_value = true) {
+					writef("%s [%s]: ", caption, default_value ? "Y/n" : "y/N");
+					auto inp = readln();
+					string userInput = "y";
+					if (inp.length > 1)
+						userInput = inp[0 .. $ - 1].toLower;
+
+					switch (userInput) {
+						case "no", "n", "0":
+							return false;
+						case "yes", "y", "1":
+						default:
+							return true;
+					}
+				}
+
+				// search for the package and filter for exact matches
+				auto search = dub.searchPackages(package_name)
+					.map!(tup => tup[1].find!(p => p.name == package_name))
+					.filter!(ps => !ps.empty);
+				if (search.empty)
+					return 2;
+
+				auto p = search.front.front;
+				writefln("%s wasn't found locally, but it's available online:", package_name);
+				writefln("Description: %s", p.description);
+				writefln("Version: %s", p.version_);
+
+				bool answer = input("Do you want to fetch %s?".format(package_name));
+				auto dep = Dependency(p.version_);
+				dub.fetch(package_name, dep, dub.defaultPlacementLocation, FetchOptions.none);
+			}
+		}
+
 		return super.execute(dub, free_args, app_args);
 	}
 }
--- a/test/issue877-auto-fetch-package-on-run.sh
+++ b/test/issue877-auto-fetch-package-on-run.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -e -o pipefail
+$DUB remove --version="*" gitcompatibledubpackage || true
+
+echo "y" | $DUB run gitcompatibledubpackage -c exe
+
+$DUB remove gitcompatibledubpackage


s-ludwig:
What about making this slightly interactive and requiring the user to confirm the download. This would also be a chance to print the package description and maybe the github address, to make it less likely to make this en entry point for malicious code.

There is http://code.dlang.org/packages/gitcompatibledubpackage which is used in some of the tests already. We'd just have to add an executable configuration (in this case just a `source/app.d`).

wilzbach:
> What about making this slightly interactive and requiring the user to confirm the download. This would also be a chance to print the package description and maybe the github address, to make it less likely to make this en entry point for malicious code.

Good idea - I gave it a shot. It's not perfect because without more refactoring the `info` payload will get downloaded twice from the registry.

> There is http://code.dlang.org/packages/gitcompatibledubpackage which is used in some of the tests already. We'd just have to add an executable configuration 

https://github.com/MartinNowak/gitcompatibledubpackage/pull/1/files

I also added a test for this -> Travis will fail for a bit.

MartinNowak:
> https://github.com/MartinNowak/gitcompatibledubpackage/pull/1/files
> 
> I also added a test for this -> Travis will fail for a bit.

Merged

s-ludwig:
Appears that `issue877-auto-download-package-if-not-found-on-error.sh` is still missing an `echo "y" |`, or maybe a command line option that does the same. Ideally, it would also detect that stdin is empty (does that work using an EOF check?) and assume "n" in that case (would be backwards compatible and avoids unsolicited downloads).

test/issue877-auto-download-package-if-not-found-on-error.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e -o pipefail
+
+$DUB remove --version='*' cpuid > /dev/null 2>&1 || true
+$DUB run dfmt
```
MartinNowak: How is that different from the test below?

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+					logInfo("Getting a release version for %s failed: %s", package_name, e.msg);
+					logInfo("Retry with ~master...");
+					dep = Dependency("~master");
+					json = dub.info(package_name, dep, fetchOpts, supplier);
+				}
+
+				if (json == Json.emptyObject)
+					return 2;
+
+				writefln("%s wasn't found online, but there's version available online:", package_name);
```
MartinNowak: `dfmt wasn't found online, but there's version available online`???

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+				if ("authors" in json)
+					writefln("Authors: %-(%s, %)", json["authors"][].map!(v => v.get!string));
+				if ("license" in json)
```
MartinNowak: How about
```d
if (auto p = "license" in json)
    writefln("Description: %s", p.get!string);
```

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+					writefln("Description: %s", json["description"].get!string);
+				if ("version" in json)
+					writefln("Version: %s", json["version"].get!string);
+				if ("description" in json)
+					writefln("License: %s", json["license"].get!string);
```
MartinNowak: Description and License are mixed up.

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+					{
+						switch (inp[0 .. $ - 1].toLower) {
+						  case "yes", "y":
+							nb = true;
+						  case "no", "n":
+							nb = false;
+						  default:
+						}
+					} else {
+						nb = true;
```
MartinNowak: Should be the default value, no?

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+					writef("%s [%s]: ", caption, default_value ? "Y/n" : "y/N");
+					auto inp = readln();
+					if (inp.length > 1)
+					{
+						switch (inp[0 .. $ - 1].toLower) {
+						  case "yes", "y":
+							nb = true;
+						  case "no", "n":
+							nb = false;
+						  default:
```
MartinNowak: I'd rather loop in here, than on the outside. Also allows you to drop Nullable.

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
+				try {
+					json = dub.info(package_name, dep, fetchOpts, supplier);
+					if (json == Json.emptyObject)
+						throw new Exception("No version found.");
+				} catch(Exception e){
+					logInfo("Getting a release version for %s failed: %s", package_name, e.msg);
+					logInfo("Retry with ~master...");
+					dep = Dependency("~master");
+					json = dub.info(package_name, dep, fetchOpts, supplier);
+				}
```
MartinNowak: Could this maybe reuse `dub.searchPackages`?

source/dub/commandline.d
```diff
@@ -825,6 +825,83 @@ class RunCommand : BuildCommand {
...
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		import std.typecons : Nullable;
+
+		if (free_args.length >= 1)
+		{
+			auto package_name = free_args[0];
+			auto pack = dub.packageManager.getFirstPackage(package_name);
+
+			if (!pack)
```
MartinNowak: Please check some `--non-interactive` argument as well to prevent hanging scripts.

wilzbach:
Okay I finally found time to rebase and rework this. Thanks to @MartinNowak's feedback and the time passed since I submitted this, I managed to cut the size of this PR in half :)

This only needs a new release of `gitcompatibledubpackage` as https://github.com/MartinNowak/gitcompatibledubpackage/pull/1 isn't part of the dummy package's 1.0.2 release (@MartinNowak: how about moving this package to dlang-community?)

source/dub/commandline.d
```diff
@@ -819,6 +819,48 @@ class BuildCommand : GenerateCommand {
...
+					return 2;
+
+				auto p = search.front.front;
+				writefln("%s wasn't found locally, but it's available online:", package_name);
+				writefln("Description: %s", p.description);
+				writefln("Version: %s", p.version_);
+
+				bool answer = input("Do you want to fetch %s?".format(package_name));
+				auto dep = Dependency(p.version_);
+				dub.fetch(package_name, dep, dub.defaultPlacementLocation, fetchOpts);
```
s-ludwig: Nitpick: since the package doesn't exist locally, `FetchOptions.forceBranchUpgrade` shouldn't have an effect and `FetchOptions.none` would be sufficient here.

wilzbach:
I accidentally closed this PR because I wanted to trigger a notification to Semaphore while I pushed and GitHub doesn't seem to like this (even restoring the old commit sha still doesn't allow me to reopen).

Anyway, I opened a new PR -> https://github.com/dlang/dub/pull/1428
Sorry for the noise.

PR #1083, opened 2017-02-23
Add dub-test-library to .gitignore

wilzbach:
It would be great if we can add the `_version` also to the `.gitignore` - however IIRC someone wanted to build DUB without `git`?


s-ludwig:
Yeah, `version_.d` is rather nasty, to only for DUB itself. I'd like to integrate some kind of standard solution into DUB for that, which would just leave the bootstrapping case to solve manually.

PR #1093, opened 2017-03-10
Fix bogus rebuild of packages without dependencies.

### LINKED ISSUE #1091: What did happen to the "already up-to-date" message ?
Since a few weeks I don't see it anymore. Was it removed on purpose ?

s-ludwig:
Packages without dependencies don't have a dub.selections.json file. Previously the absence of this file was deemed to be a sign that the build output is not up to date, thus always unconditionally rebuilding such a package even if nothing changed. Fixes #1091.


BBasile:
thanks.

PR #1094, opened 2017-03-18
sublimetext: add dependencies as folders

AndrejMitrovic:
When generating the sublime-text project file, add the dependencies as the list of folders to
open in Sublime editor when the project is open.

This allows quick lookup of symbols in the libraries from within Sublime.

Note: Although these are already included in the `include_paths`, this doesn't make the symbols in these paths searchable within the editor. I'm not sure if there's a better way to do this, but the approach in this PR works.

### DIFF
--- a/source/dub/generators/sublimetext.d
+++ b/source/dub/generators/sublimetext.d
@@ -36,7 +36,7 @@ class SublimeTextGenerator : ProjectGenerator {
 		logDebug("About to generate sublime project for %s.", m_project.rootPackage.name);
 
 		auto root = Json([
-			"folders": targets.byValue.map!(f => targetFolderJson(f)).array.Json,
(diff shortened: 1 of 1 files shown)


s-ludwig:
I wonder if something like

    targets
        .map!(t => t.packages)
        .joiner
        .map!(p => p.basePackage)
        .array.sort.uniq
        .map!(p => p.targetFolderJson)
        .array.Json

wouldn't be the better approach (i.e. register all involved package root folders). Although I regularly work with Sublime Text, I didn't work with projects, so I'm not certain what possible implications there might be.

AndrejMitrovic:
> Although I regularly work with Sublime Text, I didn't work with projects

Neither do I, but it feels like this is the only way to make symbol lookup work with dependencies set by dub. When I work with submodules I don't have those issues (since everything is in `project/submodules/`), but I want to move to dub for good.

> I wonder if something like <code> wouldn't be the better approach

I'll test it out!

PR #1101, opened 2017-03-26
interactively query for license

MartinNowak:
- add select input (also for sdl/json)
- provide common license defaults (choosealicense.com options + Boost)
- order options by popularity

licenses
Suggestion: https://twitter.com/WebFreak001/status/845731612478332929

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -470,11 +470,51 @@ class InitCommand : Command {
 			free_args = free_args[1 .. $];
 		}
 
-		string input(string caption, string default_value)
+		static string input(string caption, string default_value)
 		{
 			writef("%s [%s]: ", caption, default_value);
 			auto inp = readln();
 			return inp.length > 1 ? inp[0 .. $-1] : default_value;
+		}
+
+		static string select(string caption, bool free_choice, string default_value, string[] options...)
+		{
+			assert(options.length);
+			import std.math : floor, log10;
+			auto ndigits = (size_t val) => val.log10.floor.to!uint + 1;
+
+			immutable default_idx = options.countUntil(default_value);
+			immutable max_width = options.map!(s => s.length).reduce!max + ndigits(options.length) + "  ".length;
+			assert(default_idx >= 0);
+			writeln(free_choice ? "Select or enter " : "Select ", caption);
+			foreach (i, option; options)
+			{
+				if (i && !(i % 4)) writeln();
+				writef("%u) %s", i, leftJustifier(option, max_width - ndigits(i + 1)));
+			}
+			writeln();
+			immutable default_choice = default_idx.to!string;
+			while (true)
+			{
+				auto choice = input(free_choice ? "?" : "#?", default_choice);
+				if (choice is default_choice)
+					return default_value;
+				choice = choice.strip;
+				uint option_idx = uint.max;
+				try
+					option_idx = to!uint(choice);
+				catch (ConvException)
+				{}
+				if (option_idx != uint.max)
+				{
+					if (option_idx < options.length)
+						return options[option_idx];
+				}
+				else if (free_choice)
+					return choice;
+				logError("Select an option between 0 and %u%s.", options.length - 1,
+						 free_choice ? " or enter a custom value" : null);
+			}
 		}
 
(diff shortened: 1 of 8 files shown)


s-ludwig:
I think this would be a mistake. Choosing a license needs to be a conscious process, and just assuming a license silently is likely to cause bad assumptions at some point (code.dlang.org would display "MIT", while in reality the author chose GPL 3, but never updated the entry, leading to license infringement). Couldn't we instead present a list of known licenses to choose from?

MartinNowak:
Puh :sweat:, done.

source/dub/commandline.d
```diff
@@ -470,31 +470,68 @@ class InitCommand : Command {
...
 			p.name = input("Name", p.name);
 			p.description = input("Description", p.description);
 			p.authors = input("Author name", author).split(",").map!(a => a.strip).array;
-			p.license = input("License", p.license);
+			p.license = select(
+				"a license (https://choosealicense.com/)",
+				free_choice,
+				p.license,
+				"MIT", "BSL-1.0 (Boost)", "public domain", "GPL-3.0", "BSD 3-clause", "BSD", "GPL-2.0", "LGPL-3.0", "APL-2.0 (Apache)", "MPL-2.0 (Mozilla)", "AGPL-3.0", "proprietary"
+			);
```
s-ludwig: If there was a `.splitter(' ').front` at the end and two adjustments ("APL-2.0 (Apache)"->"Apache-2.0", "BSD"->"BSD 2-clause") then this would match exactly what is listed in the package format specification. BTW, the short name of the Apache license according to Wikipedia is "ASL", not sure if we should use that or not.

s-ludwig:
That looks really nice, the single-input choice/custom value entry is a good idea. See no issues in the implementation.

source/dub/commandline.d
MartinNowak: Updated, not sure what you mean with `.splitter(' ').front`.

MartinNowak:
Could you please disable coveralls comments under https://coveralls.io/github/dlang/dub/settings? The status API should be enough.

s-ludwig:
For some reason this currently fails just for the older frontend versions:
```
[INFO] Running /home/travis/build/dlang/dub/test/0-init-interactive.sh...
Select a package recipe format
0) sdl   1) json  
#? [1]: Name [0-init-interactive]: Description [A minimal D application.]: Author name [travis]: Select or enter a license (https://choosealicense.com/)
0) MIT                 1) BSL-1.0 (Boost)     2) public domain       3) GPL-3.0             
4) BSD 3-clause        5) BSD 2-clause        6) GPL-2.0             7) LGPL-3.0            
8) Apache-2.0          9) MPL-2.0 (Mozilla)  10) AGPL-3.0           11) proprietary        
? [11]: Copyright string [Copyright © 2017, travis]: Add dependency (leave empty to skip) []: Successfully created an empty project in '/home/travis/build/dlang/dub/0-init-interactive'.
Package successfully created in 0-init-interactive
5c5
< license "pl"
---
> license "gpl"
Contents of generated dub.sdl not as expected.
```

MartinNowak:
> For some reason this currently fails just for the older frontend versions:

It's apparently because `parse!uint` ate the first char, even though it wasn't a digit.

MartinNowak:
Fixed and rebased @s-ludwig

MartinNowak:
Is this a candidate for master?

s-ludwig:
Usually, I'd say master, but in this case, if some careful manual testing also shows no issues, the scope is IMO limited enough to make an exception if we want to (I don't particularly mind the small release, but it would help to balance the 1.4.0 and 1.5.0 releases a bit). It's currently still based on master, though.

MartinNowak:
No need to hurry with this, retargeted to master and fixed 2.069 compat.

WebFreak001:
rebase pls, useful feature!

MartinNowak:
I don't have time to work on dlang atm., but feel invited to take over this PR @WebFreak001.

WebFreak001:
I don't have permissions to write to your branch

MartinNowak:
Just clone and make a new PR @WebFreak001

PR #1104, opened 2017-03-29
download / registry requests silently fail

MartinNowak:
Took me an extra hour of debugging why our builds fails ;).
At least for search the exception that libcurl couldn't be loaded is silently ignored, https://github.com/dlang/dub/blob/756240240e980f0945988ad77e0301b389b06db5/source/dub/packagesupplier.d#L229. But I'm also seeing `Root package assert_writeln_magic references unknown package libdparse` with no further indication of the underlying problem.

Based on v1.1.0, but still seems to be present.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -16,7 +16,7 @@ import dub.version_;
 
 // todo: cleanup imports.
 import core.thread;
-import std.algorithm : startsWith;
+import std.algorithm : canFind, startsWith;
 import std.array;
 import std.conv;
 import std.exception;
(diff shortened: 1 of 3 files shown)


ZombineDev:
There many occurrences of `catch (Exception e)` in DUB's codebase and we should eventually get rid of them, but for now this should help explain some silent failures.

MartinNowak:
Seems to require a bit more work https://travis-ci.org/dlang/dub/jobs/216294639#L724

wilzbach:
I am observing "X references unknown package Y" very often on AppVeyor, e.g.
https://ci.appveyor.com/project/greenify/dcd/build/1.0.13/job/6uw39f0u185qjeun (and the regarding [discussion](https://github.com/dlang-community/discussions/issues/13)), so more logging is very welcome and obviously this LGTM!

> Seems to require a bit more work

Hmm I see:

```
/home/travis/build/dlang/dub/test/issue616-describe-vs-generate-commands.sh: logError: command not found
Script failure.
```

But this function is properly defined here:

https://github.com/dlang/dub/blob/stable/test/issue616-describe-vs-generate-commands.sh
https://github.com/dlang/dub/blob/master/test/run-unittest.sh

(I tried restarting Travis)

source/dub/internal/utils.d
```diff
@@ -199,6 +199,38 @@ void runCommands(in string[] commands, string[string] env = null)
 	}
 }
 
+version(DubUseCurl) {
+	/++
+	 Exception thrown on HTTP request failures, e.g. 404 Not Found.
+	 +/
+	static if (__VERSION__ <= 2075) class HTTPStatusException : CurlException
```
s-ludwig: Will there be a `HTTPStatusException` in 2.076.0?

s-ludwig:
Have you tried to use this with a second (empty) registry? I didn't go through the individual errors in detail, but it could very well be that this could lead to lots of bogus error messages when a package exists in one registry but not in another.

source/dub/internal/utils.d
MartinNowak: Yes https://github.com/dlang/phobos/pull/5551.

MartinNowak:
Looks OK to me.
Basically we report any non-404 error on any of the provided registry, even if sth. succeeds on a different registry. So it is transparent for people if e.g. their configured registry isn't used and dub falls back to the standard registry.

MartinNowak:
I've changed this to only handle 404's in getMetadata, everything else relies on that method and checks for an undefined JSON as return value.

s-ludwig:
LGTM, but it should be retargeted to master to get some room for testing.

MartinNowak:
Done

s-ludwig:
@MartinNowak if you have mange access here (BTW, would sometimes be useful if I had), you'll have to override the merge due to the dmd-beta failure.

s-ludwig:
Restarted with the latest DMD beta.

PR #1122, opened 2017-05-07
Ask the user again in init when the name given isn't valid

NotSpooky:


source/dub/commandline.d
```diff
@@ -491,7 +491,18 @@ class InitCommand : Command {
 				}
 			}
 			auto author = p.authors.join(", ");
-			p.name = input("Name", p.name);
+			while (true) {
+				// Tries getting the name until a valid one is given.
+				import std.regex;
+				auto nameRegex = ctRegex!`^[a-z\-_]+$`;
```
s-ludwig: I'd usually avoid `ctRegex` except for performance critical cases, due to its heavy impact on compile time/memory use. The expression is so simple that it might be negligible in this case (I can't really tell without measuring), but since this is just an interactive loop, a runtime regex should definitely be fine.

s-ludwig:
Thanks!

PR #1125, opened 2017-05-09
some speedups of dub

MartinNowak:
- replace O(N^2) removeImportFiles w/ setDifference
- use de.isDir instead of isDir(de.name)


s-ludwig:
There is a subtle change of semantics here. To make it equivalent, instead of removing `sourceFiles` from `importFiles` directly, `dst.addSourceFiles(sourceFiles);` has to be called first, and then `dst.sourceFiles` needs to be removed from `importFiles`.

PR #1128, opened 2017-05-11
redo buildsettings inheritance

John-Colvin:
the previous system had some flaws that this pr addresses:

* Everything was inherited upwards first, meaning that all settings that were upwards inherited were in fact global. This causes a lot of churn with common dependencies between projects.\*

* ~~[minor] Needed two passes on the dependency tree~~ so does this pr, it passes through every target package to "commit" downwards inheritance at the end.

This pr ~~combines the two inheritance passes while also isolating~~ isolates downwards inheritance from upwards: nothing that is upwards inherited will subsequently be downwards inherited or vice-versa.

\* E.g. p0 depends on p2, p1 depends on p2 and p3, p3 defines version "p3ver". p2 is compiled with "p3ver" when building p1, meaning that the build can't be shared between p0 and p1.


John-Colvin:
I recommend reading the diff with `?w=1` i.e. https://github.com/dlang/dub/pull/1128/files?w=1

John-Colvin:
Are those old compiler versions needed any more? Should I remove the more recent code from my pr or remove the old compilers from the `.travis.yml` (the failing ones are from late 2014 and before)?

s-ludwig:
The problem is mainly with the latest GDC release. Apart from that, removing 2.066 support should be fine.

source/dub/generators/generator.d
```diff
@@ -143,126 +142,140 @@ class ProjectGenerator
...
-		logDiagnostic("Generate target %s (%s %s %s)", pack.name, buildsettings.targetType, buildsettings.targetPath, buildsettings.targetName);
 		if (is_target)
-			targets[pack.name] = TargetInfo(pack, [pack], configs[pack.name], buildsettings, null);
+			downSettingsMap[pack.name] = downBuildSettings.dup;
+
+		buildsettings.addVersions(["Have_" ~ stripDlangSpecialChars(pack.name)]);
 
 		auto deps = pack.getDependencies(configs[pack.name]);
-		foreach (depname; deps.keys.sort()) {
+		foreach (depname; deps.keys.dup.sort()) {
```
MartinNowak: `aa.keys` returns a fresh array, shouldn't be necessary.

source/dub/generators/generator.d
```diff
@@ -143,126 +142,140 @@ class ProjectGenerator
 	*/
 	protected void performPostGenerateActions(GeneratorSettings settings, in TargetInfo[string] targets) {}
 
-	private BuildSettings collect(GeneratorSettings settings, Package pack, ref TargetInfo[string] targets, in string[string] configs, ref string[] main_files, string bin_pack)
+	private BuildSettings collect(GeneratorSettings settings, Package pack, BuildSettings parentSettings, ref TargetInfo[string] targets, in string[string] configs, ref string[] main_files, string bin_pack, BuildSettings[string] downSettingsMap = null)
```
MartinNowak: Up and down are somewhat subjective terms for dependency trees.
Using parent and child should have a bit less potential for confusion.

source/dub/generators/generator.d
John-Colvin: done

source/dub/generators/generator.d
John-Colvin: I've changed it to specify "for dependents", is that ok?

John-Colvin:
Failures are spurious, could someone restart them please?

s-ludwig:
(Just noting that I still haven't had the chance to understand the changes in detail and that this potentially breaks an unknown amount of projects that may depend on the current behavior.)

MartinNowak:
I think one example of the described issue is rebuilding of vibe-d for each of it's test-cases (https://travis-ci.org/vibe-d/vibe.d/jobs/275883886#L719).

s-ludwig:
> I think one example of the described issue is rebuilding of vibe-d for each of it's test-cases (https://travis-ci.org/vibe-d/vibe.d/jobs/275883886#L719).

It doesn't do that, basically only once for normal `main()` and once for `VibeDefaultMain`.

John-Colvin:
I've started writing tests that serve as concrete example.

As a consequence, I am finding bugs, so

### DO NOT MERGE

MartinNowak:
Weird, I'm definitely seeing too many rebuilds for my taste here https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/PR-72/1/pipeline/163#step-313-log-1438.

source/dub/generators/generator.d
```diff
@@ -226,17 +227,22 @@ class ProjectGenerator
...
 
-		if (is_target)
-			forDependentsSettingsMap[pack.name] = forDependentsBuildSettings.dup;
+		if (auto p = pack.name in forDependenciesSettingsMap)
+		{
+			mergeFromDependents(forDependenciesBuildSettings, *p);
+			forDependenciesBuildSettings = *p;
+		}
+		else
+			forDependenciesSettingsMap[pack.name] = forDependenciesBuildSettings.dup;
```
MartinNowak: Mucho better.

MartinNowak:
Lots of guesswork below, lack the time to check the facts :/, so take this with a grain of salt.

Take a look at https://travis-ci.org/vibe-d/vibe.d/jobs/275883881#L1745, dub is rebuilding vibe-d for every example.
IIUC then it's because of the effect @John-Colvin tries to solve here, vibe.d gets rebuild with `--version=Have_Example1Name` and then again with `--version=Have_Example2Name`.
I agree with John that we should fix this.

With optional dependency support fix a while ago, I think that it's reasonable to only support `version (Have_vibe_d)` when an optional dependency exists.

We could scrape all packages on the registry and search for usages of versions from dependents, shouldn't be too many.

I'm still concerned about complexity somewhat, but the last iterations look a lot better.

s-ludwig:
> With optional dependency support fix a while ago, I think that it's reasonable to only support version (Have_vibe_d) when an optional dependency exists.

`Have_xxx` must be defined for any package that directly or indirectly depends on that library, because versions also affect modules that are merely imported by an intermediate dependency. But they should never be inherited upwards like normal versions, so that the mentioned `Have_Example1Name` should not affect how the vibe-d packages are built.

source/dub/generators/generator.d
```diff
@@ -143,130 +142,151 @@ class ProjectGenerator
...
-		bool generates_binary = tt != TargetType.sourceLibrary && tt != TargetType.none;
-		bool is_target = generates_binary || pack is m_project.rootPackage;
+		BuildSettings buildsettings, forDependenciesBuildSettings;
+		bool is_target;
 
-		if (tt == TargetType.none) {
-			// ignore any build settings for targetType none (only dependencies will be processed)
-			shallowbs = BuildSettings.init;
-			shallowbs.targetType = TargetType.none;
+		void setIs_target(TargetType tt)
```
MartinNowak: Nit: `setIsTarget`

source/dub/generators/generator.d
```diff
@@ -143,130 +142,151 @@ class ProjectGenerator
...
 		}
+		else {
+			buildsettings = packInTargets.buildSettings;
+			forDependenciesBuildSettings = buildsettings.dup;
+			setIs_target(buildsettings.targetType);
+		}
+		mergeFromDependents(parentSettings, forDependenciesBuildSettings);
 
-		if (is_target) targets[pack.name].buildSettings = buildsettings.dup;
+		if (auto p = pack.name in forDependenciesSettingsMap)
```
MartinNowak: What's the point of this `forDependenciesSettingsMap`? Are you caching data flows from top to bottom, so it can be reused for multiple children?
I guess we're doing a depth-first-order traversal, why isn't it enough to dup before recursing? Diamond dependencies?

test/buildSettingsInheritanceCheck/diamond/p1/dub.sdl
```diff
@@ -0,0 +1,5 @@
+name "p1"
+versions "B"
```
MartinNowak: Better use a matching version `P1`, so it's easier to understand where the version comes from.
How about using `{grand,}{parent,mother,father}`, `daughter`, and/or `son` as terminology. Might simplify understanding.

MartinNowak:
Cyclomatic complexity is thriving ;).
https://github.com/John-Colvin/dub/blob/21c0d0a78a9f57fbf267b0811db7bc41063ff2c5/source/dub/generators/generator.d#L145-L290

MartinNowak:
> Have_xxx must be defined for any package that directly or indirectly depends on that library, because versions also affect modules that are merely imported by an intermediate dependency. But they should never be inherited upwards like normal versions, so that the mentioned Have_Example1Name should not affect how the vibe-d packages are built.

Yes, my understanding as well, but apparently not what dub does.

> E.g. p0 depends on p2, p1 depends on p2 and p3, p3 defines version "p3ver". p2 is compiled with "p3ver" when building p1, meaning that the build can't be shared between p0 and p1.

Would explain the hefty test time increase after splitting vibe-d into packages.

source/dub/generators/generator.d
```diff
@@ -143,130 +142,151 @@ class ProjectGenerator
...
-		processVars(buildsettings, m_project, pack, shallowbs, true);
+		static auto mergeFromDependents(BuildSettings parent, ref BuildSettings child) {
+			child.addVersions(parent.versions);
+			child.addDebugVersions(parent.debugVersions);
+			child.addOptions(BuildOptions(cast(BuildOptions)parent.options & inheritedBuildOptions));
+
+			// special support for overriding string imports in parent packages
+			// this is a candidate for deprecation, once an alternative approach
+			// has been found
+			if (child.stringImportPaths.length) {
```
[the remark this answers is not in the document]
s-ludwig: Since when has that been? There are large piles of code relying on this. Before removing this, there needs to be some form of replacement established. I wouldn't expect this to be possible earlier than two years from now.

source/dub/generators/generator.d
```diff
@@ -143,130 +142,151 @@ class ProjectGenerator
...
+			return child;
 		}
 
-		// set pic for dynamic library builds.
-		if (buildsettings.targetType == TargetType.dynamicLibrary)
-			buildsettings.addOptions(BuildOption.pic);
+		auto packInTargets = pack.name in targets;
+		if (packInTargets is null) {
+			// determine the actual target type
+			auto shallowbs = pack.getBuildSettings(settings.platform, configs[pack.name]);
```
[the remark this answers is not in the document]
s-ludwig: This is the raw build settings as defined in the package recipe, filtered for the actual platform/config. It doesn't include settings inherited from other packages or settings based on the current build mode.

source/dub/generators/generator.d
John-Colvin: Diamonds. Packages get visited more than once, must remember and re-use results from the previous visit(s).

John-Colvin:
Al the rebuilds are probably because I made a lot of mistakes and force pushes

MartinNowak:
> But they should never be inherited upwards like normal versions

What do you mean by that? Setting `version "VibeCustomMain"` in an app will be visibie to all of the app's dependencies (children, downwards), but a `version "UseNativeSQLDriver"` in one of the dependencies will not bubble up to the application or affect siblings.
That's my understanding at least.

source/dub/generators/generator.d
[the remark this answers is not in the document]
MartinNowak: Kind of guessed that, there is also [BuildSettingsTemplate](https://github.com/dlang/dub/blob/255209da6491f98df602b121a8effe2c1fc94e69/source/dub/recipe/packagerecipe.d#L136).
Somewhat confusing to use that term both in the receipt and in the compiler domain.

Would (Abstract/Generic)CompilerArgs or CompilerSettings be a more appropriate name?
The term Target is also fairly well know in build tools and seems to be a good fit.

I'm stopping here as this is off-topic, but unclear naming is a major hurdle for me to work on dub or review complex PRs.

s-ludwig:
>> But they should never be inherited upwards like normal versions
>
> What do you mean by that? Setting version "VibeCustomMain" in an app will be visibie to all of the app's dependencies (children, downwards), but a version "UseNativeSQLDriver" in one of the dependencies will not bubble up to the application or affect siblings.
> That's my understanding at least.

I messed up the nomenclature there. I'd be glad to have something that isn't so ambiguous as up/down. Anyway, the correct picture is this:
grafik


- `VibeCustomMain` - "down": Right, we need to think about changing this inheritance to be explicit at one point to avoid affecting the whole dependency graph in cases where this is not intended. Could be declared specific to a certain dependency, or maybe just a binary choice - inherit to all dependencies or none. Otherwise, any version will affect the whole partical dependency graph of a package (not siblings, though, because it happens *after* upwards inheritance)

- `UseNativeSQLDriver` - "up": This always needs to be inherited to the dependees, because otherwise a `version` statement imported by a dependee would behave differently than the same one compiled as part of the dependency

- `Have_xxx` - "up": These should only ever be inherited by dependees, but should never be pushed down to the dependencies, as otherwise there would be an inherent dependency cycle (dependecies being able to do things differently depending on the dependees and vice-versa). This is why these are added last in the current code (didn't investigate yet if there really is a bug there)

So, the idea of the existing algorithm is a three step approach:
- Inherit all versions upwards while collecting dependencies
- Perform a single downwards inheritance pass to support configuring dependencies
- Finally, add `Have_xxx` versions for each package with the exact set of direct or indirect dependencies (indirect ones are dictated by the always-upwards rule)

source/dub/generators/generator.d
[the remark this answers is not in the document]
s-ludwig: I'd say not really, because it really does include all build settings, not just compiler specific ones. But `BuildSettingsRecipe` would actually be a better fit than "template" (the term "recipe" was introduced later but arguably does a much better job explaining what it is).

source/dub/generators/generator.d
s-ludwig: I don't see a reason why versions shouldn't simply always be additive. That's how they are designed in the language and the library can easily check for conflicting semantics. IMO it would unnecessarily complicate DUB to start assigning semantics on that level.

MartinNowak:
> I messed up the nomenclature there. I'd be glad to have something that isn't so ambiguous as up/down.

It's a common misunderstanding when talking about trees and roots, because the tree metaphor doesn't work too well ;).
tree_up

MartinNowak:
> VibeCustomMain - "down": Right, we need to think about changing this inheritance to be explicit at one point to avoid affecting the whole dependency graph in cases where this is not intended. Could be declared specific to a certain dependency, or maybe just a binary choice - inherit to all dependencies or none. Otherwise, any version will affect the whole partical dependency graph of a package (not siblings, though, because it happens after upwards inheritance)

As said above https://github.com/dlang/dub/pull/1128#discussion_r141041219, it seems that sub-configurations are the better tool to achieve this.

> UseNativeSQLDriver - "up": This always needs to be inherited to the dependees, because otherwise a version statement imported by a dependee would behave differently than the same one compiled as part of the dependency

Right, this is mostly due to templates and inlining, which smears such compiler settings across APIs.

So the change that John wants to implement is to not inherit settings that bubbled up from dependencies. Sounds reasonable and shouldn't cause problems as `p2` and `p3` in the op example have no relation (cannot import each other).

Shouldn't the following fix all of those problems?

1. Collect all dependencies/descendants
2. Inherit versions from dependents/ancestors to their dependencies/descendants (capture phase, downwards)
3. Add Have_xxx versions for each package with the exact set of **direct** depdendencies/children
4. Bubble up versions/settings from dependencies/descendants to dependents/ancestors (bubble phase, upwards)

While many parts could be merged into a single tree traversal, I'd be surprised if the tree is huge enough to justify any read-/maintainability trade-offs.
So personally I'd rather do 3 traversals (1., 2., 4.) and 1 loop (3.) than to mash everything into a single recursive function.

This obviously requires to cache buildsettings for each individual package.

Furthermore it seems we only need a single mergeSettings primitive, as the operation performed during up and down traversal are the same.

source/dub/generators/generator.d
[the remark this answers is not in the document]
MartinNowak: Sorry, seems like I remembered this incorrectly or confused the issue. Thought this was a bug, but it's actually handled correctly. Dub determines the highest package for each string import file.
What remains is that this pollutes the string imports. If a parent overrides a single file, dependencies get `-Jparentpath` and can import anything from the parent.

source/dub/generators/generator.d
MartinNowak: What I meant is *being used additively*, so EnableFeatureX instead of DisableFeatureX, that way conflicts in diamond dependencies should be less of a problem.
Indeed I wouldn't want dub to have anything to do with this, but it might be a best practice for library developers. But even more I'd suggest people to use configurations as a cleaner mechanism instead.

s-ludwig:
> Collect all dependencies/descendants
> 1. Inherit versions from dependents/ancestors to their dependencies/descendants (capture phase, downwards)
> 2. Add Have_xxx versions for each package with the exact set of direct depdendencies/children
> 3. Bubble up versions/settings from dependencies/descendants to dependents/ancestors (bubble phase, upwards)

Yes, that should improve the situation for unrelated sub graphs. It will still not change the situation in case of diamond(like) dependencies, but it should still be correct/safe.

To make any further improvements, as far as I see, we need support in the recipe for more specific downwards version definitions (e.g. `versions "VibeDefaultMain" package="vibe-d"` or `versions "Foo" inherit="downwards"`/`"upwards"`/`"none"`). The interesting part is how to make the transition to `inherit="upwards"` being the default.

> Furthermore it seems we only need a single mergeSettings primitive, as the operation performed during up and down traversal are the same.

This sounds like `BuildSettings.add(BuildSettings)`, but not all settings may be inherited in all cases. Settings that affect the linker often need to be handled separately from compiler settings and some compiler settings don't have to be inherited at all, because they don't affect the ABI or compiler semantics (e.g. `-O`). In case of downwards inheritance, only the (debug) versions and string imports, as well as a few select compiler options are affected.

MartinNowak:
> To make any further improvements, as far as I see, we need support in the recipe for more specific downwards version definitions (e.g. versions "VibeDefaultMain" package="vibe-d" or versions "Foo" inherit="downwards"/"upwards"/"none"). The interesting part is how to make the transition to inherit="upwards" being the default.

I'd rather not complicate packages (and dub) even further. Let's defer this for a while, maybe we can find a better solution.
Would (sub-)configurations be an option? Their main drawback is that they cannot be combined atm. (https://github.com/dlang/dub/issues/1258), but they might be a cleaner solution for downward-inherited versions, as that's basically just configuring dependencies.

s-ludwig:
Yes, I was thinking about a "features" variant of "sub configurations" that is additive. Sub configurations are exclusive by nature and trying to change that is going to result in really messy semantics, but it's obvious that an additive variant is also a common use case. In theory, sub packages and optional dependencies could be used instead, but I'd rather not facilitate their broad use with the problems that they have.

MartinNowak:
One implication is that it cannot be done with a single tree traversal, as we first need to aggregate settings from *all* parents during the capture/downwards phase to correctly support diamond dependencies, otherwise one of the parents will use the library with different version flags than the ones used to compile the library.

I'm having a go at rewriting the implementation and *this* PR with the algorithm discussed above.

source/dub/generators/generator.d
MartinNowak: Wouldn't work that way as it has to support overriding string imports of transitive dependencies. The original version was checking against the root package, but after all import paths had been upwards inherited.

MartinNowak:
see #1263

Issue #1130, opened 2017-05-15
 The directory name is invalid.


andre2007:
specifying sub packages and single file execution does not work together.
Dub tries to create a directory with the name vibe-d:http.
As colons are invalid in file system, this will fail:

```
/+ dub.sdl:

name "app"
dependency "vibe-d:http" version=">=0.0.1"
+/

void main(){}
```

dub app.d
....dub\build\vibe-d:utils-0.7.31: The directory name is invalid.

MartinNowak:
Looks like on Windows, right?

andre2007:
Actually the issue is caused by the temporary folder argument and the bug is located here:
https://github.com/dlang/dub/blob/master/source/dub/generators/build.d#L162

If the temporary folder argument is set, the folder path is build by the pack.name.
To solve the issue just replace bad characters (:) with other characters.

it is a windows only bug if ":" is allowed on linux paths

PR #1137, opened 2017-05-26
temp-build: Fix colon in folder path

andre2007:
Fix issue #1130 

While using the --temp-build argument the folder path is build by using package name. This will fail if the package is a sub package as it containts a colon ":" in its name attribute.
The fix is, to use the base package name instead of the sub package name.


MartinNowak:
Thx, looks reasonable.

PR #1055, opened 2017-01-26
add osx testing

John-Colvin:

PR #1138, opened 2017-05-29
Update Travis config file to test the latest compiler releases

wilzbach:
Follow-up to https://github.com/dlang/dub-registry/issues/213

Imho it's very important that DUB is always compilable with the latest releases (and even betas or nightlies). As updating the `travis.yml` tends to be forgotten (as it has been here), I added non-hardcoded versions of all compilers and "beta" where available.
AFAICT `FRONTEND` versions is hard-coded into the build executable, s.t. builds can access it as environment variable. Hence it shouldn't matter for the CI and if, the daily cron job would be a gentle reminder ;-)
So in any case, @s-ludwig could you active daily cron jobs for `dub` and `dub-registry`? Thanks!

https://travis-ci.org/dlang/dub/settings
https://travis-ci.org/dlang/dub-registry/settings

image


wilzbach:
There's already an outstanding PR for OSX: https://github.com/dlang/dub/pull/1055/files

wilzbach:
Hmm, this already uncovered an upcoming issue with the newest gdc release:

```
/home/travis/dlang/gdc-4.8.5/lib/gcc/x86_64-unknown-linux-gnu/4.8.5/include/d/core/atomic.d:1511:13: error: static assert  "Invalid template type specified."
             static assert(0, "Invalid template type specified.");
             ^
../../../.dub/packages/vibe-d-0.7.31/vibe-d/source/vibe/http/server.d:1597:14: note: instantiated from here: atomicStore!(cast(MemoryOrder)5, shared(HTTPServerContext)[], shared(HTTPServerContext)[])
   atomicStore(g_contexts, newarr);
              ^
gdc failed with exit code 1.
```

s-ludwig:
> Imho it's very important that DUB is always compilable with the latest releases (and even betas or nightlies)

Since DUB is tested through the Jenkins run, it's at least not critical anymore to have Travis also test it, but even just for the other compilers of course that is nice to have. I've enabled the cron feature now.

> There's already an outstanding PR for OSX: https://github.com/dlang/dub/pull/1055/files

I'm not sure here if the trade-off is worth it. OS X builds tend to sit there for multiple hours depending on the time of day, which would seriously slow down the workflow and on the other hand I don't remember an OS X specific bug for DUB.

dub.sdl
```diff
@@ -22,13 +22,13 @@ configuration "library" {
 }
 
 configuration "library-nonet" {
-	dependency "vibe-d:http" version=">=0.7.30 <=0.9.0" optional=true
+	dependency "vibe-d:http" version=">=0.7.31 <=0.9.0" optional=true
```
s-ludwig: This is probably not really necessary, as DUB only uses very basic functionality (it should be compatible with very old versions actually). The interesting place is just in dub.selections.json.

wilzbach: It is necessary for 2.074.0 and higher due to https://issues.dlang.org/show_bug.cgi?id=17131
I am perfectly fine with just editing the `dub.selections.json`, but it seemed to me that these two files were kept in sync?

wilzbach:
> Since DUB is tested through the Jenkins run, it's at least not critical anymore to have Travis also test it,

Well while in theory that's true, unfortunately in practice the Jenkins instance is broken since more than one month and it never worked for more than a couple of days so far.

> I've enabled the cron feature now.

:+1:

dub.sdl
s-ludwig: Hitting 17131 would pose a natural limit, but it could still be that someone, for some reason, is stuck on vibe.d 0.7.30 while using an older compiler. While it would be even nicer to actually test the compatibility, it seems just like an unnecessary restriction to *require* 0.7.31.

The recipe is not strictly kept in sync with the selections, but I did bump the numbers in the recipe from time to time, basically when the numbers became too old to be sure that it would still work without testing it.

s-ludwig:
> Well while in theory that's true, unfortunately in practice the Jenkins instance is broken since more than one month and it never worked for more than a couple of days so far.

I just noticed that when I wanted to make sure that DUB was indeed on the list of tested proects :)

wilzbach:
> Hitting 17131 would pose a natural limit, but it could still be that someone, for some reason, is stuck on vibe.d 0.7.30 while using an older compiler. While it would be even nicer to actually test the compatibility, it seems just like an unnecessary restriction to require 0.7.31.

Okay I rebased the PR to only modify the `dub.selection.json` and as mentioned before allow failures with the latest `gdc`. I also squashed everything into one commit.

wilzbach:
> I just noticed that when I wanted to make sure that DUB was indeed on the list of tested proects :)

It is:

https://github.com/dlang/ci/blob/master/pipeline.groovy#L173

but as seen on the other PRs Jenkins wasn't able to catch the regressions manifesting in Vibe.d.

s-ludwig:
Okay, thanks!

> but as seen on the other PRs Jenkins wasn't able to catch the regressions manifesting in Vibe.d.

Good point.

PR #1139, opened 2017-05-29
Add a simple Windows CI

wilzbach:
Analog to https://github.com/rejectedsoftware/vibe.d/pull/1773 and https://github.com/dlang-community/libdparse/pull/141

Daily crons to test the latest releases for the win!

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,124 @@
+platform: x64
+environment:
+ matrix:
+  - DC: dmd
+    DVersion: nightly
+    arch: x64
+  - DC: dmd
+    DVersion: nightly
+    arch: x86
+  - DC: dmd
+    DVersion: beta
+    arch: x64
+  - DC: dmd
+    DVersion: beta
+    arch: x86
+  - DC: dmd
+    DVersion: stable
+    arch: x64
+  - DC: dmd
+    DVersion: 2.072.2
+    arch: x86
+  - DC: dmd
+    DVersion: 2.072.2
+    arch: x64
+  - DC: dmd
+    DVersion: stable
+    arch: x86
+  - DC: ldc
+    DVersion: beta
+    arch: x64
+  - DC: ldc
+    DVersion: stable
+    arch: x64
+  - DC: ldc
+    DVersion: 1.2.0
+    arch: x64
+
+skip_tags: false
+branches:
+  only:
+    - master
+
+install:
+  - ps: function ResolveLatestDMD
+        {
+            $version = $env:DVersion;
+            if($version -eq "stable") {
+                $latest = (Invoke-WebRequest "http://downloads.dlang.org/releases/LATEST").toString();
+                $url = "http://downloads.dlang.org/releases/2.x/$($latest)/dmd.$($latest).windows.7z";
+            }elseif($version -eq "beta") {
+                $latest = (Invoke-WebRequest "http://downloads.dlang.org/pre-releases/LATEST").toString();
+                $latestVersion = $latest.split("-")[0].split("~")[0];
+                $url = "http://downloads.dlang.org/pre-releases/2.x/$($latestVersion)/dmd.$($latest).windows.7z";
+            }elseif($version -eq "nightly") {
+                $url = "http://nightlies.dlang.org/dmd-master-2017-05-20/dmd.master.windows.7z"
+            }else {
+                $url = "http://downloads.dlang.org/releases/2.x/$($version)/dmd.$($version).windows.7z";
+            }
+            $env:PATH += ";C:\dmd2\windows\bin;";
+            return $url;
+        }
+  - ps: function ResolveLatestLDC
+        {
+            $version = $env:DVersion;
+            if($version -eq "stable") {
+                $latest = (Invoke-WebRequest "https://ldc-developers.github.io/LATEST").toString().replace("`n","").replace("`r","");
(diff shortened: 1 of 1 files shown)


wilzbach:
This PR needs AppVeyor to be enabled.

wilzbach:
Hmm DUB should really implement more sophisticated fallback and retry mechanism for errors as its own test-suite randomly fails due to temporary network errors:

```
[INFO] Running /home/travis/build/dlang/dub/test/issue672-upgrade-optional.sh...
Upgrading project in /home/travis/build/dlang/dub/test/issue672-upgrade-optional
Fetching dub 1.1.0 (getting selected version)...
Couldn't connect to server on handle 1DCFDC0
Script failure.
```

wilzbach:
> This PR needs AppVeyor to be enabled.

@s-ludwig @MartinNowak r u okay with this?

ohdatboi:
I don't know why this isn't merged yet, because testing on windows is very important for cross-platform project like dub.
Also, who builds binaries for http://code.dlang.org/download? It is possible to set up building and deployment of binary releases with AppVeyor and Travis CI.

s-ludwig:
I've enabled AppVeyor for the repository now. Let's see what happens.

wilzbach:
Oh the idea was to enable AppVeyor and then play with this PR until it's green. I will do this dog fooding on the weekend, but in case anyone wants to pick this up earlier - I definitely won't stop you (:

s-ludwig:
The only issue appears to be that the OMF curl.lib is used when building with LDC: https://ci.appveyor.com/project/s-ludwig/dub

BTW, kind of stupid that AppVeyor binds repositories to the user that registered them instead of the organization that they belong to.

MartinNowak:
Good start, hope this will evolve to a full install script at some point.
```ps1
(new-object Net.WebClient).DownloadString("https://dlang.org/install.ps1") | iex
```
It seems that `Net.WebClient` is slightly more recommended than `Invoke-Webrequest`.
https://blog.jourdant.me/post/3-ways-to-download-files-with-powershell

appveyor.yml
```diff
@@ -0,0 +1,124 @@
...
+            }
+            elseif($env:DC -eq "ldc"){
+              echo "downloading ...";
+              $url = ResolveLatestLDC;
+              echo $url;
+              Invoke-WebRequest $url -OutFile "c:\ldc.zip";
+              echo "finished.";
+              pushd c:\\;
+              7z x ldc.zip > $null;
+              popd;
```
MartinNowak: Windows can unpack zip files natively (they are treated like folders).
See https://gist.github.com/MartinNowak/8270666#file-setup-ps1-L32-L35, 0x614 disables UI.

appveyor.yml
```diff
@@ -0,0 +1,124 @@
...
+            }elseif($version -eq "beta") {
+                $latest = (Invoke-WebRequest "http://downloads.dlang.org/pre-releases/LATEST").toString();
+                $latestVersion = $latest.split("-")[0].split("~")[0];
+                $url = "http://downloads.dlang.org/pre-releases/2.x/$($latestVersion)/dmd.$($latest).windows.7z";
+            }elseif($version -eq "nightly") {
+                $url = "http://nightlies.dlang.org/dmd-master-2017-05-20/dmd.master.windows.7z"
+            }else {
+                $url = "http://downloads.dlang.org/releases/2.x/$($version)/dmd.$($version).windows.7z";
+            }
+            $env:PATH += ";C:\dmd2\windows\bin;";
```
MartinNowak: Wouldn't prepending be a better choice here?

appveyor.yml
```diff
@@ -0,0 +1,124 @@
...
+            $version = $env:DVersion;
+            if($version -eq "stable") {
+                $latest = (Invoke-WebRequest "http://downloads.dlang.org/releases/LATEST").toString();
+                $url = "http://downloads.dlang.org/releases/2.x/$($latest)/dmd.$($latest).windows.7z";
+            }elseif($version -eq "beta") {
+                $latest = (Invoke-WebRequest "http://downloads.dlang.org/pre-releases/LATEST").toString();
+                $latestVersion = $latest.split("-")[0].split("~")[0];
+                $url = "http://downloads.dlang.org/pre-releases/2.x/$($latestVersion)/dmd.$($latest).windows.7z";
+            }elseif($version -eq "nightly") {
+                $url = "http://nightlies.dlang.org/dmd-master-2017-05-20/dmd.master.windows.7z"
```
MartinNowak: http://nightlies.dlang.org/dmd-nightly/LATEST.
In fact you want to redirect any named branch to http://nightlies.dlang.org/dmd-<name>/LATEST, then download http://nightlies.dlang.org/dmd-<result>/dmd.<result-branch>.windows.7z.
A 2-level lookup as in install.sh should work best.
https://github.com/dlang/installer/blob/efbba50ad7102bf1eb5f25c5886b5cf128cbcaea/script/install.sh#L397
First resolving the specific version, the downloading that.
That also allows to directly download a specific build.

appveyor.yml
```diff
@@ -0,0 +1,124 @@
+platform: x64
+environment:
+ matrix:
+  - DC: dmd
+    DVersion: nightly
```
MartinNowak: Would be great if we could stick to the established `dmd-nightly` et.al. names.

MartinNowak: Seems like powershell has the necessary wildcard switch.
https://technet.microsoft.com/en-us/library/ff730937.aspx#EGC

Issue #855, opened 2016-05-27
provide a man page


wilzbach:
As dub is getting bigger providing a Linux man page would be a nice to have. Ideally there would be a nice (generic) D tool that can transform all flags into the man page format.

AFAIK it's even mandatory on Debian to provide a man page? - ping @ximion

ximion:
Not mandatory, but a strong "should". Writing manual pages in Troff isn't really much fun, most modern manpages are written in Docbook XML and then transformed to a manpage at compile-time.
That also allows for easily translatable manpages. Creating the manpage from XML is then as easy as running `xmlto man *.xml`.
For Debian, I went the lazy route and auto-generated a manpage from the --help output, but for maintaining this upstream I wouldn't encourage doing that, because first of all the auto-generated output isn't a good manpage and I needed to edit it again, and it also can break at random.
A manpage written directly by humans with useful examples is much more useful.
Maybe dub could even automate the manpage generation via xmlto and install it correctly if `dub install` (issue #839) is implemented.

s-ludwig:
For auto-generation, `dub.commandline` provides an API that can be used to query all the help text in semantic pieces (used to provide the [online docs](http://code.dlang.org/docs/commandline)).

ximion:
That is pretty awesome! But I fear a manpage that is a 1:1 copy of `dub --help` wouldn't be that helpful ^^

s-ludwig:
Probably true ;) But maybe we can combine that in a smart way with some more in-depth documentation. The online documentation still needs that, too.

ximion:
Can Ddoc generate manpages?
FTR, the current manpage we use at Debian can be found here: https://sources.debian.net/src/dub/0.9.24-2/debian/dub.1/ (but as said, it's nothing I would take upstream in that form ^^)

EmTee70:
I have installed via: sudo curl -fsS https://dlang.org/install.sh | bash -s dmd
Is there a manpage now or not? 
Better a short manpage with a link to only documentation than nothing!
Even just the help text is better than:  no manpage found for dub.

wilzbach:
FYI @ all packagers, we now generate man pages automatically - see https://github.com/dlang/dub/pull/1342

> Is there a manpage now or not?

There's, but not shipped with the install scripts at the moment.

ximion:
@wilzbach Good to know, I wasn't aware of that yet - will be added with a later release of the Debian package.

Geod24:
[...]

PR #1143, opened 2017-06-05
Manpages

mpevnev:
I've adapted --help messages and recipe file specifications into manpages. Please check them for correctness, especially the *-override, dub.json and dub.sdl pages. With *-overrides it's mostly a guesswork, since --help message is empty apart of options, and dub.json and dub.sdl pages are huge, something might have slipped under the radar. If anything needs to be added/reworded, please inform me.

### DIFF
--- a/man/dub-add-local.1
+++ b/man/dub-add-local.1
@@ -1,0 +1,41 @@
+
+.TH DUB-ADD-LOCAL "1"
+
+.SH NAME
+
+dub\-add\-local \- add a local package directory for dependency resolution
+
+.SH SYNOPSIS
+
+.B dub add\-local
+.I PATH
+[\fIVERSION\fR]
+[\fIOPTIONS\&.\&.\&.\fR]
+
+.SH DESCRIPTION
+
+Adds a local package directory to be used during dependency resolution\&. This
+command is useful for registering local packages, such as GIT working copies,
+that are either not available in the package registry, or are supposed to be
+overwritten\&.
+
+The version of the package is either determined automatically or can be
+explicitly overwritten by passing a version on the command line\&.
+
+See \fIdub-add-path\fR(1) for a way to register multiple local packages at
+once\&.
+
+
+.SH OPTIONS
+
+Apart of the options listed below, this command also honours common options 
+listed in \fIdub\fR(1)\&.
+
+.TP
+.B \-\-system
+Register system\-wide instead of user\-wide\&.
+
+.SH SEE ALSO
+
+\fIdub\fR(1), \fIdub\-add\-path\fR(1), \fIdub\-remove\-path\fR(1),
+\fIdub\-remove\-local\fR(1)
--- a/man/dub-add-override.1
+++ b/man/dub-add-override.1
@@ -1,0 +1,32 @@
+
+.TH DUB-ADD-OVERRIDE "1"
+
+.SH NAME
+
+dub\-add\-override \- use one package instead of another
+
+.SH SYNOPSIS
+
+.B dub add\-override
+.I PACKAGE
+.I VERSION\-SPEC
+.I TARGET\-PATH/TARGET\-VERSION
+[\fIOPTIONS\&.\&.\&.\fR]
+
+.SH DESCRIPTION
+
+Tells DUB to use one package instead of another when resolving dependencies and
+building\&.
+
+.SH OPTIONS
+
+Apart of the options listed below, this command also honours common options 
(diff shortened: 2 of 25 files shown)


wilzbach:
Hmm while this is pretty cool work, I worry quite a bit that this will be outdated soon.
Would it be possible to automatically generate the Manpages, maybe analogous to how the DUB registry displays the docs? 

https://github.com/dlang/dub-registry/blob/master/source/dubregistry/web.d#L591
https://github.com/dlang/dub-registry/blob/master/views/docs.commandline.dt

You might also want to have a look at `showHelp` directly at DUB:

https://github.com/dlang/dub/blob/master/source/dub/commandline.d#L1881

mpevnev:
I've tried help2man, but results were not pretty. However, the structure of help messages is quite regular, so it should be possible to write a parser with some regexes. Any such program would require a bit of manual work to make its output look nicer, but it should help with keeping manpages up-to-date. Give me a few days, this sounds fun.

MartinNowak:
Thanks, nice initiative.

Well the commands and help are already available programmatically, so no parser is needed. Just a tool (in `genmanpages/`) that links against dub (`dependency "dub" path="../"`), calls [`getCommands()`](https://github.com/dlang/dub/blob/b7e48d3f8bb7d0d3d912682f8cff95afc866fd96/source/dub/commandline.d#L43), and produces the man pages from those.
The result in the dub-registry isn't far off from a man page http://code.dlang.org/docs/commandline.

mpevnev:
@MartinNowak Well, by the time you've written this I've nearly finished a parser-based generator. It does most of the work well, but some things would have to be done manually: short descriptions in NAME section are left blank because I have no idea how to extract them from long descriptions and lists in option descriptions are clumped together rather unelegantly. The tool repo is located [here](https://github.com/mpevnev/dub-genmanpage). I'll see what I can do about lists, though.

wilzbach:
> I've nearly finished a parser-based generator. It does most of the work wel

If you are searching for an application of your parser, generating the Manpages and/or the resulting web documentation for DMD frontend, would be awesome.
It's currently entirely maintained by hand and a huge pain as it gets outdated easily...
A couple of pointers:

- https://github.com/dlang/dmd/blob/master/src/ddmd/mars.d#L98
- https://github.com/dlang/dmd/tree/master/docs/man
- https://github.com/dlang/dlang.org/blob/master/dcompiler.dd
- http://dlang.org/dmd.html (redirects based on your OS)

BBasile:
You Must start your commit message with fix `#855 - ...` so that this get closed: https://github.com/dlang/dub/issues/855

s-ludwig:
@mpevnev: If you use the [`getCommands`](https://github.com/dlang/dub/blob/3ecb1e3150806bf01bcb0c2fc7c9be4dd92e8f44/source/dub/commandline.d#L43) function, you can get the short descriptions isolated from the rest. 

The actual options need to be extracted by instantiating the `CommandArgs` struct:
```
// options common to all commands
CommonOptions options;
auto common_args = new CommandArgs(null);
options.prepare(common_args);
foreach (arg; common_args.recognizedArgs) { ... }

// options for each specific command
foreach (grp; getCommands()) {
	foreach (cmd; grp.commands) {
		auto args = new CommandArgs(null);
		cmd.prepare(args);
		foreach (arg; args.recognizedArgs) { ... }
	}
}
```

Command groups also group the individual commands into semantic groups (they have a `.caption` property).

Going this route instead of parsing the output would also simplify the generation process somewhat, as it would work independently of the dub binary.

An possible step would then be to add an additional field with more in-depth documentation to each command, which is only shown in the high-level documentation (man pages, but also on the website). Then only the package recipe format documentation would be left as redundant (of course it is already redundant now, because SDL and JSON are in different files). But I wanted to so something about that at some point anyway.

wilzbach:
FYI: We now generate the DMD MAN page automatically - https://github.com/dlang/dmd/pull/7507. 

The man generation script is pretty simple actually - https://github.com/dlang/dmd/blob/master/docs/gen_man.d

wilzbach:
> Going this route instead of parsing the output would also simplify the generation process somewhat, as it would work independently of the dub binary.

A first step: https://github.com/dlang/dub/pull/1342

wilzbach:
Opened this as an issue for the release archives, https://issues.dlang.org/show_bug.cgi?id=18424

Not so sure how to ping the package maintainers.

Geod24:
Superseded by #1342. Thanks for your work!

PR #1152, opened 2017-06-14
fix #1144 - default targetType changes when adding configuration

### LINKED ISSUE #1144: dub starts to build a newly created package as library after "configurations" list is added
```
dub init # accept all defaults
dub build # this creates executable
# add "configurations" list to dub.json
dub build # this builds library instead of executable
```

That was very annoying because I first thought dub stopped seeing file changes for some reason.

Happens on Linux, dub 1.30

MartinNowak:
- from autodetected to library
- fixed by warning about implicit targetTypes

### DIFF
--- a/source/dub/recipe/json.d
+++ b/source/dub/recipe/json.d
@@ -11,6 +11,7 @@ import dub.compilers.compiler;
 import dub.dependency;
 import dub.recipe.packagerecipe;
 
+import dub.internal.vibecompat.core.log;
 import dub.internal.vibecompat.data.json;
 
 import std.algorithm : canFind, startsWith;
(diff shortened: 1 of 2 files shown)


source/dub/recipe/json.d
```diff
@@ -54,13 +55,16 @@ void parseJson(ref PackageRecipe recipe, Json json, string parent_name)
...
-			deftargettp = recipe.buildSettings.targetType;
-
 		foreach (settings; *pv) {
 			ConfigurationInfo ci;
-			ci.parseJson(settings, recipe.name, deftargettp);
+			ci.parseJson(settings, recipe.name);
+
+			if (recipe.buildSettings.targetType == TargetType.autodetect &&
+				ci.buildSettings.targetType == TargetType.autodetect) {
+				logWarn("Warning: Configuration '%s' has no targetType, using 'library' as default.", ci.name);
```
s-ludwig: This warning should be output in `Package.simpleLint` instead, because otherwise every build will be spammed with warnings if any available package on the system runs into this case, no matter if used or not.

s-ludwig:
The changed code currently misses the case where the target type is inherited from the global setting, so basically this:
```
if (recipe.buildSettings.targetType == TargetType.autodetect &&
    ci.buildSettings.targetType == TargetType.autodetect)
```
should be changed to this:

PR #1146, opened 2017-06-14
Error message: Getting head of empty path.

MartinNowak:
Create a file with name "appTest.d" and this content:

```
/+ dub.json:
{
	"name": "appTest",
}
+/

void main(){}
```

Execute dub with command: dub appTest.d
Error is shown: Getting head of empty path.

The issue is the use of upper case letters in the name. The issue disappears by changing the name to apptest:

```
/+ dub.json:
{
	"name": "apptest",
}
+/
```

### DIFF
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -72,6 +72,7 @@ unittest
 	For higher level package handling, see the $(D Package) class.
 */
 struct PackageRecipe {
+	RecipeFormat format; /// format of the original recipe file (json, sdl)
 	string name;
 	string version_;
 	string description;
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -106,6 +106,11 @@ struct SubPackage
 	PackageRecipe recipe;
 }
 
+enum RecipeFormat
+{
+	json,
+	sdl,
+}
 
 /// Bundles information about a build configuration.
 struct ConfigurationInfo {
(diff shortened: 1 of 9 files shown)


MartinNowak:
I fiddled around a bit with giving the recipe a fake path (e.g. `path.d-dub.sdl`), but it was too much churn for such a minor issue.

s-ludwig:
It may make sense to add a field `PackageFormat format` to `BuildSettingsTemplate` that stores the format of the source and use that for the decision instead of the path. Should have a very small impact on the source and still produces the right error message for single-file packages.

MartinNowak:
> It may make sense to add a field PackageFormat format to BuildSettingsTemplate that stores the format of the source and use that for the decision instead of the path.

Done, I was a bit hesitant to do this, at some point we might support more formats or even make this pluggable, but chances for extensibility aren't too big.

source/dub/recipe/packagerecipe.d
```diff
@@ -106,6 +107,11 @@ struct SubPackage
 	PackageRecipe recipe;
 }
 
+enum RecipeFormat
+{
+	JSON,
+	SDL,
+}
```
s-ludwig: There is already a `PackageFormat` at package_.d:35, that one should be moved to here. I wouldn't mind renaming it to `RecipeFormat` (with a compatibility alias in package_.d), though, as the current name doesn't fit so well anymore (it predates `PackageRecipe).

MartinNowak:
Done @s-ludwig

s-ludwig:
This should just need a rebase, so that the dmd-beta failure goes away.

Geod24:
This does not seem to trigger in master and can be closed.

andre2007:
I can confirm, issue is reproducible on master.

PR #1153, opened 2017-06-18
Add dub.selections.json to the autogenerated .gitignore

Elronnd:


s-ludwig:
dub.selections.json is actually supposed to be checked in for application projects, so that builds are repeatable. Ignoring it just for pure library packages would be okay, but currently we don't have that information up front.

Another idea would be to create a dub.selections.json file only if a package is built with a (non-test) executable configuration. Then dub.selections.json wouldn't even exist locally for pure library packages, so that they'd need no special treatment in .gitignore.

ZombineDev:
> Another idea would be to create a dub.selections.json file only if a package is built with a (non-test) executable configuration. Then dub.selections.json wouldn't even exist locally for pure library packages, so that they'd need no special treatment in .gitignore.

Agreed, I think that's the best way forward.

MartinNowak:
I've sometimes used dub.selections.json in libraries to select working dependencies for running tests, so even if we wouldn't generate one, we should still obey present configurations (leaving the question how to get one initially).
Wouldn't we have to rerun dependency resolution for every test run without this file? Doesn't sound like a good idea either.

s-ludwig:
Agreed, reading should stay unaffected.

Repeated dependency resolution is indeed sub optimal. At least non-upgrade resolution could be limited to use only locally available packages, as long as no dependency is missing. As an optimization it could then also store a hidden selections file in ".dub/", together with a hash of the possible dependency version candidates and only re-resolve if that changes.

s-ludwig:
Converted to an enhancement request: #1245

PR #1154, opened 2017-06-18
Fix #1037

### LINKED ISSUE #1037: Root package X reference Y cannot be satisfied: need to show why

WebFreak001:
I grouped the arrays all_configs and any_config together into one struct for better maintainability and readability (and eventual segfaults where the index hasn't been checked)

Previously it only showed
```
Root package issue1037-better-dependency-messages reference gitcompatibledubpackage 1.0.1 cannot be satisfied.
```

Now it additionally shows
```
Packages causing the conflict:
	issue1037-better-dependency-messages depends on 1.0.1
	b depends on ~>1.0.2
```


WebFreak001:
I don't get why it complains about logError in the travis build (a shell function I'm not even using) missing and why it even randomly broke there on these old compilers

s-ludwig:
"logError" is called from "die" - my guess would be that the export for that is missing in "run-unittest.sh". But it's strange that this gets called at all, since the diff looks identical.

source/dub/dependencyresolver.d
```diff
@@ -53,15 +53,27 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		}
 	}
 
+	static struct ConfigEntry
```
s-ludwig: Should be `private`

source/dub/dependencyresolver.d
```diff
@@ -53,15 +53,27 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
+	{
+		static struct Depender
+		{
+			TreeNode origin;
+			TreeNodes dependency;
+		}
+
+		CONFIG[] all_configs;
+		bool any_config;
```
s-ludwig: Naming convention: `allConfigs`, `anyConfig`.

test/expected-issue1037-output
```diff
@@ -0,0 +1,4 @@
+Root package issue1037-better-dependency-messages reference gitcompatibledubpackage 1.0.1 cannot be satisfied.
+Packages causing the conflict:
+	issue1037-better-dependency-messages depends on 1.0.1
+	b depends on ~>1.0.2
```
s-ludwig: For better clarity of the "test" folder, this should be named "issue1037-expected-output".

source/dub/dependencyresolver.d
WebFreak001: I adapted to the original naming conventions of the file which used exactly such names

test/expected-issue1037-output
WebFreak001: but all the other 'expected output files' are also named this way, it would be weird to only name this one like that

s-ludwig: Missed that. I'll change those too then.

timotheecour:
ping on this?  issue #1037 keeps popping up

WebFreak001:
ping x10

timotheecour:
@WebFreak001 I just msged you on slack

timotheecour:
@WebFreak001 i just rebased yours against master in https://github.com/dlang/dub/pull/1377
can we get 1154 or 1377 merged in master? this is a big improvement

WebFreak001:
got merged in #1377

Issue #1159, opened 2017-06-21
Environment variable HTTP_PROXY is used but not NO_PROXY


andre2007:
The environment variable HTTP_PROXY is read here and used for cUrl download:
https://github.com/dlang/dub/blob/53f415caf6ade3467efe112443d41fc10b43bab1/source/dub/internal/utils.d#L266

NO_PROXY is not read. This is an issue if you have a proxy set and want to run a local dub registry (localhost). If have specified localhost in environment variable NO_PROXY
but as it is not used dub is not able to connect to the local dub registry as it always use the proxy.

Unfortunatelly std.net.curl: HTTP does not provide any possibility to set NO_PROXY. I created an issue here: https://issues.dlang.org/show_bug.cgi?id=17540

PR #1162, opened 2017-06-22
Fix/issue1159

andre2007:
Fix issue #1159 
Environment variable no_proxy is read and set in curl using the low level function handle.set.
At the moment there is no curl function to set the no proxy value.


s-ludwig:
LGTM, will just have to be rebased once master is back to green so that it can be merged.

andre2007:
Rebase done

s-ludwig:
Okay, thanks!

PR #1164, opened 2017-06-22
Merge remote-tracking branch 'upstream/master' into merge_master

MartinNowak:


wilzbach:
> Bugzilla references

Hmm, it seems like we should have a hard-coded white-list for Bugzilla references:

https://github.com/dlang-bots/dlang-bot/issues/88

PR #1165, opened 2017-06-23
Allow digits in package names.

chadjoan:
Previously, attempting to use digits in a package name during
"dub init ..." would give this error message:

Invalid name, "abc123", names should consist only of lowercase alphanumeric characters, - and _.

Thus it was not honoring the "numeric" part of "alphanumeric",
but now it does.


s-ludwig:
Thanks!

chadjoan:
You're welcome!

PR #1166, opened 2017-06-23
Improve build options of the default build script.

s-ludwig:
Uses optimizations, but leaves assertions/contracts and adds debug information to not impede debugging on the user side. This can be considered progress in the direction of #862. It remains to be seen how much of a performance difference -release would make.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,7 +56,7 @@ fi
 MACOSX_DEPLOYMENT_TARGET=10.7
 
 echo Running $DMD...
-$DMD -ofbin/dub -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
+$DMD -ofbin/dub -g -O -inline -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
 bin/dub --version
 echo DUB has been built as bin/dub.
 echo


wilzbach:
In my experience, only `invariants` are measurable costly in production, so I am a huge fan of [DIP 1006](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1006.md) which should hopefully get approved soon :)

wilzbach:
> dlang-bot merged commit 0ece25a into master a minute ago
> 0 of 3 checks passed

Huh - that wasn't on purpose. I didn't realize that no CIs are set to be required here.
The CI failures look unrelated though:

```
Fetching vibe-d 0.7.31...
Server returned nothing (no headers, no data) on handle 2BE2B10
```

That's unfortunately a typical error on DUB. Maybe some logic to retry downloads and/or try fallback registry mirror could be added to prevent this from happening?

s-ludwig:
Retry-logic is definitely a good idea, I'd say (opened #1167). Where can the required tests be configured for dlang-bot (or probably rather, who can do that)?

wilzbach:
> Retry-logic is definitely a good idea, I'd say (opened #1167).

Yes, that would help a lot!

> Where can the required tests be configured for dlang-bot (or probably rather, who can do that)?

Dlang-Bot uses the protected branch configuration from GH (https://github.com/dlang/dub/settings/branches/master)

It looks like I have the permissions to do so, so I just set the requirement for Travis to pass before auto-merging.

image

s-ludwig:
Great, thanks! Not sure if it is caused by this PR, or maybe is something temporary, but there are lots of out-of-memory errors in the build results on master. But at least it would be strange if the optimized/debug version would use considerably more memory.

wilzbach:
> Not sure if it is caused by this PR, or maybe is something temporary, but there are lots of out-of-memory errors in the build results on master. But at least it would be strange if the optimized/debug version would use considerably more memory.

Yeah I saw that as well, but it didn't appear on the PR. It's pretty weird :S
Let's try a revert and see whether the same errors are manifesting there ....

wilzbach:
> but there are lots of out-of-memory errors in the build results on master

Restarted the build and we already have a lot less errors, so it seems transient:

image

s-ludwig:
But at least on the older versions there also seem to be some genuine issues - segfaults, probably caused by inliner bugs. I'll have to see whether it makes sense to invest the time to find workarounds.

s-ludwig:
Okay, so looks like even the memory allocation failures are consistent. Next step could be to remove just `-O` or just `-inline` and see if that works at least. Somehow it feels unnecessary to invest a lot of time into tracking down codegen issues in older frontends for this. The out-of-memory issue could still be worthwhile to investigate, though.

s-ludwig:
Let's see what happens: https://travis-ci.org/dlang/dub/builds/246329046

s-ludwig:
DMD 2.065.0 still fails. IMO we can just drop support for that, though (release date Feb 2014).

PR #1168, opened 2017-06-23
Revert "Improve build options of the default build script."

wilzbach:
Reverts dlang/dub#1166 - just trying to see whether the "Memory allocation" failures are temporary.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -56,7 +56,7 @@ fi
 MACOSX_DEPLOYMENT_TARGET=10.7
 
 echo Running $DMD...
-$DMD -ofbin/dub -g -O -inline -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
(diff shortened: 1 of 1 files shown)


ZombineDev:
Jenkins checks out dub v1.3.0 [specifically](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-5455/4/pipeline/173#step-485-log-2) and furthermore there are some very simple libraries like Derelict* that shouldn't be a challenge for the compiler.
Also, as far as I know, unlike the frontend, dmd's backend has been optimized for low memory usage because when dmc was actively developed memory was not as abundant. On the other hand, I wouldn't be surprised if the build takes much longer, as there were some reports on pathological cases taking more time because of some algorithms taking quadratic time relative to the input. That said, I think most of the problems in that are have been fixed in the last 1-2 years.

Issue #1161, opened 2017-06-22
Mandatory suffix "/" in registry argument


andre2007:
If you pass a custom registry to dub --registry=http://localhost:8080 
and you forget the mandatory suffix "/" like in the example, the registry is not accepted.
It isn't obvious that the misssing suffix causes this issue.

I propose that suffix "/" is added automatically to the registry url if missing

andre2007:
Issue is solved

PR #1169, opened 2017-06-23
Add suffix / to registry urls

andre2007:
Fixes issue #1161 
The functionality can be tested with: dub build --skip-registry=standard --registry=http://code.dlang.org

### DIFF
--- a/source/dub/internal/vibecompat/inet/url.d
+++ b/source/dub/internal/vibecompat/inet/url.d
@@ -107,7 +107,7 @@ struct URL {
 			}
 		}
 
-		this.localURI = str;
+		this.localURI = (str == "") ? "/" : str;
 	}
 	/// ditto
 	static URL parse(string url_string)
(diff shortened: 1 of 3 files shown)


s-ludwig:
Also LGTM! Master is green again, like #1162, this needs a rebase so that the tests pass and merging is possible

jacob-carlborg:
Isn't the whole point of having a URL type so that it can handle adding/removing slashes where it should be automatically?

To me, this looks like the fix is in the wrong place. In my opinion, the URL struct should handle this automatically.

source/dub/commandline.d
```diff
@@ -216,7 +216,7 @@ int runDubCommandLine(string[] args)
 			dub.defaultPlacementLocation = options.placementLocation;
 		} else {
 			// initialize DUB
-			auto package_suppliers = options.registry_urls.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url))).array;
+			auto package_suppliers = options.registry_urls.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url.chomp("/")~"/"))).array;
```
jacob-carlborg: Perhaps just something minor, but if the trailing slash is already there, this isn't the most efficient implementation.

andre2007: I fully agree. The custom type URL is reponsible for adding the slash. I will rework this pull request.

source/dub/internal/vibecompat/inet/url.d
```diff
@@ -107,7 +107,7 @@ struct URL {
 			}
 		}
 
-		this.localURI = str;
+		this.localURI = (str == "") ? "/" : str;
```
jacob-carlborg: I would think this would be more natural to be handled when appending, but I noticed that it's not possible to append an absolute path, for some reason. So perhaps it doesn't matter.

jacob-carlborg:
Please squash commits.

PR #1172, opened 2017-06-26
Add missing / suffix in URL type

andre2007:
Fixes issue #1161
The functionality can be tested with: dub build --skip-registry=standard --registry=http://code.dlang.org

Squashing was not working, therefore I created a new clean pull request
@jacob-carlborg

### DIFF
--- a/source/dub/internal/vibecompat/inet/url.d
+++ b/source/dub/internal/vibecompat/inet/url.d
@@ -107,7 +107,7 @@ struct URL {
 			}
 		}
 
-		this.localURI = str;
+		this.localURI = (str == "") ? "/" : str;
 	}
 	/// ditto
 	static URL parse(string url_string)
@@ -277,4 +277,10 @@ unittest {
 	assert(url.path.toString() == "/sub2/index.html", url.path.toString());
 	assert(url.queryString == "query", url.queryString);
 	assert(url.anchor == "anchor", url.anchor);
+
+	url = URL("http://localhost")~Path("packages");
+	assert(url.toString() == "http://localhost/packages", url.toString());
+
+	url = URL("http://localhost/")~Path("packages");
+	assert(url.toString() == "http://localhost/packages", url.toString());
 }


andre2007:
rebased

s-ludwig:
I think the fully correct fix would be to add the trailing slash where the append operation happens (`opBinary`), because a URL with only a host name and without a trailing slash is actually valid. However, in the context of DUB alone, this shouldn't make a difference. Just the analogous change in vibe.d has to take this into account, as the URL struct is used in a much broader context there.

In fact I'm still not 100% decided whether this should be corrected automatically, or if it should instead throw an exception or even assert on a leading slash in the path component. But in any case, I'll merge this PR, because it fixes the issue and this is just internal code that has the same effect. Once the vibe.d side is fixed, I'd then potentially come back and adjust the code in DUB accordingly.

jacob-carlborg:
The first part I agree with Sönke that it should be fixed in `opBinary`.

The second part, I think it should be fixed automatically. Is there a use case of concatenating two paths and it should not add a `/` in between? In that case a regular string concatenation would suffice.

On the other hand the existing implementation is a problem since issue #1161 did occur. It's an implementation detail if the URL needs to end with a slash or not, the user shouldn't need to care.

s-ludwig:
> Is there a use case of concatenating two paths and it should not add a `/` in between?

I'd say my problem with this stems from the fact that probably the most obvious definition for an empty path is to be semantically equal to `./`, similar to how empty path segments are equal to `.`. Going with that assumption, especially outside of the URL context, if a `/` would always be added, that would suddenly turn the concatenation of two relative paths into an absolute path, which could be dangerous.

The URL case is also special because the path can either be empty, or otherwise it must be absolute. On the other hand, the host part of the URL could simply be considered as a path segment and then the concatenation behavior would be completely normal. Sounds convincing enough to drop the assertion/exception idea...

jacob-carlborg:
Not sure I follow. Could you provide some examples?

PR #1173, opened 2017-06-26
read additional registry URLs from DUB_REGISTRY env var

MartinNowak:
- multiple URLs separated by semicolon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+v1.5.0 - 2017-
+-------------------
+
+- Read additional registry URLs (semicolon separated) from DUB_REGISTRY env var - [pull #1173][issue1173]
+
+[issue1173]: https://github.com/dlang/dub/issues/1173
+
 v1.4.0 - 2017-
 -------------------
 
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -133,9 +133,11 @@ class Dub {
 		PackageSupplier[] ps = additional_package_suppliers;
 
 		if (skip_registry < SkipPackageSuppliers.all)
-			ps ~= m_config.registryURLs
+		{
+			ps ~= (environment.get("DUB_REGISTRY", null).split(";") ~ m_config.registryURLs)
 				.map!(url => cast(PackageSupplier)new RegistryPackageSupplier(URL(url)))
 				.array;
+		}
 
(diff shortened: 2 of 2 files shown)


MartinNowak:
Can you change the settings of Coveralls to not leave comments btw?
https://coveralls.io/github/dlang/dub/settings

s-ludwig:
> Can you change the settings of Coveralls to not leave comments btw?

Done.

source/dub/commandline.d
```diff
@@ -257,6 +257,7 @@ struct CommonOptions {
 	/// Parses all common options and stores the result in the struct instance.
 	void prepare(CommandArgs args)
 	{
+		registry_urls = environment.get("DUB_REGISTRY", null).split(";");
```
s-ludwig: I think this should rather be done in the `Dub` class, so that the command line module ideally performs pure command line parsing (I realize that it already reads `DFLAGS`, which should really be refactored). A new mode for `SkipPackageSuppliers` *could* then also be added to enable choosing between using only those specified on the command line, or also those coming from the environment.

MartinNowak: It's weird to pass package suppliers to `new Dub` only to have it add more package suppliers internally.
Also the env variables are simple alternatives to the command line options, so treating them somewhere else doesn't seem too sensible.

MartinNowak: We could add env var support to `CommandArgs` if something more structured seems appropriate.

s-ludwig: > It's weird to pass package suppliers to new Dub only to have it add more package suppliers internally.

But it will add more anyway, the default one and the ones from the config file. Also, `Dub` already reads `DUBPATH` for example. So just based on the current state, adding it there would cause a bit less additional confusion - but in general I'd say all of this should be controllable from outside of the `Dub` class and a more principled separation would of course be even better.

Adding a generic way to handle env vars as an alternative to certain command line arguments also open up the possibility for better auto-generated command line help, so that sounds like a good idea. The only thing that should be there, and this is kind of orthogonal to this PR, is a way to pass a custom environment, so that it's possible to opt-out when the package is used as a library. This actually applies to the whole library, though, and not just to the command line module.

MartinNowak: OK, makes sense to move to dub.d then. Abstracting out environment could indeed make sense, but having a library react to env vars has it's use-case as well (e.g. CURL).

s-ludwig:
Okay, I've opened two issues to remember this.

PR #1176, opened 2017-06-27
Update .travis.yml

John-Colvin:
boring version bumps

also, moving to trusty because precise is EOL according to canonical.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,21 @@
 language: d
+dist: trusty
 sudo: false
 
 matrix:
   include:
     - d: dmd-nightly
-      env: [FRONTEND=2.074]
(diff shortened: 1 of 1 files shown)


John-Colvin:
had to add something to install libevent because apparently it's not there by default on trusty

wilzbach:
Couldn't we do sth. like ` FRONTEND=$(dmd --version | cut -d' ' -f4 | head -n1 | cut -c 2-)` to avoid the need for hard-coding?

s-ludwig:
@wilzbach: Of course also for GDC and LDC. May or may not be easier to do something along the lines of:

    echo "pragma(msg, __VERSION__);" > ver.d
    FRONTEND=`dmd -o- ver.d`

PR #1177, opened 2017-06-28
Perform stricter test for single-file recipe comments.

s-ludwig:
The recipe file name must now be on the first line. The error message also mentions the required format, including the trailing ":" so that it becomes clear what needs to be changed.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -313,8 +313,11 @@ class Dub {
 			recipe_content = file_content[0 .. idx].strip();
 		} else throw new Exception("The source file must start with a recipe comment.");
 
+		auto nidx = recipe_content.indexOf('\n');
+
 		auto idx = recipe_content.indexOf(':');
-		enforce(idx > 0, "Missing recipe file name (e.g. \"dub.sdl:\") in recipe comment");
+		enforce(idx > 0 && (nidx < 0 || nidx > idx),
+			"The first line of the recipe comment must list the recipe file name followed by a colon (e.g. \"/+ dub.sdl:\").");
 		auto recipe_filename = recipe_content[0 .. idx];
 		recipe_content = recipe_content[idx+1 .. $];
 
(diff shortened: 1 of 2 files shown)


source/dub/dub.d
```diff
@@ -313,8 +313,11 @@ class Dub {
 			recipe_content = file_content[0 .. idx].strip();
 		} else throw new Exception("The source file must start with a recipe comment.");
 
+		auto nidx = recipe_content.indexOf('\n');
+
 		auto idx = recipe_content.indexOf(':');
-		enforce(idx > 0, "Missing recipe file name (e.g. \"dub.sdl:\") in recipe comment");
+		enforce(idx > 0 && (nidx < 0 || nidx > idx),
```
wilzbach: What's the reason for `nidx < 0`?
Currently a package format isn't valid without a `name`.
Or are you silently considering merging https://github.com/dlang/dub/pull/1081? :)

wilzbach:
> continuous-integration/travis-ci/push — The Travis CI build could not complete

FYI: https://www.traviscistatus.com/incidents/fyskznm7wg2c

source/dub/dub.d
s-ludwig: In theory someone could also do something like this `/+ dub.sdl: name "foo"; dependency "bar" +/` in one line. But incidentally I was indeed also thinking of #1081. Will have a look at it in a moment.

wilzbach:
> dlang-bot merged commit 841a0d0 into master 8 minutes ago
> 3 checks passed

Btw @s-ludwig if I should set up dlang-bot for the Vibe.d repos, just give Martin or me a ping. 
We have to maintain and run the bot anyways, so if we are happy if we can save a bit of your time by avoiding the need for you to wait until all Travis builds pass.

s-ludwig:
> Btw @s-ludwig if I should set up dlang-bot for the Vibe.d repos, just give Martin or me a ping.
We have to maintain and run the bot anyways, so if we are happy if we can save a bit of your time by avoiding the need for you to wait until all Travis builds pass.

That would actually be pretty nice! I'm usually pretty lazy adopting new features in this area (PR review functionality or the Sociomantic GitHub tool comes to mind), but auto-merging is simply awesome.

PR #1184, opened 2017-07-03
silence confusing test output

MartinNowak:
- failure is expected and should not appear in logs as it's easily
  confused to be the cause for a failed test


wilzbach:
>  wilzbach added the auto-merge label 40 seconds ago
> dlang-bot merged commit 2e83112 into dlang:stable 38 seconds ago
> 1 check was pending

Ehm sorry - I though we had configured branch protections here.
@MartinNowak didn't we set branch protections for `stable` as well?
I just did to avoid this problem from happening in the future:

image

MartinNowak:
Thx, hope we can soon fix the source of that problem though ;).

PR #1190, opened 2017-07-09
add code-mirror.dlang.io as fallback registry

MartinNowak:
- to handle any code.dlang.org failures
- depends on #1104

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -65,6 +65,7 @@ static this()
 
 /// The URL to the official package registry.
 enum defaultRegistryURL = "http://code.dlang.org/";
+enum fallbackRegistryURL = "https://code-mirror.dlang.io/";
 
 /** Returns a default list of package suppliers.
 
@@ -76,7 +77,12 @@ enum defaultRegistryURL = "http://code.dlang.org/";
 PackageSupplier[] defaultPackageSuppliers()
 {
 	logDiagnostic("Using dub registry url '%s'", defaultRegistryURL);
-	return [new RegistryPackageSupplier(URL(defaultRegistryURL))];
+	return [
+		new FallbackPackageSupplier(
+			new RegistryPackageSupplier(URL(defaultRegistryURL)),
+			new RegistryPackageSupplier(URL(fallbackRegistryURL))
+		)
+	];
 }
 
 
--- a/source/dub/packagesupplier.d
+++ b/source/dub/packagesupplier.d
@@ -21,6 +21,7 @@ import std.datetime;
 import std.exception;
 import std.file;
 import std.string : format;
+import std.typecons : AutoImplement;
 import std.zip;
 
 // TODO: Could drop the "best package" behavior and let retrievePackage/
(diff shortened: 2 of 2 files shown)


source/dub/dub.d
```diff
@@ -76,7 +77,12 @@ enum defaultRegistryURL = "http://code.dlang.org/";
 PackageSupplier[] defaultPackageSuppliers()
 {
 	logDiagnostic("Using dub registry url '%s'", defaultRegistryURL);
-	return [new RegistryPackageSupplier(URL(defaultRegistryURL))];
+	return [
+		new FallbackPackageSupplier(
+			new RegistryPackageSupplier(URL(defaultRegistryURL)),
+			new RegistryPackageSupplier(URL(fallbackRegistryURL))
+		)
+	];
```
s-ludwig: Why not simply return an array of the two? The intent is already that later array entries act as fallbacks to the earlier ones.

source/dub/dub.d
```diff
@@ -65,6 +65,7 @@ static this()
 
 /// The URL to the official package registry.
 enum defaultRegistryURL = "http://code.dlang.org/";
+enum fallbackRegistryURL = "https://code-mirror.dlang.io/";
```
wilzbach: > this week will I hopefully finally get around setting up a stable mirror myself).

GitHub didn't send my review, so I finally got around setting up a stable mirror myself:

-> https://code-mirror2.dlang.io

Heroku Deploy is a bit more difficult to setup due to https://github.com/rejectedsoftware/vibe.d/issues/1785, but I will get around this as well.

wilzbach: > Heroku Deploy is a bit more difficult to setup due to rejectedsoftware/vibe.d#1785, but I will get around this as well.

-> https://dub-registry.herokuapp.com/

With https://github.com/dlang/dub-registry/pull/231 setting up a local mirror should be absolutely easy (and doesn't even require a CLI nor money).

source/dub/dub.d
MartinNowak: To avoid duplicates, e.g. in `dub search` and to correctly deal with user-facing error reporting. The current mechanism adds another registry which can be used e.g. to serve private projects, very different use-case than a mirror registry which is expected to serve identical projects.

source/dub/packagesupplier.d
```diff
@@ -254,4 +267,41 @@ class RegistryPackageSupplier : PackageSupplier {
 	}
 }
 
+abstract class AbstractFallbackPackageSupplier : PackageSupplier
+{
+	protected PackageSupplier m_default, m_fallback;
```
MartinNowak: Simply nest them like a list `new AbstractFallbackPackageSupplier(new AbstractFallbackPackageSupplier(...` ;).

MartinNowak: I'd hope though that a third site wouldn't be necessary. We're also far from needing some round-robing load balancing, and should look at server side reliability/DNS etc. if that ever becomes a topic.

source/dub/dub.d
MartinNowak: Hope that the documentation on FallbackPackageRegistry is good enough @s-ludwig.

source/dub/dub.d
MartinNowak: http://alpha.dub.pm/ says bad gateway btw.

s-ludwig:
Can we make `AbstractFallbackPackageSupplier` and `FallbackPackageSupplier` `package` for now? That's the only thing I'm a bit worried about. Since we are now in the stable phase, I'd like to be very conscious about API additions and making one isn't really the primary goal of this change.

It would be nice to allow users to add fallback registries without caring for duplicates, too. So a duplicate detection logic for `dub search` might be an alternative here. But this doesn't have to be worked out within the context of this PR.

MartinNowak:
Done @s-ludwig.

> It would be nice to allow users to add fallback registries without caring for duplicates, too. So a duplicate detection logic for dub search might be an alternative here. But this doesn't have to be worked out within the context of this PR.

Yes, we could add some syntax, e.g. `--registry=http://reg1.example.com,http://reg2.example.com --registry=http://someother.registry.com` to be used for fallbacks.
Indeed looks like all operations but search would stop on first success, so dedup in search might work as well.

MartinNowak:
It might actually be simpler to just implement the dedup feature right away.

PR #1193, opened 2017-07-10
Add SkipPackageSuppliers.configured and fix unit test.

s-ludwig:
With only .all and .standard, an intermediate option was missing to make the unit test pass on systems with additional registries configured in the DUB configuration.


s-ludwig:
Will merge in a few hours, unless anyone has any doubts about the change.

PR #1197, opened 2017-07-14
use codecov integration

MartinNowak:
- patch coverage is highly useful

Still needs to be enabled under https://github.com/apps/codecov, might need to ask Andrei for that.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,17 +1,14 @@
 # Documentation: https://docs.codecov.io/docs/codecov-yaml
 
-codecov:
-  bot: dlang-bot
-
 coverage:
   precision: 3
-  # round: down
-  # range: "70...100"
+  round: down
+  range: "80...100"
 
   status:
     # Learn more at https://docs.codecov.io/docs/commit-status
     project: true
     patch: true
-    changes: false
+    changes: true
 
 comment: false
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -18,8 +18,8 @@ fi
 DUB=`pwd`/bin/dub DC=${DC} test/run-unittest.sh
 
 if [ "$COVERAGE" = true ]; then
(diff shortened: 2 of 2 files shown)


wilzbach:
> might need to ask Andrei for that

CC @andralex (dlang/tools would be another good candidate)

.codecov.yml
```diff
@@ -1,17 +1,14 @@
...
-  # range: "70...100"
+  round: down
+  range: "80...100"
 
   status:
     # Learn more at https://docs.codecov.io/docs/commit-status
     project: true
     patch: true
-    changes: false
+    changes: true
```
wilzbach: Not sure about the effects/usefulness of this one, but I guess it doesn't hurt to try.

.codecov.yml
```diff
@@ -1,17 +1,14 @@
 # Documentation: https://docs.codecov.io/docs/codecov-yaml
 
-codecov:
-  bot: dlang-bot
-
```
wilzbach:
```
codecov:
  ci:
  notify:
    after_n_builds: 1
```

Seems to be a sensible default as well.

andralex:
noice...

andralex:
Added dub to codecov.

.codecov.yml
MartinNowak: Let's check, a patch shouldn't reduce coverage in previously covered areas.

.codecov.yml
MartinNowak: Not sure it's seems sensible to me to enable an advanced configuration by default?
Don't they know how to deal with CI? I mean eventually they should just react to us sending sth. from a CI run.

wilzbach: Yes eventually they will, but by default they will wait for all CIs. Those four lines will trigger the notification on the first submission - don't ask me why they don't send it by default.

Issue #1167, opened 2017-06-23
Retry failed package downloads once or twice


s-ludwig:
Sometimes either the dub registry or github.com fails to deliver the zip file for a package. For normal builds this is acceptable, because dependencies are usually downloaded much less frequently than a build is invoked and restarting the build is trivial. But for CI builds this can be annoying. Usually these failures are of very temporary nature, so retrying once or maybe twice should eliminate most cases.

andre2007:
Retry was implemented long time ago. The new pr https://github.com/dlang/dub/pull/1566 refactors the retry logic as it was duplicated to maven supplier.

This issue can be closed?

PR #1198, opened 2017-07-15
Retry downloading a package

grogancolin:
Try 3 times to download a package/metadata

Packagesupplier.d now tries 3 times to download a package and metadata.
This should aid travis to not be as flaky as it is currently.

Implements enhancement request #1167


wilzbach:
See also: https://github.com/dlang/dub/pull/1190 though not conflicting with this PR.

source/dub/packagesupplier.d
```diff
@@ -191,7 +191,17 @@ class RegistryPackageSupplier : PackageSupplier {
 		auto vers = best["version"].get!string;
 		auto url = m_registryUrl ~ Path(PackagesPath~"/"~packageId~"/"~vers~".zip");
 		logDiagnostic("Downloading from '%s'", url);
-		download(url, path);
+		foreach(i; 0..3) {
+			try{
+				download(url, path);
+				break;
```
s-ludwig: `break;` <-> `return;`

source/dub/packagesupplier.d
```diff
@@ -191,7 +191,17 @@ class RegistryPackageSupplier : PackageSupplier {
...
+			}
+			catch(HTTPStatusException e) {
+				logDebug("Failed to download package %s from %s (Attempt %s of 3)", packageId, url, i + 1);
+				continue;
+			}
+			throw new Exception("Failed to download package %s from %s".format(packageId, url));
```
s-ludwig: This needs to be moved outside of the loop.

source/dub/packagesupplier.d
```diff
@@ -191,7 +191,17 @@ class RegistryPackageSupplier : PackageSupplier {
...
 		auto url = m_registryUrl ~ Path(PackagesPath~"/"~packageId~"/"~vers~".zip");
 		logDiagnostic("Downloading from '%s'", url);
-		download(url, path);
+		foreach(i; 0..3) {
+			try{
+				download(url, path);
+				break;
+			}
+			catch(HTTPStatusException e) {
+				logDebug("Failed to download package %s from %s (Attempt %s of 3)", packageId, url, i + 1);
```
s-ludwig: There should be a check for 404 errors here, or maybe even limit the retry behavior to 500 errors and connection failures. Retrying for 404 errors would mean that package download in scenarios with multiple registries (that contain different sets of packages) potentially gets slowed down considerably.

source/dub/packagesupplier.d
```diff
@@ -185,13 +185,30 @@ class RegistryPackageSupplier : PackageSupplier {
...
-		download(url, path);
+		foreach(i; 0..3) {
+			try{
+				download(url, path);
+				return;
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) {
+					logDebug("Failed to download package %s from %s (404)", packageId, url);
+					return;
```
s-ludwig: I'm not sure how the outer code will behave if nothing downloaded, but to be consistent with the previous behavior, this would also have to throw instead of returning.

grogancolin:
Thanks!
Apologies - I never got time to get back to this quickly :-)

ZombineDev:
@s-ludwig maybe it's time to drop support for accient frontends like 2.068 which currently prevents the merging of this PR?

wilzbach:
Looks more like a random failure - let's have a look whether restarting works

s-ludwig:
AFAIR, 2.068 is the frontend that the latest released version of GDC uses, so I think we should keep it for a while. But it's indeed strange that it fails for this PR, I already restarted it, but it still failed.

PR #1200, opened 2017-07-17
Let fuzzy version matches exclude pre-release variants of the bumped …

s-ludwig:
…version.

This changes ~>1.2.3 to not match 1.3.0-beta.1 anymore as it currently does. To achieve this, ~>1.2.3 gets expanded to ">=1.2.3 <1.3.0-0", which filters out the smallest representable pre-release version of 1.3.0.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -153,7 +153,7 @@ struct Dependency {
 			m_inclusiveB = false;
 			ves = ves[2..$];
 			m_versA = Version(expandVersion(ves));
-			m_versB = Version(bumpVersion(ves));
+			m_versB = Version(bumpVersion(ves) ~ "-0");
 		} else if (ves[0] == Version.branchPrefix) {
 			m_inclusiveA = true;
 			m_inclusiveB = true;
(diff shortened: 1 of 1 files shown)


source/dub/dependency.d
```diff
@@ -555,16 +555,17 @@ unittest {
 
 	// Approximate versions.
 	a = Dependency("~>3.0");
-	b = Dependency(">=3.0.0 <4.0.0");
+	b = Dependency(">=3.0.0 <4.0.0-0");
 	assert(a == b, "Testing failed: " ~ a.toString());
 	assert(a.matches(Version("3.1.146")), "Failed: Match 3.1.146 with ~>0.1.2");
 	assert(!a.matches(Version("0.2.0")), "Failed: Match 0.2.0 with ~>0.1.2");
+	assert(!a.matches(Version("4.0.0-beta.1")));
```
John-Colvin: would be nice to have a check that checked it works for the minor version number, e.g. checking `"3.1.0-beta.1"` doesn't match `"~>3.0.0"`

s-ludwig: Added in line 569.

John-Colvin:
LGTM. I consider this a bug fix as the previous behaviour didn't really make sense. Will merge shortly unless there's any objections

s-ludwig:
@wilzbach @MartinNowak any objections? Otherwise let's merge this in a few hours. The current behavior indeed simply doesn't make sense, and the `~>` operator was the only place that was reasonable to approach for fixing it, as fiddling with the comparison operators would have made this non-compliant with the SemVer spec.

PR #1208, opened 2017-07-27
Fix #895, so that dub looks for a compiler next to itself first

### LINKED ISSUE #895: Search dub's own path first to find compiler

joakim-noah:
Since [I didn't get a response about how checking for a compiler next to dub should interact with the one specified in a DUB config file](https://github.com/dlang/dub/issues/895#issuecomment-317688828), I went ahead and implemented what I thought makes sense.  I can add some tests once the behavior is approved.

Also puts some restrictions on the defaultCompiler specified in a DUB `settings.json`, such that it must be a compiler name or an absolute path, including a tilde-prefixed local path.  Relative paths are no longer
allowed, as they make no sense for how dub invokes the compiler.  You could argue that even a compiler name should be disallowed, as it's unlikely anyone will specify one that's not already on the default list that dub tries, but I suppose the sdc devs may want to try out dub some day. ;)

If a compiler name like `sdc` is specified in a DUB `settings.json`, I have dub check for other common compilers in the `PATH` only if `sdc` isn't found next to dub or on the `PATH` (__Update:__ which is probably why the test for #895 now fails on Travis CI, though the test will be changed), though you could argue the user only wants `sdc` used.  However, dub always says what compiler it's using and if the user hasn't provided an `sdc`, it's a convenience to search for another compiler.  I could change it to only check the `PATH` for `sdc` in that case, if wanted, as I do when checking next to dub.

Finally, all this changes nothing about specifying a compiler with `--compiler`, where none of this is used and you can specify any D compiler you want, including one with a relative path.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1182,17 +1182,56 @@ class Dub {
 
 	private void determineDefaultCompiler()
 	{
-		import std.range : front;
+		import std.file : thisExePath;
+		import std.path : buildPath, dirName, expandTilde, isAbsolute, isDirSeparator;
 		import std.process : environment;
+		import std.range : front;
+		import std.regex : ctRegex, matchFirst;
 
-		m_defaultCompiler = m_config.defaultCompiler;
-		if (m_defaultCompiler.length) return;
+		m_defaultCompiler = m_config.defaultCompiler.expandTilde;
+		if (m_defaultCompiler.length && m_defaultCompiler.isAbsolute)
+			return;
+
+		auto dubPrefix = m_defaultCompiler.matchFirst(ctRegex!(`^\$DUB_BINARY_PATH`));
+		if(!dubPrefix.empty)
+		{
+			m_defaultCompiler = thisExePath().dirName() ~ dubPrefix.post;
+			return;
+		}
+
+		if (!find!isDirSeparator(m_defaultCompiler).empty)
+			throw new Exception("defaultCompiler specified in a DUB config file cannot use an unqualified relative path:\n\n" ~ m_defaultCompiler ~
+			"\n\nUse \"$DUB_BINARY_PATH/../path/you/want\" instead.");
 
 		version (Windows) enum sep = ";", exe = ".exe";
 		version (Posix) enum sep = ":", exe = "";
 
 		auto compilers = ["dmd", "gdc", "gdmd", "ldc2", "ldmd2"];
+		// If a compiler name is specified, look for it next to dub.
+		// Otherwise, look for any of the common compilers adjacent to dub.
+		if (m_defaultCompiler.length)
+		{
+			string compilerPath = buildPath(thisExePath().dirName(), m_defaultCompiler ~ exe);
+			if (existsFile(compilerPath))
+			{
+				m_defaultCompiler = compilerPath;
+				return;
+			}
+		}
+		else
+		{
+			auto nextFound = compilers.find!(bin => existsFile(buildPath(thisExePath().dirName(), bin ~ exe)));
+			if (!nextFound.empty)
+			{
+				m_defaultCompiler = buildPath(thisExePath().dirName(),  nextFound.front ~ exe);
+				return;
+			}
+		}
 
+		// If nothing found next to dub, search the user's PATH, starting
+		// with the compiler name from their DUB config file, if specified.
+		if (m_defaultCompiler.length)
+			compilers = m_defaultCompiler ~ compilers;
 		auto paths = environment.get("PATH", "").splitter(sep).map!NativePath;
 		auto res = compilers.find!(bin => paths.canFind!(p => existsFile(p ~ (bin~exe))));
 		m_defaultCompiler = res.empty ? compilers[0] : res.front;
--- a/test/issue895-local-configuration.sh
+++ b/test/issue895-local-configuration.sh
@@ -1,11 +1,6 @@
 #!/usr/bin/env bash
 . $(dirname "${BASH_SOURCE[0]}")/common.sh
 
-cd ${CURR_DIR}
-mkdir ../etc
-mkdir ../etc/dub
-echo "{\"defaultCompiler\": \"foo\"}" > ../etc/dub/settings.json
-
 if [ -e /var/lib/dub/settings.json ]; then
 	die $LINENO 'Found existing system wide DUB configuration. Aborting.'
 fi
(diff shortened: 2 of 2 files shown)


MartinNowak:
Why not interpret relative paths in `settings.json` as relative to dir containing the dub executable?

joakim-noah:
Possible, but then inconsistent with `--compiler`, which I believe will invoke relative to `getcwd`.  Could still do that, as we assume someone writing a config file is sane, ;) just as long as we document the difference.

MartinNowak:
> Possible, but then inconsistent with --compiler, which I believe will invoke relative to getcwd. Could still do that, as we assume someone writing a config file is sane, ;) just as long as we document the difference.

I'd go for it, as it allows more flexibility for packaging (e.g. different bin folders).
We could use some `$DUB_BIN_PATH` variable instead of relative paths to make it more explicit.
Can the settings.json be located next to dub or somehow be part of the packaging?

joakim-noah:
>We could use some $DUB_BIN_PATH variable instead of relative paths to make it more explicit.

Good idea, will add that.

>Can the settings.json be located next to dub or somehow be part of the packaging?

dub looks for it [first in the system (`/var/lib/dub` on Posix, scroll up a bit to see), then relative to the binary at `../etc/dub/settings.json`, and finally in `~/.dub`](https://github.com/dlang/dub/blob/master/source/dub/dub.d#L195).  ldc uses the second to change the default compiler name to ldc, but that's it.

joakim-noah:
Alright, modified this pull to check for `%%dubbinarypath%%` also, similar to the way ldc does it.  I'll fix and write more tests once this behavior is approved.

joakim-noah:
@s-ludwig, let me know if this pull is okay, would be good to get this in before the next dmd/ldc releases.

s-ludwig:
Sorry for the late reply, this slipped past me. Two notes:
- `DUB_BIN_PATH` or `DUB_BINARY_PATH` would be more consistent with the other variables that DUB defines
- Is the changed lookup logic still needed with the new possibility to specify an executable-relative path in settings.json?

joakim-noah:
>DUB_BIN_PATH or DUB_BINARY_PATH would be more consistent with the other variables that DUB defines

Those are all environment variables normally though, right?  I thought this should be different because it's specific to the config file.  Doesn't really matter to me though, just let me know what you want.

>Is the changed lookup logic still needed with the new possibility to specify an executable-relative path in settings.json?

The only changed lookup logic is to first search for a compiler name specified in the config file in the `PATH`.  The newly added lookup in the directory next to dub could just be replaced by specifying an executable-relative path in the config file instead, as you say, but it's a convenience to not have to provide a config file at all, ie it always looks next to dub.

s-ludwig:
> Those are all environment variables normally though, right? I thought this should be different because it's specific to the config file. Doesn't really matter to me though, just let me know what you want.

My idea was that we could support real environment variables, too, at some point, and this special case could then just naturally merge with that.

> The newly added lookup in the directory next to dub could just be replaced by specifying an executable-relative path in the config file instead, as you say, but it's a convenience to not have to provide a config file at all, ie it always looks next to dub.

Okay, it's mostly just a fuzzy concern on my side right now, because I haven't really thought through the possible implications/issues. One thing that gets a bit hairy is when one explicitly wants to select the compiler in PATH instead of the one in the local directory, but maybe that's a bit superficial. Generally I'd just try to avoid changes that serve no concrete need (i.e. if a config file in the compiler distribution already works for the initial use case).

joakim-noah:
>My idea was that we could support real environment variables, too, at some point, and this special case could then just naturally merge with that.

OK, will change it.

>One thing that gets a bit hairy is when one explicitly wants to select the compiler in PATH instead of the one in the local directory, but maybe that's a bit superficial.

Just specify `--compiler=gdc` then, [as noted for rdmd with its equivalent pull](https://dlang.org/changelog/2.076.0_pre.html#b11997).  You're right that you couldn't do that from the config file anymore though.

>Generally I'd just try to avoid changes that serve no concrete need (i.e. if a config file in the compiler distribution already works for the initial use case).

It is true that adding that option to use a compiler relative to dub's path to the config file, which wasn't there before this pull, is potentially redundant with checking for a compiler next to dub automatically, but only if every compiler that ships dub adds a config file that explicitly says where dub is relative to the compiler.

Since most will stick dub right next to the compiler, this saves them that trouble.  It's sufficiently common that I think it's worth having both.

s-ludwig:
Having rdmd now behave like this is a good argument. It makes sense to stay consistent within the tools ecosystem.

source/dub/dub.d
```diff
@@ -1168,17 +1168,56 @@ class Dub {
...
+		import std.path : buildPath, dirName, expandTilde, isAbsolute, isDirSeparator;
 		import std.process : environment;
+		import std.range : front;
+		import std.regex : ctRegex, matchFirst;
+
+		m_defaultCompiler = m_config.defaultCompiler.expandTilde;
+		if (m_defaultCompiler.length && m_defaultCompiler.isAbsolute)
+			return;
+
+		auto dubPrefix = m_defaultCompiler.matchFirst(ctRegex!(`^DUB_BINARY_PATH`));
```
joakim-noah: Switched to `DUB_BINARY_PATH` liked you asked.

test/issue895-local-configuration.sh
```diff
@@ -14,9 +15,11 @@ if [ -e ~/.dub/settings.json ]; then
 	die $LINENO 'Found existing user wide DUB configuration. Aborting.'
 fi
 
-if ! { ${DUB} describe --single issue103-single-file-package.d 2>&1 || true; } | grep -cF 'Unknown compiler: foo'; then
+if ! { ${DUB} describe --single issue103-single-file-package.d 2>&1 || true; } | grep -cF 'Unknown compiler: '; then
```
joakim-noah: This test should now pass, but removed the compiler name `foo` because my new code passes the full path of a compiler next to `dub` to the settings and this error output.  I'm adding more tests for the new behavior and will update the patch soon.

joakim-noah:
Finally spent an hour and updated this pull to test all new code paths, should be ready to merge.

joakim-noah:
Ping, @s-ludwig or @MartinNowak, PR passes most CI, except for some scenarios because of the upstream druntime headers, which are broken unrelated to this pull.

joakim-noah:
Ping, would be good to get this in for the next release, so dub works out of the box with dmd.

s-ludwig:
The only issue I can see is that the "$" is missing before "DUB_BINARY_PATH". Otherwise it looks good to merge, the Windows failure is unrelated (we should really clean this up at some point, so that AppVeyor can also be marked as a required check).

joakim-noah:
OK, do you want me to add the `$` prefix for the path setting?  I just followed what you wrote above.

s-ludwig:
Yeah, sorry for that, I was indeed writing ambiguously there. So in the path, it should be written as `"$DUB_BINARY_PATH/../foo"`, so that it looks like an sh style environment variable basically.

joakim-noah:
Will do, in the meantime, I rebased to get the CI to run again.

joakim-noah:
Is codecov broken, @wilzbach?  It says I only cover 45% of my patch with the tests, when I tried to hit all of it.

Issue #824, opened 2016-04-25
dub init – A more comprehensive .gitignore file


Kelet:
Hi all,

GitHub maintains a set of `.gitignore` files for common languages. It looks like @MartinNowak has created a slightly more comprehensive (and commented) version of the one generated by `dub init`<sup>[1]</sup>. Perhaps dub should use that?

<sup>[1]:</sup> https://github.com/github/gitignore/blob/master/D.gitignore

Regards,
Kelet

wilzbach:
Btw shouldn't the `dub.selections.json` also be part of the gitignore?

s-ludwig:
Usually it makes a lot of sense to put `dub.selections.json` under version control, so that the build becomes repeatable. Packages that don't have dependencies can of course exclude it, but even for pure library packages it can make sense, so that the CI environment receives a well-defined list of dependencies (or maybe it's desired to always use the latest dependencies, so that's not so clear).

MartinNowak:
Well, here is the code that writes the .gitignore file.
https://github.com/dlang/dub/blob/9a37109faa3a956d625a95a5946314a71da0ffd7/source/dub/init.d#L146
Just make a pull that adds the nicer content and we're done :smile:.
Could use a delimited (heredoc) string, see https://dlang.org/spec/lex.html#DelimitedString.

jmdavis:
It would be nice if when generating a project for an executable, `dub init` added the name of the executable to .gitignore, since it clearly shouldn't be part of source control.

    # Executables
    *.exe

in the example would take care of it on Windows, but on other platforms, it would need to be explicitly added, since there typically is no extension for executables on *nix systems.

wilzbach:
This will be solved by https://github.com/dlang/dub/pull/1214

PR #1214, opened 2017-08-07
Better shipped gitignore

WebFreak001:
Based off githubs .gitignore & added profiling files

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -156,20 +156,37 @@ private void initDeimosPackage(Path root_path, ref PackageRecipe p, scope void d
 private void writeGitignore(Path root_path, PackageRecipe p)
 {
 	write((root_path ~ ".gitignore").toNativeString(),
-q"{.dub
+q"{# Compiled Object files
+*.o
+*.obj
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Compiled Static libraries
+*.a
+*.lib
+
+# Executables
+*.exe
+%1$s-test-*
+/%1$s
+
+# DUB
+.dub
 docs.json
 __dummy.html
 docs/
-%1$s.so
-%1$s.dylib
-%1$s.dll
-%1$s.a
-%1$s.lib
-%1$s-test-*
-*.exe
-*.o
-*.obj
+
+# Code coverage
 *.lst
+
+# Profiling
+profilegc.log
+trace.def
+trace.log
 }".format(p.name));
 }
 


s-ludwig:
I think this is a bit too broad regarding library and executable files, which are sometimes actually checked in. IMO the default should generally be as conservative as possible, because git is already hard enough and leaving a beginner wondering why certain files don't appear is not a good idea.

I'd suggest just adding the profile/trace files and leaving the rest as-is (modulo comments).

MartinNowak:
> Based off githubs .gitignore & added profiling files

Please make sure to update upstream as well, e.g. https://github.com/github/gitignore/pull/1676.

MartinNowak:
Adding static libraries is rather rare and chances that someone adds them accidentally seem higher than the someone being unable to use `git add -f`. Might be a bit more tricky on Windows, but still.

WebFreak001:
they can always add an invert file or directory to the gitignore for the libraries, that's a lot cleaner than allowing all libraries and then adding them to the gitignore imo

s-ludwig:
I have static libraries in several of my repositories (for Optlink). When would a static library even be created that doesn't fall under the `<projectname>.a/lib` pattern? That seems like a pretty niche case looking at existing packages.

BTW, I have no idea how to add an inverted rule to gitignore. I could look it up, but the point is that having the wrong files ignored by default is bound to be very confusing for beginners, whereas accidentally checking in unwanted files is something obvious even for a git newcomer and also something explicitly controlled by be user.

WebFreak001:
Well it's not going to alter existing packages and when new ones are created they will see that the files are missing when they try to add them. In general I think hosting binary files on git is a bad idea because things like merging don't work on them.

Yes it can be confusing to git newcomers if a file is not checked in, but I think it's better than their projects having lots of junk in them. A git newcomer typically won't put in library files before he knows about the gitignore.

Inverting is just prepending `!` to the file

s-ludwig:
How often will this happen with the pattern that is already in now? Of course it won't affect current packages, the point is that there will probably be more packages like that. So the question is which problem do we actually solve with this? It potentially creates new ones, so it would be good to verify this first.

s-ludwig:
> Inverting is just prepending ! to the entry

My point was that, while I can say with good conscience that I'm pretty fluent with git, I didn't know this. Just assuming that people know about anything w.r.t. git is usually wrong.

wilzbach:
> Please make sure to update upstream as well, e.g. github/gitignore#1676.

There is also a pending one:

https://github.com/github/gitignore/pull/2300

wilzbach:
This will close https://github.com/dlang/dub/issues/824

PR #1233, opened 2017-08-26
Upgrade dependencies and fix download() with vibe.d 0.8.x.

s-ludwig:


s-ludwig:
@MartinNowak any objections to still include this for the upcoming release?

s-ludwig:
Going to pull this in, as it doesn't interfere with the normal build at all (no dependencies involved) and it's quite an oversight in the vibe.d 0.8.x transition.

Issue #591, opened 2015-06-17
Allow libs when target is a library


Temtaime:
Currently dub ignores all libraries used in project that targets a library.
I think it's wrong because we can pass static libraries to dmd as we pass the object files and those libraries will be merged with objects files resulting in one composite library.

Temtaime:
BTW C++ works in the same way : it allows to pass other libraries when linking a library.

s-ludwig:
Instead of merging it into the library file, it will link it to the final executable or shared library. This is necessary to avoid issues in diamond hierarchies (multiple definition of symbol X).

But, in cases where the developer knows that there will be no issues and where the built static library is supposed to be used outside of the DUB ecosystem, we could add some kind of option to force it being merged.

As a workaround it should work to put the libraries into the "dflags" field.

WebFreak001:
`"libs"` can be used to specify paths to statically built libraries to link in at the end.

`"sourceFiles"` can be used (and has been used) to pass compiled object or archive files into the target.

If there is something I missed here, feel free to reopen.

PR #1236, opened 2017-08-27
Avoid passing linker files to a static library build command line

### LINKED ISSUE #1235: Linker files are passed to the compiler command line when building static lib
When building a static library, no linker files should be passed to the compiler, as those will instead be passed to the executable/dynamic library that is built in the last step. Currently, linker files specified as `"sourceFiles"`, however, will end up on the command line.

s-ludwig:
Fixes #1235.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -461,12 +461,12 @@ class BuildGenerator : ProjectGenerator {
 			      (either in the dub.json, or using a command line flag)
 		*/
 		} else if (generate_binary && (settings.buildMode == BuildMode.allAtOnce || settings.compiler.name != "dmd" || is_static_library)) {
+			// don't include symbols of dependencies (will be included by the top level target)
+			if (is_static_library) buildsettings.sourceFiles = buildsettings.sourceFiles.filter!(f => !f.isLinkerFile()).array;
+
 			// setup for command line
 			settings.compiler.setTarget(buildsettings, settings.platform);
 			settings.compiler.prepareBuildSettings(buildsettings, BuildSetting.commandLine);
-
-			// don't include symbols of dependencies (will be included by the top level target)
-			if (is_static_library) buildsettings.sourceFiles = buildsettings.sourceFiles.filter!(f => !f.isLinkerFile()).array;
 
 			// invoke the compiler
 			settings.compiler.invoke(buildsettings, settings.platform, settings.compileCallback);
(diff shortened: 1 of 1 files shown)


John-Colvin:
After this, will there be any way to embed another static library inside the target static library? I've needed to do this in order to avoid linking order problems before.

s-ludwig:
Do you remember the scenario in which the linking order was wrong? The idea is that DUB handles this automatically in all usual cases. There is a related issue (#591) for "libs". Using "dflags" instead of "sourceFiles" should work as an alternative, although I didn't verify this.

source/dub/generators/build.d
```diff
@@ -565,3 +565,35 @@ private Path getTargetPath(in ref BuildSettings bs, in ref GeneratorSettings set
...
+	import dub.compilers.gdc : GDCCompiler;
+
+	auto desc = parseJsonString(`{"name": "test", "targetType": "library", "sourceFiles": ["foo.d", "bar.a"]}`);
+	auto pack = new Package(desc, Path("/tmp/fooproject"));
+	auto pman = new PackageManager(Path("/tmp/foo/"), Path("/tmp/foo/"), false);
+	auto prj = new Project(pman, pack);
+
+	final static class TestCompiler : GDCCompiler {
+		override void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback) {
+			assert(!settings.dflags[].any!(f => f.endsWith("bar.a")));
```
wilzbach: canFind would be safer here.

wilzbach: Yes. For example if the Generator adds an extra whitespace `endsWith("bar.a")` would be true..

s-ludwig: Hm... can you give a full example of what case you have in mind and how it would go wrong? I guess shouldn't have asked an "or" question ;)

wilzbach: Buggy implementation returns `["bar.a "]` - sorry for being so vague. I just wanted to say that using `canFind` is safer for the future.

s-ludwig: Mh.. okay :-)

s-ludwig:
Also fixed the test for Windows, because I noticed that `isLinkerFile` currently decides only based on the host operating system what is a linker file ending. Instead this clearly has to be for the compiler to decide. I'll open a ticket for this.

PR #1239, opened 2017-09-01
Merge remote-tracking branch 'upstream/stable' into merge_stable

MartinNowak:


ZombineDev:
Now that 1.5.0 tagged, it broke Jenkins - https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-5717/1/pipeline. @MartinNowak can we remove the restriction on Jenkins for only testing official releases? That way we'll be able to catch such errors much sooner.

PR #1243, opened 2017-09-05
Fix #555, tweak sed pattern in build.sh to check dmd version properly for ldc

### LINKED ISSUE #555: Fails to build with LDC on Arch GNU/Linux
When I try to build this with LDC, I get the following:

```
[koz@Emi dub]$ ./build.sh
Generating version file...
Running ldmd2...
/usr/bin/ld: cannot find -lphobos2-ldc
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
```

What do I need to change to make this build with LDC? Is this even possible?

joakim-noah:
After this, it shouldn't apply the old fix for dmd before 2.069.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -16,7 +16,7 @@ if [ "$DMD" = "" ]; then
 	exit 1
 fi
 
-VERSION=$($DMD --version 2>/dev/null | sed -n 's|DMD.* v||p')
+VERSION=$($DMD --version 2>/dev/null | sed -En 's|.*DMD.* v([[:digit:]\.]+).*|\1|p')
 # workaround for link order issues with libcurl (phobos needs to come before curl)
 if [[ $VERSION < 2.069.0 ]]; then
 	# link against libcurl


John-Colvin:
I'm pretty sure sed on macOS doesn't support `-r`, not sure about `:digit:`

s-ludwig:
Good point. I thought we had OSX tests enabled on Travis-CI, but obviously we don't.

s-ludwig:
Just tested and using `-En` instead of `-rn` should do the trick on mac OS (including the `[:digit:]`).

John-Colvin:
w.r.t. `macOS` tests on travis, I tried but didn't quite get there, a lot of GNU dependencies: https://github.com/dlang/dub/pull/1055

joakim-noah:
Updated to use `-En`, works fine for me on Arch linux/x64 and Termux for Android/ARM.

PR #1246, opened 2017-09-10
Use multiple registry fallbacks instead of one

wilzbach:
As we already see sometimes errors on Travis while downloading the installer script due to dlang.org and nightlies.dlang.org being down, it's imho a good idea to have multiple fallbacks by default for DUB.

https://code-mirror2.dlang.io is maintained by myself and the Heroku-App is based on https://github.com/dlang/dub-registry/pull/231

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -65,7 +65,7 @@ static this()
 
 /// The URL to the official package registry.
 enum defaultRegistryURL = "http://code.dlang.org/";
-enum fallbackRegistryURL = "https://code-mirror.dlang.io/";
+enum fallbackRegistryURLs = ["https://code-mirror.dlang.io/"];
 
 /** Returns a default list of package suppliers.
 
@@ -80,7 +80,7 @@ PackageSupplier[] defaultPackageSuppliers()
 	return [
 		new FallbackPackageSupplier(
 			new RegistryPackageSupplier(URL(defaultRegistryURL)),
-			new RegistryPackageSupplier(URL(fallbackRegistryURL))
+			fallbackRegistryURLs.map!(x => cast(PackageSupplier) new RegistryPackageSupplier(URL(x))).array
 		)
 	];
 }
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -64,8 +64,14 @@ static this()
 }
 
 /// The URL to the official package registry.
-enum defaultRegistryURL = "http://code.dlang.org/";
-enum fallbackRegistryURLs = ["https://code-mirror.dlang.io/"];
+enum defaultRegistryURL = "https://code.dlang.org/";
+enum fallbackRegistryURLs = [
+	// fallback in case of HTTPS problems
+	"http://code.dlang.org/",
+	"https://code-mirror.dlang.io/",
+	"https://code-mirror2.dlang.io/",
+	"https://dub-registry.herokuapp.com/",
+];
 
 /** Returns a default list of package suppliers.
 
--- a/source/dub/packagesupplier.d
+++ b/source/dub/packagesupplier.d
@@ -303,7 +303,7 @@ package abstract class AbstractFallbackPackageSupplier : PackageSupplier
 
 	override @property string description()
 	{
-		import std.algorithm.iteration : map;
+		import std.algorithm : map;
 		return format("%s (fallback %s)", m_default.description, m_fallbacks.map!(x => x.description));
 	}
 
--- a/source/dub/packagesupplier.d
+++ b/source/dub/packagesupplier.d
@@ -325,14 +325,15 @@ private template fallback(T, alias func)
 {
 	enum fallback = q{
 		import std.range : back, dropBackOne;
+		import dub.internal.vibecompat.core.log : logDebug;
 		scope (failure)
 		{
 			foreach (m_fallback; m_fallbacks.dropBackOne)
 			{
 				try
 					return m_fallback.%1$s(args);
-				catch(Throwable)
-				    assert(1);
+				catch(Exception)
+					logDebug("Package supplier %s failed. Trying next fallback.", m_fallback);
 			}
 			return m_fallbacks.back.%1$s(args);
 		}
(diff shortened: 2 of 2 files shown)


wilzbach:
Btw we didn't we simply add a list of RegistryPackageSuppliers as default?
A lot of the methods of `Dub` look like this:

```d
foreach(ps; m_packageSuppliers){
	try {
		auto versions = ps.getVersions(p);
```

Doing the fallbacking as "Meta" supplier as done now and improved by this PR _and_ within the class itself seems redundant.

MartinNowak:
> Btw we didn't we simply add a list of RegistryPackageSuppliers as default?

This was discussed in the original PR, https://github.com/dlang/dub/pull/1190#issuecomment-320616565.

s-ludwig:
Is it possible to set up a CNAME record for codemirror3.dlang.io that points to the Heroku app, or do they do any internal dispatching based on the original name? It would be good to keep a central place where the mirrors can be replaced/removed, so that we don't have to rely on application updates until such modifications take effect.

wilzbach:
> Is it possible to set up a CNAME record for codemirror3.dlang.io that points to the Heroku app, or do they do any internal dispatching based on the original name?

Custom CNames are free at Heroku, but SSL for them isn't. We could, however, setup a wildcard redirect. dlang.io can be managed via Namecheap - I will invite you later today.

s-ludwig:
A wildcard redirect would work. Regarding switching to HTTPS for code.dlang.org. As a simple first step we could have `["https://code.dlang.org/", "http://code.dlang.org/", ...]` to provide the fallback in case any client has problems with HTTPS.

wilzbach:
> but uses the same DNS as code-mirror.dlang.io (another failure point, e.g. misconfiguration)

@s-ludwig I added the Heroku App as code-mirror3.dlang.io, but I have to agree with @MartinNowak: by using the Heroku DNS, we have another fallback in case NameCheap has some issues (also wildcard redirect don't seem to be possible directly with NameCheap)

But I did add "http://code.dlang.org/" as a fallback if SSL isn't supported and fixed the Travis errors for 2.066 (as GDC is at 2.074.0 in master and 2.068.0 with their last release, we could drop the support for 2.066)

s-ludwig:
Alternatively, we could add some other mechanism for maintaining the mirror list in the future, such as requesting an up-to-date one from code.dlang.org, which would ideally be signed to avoid a worthwhile security target.  BTW, package signing is also something we should also put on the (virtual) higher priority list as the eco system keeps growing.

MartinNowak:
Let's not sweat this too much, adding a few mirrors is nice as it might help with certain issues.
But this mirror thing isn't a replacement for a proper HA setup (or PaaS) in the long-term.

source/dub/packagesupplier.d
```diff
@@ -292,17 +292,19 @@ class RegistryPackageSupplier : PackageSupplier {
...
 		m_default = default_;
-		m_fallback = fallback;
+		m_fallbacks = fallbacks;
 	}
 
 	override @property string description()
 	{
-		return format("%s (fallback %s)", m_default.description, m_fallback.description);
+		import std.algorithm : map;
+		return format("%s (fallback %s)", m_default.description, m_fallbacks.map!(x => x.description));
```
MartinNowak: Not a blocker, but this is fairly long and detailed.
```
==== registry at https://code.dlang.org/ (fallback ["registry at http://code.dlang.org/", "registry at https://code-mirror.dlang.io/", "registry at https://code-mirror2.dlang.io/", "registry at https://dub-registry.herokuapp.com/"]) ====
```

source/dub/packagesupplier.d
```diff
@@ -322,7 +324,18 @@ package alias FallbackPackageSupplier = AutoImplement!(AbstractFallbackPackageSu
...
-		scope (failure) return m_fallback.%1$s(args);
+		import std.range : back, dropBackOne;
+		scope (failure)
+		{
+			foreach (m_fallback; m_fallbacks.dropBackOne)
+			{
+				try
+					return m_fallback.%1$s(args);
+				catch(Throwable)
+				    assert(1);
```
MartinNowak: Isn't `{}` what you should use for an empty statement?

source/dub/packagesupplier.d
```diff
@@ -322,7 +324,18 @@ package alias FallbackPackageSupplier = AutoImplement!(AbstractFallbackPackageSu
...
+			}
+			return m_fallbacks.back.%1$s(args);
```
MartinNowak: The recursive structure was setup to chain exceptions, so that users get a list of unreachable sites. Not too important though.

source/dub/dub.d
```diff
@@ -80,7 +86,7 @@ PackageSupplier[] defaultPackageSuppliers()
 	return [
 		new FallbackPackageSupplier(
 			new RegistryPackageSupplier(URL(defaultRegistryURL)),
-			new RegistryPackageSupplier(URL(fallbackRegistryURL))
+			fallbackRegistryURLs.map!(x => cast(PackageSupplier) new RegistryPackageSupplier(URL(x))).array
```
MartinNowak: Could have stayed with the recursive structure.
```d
fallbackRegistryURLs.reduce!((agg, url) => new FallbackPackageSupplier(agg, new RegistryPackageSupplier(URL(x))))
```

source/dub/packagesupplier.d
s-ludwig: More importantly, `Throwable` is a bad idea, must be `Exception` - logging the error with `logDiagnostic` or maybe just `logDebug` could also be important for remote debugging .

source/dub/packagesupplier.d
wilzbach: I know, but the `PackageSupplier` interface currently only supports a `description` attribute. We could add sth. like `location` in a follow-up?

source/dub/dub.d
wilzbach: Hmm, I thought the "linear" way is more intuitive, but I don't really mind. If you feel strongly the recursive structure, feel free to change it back.

wilzbach:
> Just nits, lets move on with this.

Addressed the nits.

> The recursive structure was setup to chain exceptions, so that users get a list of unreachable sites. Not too important though.

Don't they get this with `logDebug` as well?
Anyhow as mentioned before if you prefer the recursive way, feel free to change.
In any case it would be great to get this in before the next release.

Issue #1210, opened 2017-08-01
dub doesn't handle gzip encoding from code.dlang.org


dazoe:
I couldn't find this reported anywhere yet.

### System information
Windows 10
DUB version 1.4.0, built on Jul 19 2017
DMD32 D Compiler v2.075.0

Archlinux
DUB version 1.4.0, built on Jul 31 2017
DMD32 D Compiler v2.075.0

### Bug Description
dub fails to fetch packages from code.dlang.org

```
C:\tmp>dub fetch dub
Getting a release version failed: No package dub was found matching the dependency >=0.0.0
Retry with ~master...
No package dub was found matching the dependency ~master

C:\tmp>dub fetch dub -v
Using dub registry url 'http://code.dlang.org/'
Refreshing local packages (refresh existing: true)...
Looking for local package map at C:\ProgramData\dub\packages\local-packages.json
Looking for local package map at C:\Users\Dave\AppData\Roaming\dub\packages\local-packages.json
No package found in current working directory.
Package dub not found for registry at http://code.dlang.org/: http://code.dlang.org/packages/dub.json(1): Error: Expected valid JSON token, got '      φ¥'.
Getting a release version failed: No package dub was found matching the dependency >=0.0.0
Retry with ~master...
Package dub not found for registry at http://code.dlang.org/: http://code.dlang.org/packages/dub.json(1): Error: Expected valid JSON token, got '      φ¥'.
No package dub was found matching the dependency ~master
```

This boils down to dub.internal.utils.download not handling the gzip content-encoding from code.dlang.org.
This can be fixed by enabling libcurl's internal encoding handling adding the following to dub.internal.utils.setupHTTPClient()
```
		conn.handle.set(CurlOption.encoding, "");
```

ohdatboi:
It's fixed. Could somebody close this issue? Thanks.

PR #1247, opened 2017-09-10
Handle gzip compression from code.dlang.org

dazoe:
fix for #1210

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -332,6 +332,8 @@ version(DubUseCurl) {
 		auto noProxy = environment.get("no_proxy", null);
 		if (noProxy.length) conn.handle.set(CurlOption.noproxy, noProxy);
 
+		conn.handle.set(CurlOption.encoding, "");
+
 		conn.addRequestHeader("User-Agent", "dub/"~getDUBVersion()~" (std.net.curl; +https://github.com/rejectedsoftware/dub)");
 	}
 }


s-ludwig:
Interesting, code.dlang.org doesn't do any gzip encoding, except for a single URL, do you possibly have a proxy in-between?

If I understand it correctly, the `set` call will cause cURL to send an empty "Accepts-Encoding" header, thus disabling any gzip encoding, right? IMO that would be a little too much, as we might have to re-enable gzip encoding in general at some point to save some bandwidth.

MartinNowak:
No, this enables encoding handling by curl, see https://curl.haxx.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html. Setting it to `null` would disable any encoding.

dazoe:
Ah, I didn't think about the whole proxy thing. I use a mobile connection and apparently they MTIM my http traffic and force gzip compression. I confirmed this by running some curl fetches from a remote server.

Calling set with an empty string causes libcurl to send an Accept-Encoding header containing libcurl's built-in supported encodings which libcurl handles transparently. Reference: https://curl.haxx.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html

MartinNowak:
Most MNOs do such things, better we'd move to HTTPS at some point, and maybe also add gzip encoding to dub-registry or it's reverse proxy (taking care of [BREACH](https://en.wikipedia.org/wiki/BREACH)).

s-ludwig:
Okay, I didn't find the official documentation page and just guessed based on other sources, sounds fine then.

We could switch to HTTPS at any time, at least for code.dlang.org. There would just have to be a configuration option to force plain HTTP to avoid breakage on more niche systems. Taking care of BREACH attacks is of course important, but since we are coming from a completely unencrypted connection and there is no confidential information in the requests that DUB itself performs, this fortunately isn't very critical in the beginning.

MartinNowak:
Was thinking more of a http->https redirect on code.dlang.org. But even for that, there shouldn't be any secrets in the payloads (only uncompressed session cookies in HTTP headers).

PR #1248, opened 2017-09-12
Allow testing unittests in main() module

### LINKED ISSUE #1118: The unittest doesn't work for an application
Here is my test environment:
Build tool: DUB version 1.3.0
Compiler : DMD32 D Compiler v2.074.0
OS: Windows 10, Ubuntu
Build type: unittest

Test code:
```D
import std.stdio;

void main()
{
	writeln("Edit source/app.d to start your project.");
}

unittest
{
	assert(false, "It doesn't work.");
}
```

dhasenan:
We avoid defining our own main() for executables. In order to avoid
running main(), we provide our own test runner that exits when done
instead of ceding control back to the runtime.

The downside is that module destructors are not called.

Fixes #1118.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -589,11 +589,13 @@ class Dub {
 				scope(exit) fil.close();
 				fil.write("module dub_test_root;\n");
 				fil.write("import std.typetuple;\n");
-				foreach (mod; import_modules) fil.write(format("static import %s;\n", mod));
+				// Rename imports to avoid name collisions.
+				foreach (i, mod; import_modules) fil.write(format("static import m%s = %s;\n", i, mod));
 				fil.write("alias allModules = TypeTuple!(");
 				foreach (i, mod; import_modules) {
 					if (i > 0) fil.write(", ");
-					fil.write(mod);
+					fil.write("m");
+					fil.write(i.to!string);
 				}
 				fil.write(");\n");
 				if (custommodname.length) {
@@ -608,10 +610,13 @@ class Dub {
 
 						bool defaultUnitTester() {
 							import core.stdc.stdlib : exit;
+							import core.runtime : Runtime;
 							foreach (modul; allModules)
 								foreach (test; __traits(getUnitTests, modul))
 									test();
 							writeln("All unit tests have been run successfully.");
+							Runtime.terminate;
+							exit(0);
 							return false;
 						}
 
@@ -621,7 +626,7 @@ class Dub {
 								import tested;
 								import core.runtime;
 								import std.exception;
-								Runtime.moduleUnitTester = () => false;
+								Runtime.moduleUnitTester = () => true;
 								//runUnitTests!app(new JsonTestResultWriter("results.json"));
 								enforce(runUnitTests!allModules(new ConsoleTestResultWriter), "Unit tests failed.");
 							} else {
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -548,8 +548,12 @@ class Dub {
 			tcinfo.targetName = test_config;
 
 			if (lbuildsettings.targetType != TargetType.executable) {
-				logInfo("target type is %s, which doesn't have a main(); we're providing one");
-				tcinfo.versions[""] ~= "dub_test_main";
+				// HACK for vibe.d's legacy main() behavior (for libraries that depend on an old
+				// vibe version).
+				m_project.rootPackage.recipe.buildSettings.versions[""] = m_project.rootPackage.recipe.buildSettings.versions.get("", null).remove!(v => v == "VibeDefaultMain");
+				tcinfo.versions[""] ~= "VibeCustomMain";
+				// TODO: remove this ^ once vibe.d has removed the default main implementation
+				tcinfo.versions[""] ~= "DubTestMain";
 			}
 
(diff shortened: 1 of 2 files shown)


source/dub/dub.d
```diff
@@ -546,10 +546,11 @@ class Dub {
 			BuildSettingsTemplate tcinfo = m_project.rootPackage.recipe.getConfiguration(config).buildSettings;
 			tcinfo.targetType = TargetType.executable;
 			tcinfo.targetName = test_config;
-			// HACK for vibe.d's legacy main() behavior:
-			tcinfo.versions[""] ~= "VibeCustomMain";
-			m_project.rootPackage.recipe.buildSettings.versions[""] = m_project.rootPackage.recipe.buildSettings.versions.get("", null).remove!(v => v == "VibeDefaultMain");
-			// TODO: remove this ^ once vibe.d has removed the default main implementation
```
s-ludwig: This block still needs to stay for a while, since although `VibeDefaultMain` isn't endorsed since a few releases anymore, there is probably still a lot of code out there that relies on it.

source/dub/dub.d
```diff
@@ -546,10 +546,11 @@ class Dub {
...
+
+			if (lbuildsettings.targetType != TargetType.executable) {
+				logInfo("target type is %s, which doesn't have a main(); we're providing one");
+				tcinfo.versions[""] ~= "dub_test_main";
```
s-ludwig: Not sure whether there is a documented convention for this, but except for the `Have_xxx` versions that DUB generates, versions are usually `CamelCase` or a mixture of camel case and underscores.

source/dub/dub.d
```diff
@@ -603,15 +599,33 @@ class Dub {
...
+						version (dub_test_main) void main() {}
+
+						bool defaultUnitTester() {
+							import core.stdc.stdlib : exit;
+							foreach (modul; allModules)
+								foreach (test; __traits(getUnitTests, modul))
+									test();
+							writeln("All unit tests have been run successfully.");
+							exit(0);
+							return true;
```
s-ludwig: Should be `return false;`, so that the application main function isn't executed afterwards. That makes it possible to remove the `exit(0);` and solves the module destructor issue.

source/dub/dub.d
```diff
@@ -603,15 +599,33 @@ class Dub {
...
+						}
+
+						shared static this()
+						{
 							version (Have_tested) {
 								import tested;
 								import core.runtime;
 								import std.exception;
 								Runtime.moduleUnitTester = () => true;
```
s-ludwig: Again, `() => false` due to the applications `main()` getting called otherwise.

source/dub/dub.d
```diff
@@ -546,10 +546,11 @@ class Dub {
 			BuildSettingsTemplate tcinfo = m_project.rootPackage.recipe.getConfiguration(config).buildSettings;
 			tcinfo.targetType = TargetType.executable;
 			tcinfo.targetName = test_config;
-			// HACK for vibe.d's legacy main() behavior:
-			tcinfo.versions[""] ~= "VibeCustomMain";
-			m_project.rootPackage.recipe.buildSettings.versions[""] = m_project.rootPackage.recipe.buildSettings.versions.get("", null).remove!(v => v == "VibeDefaultMain");
-			// TODO: remove this ^ once vibe.d has removed the default main implementation
+
+			if (lbuildsettings.targetType != TargetType.executable) {
+				logInfo("target type is %s, which doesn't have a main(); we're providing one");
```
s-ludwig: Not sure whether the user usually needs to care about this. I'd change this to `logDiagnostic` to avoid polluting the output unnecessarily.

source/dub/dub.d
dhasenan: `return false` means the program exits with code 2.

This means I need to run the module destructors manually and then exit with code 0.

source/dub/dub.d
dhasenan: Specifically a library with a dependency on an ancient vibe version. Done.

source/dub/dub.d
s-ludwig: Hm, you are right, that sucks.... I was sure that something like this was mentioned as a way to skip `main` during the last discussion about the unit testing topic. There was also supposed to be some improvement in this area in a recent DMD release, but I could neither find anything in the change log, nor the discussion.

An alternative possibility could be to define a C `main()`, but the question is whether that is less risky overall than skipping the module finalizers. Speaking of which, what about just calling `Runtime.terminate();` just before `exit(0);`?

dhasenan: Bit ahead of you there.

dhasenan:
The test failure for `issue502-root-import` is strange. When building locally, it generates a different executable with `dub test` than `dub build` -- the md5sums differ. But the filenames are the same, so it clobbers your exe with your unittests and vice versa. Whichever way it's built, it runs the app's main and doesn't run unittests.

I even put an `assert(0) at the start of `testProject` to see if it would run, and that assert wasn't tripped. So there's something weird going on there, and I'm not sure what.

Geod24:
I don't think we need this anymore. First we now have `mainSourceFile` for the file to exclude from the configuration (by design), but we also don't have the issue of main running after unittests since druntime changed. The `dub test` command still need a bit of love, but I'm fairly sure it's safe to close this.

PR #1252, opened 2017-09-16
Automatically generate source file list

### LINKED ISSUE #953: build.sh errors
```
sh build.sh
build.sh: 20: build.sh: cannot open 2.069.0: No such file
build.sh: 20: build.sh: [[: not found
build.sh: 26: build.sh: [[: not found
build.sh: 32: build.sh: [[: not found
Generating version file...
Running dmd...
```
- These are not fatal errors, but why not just replace `sh build.sh` by some `rdmd build.d`, which is much easier to read and maintain?
- Furthermore, inside the script, why not use rdmd instead of `build-files.txt` (DRY)

dhasenan:
It's cheap and easy to do this, so why not?

Confirmed that it should work on FreeBSD and Linux.

Fixes #953 (the second part of it, anyway)


MartinNowak:
Thx, extended that to Windows to and removed build-files.txt.

s-ludwig:
Unfortunately triggers an issue in GDC. Maybe try to reverse the sort order?

Geod24:
Closing since this PR has been stale for 2 years.
It's a good idea and I'll add it as a follow up to #1821 .

PR #1254, opened 2017-09-18
change default registry to use https

MartinNowak:


s-ludwig:
Needs to be adjusted/rebased for the latest master changes.

wilzbach:
Master now contains this list:

```
enum defaultRegistryURL = "https://code.dlang.org/";
enum fallbackRegistryURLs = [
	// fallback in case of HTTPS problems
	"http://code.dlang.org/",
	"https://code-mirror.dlang.io/",
	"https://code-mirror2.dlang.io/",
	"https://dub-registry.herokuapp.com/",
];
```

As the HTTPS problems are unlikely, we might want to move that one to the end of the list, but apart from that I hope we are good with mirrors for now.
Anyhow as this has been integrated with master -> closing.

Issue #1206, opened 2017-07-26
Using ldmd2 causes a warning about possible "build errors" to be spit out


joakim-noah:
If I use ldmd2 with dub, it always spits out the following warning, despite not having any actual problems:
```
The determined compiler type "ldc" doesn't match the expected type "dmd". This will probably result in build errors.
```
What is the point of this warning?  Looking at the code, it appears that it sets [the `Compiler` to dmd here](https://github.com/dlang/dub/blob/master/source/dub/compilers/compiler.d#L39), since ldmd2 has dmd in its name and must be passed dmd-style flags.  However, [the `BuildPlatform.compiler` is set using the inbuilt version flag to ldc2](https://github.com/dlang/dub/blob/master/source/dub/platform.d#L134), so [that warning is always generated for ldmd2](https://github.com/dlang/dub/blob/master/source/dub/compilers/compiler.d#L134).  I imagine the same is true for gdmd, though I haven't tried it.

The warning seems useless, maybe we can get rid of it or refine it.

joakim-noah:
Hmm, it turns out the [`BuildPlatform.compiler` is actually set here to ldc, not ldc2](https://github.com/dlang/dub/blob/master/source/dub/compilers/utils.d#L348), as you can see in my warning message above.  I wonder why the unused stuff in `dub.platform` is even around.

timotheecour:
just ran into this as well
dub --version
DUB version 1.6.0, built on Nov 26 2017

PR #1255, opened 2017-09-20
Refactor determinePlatform, determineArchitecture, determineCompiler so their checks can be reused

joakim-noah:
~~Only `determinePlatform` was used anywhere, in a single unit test.  Cut-and-pasted the doc comments into `dub.compilers.utils`, where the versions of these functions were actually used.~~  [Mentioned this duplication before when I was looking at a different issue](https://github.com/dlang/dub/issues/1206#issuecomment-318161102), ~~thought I'd get rid of~~ refactored it.

### DIFF
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -8,7 +8,7 @@
 module dub.compilers.utils;
 
 import dub.compilers.buildsettings;
-import dub.platform;
+import dub.platform : BuildPlatform, archCheck, compilerCheck, platformCheck;
 import dub.internal.vibecompat.core.log;
 import dub.internal.vibecompat.inet.path;
 import std.algorithm : canFind, endsWith, filter;
@@ -255,9 +255,9 @@ Path generatePlatformProbeFile()
 		fil.close();
 	}
 
-	// NOTE: This must be kept in sync with the dub.platform module
 	fil.write(q{
 		module dub_platform_probe;
+		import std.array;
 
 		template toString(int v) { enum toString = v.stringof; }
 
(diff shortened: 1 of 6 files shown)


source/dub/platform.d
```diff
@@ -1,141 +1,17 @@
...
-	The possible values currently are "dmd", "gdc", "ldc2" or "sdc". If an
-	unknown compiler is used, this function will return an empty string.
-
-	See_Also: `determineBuildPlatform`
-*/
-string determineCompiler()
-{
-	version(DigitalMars) return "dmd";
-	else version(GNU) return "gdc";
-	else version(LDC) return "ldc2";
```
joakim-noah: Despite the comment above to keep the two in sync, [they weren't since more than three years ago](https://github.com/dlang/dub/commit/456329eb#diff-554996e078f7c2bc3fe664acc6ead97dR84).  Didn't really matter once this function stopped being used, but might as well just remove it.

s-ludwig:
This is a public API and we can't just remove it (except for DUB 2.0.0). But what we can do is to refactor the platform probe generation code, so that parts can be mixed into the functions in `dub.platform` and the manual syncing isn't necessary anymore.

BTW, I'm actually using this in an external project (a build worker implementation).

joakim-noah:
OK, wasn't sure if you had public APIs.  I should have looked at `dub.sdl`, where dub's available as a library too.

I thought about refactoring these into one place, but wasn't sure how to do it, as I was thinking in terms of keeping the definitions in `dub.platform` and somehow mixing those into `dub.compilers.utils`.  I didn't think of the other way around, mixing the function strings from `dub.compilers.utils` into `dub.platform`: that should work.

I assume you don't care that `dub.platform.determineCompiler` will return `ldc` not `ldc2` after the change?

s-ludwig:
> I assume you don't care that dub.platform.determineCompiler will return ldc, not ldc2, after the change?

Hmm, "ldc2" sounds like a bug, since the package format documentation clearly states that there is a direct correspondence to the predefined version identifiers. So I don't think we have a real choice here than to risk breaking code that relies on the old behavior (my personal code doesn't, though, and I'd not expect that there are many other people using this module, so the risk seems acceptable).

joakim-noah:
Alright, refactored into constant strings that can be mixed in or pasted, so that there's only one version of each arch/OS/compiler check in the codebase now.  As discussed, the only change in behavior that I see is the switch from `ldc2 to `ldc` in `build.platform.determineCompiler`.

joakim-noah:
Single test failure on Travis seems spurious, networking issue that causes zip file not to fully download?  Looks like this pull is good.

source/dub/platform.d
```diff
@@ -8,43 +8,18 @@
...
-	language, except that they are converted to lower case.
-
-	See_Also: `determineBuildPlatform`
-*/
-string[] determinePlatform()
-{
+// archCheck, compilerCheck, and platformCheck are used below and in
+// generatePlatformProbeFile, so they've been extracted into these strings
+// that can be reused.
+enum string platformCheck = q{
```
s-ludwig: Can you put a `/// private` comment right in from of this and the other `...Check` enums? That will keep them from being included in the documentation. It would be nicer to make them `package(dub.compilers)`, but I'm pretty sure that we still support frontend versions that don't support that syntax.

s-ludwig:
LGTM otherwise!

source/dub/platform.d
s-ludwig: Actually, just let me just do that from the GitHub UI quickly...

source/dub/compilers/utils.d
```diff
@@ -255,9 +255,9 @@ Path generatePlatformProbeFile()
 		fil.close();
 	}
 
-	// NOTE: This must be kept in sync with the dub.platform module
 	fil.write(q{
 		module dub_platform_probe;
+		import std.array;
```
MartinNowak: Importing phobos in the probe introduced a performance regression, see https://github.com/dlang/dub/pull/1318.

PR #1263, opened 2017-10-03
rewrite generator targets configuration

MartinNowak:
- as discusses in in https://github.com/dlang/dub/pull/1128 change the
  configuration algorithm to solve polluting sibling dependencies with
  compiler flags (e.g. versions), thus making it possible to better
  reuse cached builds

0. collect dependencies and shallow configuration
1. downwards inherit versions, debugVersions, and inheritable build options
2. define Have_dependency_xyz for all direct dependencies
3. upwards inherit dflags, versions, debugVersions, import paths,
    string import paths (and libs, lflags for static libs)
4. override string import files in dependencies

A rather big change here, I no longer add import paths from executable dependencies. As those don't get linked anyhow I find it hard to believe that someone relied on that.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -89,18 +89,20 @@ class ProjectGenerator
 
 		if (!settings.config.length) settings.config = m_project.getDefaultConfiguration(settings.platform);
 
-		TargetInfo[string] targets;
 		string[string] configs = m_project.getPackageConfigs(settings.platform, settings.config);
+		TargetInfo[string] targets;
 
 		foreach (pack; m_project.getTopologicalPackageList(true, null, configs)) {
-			BuildSettings buildsettings;
-			buildsettings.processVars(m_project, pack, pack.getBuildSettings(settings.platform, configs[pack.name]), true);
-			prepareGeneration(pack, m_project, settings, buildsettings);
+			BuildSettings buildSettings;
+			auto config = configs[pack.name];
+			buildSettings.processVars(m_project, pack, pack.getBuildSettings(settings.platform, config), true);
+			targets[pack.name] = TargetInfo(pack, [pack], config, buildSettings);
+
+			prepareGeneration(pack, m_project, settings, buildSettings);
 		}
 
-		string[] mainfiles;
-		collect(settings, m_project.rootPackage, targets, configs, mainfiles, null);
-		downwardsInheritSettings(m_project.rootPackage.name, targets, targets[m_project.rootPackage.name].buildSettings);
+		string[] mainfiles = configurePackages(m_project.rootPackage, targets, settings);
+
 		addBuildTypeSettings(targets, settings);
 		foreach (ref t; targets.byValue) enforceBuildRequirements(t.buildSettings);
 		auto bs = &targets[m_project.rootPackage.name].buildSettings;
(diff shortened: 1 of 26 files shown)


MartinNowak:
Could you please confirm, that this fixes your problem @John-Colvin.

source/dub/generators/generator.d
```diff
@@ -89,18 +89,20 @@ class ProjectGenerator
...
+		TargetInfo[string] targets;
 
 		foreach (pack; m_project.getTopologicalPackageList(true, null, configs)) {
-			BuildSettings buildsettings;
-			buildsettings.processVars(m_project, pack, pack.getBuildSettings(settings.platform, configs[pack.name]), true);
-			prepareGeneration(pack, m_project, settings, buildsettings);
+			BuildSettings buildSettings;
+			auto config = configs[pack.name];
+			buildSettings.processVars(m_project, pack, pack.getBuildSettings(settings.platform, config), true);
+			targets[pack.name] = TargetInfo(pack, [pack], config, buildSettings);
```
s-ludwig: If this is going to stay like this, there should be a comment here stating that all dependencies are assumed to be a target here and the non-targets are filtered out at the end of `configureTargets`. It was quite surprising to read this and the place where the map is cleaned up is not really obvious, considering the complexity of the algorithm.

source/dub/generators/generator.d
```diff
@@ -143,144 +145,278 @@ class ProjectGenerator
...
+			// use `visited` here as pkgs cannot depend on themselves
+			if (pack in visited)
+				return;
+			visited[pack] = typeof(visited[pack]).init;
+
+			auto bs = &ti.buildSettings;
+			if (isBinaryTarget[pack.name])
+				logDebug("%sConfiguring target %s (%s %s %s)", ' '.repeat(2 * level), pack.name, bs.targetType, bs.targetPath, bs.targetName);
+			else
+				logDebug("%sConfiguring non-binary target %s", ' '.repeat(2 * level), pack.name);
```
s-ludwig: Except potentially the root package, there is no such thing as a "non-binary target" - "non-target package" or similar would be more appropriate.

source/dub/generators/generator.d
```diff
@@ -143,144 +145,278 @@ class ProjectGenerator
...
-			targets[pack.name] = TargetInfo(pack, [pack], configs[pack.name], buildsettings, null);
+		configureDependencies(targets[rootPackage.name], targets);
 
-		auto deps = pack.getDependencies(configs[pack.name]);
-		foreach (depname; deps.keys.sort()) {
-			auto depspec = deps[depname];
-			auto dep = m_project.getDependency(depname, depspec.optional);
-			if (!dep) continue;
+		// 2. add Have_dependency_xyz for all direct dependencies of a target
+		// (includes incorporated non-target dependencies and their dependencies)
```
s-ludwig: Since they are inherited upwards anyway, isn't it enough to set the `Have_xyz` constant for just `ti.packages`?

s-ludwig:
> A rather big change here, I no longer add import paths from executable dependencies. As those don't get linked anyhow I find it hard to believe that someone relied on that.

Executable dependencies should have (well, not "should", but that's what I think is the current state of things) resulted in an error anyway, once a built was attempted, so this should indeed not be an issue.

source/dub/generators/generator.d
MartinNowak: Well, what you call non-targets are targets with TargetType `"sourceLibrary"` or `"none"`, so they are still processed, they just don't have any outputs.

MartinNowak:
Thx, finally revisited @s-ludwig.

> Just looking at the logic, this appears to be alright. It's also good to have the string import special case separated out like this. However, I still do have a few worries with this:
> 
>     It would have been nice if refactoring and the logical changes had been separate. It's really hard to tell if there are any subtle unintended changes in the logic just by review and I'm not confident that the test coverage is sufficient to blindly trust it, so we are going to have to take a leap of faith here

This is not such much a refactoring but more of a rewrite, with a different algorithm, split up into multiple passes over the dependency tree.

>     Mixing of nested functions with code IMO hurts readability. Having those functions separate as methods or even as free-standing functions would improve this quite a bit, because the high-level algorithm would become much easier to spot (despite the distant comments that list the high-level steps). It would make it also explicit where side-effects happen instead of burying them in a hidden context pointer.

Yes, not too nice, but there is a lot of data involved and the code is heavily entangled (e.g. removing and collection main source files happens as a side-effect of determining the target type). Most things just access most other things. Besides bundling all data to configure a package in `TargetInfo`, it's tricky to further reduce the data dependencies.

>     I don't really like the approach of (ab)using the targets map for non-targets and cleaning up later. Is there any reason why collectDependencies can't just invoke a function that performs the shallow configuration (step 0) and then sets the targets entries just for actual targets?

Well, for me it's targets with TargetType "sourceLibrary" or "none", so they are still processed but just don't have any outputs.
The main reason for this approach is that I need want to store all the required data to configure a package target in a hash table, so that I can split the different phases. Without this, I'd need to pass a lot more loose info around and/or intermingle the shallow configuration with the pre-build loop. Both of which makes the code less maintainable.

source/dub/generators/generator.d
MartinNowak: I've renamed `configureTargets -> confgurePackages` and added a note to https://github.com/dlang/dub/pull/1263/files#diff-92bd5c95e7074c6c3b0d83c3b5e15356R164. I think it's the right place to modify the target list in this function, e.g. consider a feature for package configs with multiple build targets.

MartinNowak:
BTW, we should prolly consider to grep the source code for `version (XYZ)` to determine actually used version identifiers (maybe providing some config option (features) to add non-detectable version identifiers in mixins).
Alternatively some `unused version identifier` output could be added to dmd, allowing to fixup the build cache id.
Either of those would help to improve the build cache without requiring manual listing.

s-ludwig:
Okay, maybe something can still be done about the data dependency issue in a later refactoring, I think we can move on with this for now. But the fact that I obviously missed a lot of these dependencies during careful review to me is a red flag that the algorithm needs to be further disentangled at some point.

Regarding "targets" and "packages", the idea was that a target is a desired piece of the build output. At the package level that means one or more packages. If we'd just call every package a target then we can as well drop the term altogether and stay with "package" or maybe "target package".

The version optimization would be interesting, especially with DMD cooperation. The "features" based approach still seems worthwhile as a general replacement for the current behavior, because a `version` is often not sufficient (e.g. import libraries or additional dependencies may be needed). Once we have them, we'd have to have a quite long deprecation phase, though.

MartinNowak:
> Okay, maybe something can still be done about the data dependency issue in a later refactoring, I think we can move on with this for now. But the fact that I obviously missed a lot of these dependencies during careful review to me is a red flag that the algorithm needs to be further disentangled at some point.

Thanks, and yes, more refactoring would definitely help here.

> The version optimization would be interesting, especially with DMD cooperation.

Had a simpler idea for now, adding a plumbing `usesVersionsInMixins` boolean config to dub (conservatively defaulting to true), and only grepping for `version\s(\([^)]*\))` if it's set to false.
I'll check how much that'd help with vibe-d's test times.
I've opened https://github.com/dlang/dub/issues/1312 for this.

> The "features" based approach still seems worthwhile as a general replacement for the current behavior, because a version is often not sufficient (e.g. import libraries or additional dependencies may be needed). Once we have them, we'd have to have a quite long deprecation phase, though.

Mmh, needs to be carefully designed to avoid overlapping with configs and subpackages. It seems that we/dub is already quite complex in that area.

PR #1283, opened 2017-11-06
Make the code compatible with vibe-core

s-ludwig:
vibe-core has a redesigned path module that is much more robust against accidential misuse, as well as path encoding aware. It also supports allocation-less construction from a path's string representation, which results in some non-backwards-compatible changes. This PR makes the code compile with these changes in addition to the already supported combinations.

### DIFF
--- a/source/dub/internal/vibecompat/inet/path.d
+++ b/source/dub/internal/vibecompat/inet/path.d
@@ -18,6 +18,9 @@ import std.exception;
 import std.string;
 
 
+deprecated("Use NativePath instead.")
+alias Path = NativePath;
+
 /**
 	Represents an absolute or relative file system path.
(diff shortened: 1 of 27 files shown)


source/dub/internal/vibecompat/inet/path.d
```diff
@@ -17,21 +18,28 @@ import std.exception;
 import std.string;
 
 
+deprecated("Use NativePath instead.")
```
wilzbach: Maybe it's a good idea to point out when or in which version the deprecation will be removed?

wilzbach:
Rebased to fix the merge conflict (there was no interaction needed, seems like GitHub is worse in merging than the CLI).

PR #1284, opened 2017-11-06
Let Dependency use generic interval logic.

s-ludwig:
Changes `Dependency` to rely on generic interval operations instead of manually working with version bounds all over the place. This makes it considerably easier to reason about such operations and simplifies the logic within `Dependency`.


s-ludwig:
Just noticed that there is still a "BUG" comment in the rebased code.

Geod24:
Done in #2271

PR #1286, opened 2017-11-08
Use shortened object file names in single-file build mode.

s-ludwig:
The full path was repeated twice in the object file path, resulting in excessively long paths (and failures on Windows). This change uses just the module name part (i.e. just the file *name* of the D file) and combines the path part as a single crc32 value (e.g. "somemod_0a56fb23.o").

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -401,7 +401,7 @@ class BuildGenerator : ProjectGenerator {
 
 	static string pathToObjName(string path)
 	{
-		import std.digest.crc;
+		import std.digest.crc : crc32Of;
 		import std.path : buildNormalizedPath, dirSeparator, relativePath, stripDrive;
 		if (path.endsWith(".d")) path = path[0 .. $-2];
 		auto ret = buildNormalizedPath(getcwd(), path).replace(dirSeparator, ".");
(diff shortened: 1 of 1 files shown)


source/dub/generators/build.d
```diff
@@ -401,8 +401,12 @@ class BuildGenerator : ProjectGenerator {
...
 	static string pathToObjName(string path)
 	{
-		import std.path : buildNormalizedPath, dirSeparator, stripDrive;
-		return stripDrive(buildNormalizedPath(getcwd(), path~objSuffix))[1..$].replace(dirSeparator, ".");
+		import std.digest.crc : crc32Of;
+		import std.path : buildNormalizedPath, dirSeparator, relativePath, stripDrive;
+		if (path.endsWith(".d")) path = path[0 .. $-2];
+		auto ret = buildNormalizedPath(getcwd(), path).replace(dirSeparator, ".");
+		auto idx = ret.lastIndexOf('.');
+		return idx < 0 ? ret ~ objSuffix : format("%s_%(%02x%)%s", ret[idx+1 .. $], crc32Of(ret[0 .. idx]), objSuffix);
```
wilzbach: FWIW `toHexString!(LetterCase.lower)` is imho easier to read than `%(%02x%)`

s-ludwig: Hm, that's indeed true. For some reason I thought that `toHexString` wouldn't work with CRC32.

PR #1287, opened 2017-11-08
Skip explicitly linking with libcurl for LDC on Windows.

s-ludwig:
Attempts to fix the failing AppVeyor build that currently ruins the PR overview page by making it see that all PRs are broken.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -15,7 +15,6 @@ configuration "application" {
 
 configuration "library" {
 	targetType "library"
-	libs "curl"
 	excludedSourceFiles "source/app.d"
 	copyFiles "bin/libcurl.dll" "bin/libeay32.dll" "bin/ssleay32.dll" platform="windows"
 	versions "DubUseCurl"


s-ludwig:
@MartinNowak: Can we disable the jenkins/branch test? It fails unconditionally for all PRs and presumably doesn't add much value, considering that Travis already tests branch and merge commits.

s-ludwig:
Was just about to hit "merge" when I noticed that this is accidentally using the JSON platform suffix syntax instead of `platform="..."`. Searching for the origins of `libs "curl"` revealed that it was actually added to make the Windows build work in ancient times, and obviously this is not needed at all anymore. So instead of removing this for just LDC/Windows it looks like this can be removed completely, even for the oldest supported versions.

ZombineDev:
TBH, I was surprised that you had to manually add curl to the link list. AFAIK, phobos links it internally.

s-ludwig:
I don't remember exactly, but there used to be a time when this was not the case. The latest versions link to it at runtime and the versions before that had the import symbols embedded in phobos.lib. But that was not always the case. There were also some changes for LDC in this area a while ago. But as long as our official support list works without it, it fortunately doesn't matter anymore.

ZombineDev:
Martin changed it with this PR: https://github.com/dlang/phobos/pull/3009/files

PR #1251, opened 2017-09-15
Fix AppVeyor for LDC

wilzbach:
Experiment to use the `curl.lib` shipped with LDC - maybe DUB should always try to use the vendored curl.lib and don't provide its own?

PR #1291, opened 2017-11-30
Fix LDC build on AppVeyor

wilzbach:
Continuing https://github.com/dlang/dub/pull/1251 as #1251 now just disables the LDC jobs.


wilzbach:
image


All green (:

PR #1292, opened 2017-12-01
Optimize dependency resolution and add a hard execution limit

s-ludwig:
This speeds up the dependency resolution process by a factor of around 20x (without exact measurements). On top of that, there is now an execution cap of 1 million iterations (which previously would have easily taken an hour) to avoid CI processed getting stuck and to be able to print a helpful error message pointing to the bug tracker.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -439,20 +439,19 @@ struct Dependency {
 	const {
 		if (this.matchesAny) return o;
 		if (o.matchesAny) return this;
-		if (!this.valid || !o.valid) return invalid;
 		if (m_versA.isBranch != o.m_versA.isBranch) return invalid;
 		if (m_versB.isBranch != o.m_versB.isBranch) return invalid;
 		if (m_versA.isBranch) return m_versA == o.m_versA ? this : invalid;
 		if (this.path != o.path) return invalid;
 
-		Version a = m_versA > o.m_versA ? m_versA : o.m_versA;
-		Version b = m_versB < o.m_versB ? m_versB : o.m_versB;
+		int acmp = m_versA.opCmp(o.m_versA);
+		int bcmp = m_versB.opCmp(o.m_versB);
 
 		Dependency d = this;
-		d.m_inclusiveA = !m_inclusiveA && m_versA >= o.m_versA ? false : o.m_inclusiveA;
-		d.m_versA = a;
-		d.m_inclusiveB = !m_inclusiveB && m_versB <= o.m_versB ? false : o.m_inclusiveB;
-		d.m_versB = b;
+		d.m_inclusiveA = !m_inclusiveA && acmp >= 0 ? false : o.m_inclusiveA;
+		d.m_versA = acmp > 0 ? m_versA : o.m_versA;
+		d.m_inclusiveB = !m_inclusiveB && bcmp <= 0 ? false : o.m_inclusiveB;
+		d.m_versB = bcmp < 0 ? m_versB : o.m_versB;
 		d.m_optional = m_optional && o.m_optional;
 		if (!d.valid) return invalid;
 
@@ -669,7 +668,7 @@ struct Version {
 	this(string vers)
 	{
 		enforce(vers.length > 1, "Version strings must not be empty.");
-		if (vers[0] != branchPrefix && vers != UNKNOWN_VERS)
+		if (vers[0] != branchPrefix && vers.ptr !is UNKNOWN_VERS.ptr)
 			enforce(vers.isValidVersion(), "Invalid SemVer format: " ~ vers);
 		m_version = vers;
 	}
@@ -684,7 +683,7 @@ struct Version {
 	bool opEquals(const Version oth) const { return opCmp(oth) == 0; }
 
 	/// Tests if this represents a branch instead of a version.
-	@property bool isBranch() const { return !m_version.empty && m_version[0] == branchPrefix; }
+	@property bool isBranch() const { return m_version.length > 0 && m_version[0] == branchPrefix; }
 
 	/// Tests if this represents the master branch "~master".
 	@property bool isMaster() const { return m_version == masterString; }
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -369,9 +369,14 @@ struct Dependency {
 	/// ditto
 	hash_t toHash() const nothrow @trusted  {
 		try {
-			auto strhash = &typeid(string).getHash;
-			auto str = this.toString();
-			return strhash(&str);
+			size_t hash = 0;
+			hash = m_inclusiveA.hashOf(hash);
+			hash = m_versA.toString().hashOf(hash);
+			hash = m_inclusiveB.hashOf(hash);
+			hash = m_versB.toString().hashOf(hash);
+			hash = m_optional.hashOf(hash);
+			hash = m_default.hashOf(hash);
+			return hash;
 		} catch (Exception) assert(false);
 	}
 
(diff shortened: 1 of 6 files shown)


source/dub/dependency.d
```diff
@@ -387,11 +392,16 @@ struct Dependency {
...
+	bool matchesAny()
+	const {
+		return m_inclusiveA && m_inclusiveB
+			&& m_versA.toString() == "0.0.0"
+			&& m_versB == Version.maxRelease;
+	}
+
+	unittest {
+		assert(Dependency("*").matchesAny);
+		assert(!Dependency(">0.0.0").matchesAny);
```
Geod24: Could you add `assert(!Dependency(">+0.0.0").matchesAny);` just for clarity ?

source/dub/dependency.d
```diff
@@ -647,10 +657,10 @@ struct Version {
...
 	}
 
-	static @property Version minRelease() { return Version("0.0.0"); }
-	static @property Version maxRelease() { return Version(MAX_VERS); }
-	static @property Version masterBranch() { return Version(masterString); }
-	static @property Version unknown() { return Version(UNKNOWN_VERS); }
+	static @property Version minRelease() { return Version("0.0.0", true); }
+	static @property Version maxRelease() { return Version(MAX_VERS, true); }
+	static @property Version masterBranch() { return Version(masterString, true); }
+	static @property Version unknown() { return Version(UNKNOWN_VERS, true); }
```
Geod24: Is this that much of an improvement ? Maybe you'd like to provide those as `static immutable` instead, to avoid construction an instance of the struct over and over.

source/dub/dependency.d
```diff
@@ -662,19 +672,20 @@ struct Version {
...
 	*/
 	static Version fromString(string vers) { return Version(vers); }
 
-	bool opEquals(const Version oth) const {
-		if (isUnknown || oth.isUnknown) {
-			throw new Exception("Can't compare unknown versions! (this: %s, other: %s)".format(this, oth));
-		}
-		return opCmp(oth) == 0;
-	}
+	bool opEquals(const Version oth) const { return opCmp(oth) == 0; }
```
Geod24: I assume unknown is caught at an upper level now ?

source/dub/semver.d
```diff
@@ -26,7 +26,7 @@ import std.conv;
 	Validates a version string according to the SemVer specification.
 */
 bool isValidVersion(string ver)
-{
+pure @nogc {
```
Geod24: Could you add it to `Version.toString` as well ? And maybe other methods, but this is the one I had to check in the previous commit as many `toString` are allocating.

source/dub/dependency.d
```diff
@@ -667,14 +666,15 @@ struct Version {
 	this(string vers)
 	{
 		enforce(vers.length > 1, "Version strings must not be empty.");
-		if (vers[0] != branchPrefix && vers != UNKNOWN_VERS)
+		if (vers[0] != branchPrefix && vers !is UNKNOWN_VERS)
```
Geod24: Is `UNKNOWN_VERS` internal-only ? If so you could make it `static immutable` and only compare `ptr`

source/dub/dependency.d
```diff
@@ -667,14 +666,15 @@ struct Version {
...
 			enforce(vers.isValidVersion(), "Invalid SemVer format: " ~ vers);
 		m_version = vers;
 	}
 
 	// skips validation for efficiency
 	private this(string vers, bool _trusted)
 	{
+		assert(vers.length > 0);
```
Geod24: Why this `assert` ?

source/dub/dub.d
```diff
@@ -1344,7 +1345,7 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 		}
 		auto basepack = pack.basePackage;
 
-		foreach (d; pack.getAllDependencies()) {
+		pack.getAllDependencies((d) {
```
Geod24: Sounds to me that you could / should return an `InputRange` (or maybe `ForwardRange`) from `getAllDependencies` instead of an array.

source/dub/package_.d
```diff
@@ -526,12 +526,17 @@ class Package {
 	PackageDependency[] getAllDependencies()
 	const {
 		auto ret = appender!(PackageDependency[]);
+		getAllDependencies(d => ret.put(d));
+		return ret.data;
+	}
+	/// dutti
```
Geod24: Typo. Also see my comment about the Range

source/dub/dependencyresolver.d
```diff
@@ -197,8 +197,16 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		}
 
 		string first_error;
+		size_t loop_counter = 0;
 
 		while (true) {
+			assert(loop_counter++ < 1_000_000,
+				"The dependency resolution process is taking too long. The"
+				~ " dependency graph is likely hitting a pthological case in"
```
Geod24: Typo: "pathological".

Also I *feel* there should be a way to bypass this. I would prefer to issue a warning at first instead of crashing the process, but since it is aimed at stuck CI, maybe just add an option to bypass this behavior so someone with such a pathological case still has a working project.

source/dub/dependencyresolver.d
```diff
@@ -41,7 +42,8 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		CONFIG config;
 
 		hash_t toHash() const nothrow @trusted {
-			size_t ret = typeid(string).getHash(&pack);
+			size_t ret = pack.hashOf();
+			//size_t ret = typeid(string).getHash(&pack);
```
Geod24: Dead code

source/dub/dependency.d
s-ludwig: Certainly not the top performance hog, but it was at the top of the profiler with the other things. Static immutable sounds like a good idea, not sure why that hasn't been the case.

source/dub/dependency.d
s-ludwig: Yes, turned out that `opCmp` has the exact same lines at the top of its body.

source/dub/dependencyresolver.d
s-ludwig: The problem is that when reaching this amount of iterations, it is likely that the total number lies in the hundreds of millions or billions and the process will not terminate in any reasonable amount of time. At least that is the experience so far, either it terminates after a few hundred iterations, or "never". The reason for this also lies in the simple lexicographical order in which combinations are tested

source/dub/dub.d
s-ludwig: My main worry is that this would be a breaking API change. A new method with a different name would have to be added for this, or at least some kind of struct with an `alias this`  to an array conversion function. I'll look into the latter option.

source/dub/semver.d
s-ludwig: Yeah, that should definitely be done in more places. I'd do that in a separate pass through the code base, though. I did it here mainly because I was getting bogus stack traces involving the GC in the profiler.

source/dub/dependencyresolver.d
Geod24: I can understand why `assert` by default is prefered, but I don't think we should ever allow a false positive to happen without a workaround.
If there's one single user which have a project which is working currently, that this change will break, it's not acceptable to pull this change, because for this person *you're not improving things*.

My suggestion would be to use an hidden switch here. Now if you feel strongly about it, I won't push it, 1M sound already quite high.
I just saw users' use case being overlooked too many times because "why would you do that?".

s-ludwig:
Okay, everything except a command line switch to disable the assert should be adressed now. Hopefully this didn't introduce new compile errors on older DMD versions. What do you think about the switch (or maybe envvar)?

source/dub/dependencyresolver.d
s-ludwig: Okay, just based on gut feeling I think the likelihood for a false positive is very low here, and even if there is one, nobody would probably rely on it, considering a DUB invocation would take more than an hour. But on the other hand it's not really a big deal to add a way to disable the assert. Instead of a switch I'd suggest checking for a `DUB_NO_RESOLVE_LIMIT` environment variable, because that only has local code impact and may also be easier to incorporate in situations where DUB is invoked many times through a script.

Geod24:
Envvar is even better :+1:

source/dub/dependency.d
```diff
@@ -698,7 +698,7 @@ struct Version {
 	}
 
 	/// Tests if this represents the special unknown version constant.
-	@property bool isUnknown() const { return m_version is UNKNOWN_VERS; }
+	@property bool isUnknown() const { return m_version == UNKNOWN_VERS; }
```
Geod24: You need to make `UNKNOWN_VERS` a `static immutable` otherwise it might fail

s-ludwig: It is.

s-ludwig:
I think this may be the right time to drop 2.066.x support. It's over three years old now and I just had to add multiple workarounds for that version. Now the range based `getAllDependencies` still fails because `AA.byKeyValue` was added in 2.067.0.

s-ludwig:
Rebased and squashed the fixes. Do you see anything left?

PR #1293, opened 2017-12-01
[trivial] Remove compatibility code for DMD frontends < 2.067.x.

s-ludwig:


Geod24:
Maybe time to look into https://github.com/dlang/dub/blob/78b922ef353efde08c0486ef3fc0de06b238d5ca/source/dub/dependency.d#L47 as well ?

s-ludwig:
@Geod24 Good point, there are still quite a few places that require `@trusted`, because vibe.d 0.7.x is still supported (where `Path` and `Json` are involved), but other than that there was no problem. I'll open another PR.

PR #1296, opened 2017-12-02
Make deprecation warnings compile errors

wilzbach:
`-de` is also used at Phobos, Druntime and Tools.
See also: https://github.com/dlang/phobos/pull/5546
The idea is that whenever sth. gets deprecated in Druntime or Phobos a replacement needs to exists. By always building with `-de`, we can enforce that the replacement really exists and force the person who submits the deprecation to look after the fallout.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -52,7 +52,8 @@ install:
                 $latestVersion = $latest.split("-")[0].split("~")[0];
                 $url = "http://downloads.dlang.org/pre-releases/2.x/$($latestVersion)/dmd.$($latest).windows.7z";
             }elseif($version -eq "nightly") {
-                $url = "http://nightlies.dlang.org/dmd-master-2017-05-20/dmd.master.windows.7z"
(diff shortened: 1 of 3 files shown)


wilzbach:
Needs https://github.com/dlang/dub/pull/1295

joakim-noah:
I like the idea, but looks like you have more work to do to get some D versions working.  [I build the native Android dub with this flag](https://github.com/termux/termux-packages/blob/master/packages/ldc/build.sh#L141), but I only have to worry about the latest ldc. 😉

Geod24:
I'm going to go ahead and close this. While it is a nice to have thing, I don't thing it's something that is easily achievable, just for the fact that DUB depends on Phobos and Vibe.d (and its dependencies) and support a wide range of libraries.

We have evidence that there are unsolved problems with deprecations, for example deprecating `alias this` triggers deprecations in generic code (including places that checks for range types), deprecating things like `toString()` leads to deprecation warning in `std.format`...

So `-de` is great when you control the full stack and have a limited compiler support (druntime / phobos only support HEAD), but when you throw Vibe.d into the mix, and multiple compilers, it starts to require much more manpower than we have. Add to that the fact that `dub` is currently mostly used as a binary, where deprecations don't matter as much as with a lib.

PR #1298, opened 2017-12-04
Start testing nonet DUB build

wilzbach:
It might also be interesting to test the build using the new `vibe-core` package.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -13,6 +13,13 @@ configuration "application" {
 	versions "DubUseCurl"
 }
 
+configuration "nonet" {
+	targetType "executable"
+	mainSourceFile "source/app.d"
+	dependency "vibe-d:http" version=">=0.7.30 <=0.9.0"
+	versions "VibeUseOpenSSL11"
+}
+
 configuration "library" {
 	targetType "library"
 	excludedSourceFiles "source/app.d"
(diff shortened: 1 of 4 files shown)


dub.sdl
```diff
@@ -13,6 +13,22 @@ configuration "application" {
...
 }
 
+configuration "nonet" {
+	targetType "executable"
+	mainSourceFile "source/app.d"
+	dependency "vibe-d:http" version=">=0.7.30 <=0.9.0"
+	versions "VibeUseOpenSSL11"
+}
+
+configuration "nonet-test" {
```
s-ludwig: `dub test -c nonet` should do the same here, have you tried that?

dub.sdl
```diff
@@ -13,6 +13,22 @@ configuration "application" {
 	versions "DubUseCurl"
 }
 
+configuration "nonet" {
+	targetType "executable"
+	mainSourceFile "source/app.d"
+	dependency "vibe-d:http" version=">=0.7.30 <=0.9.0"
+	versions "VibeUseOpenSSL11"
```
s-ludwig: Since this is target system specific, we shouldn't include this by default. I think this is also the reason for the Travis failure.

Geod24:
This is being tested nowadays.

PR #1306, opened 2017-12-10
Git version resolution for path-based dependecy

llucenic:
It is fairly legal in case of a git submodule that _.git_ is a regular file. Git accepts this file in its _--git-dir_ parameter in `git describe`. Thus rendering the `isDir()` check in `determineVersionWithGIT(NativePath path)` being counter-effective.

### DIFF
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -756,7 +756,7 @@ private string determineVersionWithGIT(NativePath path)
 	import dub.semver;
 
 	auto git_dir = path ~ ".git";
-	if (!existsFile(git_dir) || !isDir(git_dir.toNativeString)) return null;
+	if (!existsFile(git_dir)) return null;
 	auto git_dir_param = "--git-dir=" ~ git_dir.toNativeString();
 
 	static string exec(scope string[] params...) {


llucenic:
Please help me with the interpreting of failing checks. I do not understand what should be changed in order to pass them.

wilzbach:
Don't worry about Jenkins it used to be very unstable in the last few months thought this got better over the last weeks as we started to look into the failures more carefully.
AppVeyor is just a download failure.

Anyhow, I'm sorry that this got lost in the queue.
I see your point for removing the check, but it would be much stronger if your PR would be accompanied by a unittest for it. This would greatly help to prevent regressions in this matter ;-)

llucenic:
I will write a unit test, but frankly I have no idea how it should look like in a change like this :-)

wilzbach:
> but frankly I have no idea how it should look like in a change like this :-)

Have a look at e.g. https://github.com/dlang/dub/blob/master/test/4-describe-data-1-list.sh

Essentially you create a the layout that's needed to reproduce this in `test` (like e.g. https://github.com/dlang/dub/blob/master/test/describe-dependency-2/dub.json) and then add a test as bash script.
I think it's best to do this from `master`, s.t. your bash script fails with `master` and once it does you switch to your branch and it should pass.
For git submodules, I think the easiest is to create them in a temporary directory that's nuked with a bash trap (I'm not sure whether all CI scripts do a `--recursive` clone atm).
Otherwise there's also e.g. https://github.com/dlang-community/gitcompatibledubpackage

PR #1308, opened 2017-12-15
Fix sublime text generator file regex

tofuninja:
Errors were not displaying properly in sublime text before because of the extra empty group capture. 
That group was being interpreted as the error text when the one right after is actually the error text.

### DIFF
--- a/source/dub/generators/sublimetext.d
+++ b/source/dub/generators/sublimetext.d
@@ -85,7 +85,7 @@ private Json buildSystems(BuildPlatform buildPlatform, string workingDiretory = 
 	string fileRegex;
 
 	if (buildPlatform.frontendVersion >= 2066 && buildPlatform.compiler == "dmd")
-		fileRegex = r"^(.+)\(([0-9]+)\,([0-9]+)\)\:() (.*)$";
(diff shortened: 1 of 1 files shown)


s-ludwig:
Could it be that the extra group was added/removed at some point in ST's history? Did you test this on the latest 3 release? Maybe the original code was still for ST 2 (which I don't think makes sense to support anyway). I have to admit that altough I use ST on a daily basis, I never used the project generator so far, because that would require way too much project switching.

tofuninja:
If you look there are two versions of the regex, and the one that was wrong has a check for dmd 2066. The other one has the extra grouping as well but does not seem to take into account columns. In 2.066 -vcolumns was added. I suspect the other one had the extra grouping because dmd was not outputting columns and it was simply there to satisfy sublime text requesting 4 groupings: file, line, column, error. Then someone came in to fix it after 2.066 but left the extra grouping in getting errors to go to the right file and line and column but the error text was being omitted because of the extra grouping.

tofuninja:
Also yes, this was tested in the latest version of sublime.

s-ludwig:
Okay, placeholder for the column makes sense ideed, thanks for the clarification! Looks good to merge then.

PR #1309, opened 2017-12-17
add donationUrl & donationDetail to dub.json/dub.sdl

WebFreak001:
required for dlang/dub-registry#265

### DIFF
--- a/source/dub/recipe/json.d
+++ b/source/dub/recipe/json.d
@@ -33,6 +33,8 @@ void parseJson(ref PackageRecipe recipe, Json json, string parent_name)
 			case "authors": recipe.authors = deserializeJson!(string[])(value); break;
 			case "copyright": recipe.copyright = value.get!string; break;
 			case "license": recipe.license = value.get!string; break;
+			case "donationUrl": recipe.donationUrl = value.get!string; break;
+			case "donationDetail": recipe.donationDetail = value.get!string; break;
 			case "configurations": break; // handled below, after the global settings have been parsed
 			case "buildTypes":
 				foreach (string name, settings; value) {
@@ -80,6 +82,8 @@ Json toJson(in ref PackageRecipe recipe)
 	if (!recipe.authors.empty) ret["authors"] = serializeToJson(recipe.authors);
 	if (!recipe.copyright.empty) ret["copyright"] = recipe.copyright;
 	if (!recipe.license.empty) ret["license"] = recipe.license;
+	if (!recipe.donationUrl.empty) ret["donationUrl"] = recipe.donationUrl;
+	if (!recipe.donationDetail.empty) ret["donationDetail"] = recipe.donationDetail;
 	if (!recipe.subPackages.empty) {
 		Json[] jsonSubPackages = new Json[recipe.subPackages.length];
 		foreach (i, subPackage; recipe.subPackages) {
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -79,6 +79,8 @@ struct PackageRecipe {
 	string[] authors;
 	string copyright;
 	string license;
+	string donationUrl;
+	string donationDetail;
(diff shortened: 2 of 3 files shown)


WebFreak001:
making the fields auto deserialize/serialize using traits would be a good step forward

s-ludwig:
I'm not sure if the package recipe is the right place for information like this. There have been lot's of (legitimate) complaints already that dependency information and build instructions are in the same file (which is actually important, because build configurations need to be able to control dependencies in order to be a useful tool).

But if the conclusion is be to really put it in the package recipe, it should at least live in a different namespace than the core directives (e.g. "x:donationURL" or "dub-registry:donationURL") and there should be a generic mechanism to access them instead of putting more specific fields into `PackageRecipe` that are not concerned with the actual operation of DUB.

A consideration should also be whether donation information should be inherited by forks. Arguments can be made in both directions. If the answer is "no" or "maybe", it may make more sense to just make this a setting in dub-registry itself.

s-ludwig:
Regarding redundancy of the two parsers, it's true that it would be nice to use generic serialization at some point. But since very few fields are added these days, it wouldn't gain much with the risk of introducing bugs, so I wouldn't necessarily make it a priority at this point.

WebFreak001:
I first thought of putting this on the dub registry but then you would need to add some kind of github verification because you don't want other people to register your package and get donations for it

s-ludwig:
Hm, that's a good point. Bummer that we don't have GitHub OAuth support yet.

But of course the problem goes beyond donations and beyond GitHub authentication. Even the fact that some foreigner has exclusive access to a package can be problematic, and even when tied to the repository it's easy to clone and register the clone with modified credentials, unless the original is already registered.

WebFreak001:
ok so how about I add back the donationURL/detail to the website in the PR and in a separate PR github oauth

s-ludwig:
That sounds like a good way forward, the possible transition to a repo based solution is easier from that direction and it would also give us a chance to think about custom fields in general (e.g. [DEP3](https://github.com/dlang/dub/wiki/DEP3)) before deciding on an actual implementation.

WebFreak001:
so would you rather completely separate it out of the repository (as well as other fields like description then maybe) or would you put it all under namespaces in the dub.json/sdl?

s-ludwig:
Description is such a basic piece of information that it makes sense to include it. License compatibility is supposed to be santity checked at some point, so that is also required as a field. "authors" is a field that could be debatable, but it's an integral part by now. Apart from the `x:ddox*` fields I don't think there are currently any fields that are candidates for removal.

Generally using a namespace within the package recipe would be fine from my side, we just need to define how such custom fields are handled within the DUB code base. The registry could of course also just custom parse them, but that would still risk that the field definitions could be incompatible to a future DUB solution to custom fields (e.g. the semantics could cause DUB to error out because of an undefined field).

So the quick way to move forward without contention would be to add it as a setting to the registry now and to look into the recipe issue afterwards.

wilzbach:
> Generally using a namespace within the package recipe would be fine from my side, 

A link to the documentation like requested in https://github.com/dlang/dub-registry/issues/287 seems to be another good use case for this.

How about naming the namespace "registry" namespace as all these keys will be used by the registry?
Maybe thinking about the future another "other" namespace too?

> we just need to define how such custom fields are handled within the DUB code base.

Not looked at (except for the registry), but available as Key/Value dictionary?

ohdatboi:
Even though programmers deserve to be awarded for their work, it has nothing to do with dub as a build system. dub.json should contain bare minimum of needed information (Authors, copyright, licence, etc) and everything else should be handled by the registry.

wilzbach:
We recently moved to having a field for the documentationURL at the registry (https://github.com/dlang/dub-registry/pull/332), so I think the consensus moved to just storing such additional information in the at the registry and I guess this isn't moving forward.

@WebFreak001 maybe do sth. like https://github.com/dlang/dub-registry/pull/332 for these two fields?

PR #1314, opened 2017-12-20
refactor package configuration

MartinNowak:


MartinNowak:
If sourceLibraries weren't in the `targets` hash, I couldn't easily configure their build settings, when their occurs as a dependency.
Previously this was simpler with the single tree traversal.
Not sure how a diamond dependency on a source library is supposed to be handled anyhow, guess build settings should be handled like for normal library dependencies though.

MartinNowak:
Superseded by https://github.com/dlang/dub/pull/1263

PR #1316, opened 2017-12-25
Avoid creation of a temporary probe file for newer compilers

wilzbach:
When I chatred with @RazvanN7 today, he complained about DUB creating a lot of temporary probing files in his current working directory.
A quick look at the code revealed that DUB uses this probing hack to extract information
from DMD. Since https://github.com/dlang/dmd/pull/6880 (2.076.0), it's possible
to avoid the need of creating a temporary file for this hack.
Of course, it would be better to read this file from the compiler directly...

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -115,19 +115,53 @@ interface Compiler {
 		enforce(status == 0, format("%s failed with exit code %s.", args[0], status));
 	}
 
+	/** Detect the frontend version of a compiler_binary
+	    Returns: 0 if the detection failed, otherwise the frontend version.
+		Note: This only works for DMD and LDC.
+	 */
+	private int frontendVersion(string compiler_binary)
+	{
+		import std.range : dropOne;
+
+		auto res = execute([compiler_binary, "--version"]);
+		if (res.status == 0) {
+			auto versionRange = res.output.find("DMD64 D Compiler v").splitter(".");
+			if (!versionRange.empty && !versionRange.dropOne.empty) {
+				return versionRange.dropOne.front.to!int;
+			}
+		}
+		return 0;
+	}
+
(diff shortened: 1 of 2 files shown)


source/dub/compilers/compiler.d
```diff
@@ -115,19 +115,53 @@ interface Compiler {
...
 	/// Compiles platform probe file with the specified compiler and parses its output.
 	protected final BuildPlatform probePlatform(string compiler_binary, string[] args, string arch_override)
 	{
 		import std.string : format;
-		import dub.compilers.utils : generatePlatformProbeFile, readPlatformProbe;
+		import dub.compilers.utils : generatePlatformProbeFile, readPlatformProbe, platformProbeFile;
 
-		auto fil = generatePlatformProbeFile();
+		bool canUseStdin = this.canUseStdin(compiler_binary);
+		string fil = canUseStdin ? generatePlatformProbeFile().toNativeString() : "-";
```
wilzbach: Of course we could also remove the temp file after the exit of the current scope.

MartinNowak:
Temp files get deleted at the end of the run, would be better to make that mechanism more reliable instead of increasing the complexity for that single task.
https://github.com/dlang/dub/blob/79475c088cf90b02611818863bc835f499b9145c/source/dub/internal/utils.d#L40-L47
I'd guess the main source for the remaining temp files is dub's missing `CTRL+C` sigint handling, in particular in combination with `dub run` where you want to exit the child process, but terminate the parent instead.

wilzbach:
FYI: the new JSON API (e.g. `dmd -Xi=compilerInfo -Xf=-` has been released with 2.079 and we should be able to use it for new LDC + DMD compilers.

See also: https://dlang.org/changelog/2.079.0.html#json_includes

PR #1318, opened 2017-12-26
speed up platform probe 50ms -> 5ms by avoiding phobos imports

MartinNowak:


MartinNowak:
Should this go into stable as it fixes a performance regression?

source/dub/compilers/utils.d
```diff
@@ -248,39 +248,40 @@ NativePath generatePlatformProbeFile()
...
-		`	return '"' ~ ret.data.join("\", \"") ~ "\", "; }` ~ q{
-
-		string determineCompiler() } ~ '{' ~ compilerCheck ~ "	}");
+		string[] determinePlatform() } ~ '{' ~ platformCheck ~ '}' ~ q{
+		string[] determineArchitecture() } ~ '{' ~ archCheck ~ '}' ~ q{
+		string determineCompiler() } ~ '{' ~ compilerCheck ~ '}';
 
+	auto path = getTempFile("dub_platform_probe", ".d");
+	auto fil = openFile(path, FileMode.createTrunc);
+	scope (failure) {
```
wilzbach: Why not `scope(exit)` and remove `fil.close()` in L285?

wilzbach:
> Should this go into stable as it fixes a performance regression introduced by 61cf37f?

Sounds very reasonable.

source/dub/compilers/utils.d
s-ludwig: `File` should be ref counted anyway, so neither should be neccessary.

s-ludwig:
From my side too this can go into stable, although labeling this as a "regression" is maybe a bit too much when talking about 45ms ;)

PR #1324, opened 2018-01-03
split dub search test

MartinNowak:
- to better localize errors

Recent failure https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7576/5/pipeline#step-338-log-312

### DIFF
--- a/test/feat663-search.sh
+++ b/test/feat663-search.sh
@@ -7,6 +7,9 @@ fi
 if ${DUB} search nonexistent123456789package 2>/dev/null; then
     die $LINENO '`dub search nonexistent123456789package` succeeded'
 fi
-if ! OUTPUT=$(${DUB} search dub -v 2>&1) || ! { echo "$OUTPUT" | grep -q '^dub (.*)\s'; } then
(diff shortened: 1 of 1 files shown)


test/feat663-search.sh
```diff
@@ -7,6 +7,9 @@ fi
 if ${DUB} search nonexistent123456789package 2>/dev/null; then
     die $LINENO '`dub search nonexistent123456789package` succeeded'
 fi
-if ! OUTPUT=$(${DUB} search dub -v 2>&1) || ! { echo "$OUTPUT" | grep -q '^dub (.*)\s'; } then
+if ! OUTPUT=$(${DUB} search dub -v 2>&1); then
     die $LINENO '`dub search dub` failed' "$OUTPUT"
 fi
+if ! grep -q '^dub (.*)\s'<<<"$OUTPUT"; then
```
wilzbach: Here strings are Bash-specific, but I assume there's no one running the testsuite with something other than bash?

s-ludwig: Shouldn't be the case. There are certainly some other places as well where bash-specifc syntax is used.

PR #1326, opened 2018-01-08
Bump vibe.d version used for 'dub init'

wilzbach:
From: https://github.com/vibe-d/vibe.d/issues/2024

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -111,7 +111,7 @@ void main()
 private void initVibeDPackage(NativePath root_path, ref PackageRecipe p, scope void delegate() pre_write_callback)
 {
 	if ("vibe-d" !in p.buildSettings.dependencies)
-		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.7.30");
+		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.8.2");
 	p.description = "A simple vibe.d server application.";
 	pre_write_callback();
 


wilzbach:
@s-ludwig any chance you can look into enabling an [AppVeyor team](https://ci.appveyor.com/team). It would allow folks like myself to restart failing builds ;-)
For DMD: I did the following:
- added a role "DLang Team" 
- added all teams from GitHub to this role

I used these permissions for the DLang Team role, though just "run project builds" would be enough:

image

mathias-lang-sociomantic:
IMO versions shouldn't be hardcoded. Don't we have a mean to get the latest version ?

s-ludwig:
@wilzbach I've added a bunch of teams with project settings and build execution access now. I left out administrative rights, if you need them, let me know and I'll add you as a single user instead.

wilzbach:
> I've added a bunch of teams with project settings and build execution access now.

Thanks a lot. Works! Could you do the same for the `vibe-d` organisation?

> IMO versions shouldn't be hardcoded. Don't we have a mean to get the latest version ?

In theory there could be a version 1.0 release of vibe.d which breaks the example template provided.

s-ludwig:
@mathias-lang-sociomantic That could be implemented, but the whole `dub init -t` feature is planned to be replaced by a more general approach, so it would be rather wasteful to invest more time than necessary in this. The new approach basically would simply look for a ":example" (or similar) sub package for the first dependency specified to `dub init`, uses that as the base for the newly created package, and would then just adjust the recipe according to the remaining init parameters.

s-ludwig:
Shouldn't the team change apply to vibe-d organization projects as well? I've also invited you to the "VibeD-Devs" team now (you were just member of the vibe-core repository specifically).

PR #1332, opened 2018-01-12
Change LDC Windows binary URLs

wilzbach:
See also: https://github.com/ldc-developers/ldc/releases

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,7 +29,7 @@ environment:
     DVersion: stable
     arch: x64
   - DC: ldc
-    DVersion: 1.2.0
+    DVersion: 1.7.0
     arch: x64
 
 skip_tags: false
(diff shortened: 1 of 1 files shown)


wilzbach:
(removed auto-merge as (1) I did this change blindly and rely on AppVeyor for feedback and (2) AppVeyor isn't enforced to pass)

s-ludwig:
> AppVeyor isn't enforced to pas

We should deinitely change that! It didn't pass until recently(ish), but since it does now, there is no  reason to not enforce it.

wilzbach:
> We should deinitely change that! It didn't pass until recently(ish), but since it does now, there is no reason to not enforce it.

Okay - just did so.
BTW I'm not sure whether you have heard about our switch from Travis to Semaphore at dlang/dmd
tl;dr: a quote from Ian at GDC:

> I would say more. On a good day, gdc is built with testsuite and unittests ran in 14 minutes flat on semaphore.
> Travis barely could do the same in 50 minutes.

I have observed similar good results with DMD (not as good because the parallelism level of the DMD testsuite isn't perfect).
Anyhow seeing that Travis is pretty slow here, it might be worth thinking about doing a similar switch as well - or doing a 50:50 transition by using the same build script.

Disadvantages:
- only two concurrent builds (but with 8(!) cores)
- no synchronization with GitHub teams - members need to be added manually

PR #1333, opened 2018-01-14
Add support for DragonFlyBSD

dkgroot:
Notes:
- Switch from /bin/sh to /usr/bin/env bash
- Ignore temporary and test-build files (gitignore)
- Untrack changes to source/dub/version_.d
- Handling 'isDragonflybsd'
- Use 'tr -d' instead of 'tr --deleted' which works on Linux and BSD based platforms
- Fix FRONTEND = unbound variable (test/run-unittest.sh)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.o
 *.obj
+*~
 
 .dub
 .directory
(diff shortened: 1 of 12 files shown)


s-ludwig:
Thanks! Looks good to merge.

dkgroot:
@s-ludwig Thanks for the review and subsequent merge !

PR #1338, opened 2018-01-15
download timeouts

MartinNowak:
- 8s by default
- only for curl atm.

### DIFF
--- a/test/timeout.sh
+++ b/test/timeout.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+
+PORT=$(($$ + 1024)) # PID + 1024
+
+log '    Testing unconnectable registry'
+if timeout 1s $DUB fetch dub --skip-registry=all --registry=http://localhost:$PORT; then
+    die 'Fetching from unconnectable registry should fail.'
+elif [ $? -eq 124 ]; then
+    die 'Fetching from unconnectable registry should fail immediately.'
+fi
+
+log '    Testing non-responding registry'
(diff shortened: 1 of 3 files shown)


test/timeout.sh
```diff
@@ -0,0 +1,43 @@
...
+Server: dummy\r
+Content-Type: application/json\r
+Content-Length: 2\r
+\r
+{}')
+    for i in $(seq 0 $((${#res} - 1))); do
+        echo -n "${res:$i:1}"
+        sleep 1
+    done
+} | nc --listen $PORT >/dev/null &
```
MartinNowak: terminal

test/timeout.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+
+PORT=$(($$ + 1024)) # PID + 1024
```
wilzbach: Good, old std.socket can get you a free port too:

```
rdmd --eval='import std.socket; auto s = new Socket(AddressFamily.INET, SocketType.STREAM); scope(exit) s.close; s.bind(new InternetAddress(0));s.localAddress.toPortString.writeln;'
```

MartinNowak:
> I like the tests!

Would be nicer if it were unittests in the ms range, but that was too much work.

test/timeout.sh
MartinNowak: It's fine there won't be colliding PIDs.

MartinNowak:
Merging this as it's crucial for the mirror mechanism to actually work. We had to enjoy the recent code.dlang.org downtime party to notice ;). It'll be out with 2.078.1 soon, so we win some time to address stability.

PR #1339, opened 2018-01-15
more reliable retries and fallback mirror usage - #1336

### LINKED ISSUE #1336: Try mirrors/retry loop even on an Exception
From http://dtest.dlang.io/results/68037a013ee8965dbd7f4893a1588f57668fcade/24eb29bf325e539a622923bb537cd2fbfccbf07e/build.log

```
.generated/stable_dmd-2.077.1/dmd2/linux/bin64/dub build --single --compiler=.generated/stable_dmd-2.077.1/dmd2/linux/bin64/dmd assert_writeln_magic.d
Fetching libdparse 0.7.2-alpha.4 (getting selected version)...
ZipException: no end record
posix.mak:844: recipe for target '.generated/assert_writeln_magic' failed
```

I'm aware of https://github.com/dlang/dub/issues/623, but according to this the Phobos issue has been fixed.
Such exceptions should be caught in `fetch`, s.t. the retry logic can try it again.

Appearing [here](http://dtest.dlang.io/results/c437b30cc5b04ae18c798bc590ef0849468e8f97/d464620060bf11a33fec4b5ca6bcf97dcf739008/build.log) too:

```
Upgrading project in /dev/shm/dtest/work/repo/dlang.org/dpl-docs
Fetching ddox 0.16.7 (getting selected version)...
Fetching libevent 2.0.2+2.0.16 (getting selected version)...
Fetching diet-ng 1.4.3 (getting selected version)...
Fetching taggedalgebraic 0.10.7 (getting selected version)...
Fetching openssl 1.1.5+1.0.1g (getting selected version)...
Fetching botan 1.12.9 (getting selected version)...
Fetching libdparse 0.7.1 (getting selected version)...
Fetching memutils 0.4.9 (getting selected version)...
Fetching vibe-d 0.8.1 (getting selected version)...
Fetching vibe-core 1.2.0 (getting selected version)...
Fetching hyphenate 1.1.1 (getting selected version)...
Fetching libasync 0.8.3 (getting selected version)...
Fetching botan-math 1.0.3 (getting selected version)...
ZipException: no end record
posix.mak:781: recipe for target 'dpl-docs' failed
make: *** [dpl-docs] Error 2
```

More:
- http://dtest.dlang.io/results/a9e260f2b1577962c75d342ca416fc3d5006d377/f96faa6642be21a5b282b9193328b753eb694973/build.log

wilzbach:
See https://github.com/dlang/dub/issues/1336
I'm wondering why https://github.com/dlang/dub/commit/435fd1cd3dfc3db94434e6ea67733d136fe488fc#diff-ce350ca917a3ebde71db3a122988b8ba didn't help (it's part of 2.077).


wilzbach:
On Travis two jobs even failed with this:

```
[INFO] Building /home/travis/build/dlang/dub/test/issue502-root-import/...
Fetching gitcompatibledubpackage 1.0.4 (getting selected version)...
ZipException: no end record
[ERROR] Build failure.
```

https://travis-ci.org/dlang/dub/jobs/328895076

So apparently this doesn't fix the issue :/

source/dub/dub.d
```diff
@@ -757,12 +757,24 @@ class Dub {
...
+			import std.zip : ZipException;
+			try {
+				auto path = getTempFile(packageId, ".zip");
+				supplier.fetchPackage(path, packageId, dep, (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
+				scope(exit) std.file.remove(path.toNativeString());
+
+				logDiagnostic("Placing to %s...", placement.toNativeString());
+				return m_packageManager.storeFetchedPackage(path, pinfo, dstpath);
+			} catch (ZipException e) {
+				logInfo("Failed to extract zip archive for %s %s...", packageId, ver);
```
wilzbach: This isn't the best behavior as the PackageManager fallbacks will never be tried if the main registry returns invalid zips. However, this would require refactoring their interface a bit and I wasn't sure whether this is wanted. In any case, this should already be an measureable improvement (see the test) as it will restart the download when the registry / GH returned invalid data.

wilzbach:
This is finally passing on the CIs. It was a bit tricky to add a test for `ZipException`, but CodeCov shows that mocking the registry works.

MartinNowak:
This seems to work very much on the symptom side of things. Why do we end up silently with corrupted zip packages in the first place? Maybe sth. wrong with `std.net.curl.download` that doesn't throw an error?

wilzbach:
> This seems to work very much on the symptom side of things.

Well, the errors are pretty random and spurious, so it looks like the network stream gets interrupted.
Also I don't know how to reproduce them locally as I only see them from time to time on a CI.
However, the point is that random CI failures are annoying and a simple retry-loop should fix most of the problems for the next release and gives us time to dig into this.

> Why do we end up silently with corrupted zip packages in the first place?

If I would know, I wouldn't have proposed this workaround for now.
Maybe instead of digging into this and investing time into the old `std.net.curl`, we should simply start deploying/using the `nonet` (aka vibe.d) build?

MartinNowak:
I'd suggest to checksum the packages, but GH apparently rebuild those archives from time to time.
It'd also be possible that we're looking at corruptions on GH's side.

If it's simply an interrupted transfer, then it should be feasible to reproduce this in a test case, e.g. through a proxy.

> Maybe instead of digging into this and investing time into the old std.net.curl, we should simply start deploying/using the nonet (aka vibe.d) build?

Not until vibe-core is stable enough to replace the libevent dependency. I've checked std.net.curl.download and couldn't see anything wrong, it's just a thin wrapper around curl_easy_perform.
https://github.com/dlang/phobos/blob/fd3ae774b802a43f2f700b708f53351d07318621/std/net/curl.d#L406-L408
Also vibe's inet.urltransfer.download is fairly simply itself and doesn't handle certain issues (e.g. timeouts), and std.net.curl's bad reputation is mostly based on resolved problems. But as with anything curl (as opposed to wget), you need to explicitly configure lots of parameters.

MartinNowak:
> I'm wondering why 435fd1c#diff-ce350ca917a3ebde71db3a122988b8ba didn't help (it's part of 2.077).

Because that retries downloads, i.e. `supplier.fetchPackage` but not the unpacking that follows that.

source/dub/dub.d
MartinNowak: Packages are downloaded directly from GH though, at least that should be the case.
https://github.com/dlang/dub-registry/blob/fe18df23b32f602a7579816a801a8a4e98346b5c/source/dubregistry/web.d#L177-L187

wilzbach:
> I'd suggest to checksum the packages, but GH apparently rebuild those archives from time to time.
It'd also be possible that we're looking at corruptions on GH's side.

> If it's simply an interrupted transfer, then it should be feasible to reproduce this in a test case, e.g. through a proxy.

I already added a simple test registry based on Vibe.d in this PR that either serves a valid or invalid zip. We can probably also test interrupted downloads with this, but I'm not sure what's the best way to simulate this. Should we implement our own file serving (for a specific version) and sleep for `50ms` after every n bytes? Or is there more to it?

MartinNowak:
BTW, we should finish support for `file://` package suppliers (https://github.com/dlang/dub/blob/785a0b2fb7fee298bf42b9191c59ec80776d59ce/source/dub/packagesupplier.d#L89).

MartinNowak:
Here we go,
```d
router.get("/packages/gitcompatibledubpackage/1.0.2.zip", (req, res) @trusted {
    res.writeBody("", HTTPStatus.internalServerError);
});
```
```sh
$ curl localhost:1234/packages/gitcompatibledubpackage/1.0.2.zip                                                                                                                       
$ curl -fsSL localhost:1234/packages/gitcompatibledubpackage/1.0.2.zip                                                                                                                
curl: (22) The requested URL returned error: 500 Internal Server Error
```
OK, curl only fails with `-f`.

```sh
$ ./bin/dub fetch gitcompatibledubpackage --version=1.0.2 --skip-registry=all --registry=http://localhost:1234/ --vverbose
Refreshing local packages (refresh existing: true)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/dawg/.dub/packages/local-packages.json
Try to load local package map at /home/dawg/.dub/packages/local-packages.json
iterating dir /home/dawg/.dub/packages/
...
Determined package version using GIT: dub 1.7.1-beta.1+commit.34.gcdf18e3
Downloading metadata for gitcompatibledubpackage
Getting from http://localhost:1234/packages/gitcompatibledubpackage.json
Getting http://localhost:1234/packages/gitcompatibledubpackage.json...
Fetching gitcompatibledubpackage 1.0.2...
Acquiring package zip file
Downloading from 'http://localhost:1234/packages/gitcompatibledubpackage/1.0.2.zip'
Storing http://localhost:1234/packages/gitcompatibledubpackage/1.0.2.zip...
Placing to /home/dawg/.dub/packages/...
Placing package 'gitcompatibledubpackage' version '1.0.2' to location '/home/dawg/.dub/packages/gitcompatibledubpackage-1.0.2/gitcompatibledubpackage' from file '/tmp/gitcompatibledubpackage-0c8f66bd-cd98-4a84-819c-f9f06cc8180b.zip'
Opening file /tmp/gitcompatibledubpackage-0c8f66bd-cd98-4a84-819c-f9f06cc8180b.zip
rawRead must take a non-empty buffer
Full exception: object.Exception@/usr/include/dmd/phobos/std/stdio.d(972): rawRead must take a non-empty buffer
----------------
/usr/include/dmd/phobos/std/stdio.d:972 @safe ubyte[] std.stdio.File.rawRead!(ubyte).rawRead(ubyte[]) [0x8851dc]
source/dub/internal/vibecompat/core/file.d:46 @trusted void dub.internal.vibecompat.core.file.RangeFile.rawRead(ubyte[]) [0x86fa7c]
source/dub/internal/vibecompat/core/file.d:42 @safe ubyte[] dub.internal.vibecompat.core.file.RangeFile.readAll() [0x86fa32]
source/dub/packagemanager.d:360 dub.package_.Package dub.packagemanager.PackageManager.storeFetchedPackage(dub.internal.vibecompat.inet.path.NativePath, dub.internal.vibecompat.data.json.Json, dub.internal.vibecompat.inet.path.NativePath) [0x7513ff]
```

This should have failed in std.net.curl.download, but instead fails during unpacking. Almost as if we're missing a `-f` option for libcurl.

MartinNowak:
Think I found the culprits (yes multiple ones as always).

[18318 – std.net.curl.download silently ignores non-2xx http statuses](https://issues.dlang.org/show_bug.cgi?id=18318)

Also since #1104 we're only catching HTTPStatusErrors, but we also need to catch CurlErrors for things like timeouts, ssl, etc. pp.

MartinNowak:
Ready from my side @wilzbach. This fixes the HTTP status problem, but still keeps the zip exception for the unlikely case of corrupted downloads.
We should replace the FallbackRegistry with simply iterating over all registries at some point.
That require quite some refactoring of `Dub.fetch` though.

wilzbach:
> Ready from my side @wilzbach. 

Awesome! Travis failed because you forgot to commit the 1.0.2 symlink - I added that one.

> We should replace the FallbackRegistry with simply iterating over all registries at some point.

Agreed, there are lots of other things we could do as well:
- cache the zip binary on the server / mirrors (at the moment all the registry does is a redirect to GH)
- finally implement the git procotol (might be more stable & flexible than the zip download)
- user the slightly smaller `.tar.gz`
...

Though of all these will require quite some effort.

test/fetchzip.sh
```diff
@@ -0,0 +1,59 @@
...
+    die 'DUB should have tried to download the zip archive multiple times.'
+elif [ $rc -eq 124 ]; then
+    die 'DUB timed out unexpectedly.'
+fi
+if dub remove gitcompatibledubpackage --non-interactive --version=* 2>/dev/null; then
+    die 'DUB should not have installed a broken package.'
+fi
+
+echo "HTTP status errors on downloads should be retried - gitcompatibledubpackage (1.0.2)"
+retryOut=$(! timeout 1s "$DUB" fetch gitcompatibledubpackage --version=1.0.2 --skip-registry=all --registry=http://localhost:$PORT --vverbose 2>&1)
```
wilzbach: The Travis log shows an error and `retryCount` of `1` -  despite the log clearly showing "Bad Gateway" three times:

```
Downloading metadata for gitcompatibledubpackage
Getting from http://localhost:4934/packages/gitcompatibledubpackage.json
Getting http://localhost:4934/packages/gitcompatibledubpackage.json...
Fetching gitcompatibledubpackage 1.0.2...
Acquiring package zip file
Downloading from 'http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip'
Storing http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip...
Download http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip 502 Bad Gateway (1.1)
Failed to download package gitcompatibledubpackage from http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip (Attempt 1 of 3)
Storing http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip...
Download http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip 502 Bad Gateway (1.1)
Failed to download package gitcompatibledubpackage from http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip (Attempt 2 of 3)
Storing http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip...
Download http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip 502 Bad Gateway (1.1)
Failed to download package gitcompatibledubpackage from http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip (Attempt 3 of 3)
Failed to download package gitcompatibledubpackage from http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip
Full exception: object.Exception@source/dub/packagesupplier.d(208): Failed to download package gitcompatibledubpackage from http://localhost:4934/packages/gitcompatibledubpackage/1.0.2.zip
----------------
source/dub/packagesupplier.d:208 void dub.packagesupplier.RegistryPackageSupplier.fetchPackage(dub.internal.vibecompat.inet.path.NativePath, immutable(char)[], dub.dependency.Dependency, bool) [0x762422]
source/dub/dub.d:766 dub.package_.Package dub.dub.Dub.fetch(immutable(char)[], const(dub.dependency.Dependency), dub.project.PlacementLocation, dub.dub.FetchOptions, immutable(char)[]) [0x74bafa]
source/dub/commandline.d:1235 int dub.commandline.FetchCommand.execute(dub.dub.Dub, immutable(char)[][], immutable(char)[][]) [0x70c427]
source/dub/commandline.d:252 int dub.commandline.runDubCommandLine(immutable(char)[][]) [0x704e8a]
source/app.d:14 _Dmain [0x6fc025]
========== -Output was ==========
[ERROR] :DUB should have retried download on server error multiple times, but only tried 1 times. command failed
```

MartinNowak:
Thx for fixing this up :).

ThomasMader:
I am getting the following with 2.078.2. It used to work with 2.078.1 and missed the beta release of 2.078.2.

Fetching gitcompatibledubpackage 1.0.4...
/tmp/nix-build-dubUnittests-1.7.2.drv-0/source/test/fetchzip.sh: line 21: dub: command not found
[ERROR] :21 command failed
[main(----) INF] Received signal 15. Shutting down.
[ERROR] Script failure.

wilzbach:
How do you run the testsuite?

ThomasMader:
It's running as part of the package building process for NixOS.
For me it seems it should not be "dub" but "$DUB" as used with the "$DUB" build command inside the same script.
dub is not in the path.

wilzbach:
> dub might not be in the path.

Yes, looks like that's not the case for you.
I saw that too and currently preparing a PR that will remove the DUB environment variable from the testsuite... Sorry about the inconvenience.

wilzbach:
https://github.com/dlang/dub/pull/1368

ThomasMader:
I patched it and now getting:

```
Trying to download gitcompatibledubpackage (1.0.4)
Package gitcompatibledubpackage not found for registry at http://localhost:5593/: Couldn't connect to server on handle CFD950
No package gitcompatibledubpackage was found matching the dependency 1.0.4
[ERROR] :17 command failed
[ERROR] Script failure.
```

wilzbach:
Do you expose some limitations that prevent vibe.d from starting? E.g. no network access to fetch the binary?

Could you post the full log?

ThomasMader:
It was my mistake, it works now.

PR #1340, opened 2018-01-15
Bump versions to 2.078.0

wilzbach:
Removed `dmd-nightly` as we now have the Project-Tester and the `FRONTEND`
versions is very likely to be out of sync. The same motivation applies for
replacing `ldc` with `ldc-1.7.0`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,9 @@ sudo: false
 
 matrix:
   include:
-    - d: dmd-nightly
+    - d: dmd-2.078.0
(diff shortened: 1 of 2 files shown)


wilzbach:
Travis LDC 1.7.0 hit https://github.com/dlang/dub/issues/1336

```
Fetching vibe-d 0.8.2...
Fetching libevent 2.0.2+2.0.16 (getting selected version)...
ZipException: no end record
```

again :/

PR #1341, opened 2018-01-15
Use build.cmd on AppVeyor to test bootstrapping DUB

wilzbach:


Geod24:
We have `build.d` now, and before we are self hosted (#1972) we'll need to do this anyway.

PR #1342, opened 2018-01-15
Start to generate man pages automatically

wilzbach:
Turns out that generating man pages automatically isn't hard.
See also: https://github.com/dlang/dub/pull/1143 (manual approch).

This is far from ideal, but it does generate man pages for all commands and
thus should get us started.


s-ludwig:
Nice! Is there anything left in the manual work that we should get in somehow?

The only thing is that this should be put in the "scripts" folder instead of creating its own top-level one.

wilzbach:
> Nice! Is there anything left in the manual work that we should get in somehow?

- "see also" is more sophisticated on the manual PR, but I don't think this is crucial
- dub.sdl / dub.json page (this2 can be done in a separate PR)

BTW the CLI output for a command lists the common option too. I left them out for now as #1143 didn't add them either, but thinking about it, we should probably add them too

> The only thing is that this should be put in the "scripts" folder instead of creating its own top-level one.

Done. Other changes:
- added it to the Travis script
- made the synopsis text dynamic
- fixed newline output for `DESCRIPTION`
- check for `acceptAppArgs`
- check for single / long arguments

Well, this doesn't aim to be perfect on the first try. I am pretty sure that some small modifications will be needed, but I think this brings us to 95% of the solution.

scripts/man/gen_man.d
```diff
@@ -0,0 +1,187 @@
...
+{
+	import std.uni : toUpper;
+
+	auto args = new CommandArgs(null);
+	command.prepare(args);
+	string fileName = format("dub-%s.1", command.name);
+	auto manFile = File(config.cwd.buildPath(fileName), "w");
+	auto manName = format("DUB-%s", command.name).toUpper;
+	manFile.writeHeader(manName, config);
+	static immutable seeAlso = ["dmd(1)".br, "dub(1)"].joiner("\n").to!string;
```
wilzbach: This could be more sophisticated.

s-ludwig:
Sounds like a good start!

PR #1343, opened 2018-01-15
Move the package repository on Windows to LocalAppData

Mark-MnetCS:
Move the package repository on Windows to LocalAppData to avoid bloat in the roaming data folder (speed up log-in/-off process). Keep the userSettings directory in AppData as this config data should roam.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -159,7 +159,7 @@ class Dub {
 			ps ~= defaultPackageSuppliers();
 
 		m_packageSuppliers = ps;
-		m_packageManager = new PackageManager(m_dirs.userSettings, m_dirs.systemSettings);
+		m_packageManager = new PackageManager(m_dirs.localRepository, m_dirs.systemSettings);
 		updatePackageSearchPath();
 	}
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Good to finally make progress in this area, thanks!

There are two things that should also be done:
1. An existing `Roaming/dub/packages` folder should be moved to `Local/dub/` if none exists there already
2. The file `dub/packages/local-packages.json`, as well as `local-overrides.json` should ideally be moved/stored to `Roaming/dub/*.json` instead of residing together with the cached packages

Number 2 requires a new constructor for `PackageManager` that takes two paths per "repository", one for packages and one for configuration files. Since this complicates the PR somewhat, I'd be fine with skipping this for now.

Mark-MnetCS:
I've updated the PR to handle the cache migration.

As for #2, it looks like this could be done by just changing the Repository struct, rather than adding a new constructor to PackageManager. However, does it really make sense to do this? I thought those files defined the packages that were overridden on the local machine. If those are part of the profile's roaming data they will get copied to any machine when the user logs in. Those overridden package files might not be present on that machine. Have I misunderstood?

s-ludwig:
You are right, I didn't think about it that way. We'd have to add the machine name to the file name or path to disambiguate then. On the other hand, "settings.json" can also contain paths (defaultCompiler), so if we were really strict about it, we'd simply have to move everything over to "Local"...

On the other hand, local-packages could also be seen as an additive list of paths that could contain the combined list of local packages from all machines and local-overrides can also override versions and not just paths. Then there is also the system-wide configuration that could be used as an alternative on machines where machine-specific global paths are used.

s-ludwig:
Hm, dlang-bot not active?

wilzbach:
Martin changed the behavior recently, so that it will only merge if all status checks pass. You can set the CodeCov notifications to be always green with `informational: true` - I can do so once I am back from my weekend trip.

PR #1347, opened 2018-01-24
Always set CodeCov notifications to green/passing

wilzbach:
As the dlang-bot now won't merge a PR when there's one failure [1], it makes sense
to force CodeCov to be `informational`. The interesting bit - coverage of the diff -
is still displayed in the CI tab and more importantly the browser extension still
works too and will continue to highlight the uncovered parts in red.

See also: https://github.com/dlang/dmd/blob/master/.codecov.yml

[1] https://github.com/dlang-bots/dlang-bot/pull/69

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,14 +1,34 @@
 # Documentation: https://docs.codecov.io/docs/codecov-yaml
+# Validate with: `curl --data-binary @.codecov.yml https://codecov.io/validate`
+
+codecov:
+  notify:
+    # We don't want to wait for the CodeCov report
+    # See https://github.com/codecov/support/issues/312
+    require_ci_to_pass: false
+    after_n_builds: 1  # send notifications after the first upload
+    wait_for_ci: false
+
+  bot: dlang-bot
+
+  # At Travis, the PR is merged into `master` before the testsuite is run.
+  # This allows CodeCov to adjust the resulting coverage diff, s.t. it matches
+  # with the GitHub diff.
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -1,14 +1,50 @@
...
-    patch: true
-    changes: true
+    project:
+      default:
+        informational: true
+        # Informational doesn't work with project yet
+        threshold: 0.1
+    patch:
+      default:
+        informational: true
```
MartinNowak: Didn't we figure that only patch coverage is relevant?
Can we just use `status: { patch: { default: { threshold: 5 } } }` instead? That should allow patch coverage to be a bit (5 per-cent points) lower than project coverage, but fails (and requires manual review) if it's even lower.
All the other coverage statuses seem superfluous.

MartinNowak:
Crazy amount of configuration.

.codecov.yml
wilzbach: Yes, but there seems to be a bug at CodeCov which leads to it not showing up at all. I haven't had time to report this to the CodeCov team, but that's what's happening at dmd atm.

wilzbach:
> Didn't we figure that only patch coverage is relevant?

I re-thought this and the error we experience at DMD is actually related to us using multiple builds, so we can safely use only CodeCov `patch` here.

wilzbach:
The Jenkins failure is unrelated - see https://github.com/dlang/ci/issues/135

PR #1352, opened 2018-01-25
Bump vibe-d to 0.8.3-alpha.3

wilzbach:


dub.selections.json
```diff
@@ -13,6 +13,6 @@
 		"stdx-allocator": "2.77.0",
 		"taggedalgebraic": "0.10.8",
 		"vibe-core": "1.3.0",
```
s-ludwig: This also needs to be bumped to "1.4.0-alpha.1", and the minimum frontend version in travis-ci.sh:7 needs to be raised to 2.073.

PR #1361, opened 2018-01-31
Remove use of the deprecated -gc DMD flag for debugInfoC

wilzbach:
See also: https://github.com/dlang/dmd/pull/7815

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -28,7 +28,7 @@ class DMDCompiler : Compiler {
 		tuple(BuildOption.releaseMode, ["-release"]),
 		tuple(BuildOption.coverage, ["-cov"]),
 		tuple(BuildOption.debugInfo, ["-g"]),
-		tuple(BuildOption.debugInfoC, ["-gc"]),
+		tuple(BuildOption.debugInfoC, ["-g"]),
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
Perhaps `debugInfoC` should be deprecated? This change is a bit misleading.

wilzbach:
> Perhaps debugInfoC should be deprecated? 

Not sure as it still has an effect for other compilers, see e.g.
https://github.com/dlang/dub/blob/master/source/dub/compilers/gdc.d#L31

jacob-carlborg:
> Not sure as it still has an effect for other compiler

Aha, ok.

MartinNowak:
We should ask @ibuclaw on the future of this, as he has worked a lot to improve D support in gdb, and it has long landed in recent distributions.

ibuclaw:
The original problem with dmd was that they invented their own dwarf debug tags for dynamic and associative arrays and delegates.

https://dlang.org/spec/abi.html#symbolic_debugging

The only debugger that understood these tags was the now dead zerobugs (originally proprietary, now open source abandonware).

Common sense has long since prevailed and dmd hasn't emitted this for years.

The outdated documentation I just linked should be deleted.

Issue #819, opened 2016-04-18
Build multiple executebles (easier config)


wilzbach:
My problem is simple - I would like to build multiple scripts that use the same library with `dub`.

```
examples/script1
examples/script2
source/lib/...
```
(concrete example [here](https://github.com/DlangScience/mir/pull/156))

Creating a new configuration entry for each executable fells a bit annoy - is there a solution to this on your roadmap?

s-ludwig:
See #97 - this should definitely be implemented sooner rather than later, as it's quite frequently requested.

jondegenhardt:
The TSV command line tools I released may provide a useful example of the needs of a repository/package creating multiple executables. Repo: [tsv-utils-dlang](https://github.com/eBay/tsv-utils-dlang). I normally use `make`, but also have a `dub` build mode to make it simpler for D developers who want to download and build with `dub`. The `dub` support seems popular, but making this usage smooth took some effort. Some specifics:

- Installation - `dub fetch` assumes installation into dub package directories and using `dub run` to execute. This is inconvenient for command line utilities, especially those commonly taking command line arguments or used in unix pipes. In addition, the dub package directories do not appear well suited for inclusion the user PATH. Using `dub fetch --local` would be more suitable, but this doesn't work at present (issue #1022). Whether `dub` should support an install command is a larger issue, but having a simple form of executable installation for folks building from source may be useful.
- Recursing through sub-directories / sub-packages - The repo has a sub-directory for each tool. Building all the tools involves iterating over the sub-directories and performing a `dub build`. However, I couldn't find a way to create a dub package specification that would do this. I ended up creating a special 'dub_build.d' program at the top-level that does this in response to a `dub run` command. This works, but it's limited. For example, it would be useful to be able to use `dub test` at the top level and have it run against all sub-directories / sub-packages. That is, running arbitrary dub commands at top-level and having them run against all sub-packages would be useful.
- Running arbitrary shell commands - The tools have both unit tests and tests run against the built executables. The tests against the executables use Unix shell commands. I was unable to find a way to run these tests as part of a `dub test` command, and I did not see a way to add a new command to dub that would provide a way to run these tests.

s-ludwig:
With #1364 in, what is left for this issue is a command line switch that allows building all sub packages or possibly all root packages in sub directories, even if they are not a dependency of the base package. It could also come in the form of supporting wildcards for "dub build" (e.g. `dub build --multi examples/*`)

davispuh:
[...]

Issue #97, opened 2013-07-20
Do something useful when "building" targetType none packages


s-ludwig:
Either an informative error message should appear, or, when sub packages exist, DUB could build all sub packages instead for example See also #87.

LightBender:
*bump* I know it's a low priority but it would be very helpful when working on large scale projects with many components and dependencies.

SimonN:
_bump_ -- I'd love to build two target binaries.

Or are there idiomatic solutions, besides running dub twice e.g. from a script/makefile?

s-ludwig:
Fixed by #1364

PR #1364, opened 2018-02-06
Build all dependencies when target type is "none"

lesderid:
This builds all dependencies when 'building' a package type with target type "none". This is mostly useful to allow building multiple (executable) sub-packages at once. (see issues #97 and #819)

There is still some more work to be done:
* Should sub-packages be implicit dependencies when the parent target type is "none"?
* How should this integrate with build configurations if at all?
* What should be removed when calling `dub clean` on a package with target type "none"? (in this PR, all path-based sub-packages are cleaned)
* Building with `rdmd` is not supported yet.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -773,6 +773,13 @@ class Dub {
 
 		if (existsFile(path ~ ".dub/build")) rmdirRecurse((path ~ ".dub/build").toNativeString());
 		if (existsFile(path ~ ".dub/obj")) rmdirRecurse((path ~ ".dub/obj").toNativeString());
+
+		auto p = Package.load(path);
+		if (p.getBuildSettings().targetType == TargetType.none) {
+			foreach (sp; p.subPackages.filter!(sp => !sp.path.empty)) {
+				cleanPackage(path ~ sp.path);
+			}
+		}
 	}
 
 	/// Fetches the package matching the dependency and places it in the specified location.
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -47,6 +47,11 @@ class BuildGenerator : ProjectGenerator {
 	{
 		scope (exit) cleanupTemporaries();
 
+		auto root_ti = targets[m_project.rootPackage.name];
+
+		enforce(!(settings.rdmd && root_ti.buildSettings.targetType == TargetType.none),
+				"Building package with target type \"none\" with rdmd is not supported yet.");
+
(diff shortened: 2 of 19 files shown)


wilzbach:
(in any case this would need tests)

lesderid:
I've added some basic tests.

jcrapuchettes:
What can I do to help this along? The company I work for could really use this feature.

s-ludwig:
Hm, I thought that I commented, but it seems like I never pushed the "comment" button.  So my main question it whether the other generators (VisualD, Sublime, CMake) still output valid project files with the changes.

> Should sub-packages be implicit dependencies when the parent target type is "none"?

Sounds like this should be triggered with a separate command line switch, also since it is something that happens at a different abstraction level. That switch could also cause any nested packages to be built, including single-file packages, so that examples get built, too.

> How should this integrate with build configurations if at all?

This is an argument for not building sub packages or other packages by default. The normal configuration handling should just work with the current solution; the target type `none` package can define configurations with appropriate sub configurations, so that the `--config` switch can be used, and the `--override-config` switch of course also works normally.

>  What should be removed when calling dub clean on a package with target type "none"? (in this PR, all path-based sub-packages are cleaned)

Sounds fine to me. It's currently too difficult to clean a complete package with sub packages or nested packages anyway.

> Building with rdmd is not supported yet.

I think we could see this as an additive enhancement. It should definitely be implemented one day (or an equivalent build mode using the new `-i` switch in DMD 2.079.0), but since building normally is still possible I'd not see this as critical. There should be a new ticket that replaces #97 in this case, though.

Domain:
I need this feature! My project contains a lot of dlls.

Hackerpilot:
@s-ludwig If I were to take over this merge request what are the chances of it eventually being merged? I'm currently working on a project that involves building about 250 executable files, and calling `dub` from `make` is quickly becoming a maintenance problem.

jcrapuchettes:
@s-ludwig @wilzbach What needs to be done to this issue to get it pushed through? This is an issue that needs to be added for industry (the company @Hackerpilot and I work for).

s-ludwig:
This actually looks fine to me as an incremental improvement. Additional command line switches or rules can be added later on as needed. If nobody sees any issues, I'll merge this tomorrow. Sorry for letting this sit here for so long!

s-ludwig:
Had to re-open to trigger SemaphoreCI.

PR #1368, opened 2018-02-10
[WIP] Use $DUB instead of dub and ensure that global DUB isn't available in the testsuite

wilzbach:
See also: https://github.com/dlang/dub/pull/1339

Removing `$DMD` from the testsuite might crash a few things, so expect this to fail for now.
I will remove WIP once it's passing.

PR #1376, opened 2018-02-12
[Trivial] Need to use '/usr/bin/env bash' instead of '/bin/bash'

dkgroot:


wilzbach:
See also: https://github.com/dlang/dub/pull/1368

PR #1377, opened 2018-02-12
Fix 1037 (rebased against master)

timotheecour:
just a rebase of https://github.com/dlang/dub/pull/1154 from @WebFreak001 against master

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -54,7 +54,7 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		}
 	}
 
-	static struct ConfigEntry
+	private static struct ConfigEntry
 	{
 		static struct Depender
 		{
@@ -62,8 +62,8 @@ class DependencyResolver(CONFIGS, CONFIG) {
 			TreeNodes dependency;
 		}
 
-		CONFIG[] all_configs;
-		bool any_config;
+		CONFIG[] allConfigs;
+		bool anyConfig;
 		Depender[] origins;
 	}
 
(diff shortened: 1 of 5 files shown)


source/dub/dependencyresolver.d
```diff
@@ -142,7 +156,8 @@ class DependencyResolver(CONFIGS, CONFIG) {
 
 				// get the current config/version of the current dependency
 				sizediff_t childidx = package_indices[basepack];
-				if (all_configs[childidx].length == 1 && all_configs[childidx][0] == CONFIG.invalid) {
+				auto child = configs[childidx];
+				if (child.allConfigs == [CONFIG.invalid]) {
```
s-ludwig: Produces a GC allocation in the inner loop.

timotheecour: did you measure whether it actually had a performance impact? 
also, if it does, StaticArray(Config.invalid) could be used

WebFreak001: oh wow that actually uses the GC... who would have thought if you could simply put that array in a separate line as `static immutable` to not do that.

s-ludwig: Yes, in the sense that I got huge speedups with all allocations removed from that loop. Pretty stupid that this is not optimized away by the compiler automatically.

timotheecour: https://github.com/dlang/druntime/pull/2093

s-ludwig: Definitely an improvement, although I still think that the compiler should be smarter about array literals. It could for example simply infer the possible `const`ness of the literal, and if it is not required to be mutable, it can safely allocate a static array instead of issuing a GC allocation (marking it as `@nogc`, too).

Issue #958, opened 2016-10-06
absolute paths shown as relative #178


timotheecour:
moved from https://github.com/dlang/dub-registry/issues/178

```
"targetPath":"/abspath/bar",
```

calling dub shows:
```
Running ../../../../../abspath/bar/exe
```

it should show: `/abspath/bar/exe`

(using appropriate std.path functions)

timotheecour:
ping on this?

s-ludwig:
The above case should be fixed already: 5c051bb19e622178bf457f68051694b80454224d

The idea is to work around path length limitations. #958 would introduce a regression in that regard.

s-ludwig:
So to elaborate, a typical scanario is to have a project in `%USERPROFILE%/dev/foo` and dependencies in `%USERPROFILE%/AppData/Roaming/dub/packages`. Using a relative path allows to replace `%USERPROFILE%` with `../../`, which is usually quite a bit shorter.

Once everything in the Windows build chain is UNC path aware, we can drop this workaround and then I'd also favor the approach in #958.

PR #1379, opened 2018-02-13
fix issue #958

timotheecour:
fix https://github.com/dlang/dub/issues/958

### DIFF
--- a/source/dub/internal/vibecompat/inet/path.d
+++ b/source/dub/internal/vibecompat/inet/path.d
@@ -156,7 +156,7 @@ struct NativePath {
 	}
 
 	/// Computes the relative path from `parentPath` to this path.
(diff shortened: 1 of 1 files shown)


s-ludwig:
See the comment in #958 - with the current state of things this would introduce a path length regression on Windows. We should target supporting `\\.\C:\...` style UNC paths throughout the build chain and then afterwards switch to this new relative path scheme.

MartinNowak:
> See the comment in #958 - with the current state of things this would introduce a path length regression on Windows.

Doesn't dub use response files for compiler invocations?

s-ludwig:
Yes, its the win32 `MAX_PATH` limitation, not the command line length limit.

timotheecour:
@s-ludwig what if i add version(Windows), would u merge?

Issue #1378, opened 2018-02-12
need option to override dependency in root package (useful for local development)


timotheecour:
often times I face situation where my main dub.json has a dependency A that itself depends on B:X.Y.Z
For whatever reason I need a newer version of B (X'.Y'.Z'). Currently the only way is to git clone A and manually update its dub.json, and then change dependency A in main dub.json to a path dependency pointing to the git clone. It would be nice to simply allow overriding B:X'.Y'.Z' in main dub.json and avoid all this extra work. This would only be allowed locally, not to be pushed to dub registery.
eg syntax:

in main dub.json:
```
"B": ":=1.2.3"
```

wilzbach:
Why don't you bump the version in the dub.selections.json? That's what it's therefore after all...

timotheecour:
but `dub.selections.json` will not even get generated in the 1st place if there's a conflict (IIRC)

MartinNowak:
No, the dependency should declare what versions it is compatible with and people should use less strict requirements, e.g. `~>1.0.0`. Also dependecies (libraries) don't pin their exact version (should not check-in dub.selections.json).
Allowing to mess with dependency specs in dependents would add a lot of additional complexity.

> but dub.selections.json will not even get generated in the 1st place if there's a conflict (IIRC)

What sort of conflict do you mean here? If you package requires a newer version of B, just state so in your `dub.sdl` and the dependency resolver will (try to) find a version that suits both your's and `A` depenency spec.

PR #1380, opened 2018-02-13
allow absolute paths in sourcePaths (useful for local development)

timotheecour:
this useful for local development (workarounds are painful), but we probably shouldn't use this feature for published dub packages.

Is there a way to prevent pushing a dub package to official dub repository that would rely on this feature? Likewise if a fix for https://github.com/dlang/dub/issues/1378 were implemented, we'd want to prevent pushing to official dub package repository pacakges using that override feature.

### DIFF
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -646,7 +646,12 @@ class Package {
 		string app_main_file;
 		auto pkg_name = m_info.name.length ? m_info.name : "unknown";
 		foreach(sf; bs.sourcePaths.get("", null)){
-			auto p = m_path ~ sf;
+			NativePath p;
+			import std.path:isAbsolute;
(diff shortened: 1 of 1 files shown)


s-ludwig:
There is a method in dub-registry that performs similar validation: https://github.com/dlang/dub-registry/blob/master/source/dubregistry/registry.d#L324

The only problem is that it just takes effect when the package is initially registered. Any later commit would then circumvent the check. But I guess in this particular case that would be an acceptable trade-off.

As for the change, I'd suggest to create a separate method `makeAbsolute` that keeps absolute paths, but at the same time outputs a warning to make people aware that this should only be used as a local/temporary workaround. That method could then also be used for the other fields (views, sourceFiles etc.).

PR #1386, opened 2018-02-17
Print build types from package file.

grogancolin:
Instead of the default list...

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -772,9 +772,9 @@ class GenerateCommand : PackageBuildCommand {
 
 		setupPackage(dub, package_name);
 
-		if (m_printBuilds) { // FIXME: use actual package data
+		if (m_printBuilds) {
 			logInfo("Available build types:");
-			foreach (tp; ["debug", "release", "unittest", "profile"])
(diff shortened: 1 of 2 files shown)


source/dub/commandline.d
```diff
@@ -772,9 +772,9 @@ class GenerateCommand : PackageBuildCommand {
 
 		setupPackage(dub, package_name);
 
-		if (m_printBuilds) { // FIXME: use actual package data
+		if (m_printBuilds) {
 			logInfo("Available build types:");
-			foreach (tp; ["debug", "release", "unittest", "profile"])
+			foreach (tp; (dub.project.rootPackage.getBuildTypes() ~ ["debug", "release", "unittest", "profile"]).sort)
```
s-ludwig: Would be good to add a `.uniq` to avoid duplicates here.

source/dub/package_.d
```diff
@@ -375,6 +375,10 @@ class Package {
 		return ret;
 	}
 
+	string[] getBuildTypes(){
```
s-ludwig: This may be a bit too specific to warrant a public method, since the external code almost as concisely directly call `.recipe.buildTypes.keys` instead. But a version of this method that provides enough own value would be one that returns a range of *all* possible build types, including the pre-defined ones. That would also move the list of pre-defined version from commandline.d closer to where it is actually defined.

PR #1388, opened 2018-02-19
[auto-deploy] search for 32-bit DMD executable

wilzbach:
See also: https://github.com/dlang/dub/issues/1367

This wasn't detected before, because I used `ldc` in my testing and only changed to DMD at the end (see https://github.com/dlang/dub/pull/1369) and the release steps are only executed on new tags.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,8 @@ jobs:
           on:
             tags: true
     - d: dmd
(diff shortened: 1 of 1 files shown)


wilzbach:
Going to pull this, s.t. auto-deploy works for the next tag.

Issue #268, opened 2014-03-05
'dub add-local' does not correctly detect duplicate path on Windows, if "/" is appended to path


bruno-medeiros:
When 'dub add-local ' is performed on entries that are already added, it doesn't ignore the action and creates dublicates. Then running 'dub list' for example gives:

Packages present in the system and known to dub:
  deelib 1.1.1: D:\devel\Scratchpad.D\DeeLib
  deelib 1.1.1: D:\devel\Scratchpad.D\DeeLib
  deelib 1.1.1: D:\devel\Scratchpad.D\DeeLib

**EDITED** See my comment further ahead with cmd-line code sample.

PR #1392, opened 2018-02-22
env var replacement not supporting `${VAR_NAME}` syntax

MartinNowak:
This is commonly needed to join pathes or a suffix, e.g. `$PACKAGE_DIRmylib.a` (as unfortunately PACKAGE_DIR ends on `/`).
Expanding variables is straighforward, https://github.com/MartinNowak/alertd/blob/cbc4459bbdbcc8c799413f58a591f6d28205f147/src/alertd.d#L503.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1183,9 +1183,9 @@ private void processVars(ref Appender!(string[]) dst, in Project project, in Pac
 
 private string processVars(Project, Package)(string var, in Project project, in Package pack,in GeneratorSettings gsettings, bool is_path)
 {
-	import std.regex : ctRegex, replaceAll;
+	import std.regex : regex, replaceAll;
 
-	enum varRE = ctRegex!`\$([\w_]+)|\$\{([\w_]+)\}`;
+	auto varRE = regex(`\$([\w_]+)|\$\{([\w_]+)\}`);
 	var = var.replaceAll!(m => getVariable(m[1].length ? m[1] : m[2], project, pack, gsettings))(varRE);
 	if (is_path) {
 		auto p = NativePath(var);
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1204,14 +1204,27 @@ package(dub) immutable buildSettingsVars = [
 private string getVariable(Project, Package)(string name, in Project project, in Package pack, in GeneratorSettings gsettings)
 {
 	import std.process : environment;
-	if (name == "PACKAGE_DIR") return pack.path.toNativeString();
-	if (name == "ROOT_PACKAGE_DIR") return project.rootPackage.path.toNativeString();
+	Path path;
+	if (name == "PACKAGE_DIR")
+		path = pack.path;
+	else if (name == "ROOT_PACKAGE_DIR")
+		path = project.rootPackage.path;
 
 	if (name.endsWith("_PACKAGE_DIR")) {
 		auto pname = name[0 .. $-12];
 		foreach (prj; project.getTopologicalPackageList())
 			if (prj.name.toUpper().replace("-", "_") == pname)
-				return prj.path.toNativeString();
+			{
+				path = prj.path;
+				break;
+			}
+	}
+
+	if (!path.empty)
+	{
+		// no trailing slash for clean path concatenation (see #1392)
+		path.endsWithSlash = false;
+		return path.toNativeString();
 	}
 
 	if (name == "ARCH") {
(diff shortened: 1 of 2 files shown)


s-ludwig:
Did you check if we get any significant compile-time regressions due to `ctRegex`?

Otherwise LGTM.

BTW, I'll go through the commit history to add missing change log entries once more, but I already wanted to suggest that we switch to a mode where each significant PR must add a change log entry, because I simply didn't manage to manually author them in time recently. We should probably add a CONTRIBUTING file at some point mentioning this.

MartinNowak:
> Did you check if we get any significant compile-time regressions due to ctRegex?

Yeah, replaced it with a runtime regex (btw, nowadays even `enum re = regex("")` is slow, as it merely does the same as ctRegex (https://github.com/dlang/phobos/pull/5722)).

s-ludwig:
> Regexes are cached by std.regex, so it shouldn't be a runtime problem.

Didn't know that. Then we can probably also remove the manual caching that is done in other places within DUB.

MartinNowak:
Of course it's messier than explainable, seems like `NativePath` only preserves trailing separators on posix, so one would need `${PACKAGE_PATH}foobar` on posix but `${PACKAGE_PATH}\\foobar` for Windows. Given that concatenating suffixes is currently not possible the risk of breaking sth. by removing the slash from PACKAGE_PATH on Posix doesn't seem to big, so I'll go for that to not create an even worse mess.

BTW, I think it would be a net positive for dub to get rid of NativePath.

MartinNowak:
> Didn't know that. Then we can also remove the manual caching that is done in other places within DUB.

Depends on expectations, https://github.com/dlang/phobos/blob/b26aad3209c1adb3006ff7003c44a7f0cdde06d9/std/regex/package.d#L364, 8 entries isn't necessarily enough, but processVars calls are rather sequential.

MartinNowak:
The slash was introduced with 4d4017c14c440ef444194f04cad72a0c1e1d312b to fix #268. It also makes sense for output such as dub list to append a slash, so I'd suggest to just remove the slash in env var expansion.

MartinNowak:
Does that look reasonable? It seems that the chances for breakage are small as `$PATHsuffix` wasn't possible beforehand, so mostly `cd $PATH && ./script` were used.
Could subtly break weird path based comparisons in scripts though.

wilzbach:
>but I already wanted to suggest that we switch to a mode where each significant PR must add a change log entry, because I simply didn't manage to manually author them in time recently. We should probably add a CONTRIBUTING file at some point mentioning this.

If you add a changelog entry to `changelog` like [DMD does](https://github.com/dlang/dmd/tree/master/changelog), the files will be displayed at https://dlang.org/changelog/pending.html and automatically part of the release announcement.

I recently added support for this on tools/dlang.org side for this:
https://github.com/dlang/tools/pull/302
https://github.com/dlang/dlang.org/pull/2163

Integrating the GitHub issues (like @dlang-bot currently does) is more work and I wasn't sure whether that's wanted, so I left that one out for now.

wilzbach:
>Regexes are cached by std.regex, so it shouldn't be a runtime problem.

Only since https://github.com/dlang/phobos/pull/6268 (the upcoming 2.080) and only the most recent one.
So it might be worthwhile to use `static varRE`

source/dub/project.d
```diff
@@ -1263,6 +1256,64 @@ private string getVariable(string name, in Project project, in Package pack)
...
+
+	MockProject proj = {
+		rootPackage: MockPackage("root"),
+		_dependencies: [MockPackage("dep1"), MockPackage("dep2")]
+	};
+	auto pack = MockPackage("test");
+	enum isPath = true;
+
+	version (Posix) enum sep = "/";
+	else version (Windows) enum sep = `\`;
```
wilzbach: aka as `std.path.dirSeparator`

source/dub/project.d
```diff
@@ -1243,17 +1223,30 @@ private string processVars(string var, in Project project, in Package pack, bool
...
+	Path path;
+	if (name == "PACKAGE_DIR")
+		path = pack.path;
+	else if (name == "ROOT_PACKAGE_DIR")
+		path = project.rootPackage.path;
 
 	if (name.endsWith("_PACKAGE_DIR")) {
 		auto pname = name[0 .. $-12];
 		foreach (prj; project.getTopologicalPackageList())
 			if (prj.name.toUpper().replace("-", "_") == pname)
```
wilzbach: FWIW there's no need to allocate for this check,sth. like this should work fine and is @nogc:
`prj.name.asUpperCase.map!(a == '-' ? '_' : a).equal(pname)`

wilzbach:
Addressed my review myself.

PR #1409, opened 2018-03-12
Inherit linker sourceFiles as sourceFiles instead of linkerFiles

### LINKED ISSUE #1408: [REG 1.8] Inherited linker files not included in the build
When building a vibe.d project that uses OpenSSL on Windows for x86_mscoff or x86_64, the OpenSSL import libraries that are specified in `vibe-d:tls` as `sourceFiles` are not included in the final linker invocation, thus causing the build to fail.

Inside the DUB code, such linker files are inherited to the `BuildSettings.linkerFiles` field instead of the `BuildSettings.sourceFiles` field since #1263. However, `linkerFiles` is currently only used by the "dub describe" functionality.

s-ludwig:
Fixes #1408.

The linkerFiles field was added some time ago in a very partial way and is not used at all, except for the "dub describe" functionality. Keeping inherited linker files in the sourceFiles list fixes the immediate regression introduced by #1263 - properly using linkerFiles throughout the code base should be implemented in a later step.

Reproducing the error is possible by building any vibe.d project that uses OpenSSL on Windows on x86_mscoff or x86_64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+v1.8.1 - 2018-
+-------------------
+
+- Fixed a regression in 1.8.0 that caused linker files specified as `sourceFiles` to not get inherited properly - [issue #1408][issue1408], [pull #1409][issue1409]
+
+[issue1408]: https://github.com/dlang/dub/issues/1408
+[issue1409]: https://github.com/dlang/dub/issues/1409
+
+
 v1.8.0 - 2018-03-01
 -------------------
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -36,6 +36,7 @@ skip_tags: false
 branches:
   only:
     - master
+    - stable
 
 install:
   - ps: function ResolveLatestDMD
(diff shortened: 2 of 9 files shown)


s-ludwig:
The interesting part is that it builds correctly - maybe there could also be a runtime failure in theory, but that would be unlikely to be related to the targeted issue. For that reason I'd rather keep the test suite lean, the CI builds already take ages anyway.

wilzbach:
> the CI builds already take ages anyway.

Fair point. BTW I was thinking about shifting some load to SemaphoreCI as from the experience with DMD it's really ~2-3x faster than Travis. One downside is that new jobs can only be added via their web UI. Though that could simply be fixed to series of a few old DMD versions.

Another idea would be to just use the latest two versions on AppVeyor.

s-ludwig:
Sounds good, I think AppVeyor is fine, though, it tends to be quicker than Travis nowadays.

wilzbach:
> Sounds good, 

Okay. I setup Semaphore to build with a single job to get us started:

https://semaphoreci.com/wilzbach/dub

It should now appear on new PRs (or old ones if you close/reopen them).
At the moment there's isn't a speedup as the entire testsuite is single-threaded. A worker on Semaphore has 8 threads (i7). So once the testsuite can run in multiple threads, it could potentially run in less than a minute per job.
I created https://github.com/dlang/dub/issues/1411 to keep track of this.

PR #1410, opened 2018-03-12
Ensure mainSourceFile is deduplicated properly

### LINKED ISSUE #1407: module app specified twice on command line
### System information
- **dub version**: DUB version 1.8.0, built on Mar  2 2018
- **OS Platform and distribution**: Windows 8
- **compiler version** DMD 2.079

### Bug Description
Receiving the following error when trying to build docs through dub:
Error: module `app` from file ..\packages\ddox-0.16.8\ddox\source\app.d is specified twice on the command line

### How to reproduce?
Run> dub build --build ddox

s-ludwig:
This forces the mainSourceFile to be normalized and converted to the native path format before being added to sourceFiles, so that when adding files later on with collectFiles matches up properly.

Fixes #1407.

### DIFF
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -169,8 +169,10 @@ struct BuildSettingsTemplate {
 		if (!this.targetName.empty) dst.targetName = this.targetName;
 		if (!this.workingDirectory.empty) dst.workingDirectory = this.workingDirectory;
 		if (!this.mainSourceFile.empty) {
-			dst.mainSourceFile = this.mainSourceFile;
-			dst.addSourceFiles(this.mainSourceFile);
+			auto p = NativePath(this.mainSourceFile);
+			p.normalize();
+			dst.mainSourceFile = p.toNativeString();
+			dst.addSourceFiles(dst.mainSourceFile);
 		}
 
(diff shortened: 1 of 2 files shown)


source/dub/recipe/packagerecipe.d
```diff
@@ -169,8 +169,10 @@ struct BuildSettingsTemplate {
 		if (!this.targetName.empty) dst.targetName = this.targetName;
 		if (!this.workingDirectory.empty) dst.workingDirectory = this.workingDirectory;
 		if (!this.mainSourceFile.empty) {
-			dst.mainSourceFile = this.mainSourceFile;
-			dst.addSourceFiles(this.mainSourceFile);
+			auto p = NativePath(this.mainSourceFile);
+			p.normalize();
+			dst.mainSourceFile = p.toNativeString();
+			dst.addSourceFiles(p.toNativeString());
```
wilzbach: I doubt it matters but the result of `toNativeString` could be cached.

s-ludwig: I'll simply use `dst.mainSourceFile`.

BTW, would be good to get #1409 in, so that this one can be rebased on top before adding a change log entry.

PR #1420, opened 2018-03-24
 Fix #1419 - Use LDC by default for the released binaries

### LINKED ISSUE #1419: Compile DUB binaries with LDC
(from a recent discussion https://github.com/dlang/dub/pull/1369#discussion_r167407362 in which the resulting size of a DUB's binary was discussed)

Compiler | Size
---------|------
DMD      | 17M
LDC      | 13M
GDC      | 19M

DUB is currently compiled with DMD, because LDC doesn't support showing the stack for exceptions, but imho the increase in speed and reduced side justifies overlooking this. After all, this is a production release - not a debug build.

LDC has LTO easily available which results usually in ~10% performance improvement and also a noticeable binary size reduction.
I haven't done big tests, but a build of vibe.d without needing to rebuild goes reproducible down from 0.22s to 0.15s

So my question: how big of a performance gain would justify the missing stack trace for exceptions?
(reminder: here's what DMD generates https://run.dlang.io/is/vzByQy vs. LDC: https://run.dlang.io/is/6iG8ai)

This issue is just about the separately built binaries.
(there's https://issues.dlang.org/show_bug.cgi?id=18656 for the bundled dub with the official release archives)

wilzbach:
See https://github.com/dlang/dub/issues/1419

Travis release build: https://travis-ci.org/wilzbach/dub/builds/357736472

https://github.com/wilzbach/dub/releases/tag/v1.8.4 (<- release tag is arbitrary and just needed to kickoff a Travis build on my fork)

(for comparison: https://github.com/dlang/dub/releases/tag/v1.8.0)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
     - d: gdc-4.8.5
       env: [FRONTEND=2.068]
     - stage: deploy
-      d: dmd
+      d: ldc
       os: osx
       script: echo "Deploying to GitHub releases ..." && ./release.sh
       deploy:
@@ -69,9 +69,8 @@ jobs:
           api_key: $GH_REPO_TOKEN
           on:
             tags: true
-    - d: dmd
-      # DMD 32-bit is needed for 32-bit compilation
-      script: echo "Deploying to GitHub releases ..." && DMD=$(find $HOME/dlang | grep "dmd-.*/linux/bin32/dmd") ./release.sh
+    - d: ldc
+      script: echo "Deploying to GitHub releases ..." && ./release.sh
       env: [ARCH=32]
       addons:
         apt:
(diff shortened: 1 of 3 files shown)


wilzbach:
Ping @s-ludwig - so are you okay with doing releasing the Posix dub binaries built with ldc?

s-ludwig:
Has the stack trace situation been solved with the `--export-dynamic`? In that case let's just go ahead!

wilzbach:
Yes -> https://github.com/dlang/dub/issues/1419#issuecomment-375860460

MartinNowak:
Not a good pick for a point release IMO.

wilzbach:
Oh sorry. Targetted this on `stable`, because only `stable` gets deployed and totally forgot that a new point release might be coming too.

Should I revert this and retarget `master`?

PR #1422, opened 2018-03-25
gen_man: Add missing "-lz" linker flag to link against zlib

ximion:
We add this patch on Debian, to ensure the resulting binary is correctly linked against ZLib.

### DIFF
--- a/scripts/man/gen_man.d
+++ b/scripts/man/gen_man.d
@@ -1,6 +1,7 @@
 #!/usr/bin/env dub
 /+dub.sdl:
 dependency "dub" path="../.."
+lflags "-lz"
 +/
 
 import std.algorithm, std.conv, std.format, std.path, std.range, std.stdio;


wilzbach:
Phobos has a built-in copy of zlib (https://github.com/dlang/phobos/tree/master/etc/c/zlib) and is part of the `libphobos2.a`.
Not sure how you modify Phobos, do you patch zlib out?

ximion:
@wilzbach 
> Not sure how you modify Phobos, do you patch zlib out?

Jup, and so does any other Linux distribution, as we don't allow embedded libraries for security and maintainability reasons[1] (ZLib is probably the most common case of them all).
The ideal solution would be if the compiler would detect whether some project uses ZLib and then link in the shared library automatically, I guess ;-)

If adding the additional ldflag causes problems to other applications, I would happily carry this as a downstream patch for now (that's why this is a separate PR anyway).

[1]: https://wiki.debian.org/EmbeddedCodeCopies

wilzbach:
> The ideal solution would be if the compiler would detect whether some project uses ZLib and then link in the shared library automatically, I guess ;-)

Can't you add `pragma(lib, "z");` to your patched version of `std.zlib`?

> If adding the additional ldflag causes problems to other applications, I would happily carry this as a downstream patch for now (that's why this is a separate PR anyway).

I doubt it matters, because (1) this script isn't widely used so far and (2) it just loads `libz` which is already part of the binary.

ximion:
> Can't you add pragma(lib, "z"); to your patched version of std.zlib?

This will only do something if people do a union build (compile and link in one step), so it would only cover a small amount of cases (in Debian right now, due to most projects using Meson or hand-crafted Makefiles, I guess no project would be using it).
For union builds, I wonder if that pragma would also work if the unit is just included, but not compiled.

PR #1423, opened 2018-03-25
gen_man: Fix build with gdc-8

ximion:
Apparently, GDC 8 (or previous versions) don't have a Phobos that allows the regex or string to be immutable.
This patch allows the current gen_man to be built with GDC 8.

CC: @ibuclaw


wilzbach:
> Apparently, GDC 8 (or previous versions) don't have a Phobos that allows the regex or string to be immutable.

I just checked and there doesn't seem to be a regression on the Phobos side:

https://run.dlang.io/is/hs0EIm

Anyhow, it doesn't really matter for this script.

PR #1426, opened 2018-03-26
Add a MIT License file that can be detected by GitHub

wilzbach:
https://github.com/wilzbach/dub/blob/master/LICENSE

GH uses licensee to detect the LICENSE and it has severe problems with LICENSE_de.txt - so went with unifying to one license file.

https://blog.github.com/2016-09-21-license-now-displayed-on-repository-overview/

https://github.com/benbalter/licensee

If keeping LICENSE_DE is desired, I guess I would have to dive more into licensee and see if there's a workaround

image

image

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2012-2016 RejectedSoftware e.K.
+Copyright (c) 2016-2018 D Language Foundation
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
+
+Copyright (c) 2012-2016 RejectedSoftware e.K.
+Copyright (c) 2016-2018 D Language Foundation
```
wilzbach: Wasn't sure about this and that's just an assumption.

PR #1428, opened 2018-03-28
dub run <package> should search the registry on failure

wilzbach:
(Reopened https://github.com/dlang/dub/pull/1082 as I accidentally closed it during the push)

This has been pending in the queue for quite a while. Anything blocking this?

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -517,7 +517,7 @@ class InitCommand : Command {
 			free_args = free_args[1 .. $];
 		}
 
-		string input(string caption, string default_value)
+		static string input(string caption, string default_value)
 		{
 			writef("%s [%s]: ", caption, default_value);
 			auto inp = readln();
@@ -829,6 +829,7 @@ class GenerateCommand : PackageBuildCommand {
 }
 
 class BuildCommand : GenerateCommand {
+	bool m_yes; // automatic yes to prompts;
 	this()
 	{
 		this.name = "build";
@@ -848,12 +849,70 @@ class BuildCommand : GenerateCommand {
 		args.getopt("f|force", &m_force, [
 			"Forces a recompilation even if the target is up to date"
 		]);
+		args.getopt("y|yes", &m_yes, [
+			`Automatic yes to prompts. Assume "yes" as answer to all prompts and run non-interactively.`
+		]);
 		super.prepare(args);
 		m_generator = "build";
 	}
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		// single package files don't need to be downloaded, they are on the disk.
+		if (free_args.length < 1 || m_single)
+		    return super.execute(dub, free_args, app_args);
+
+        const package_parts = splitPackageName(free_args[0]);
+        const package_name = package_parts.name;
+
+        static bool input(string caption, bool default_value = true) {
+            writef("%s [%s]: ", caption, default_value ? "Y/n" : "y/N");
+            auto inp = readln();
+            string userInput = "y";
+            if (inp.length > 1)
+                userInput = inp[0 .. $ - 1].toLower;
+
+            switch (userInput) {
+                case "no", "n", "0":
+                    return false;
+                case "yes", "y", "1":
+                default:
+                    return true;
+            }
+        }
+
+        Dependency dep;
+
+        if (package_parts.version_.length > 0) {
+            // the user provided a version manually
+            free_args[0] = package_name;
+            dep = Dependency(package_parts.version_);
+        } else {
+			const pack = dub.packageManager.getFirstPackage(package_name);
+			if (pack)
+				return super.execute(dub, free_args, app_args);
+
+            // search for the package and filter versions for exact matches
+            auto search = dub.searchPackages(package_name)
+                .map!(tup => tup[1].find!(p => p.name == package_name))
+                .filter!(ps => !ps.empty);
+            if (search.empty)
+                return 2;
+
+            const p = search.front.front;
+            logInfo("%s wasn't found locally, but it's available online:", package_name);
+            logInfo("---");
+            logInfo("Description: %s", p.description);
+            logInfo("Version: %s", p.version_);
+            logInfo("---");
+
+            const answer = m_yes ? true : input("Do you want to fetch %s?".format(package_name));
+            if (!answer)
+                return 0;
+            dep = Dependency(p.version_);
+        }
+
+        dub.fetch(package_name, dep, dub.defaultPlacementLocation, FetchOptions.none);
 		return super.execute(dub, free_args, app_args);
 	}
 }
--- a/changelog/dub-run.dd
+++ b/changelog/dub-run.dd
@@ -1,0 +1,45 @@
+`dub run` will now automatically fetch a package if it's not found locally
+
+Starting with this release, `dub run <mypackage>` makes sure that the package is available locally.
+This means that now a `dub fetch <mypackage>` is no longer required and all a user of a library needs to run your dub package is `dub run`:
+
+$(CONSOLE
+> dub run gitcompatibledubpackage
+gitcompatibledubpackage wasn't found locally, but it's available online:
+---
+Description: Example of a DUB package also usable as git submodule. For DUB test suite.
+Version: 1.0.4
+---
+Do you want to fetch gitcompatibledubpackage? [Y/n]:
+)
+
+An optional `--yes` (`-y`) flag is provided for non-interactive use:
+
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+v1.9.0 (upcoming)
+-------------------
+
+- `dub run` will now automatically fetch a package if it's not found locally [pull #1428][pull1428]
```
wilzbach: We could also opt to switch to the "D" `changelog` format and add individual entries.
The advantage: it would be displayed at https://dlang.org/changelog/pending.html and automatically part of the release changelog
The disadvantage: more overhead for the tiny traffic that dub currently receives

wilzbach:
BTW now that we show instructions on how to build a dlang PR locally:

image

(example: https://github.com/dlang/dmd/pull/8155)

This would be even more interesting imho.
(Of course, I'm aware we would have to wait a bit until we can remove `dub fetch` from the instructions, but we have to start somewhere and this has been in the queue for more than a year already)

wilzbach:
Ping on this. I still think this would be nice to have and there haven't been any objections.
Maybe we can finally get this in for the next release?

wilzbach:
> Can we have -y command line option to assume "Yes" as answer to the prompt?

Good idea. Added.

source/dub/commandline.d
```diff
@@ -836,12 +837,54 @@ class BuildCommand : GenerateCommand {
...
+					.filter!(ps => !ps.empty);
+				if (search.empty)
+					return 2;
+
+				auto p = search.front.front;
+				writefln("%s wasn't found locally, but it's available online:", package_name);
+				writefln("Description: %s", p.description);
+				writefln("Version: %s", p.version_);
+
+				bool answer = m_yes ? true : input("Do you want to fetch %s?".format(package_name));
```
WebFreak001: I like the change but I think we should move the m_yes/input method into a more general scope where all methods could read a boolean input

andre2007:
As dub run now implicit calls fetch, I wonder whether dub run should also have an argument "version".
If I run a CI  build pipeline it must always have the same result. Therefore i would like to specify the version of the tool I ran with dub run.

On the other hand one can say that is a special case for which I can use dub fetch && dub run.

source/dub/commandline.d
wilzbach: Is there anything else that currently uses this?
We could have easily done this after this PR has been merged and this simple improvement has been in the queue now for more than 1 1/2 years (see the original PR).

wilzbach:
Rebased.

thewilsonator:
This appears to break a bunch of shell script tests.

andre2007:
E.g. test issue1158-stdin-for-single-files.sh fails because package_name (free_args[0]) contains value
```C:\Users\User\AppData\Local\Temp\app_dcb374c2_7b1e_4536_9027_d1d10347846f.d```

**Comment by andre2007** [comment]
I just realized, with this feature the custom init idea from @MartinNowak (https://github.com/dlang/dub/pull/600) becomes a low hanging fruit. Just a matter of rewriting the args[] array.

**Comment by wilzbach** [comment]
All green now.

File: source/dub/commandline.d [hunk 3]
```diff
@@ -848,12 +849,55 @@ class BuildCommand : GenerateCommand {
...
+
+				// search for the package and filter for exact matches
+				auto search = dub.searchPackages(package_name)
+					.map!(tup => tup[1].find!(p => p.name == package_name))
+					.filter!(ps => !ps.empty);
+				if (search.empty)
+					return 2;
+
+				auto p = search.front.front;
+				writefln("%s wasn't found locally, but it's available online:", package_name);
```
ZombineDev: Don't use `std.stdio` directly, as this circumvents dub's logging infrastructure. Use `logInfo` instead.

source/dub/commandline.d
```diff
@@ -848,12 +849,55 @@ class BuildCommand : GenerateCommand {
...
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		// single package files don't need to be downloaded, they are on the disk.
+		if (free_args.length >= 1 && !m_single) {
+			auto package_name = free_args[0];
+			auto pack = dub.packageManager.getFirstPackage(package_name);
+
+			if (!pack) {
+				bool input(string caption, bool default_value = true) {
```
ZombineDev: There is at least one more local definition of `input` (e.g. for `dub init`) in this file.  I'd like if we could generalize this into a reusable free function + as @WebFreak001 mentioned the `y/N` handling code.
Not a strong requirement from my side, just nice to do.

source/dub/commandline.d
```diff
@@ -848,12 +849,55 @@ class BuildCommand : GenerateCommand {
...
 	}
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		// single package files don't need to be downloaded, they are on the disk.
+		if (free_args.length >= 1 && !m_single) {
+			auto package_name = free_args[0];
+			auto pack = dub.packageManager.getFirstPackage(package_name);
+
+			if (!pack) {
```
ZombineDev: Switch the condition and do an early exit, if possible.

source/dub/commandline.d
```diff
@@ -848,12 +849,55 @@ class BuildCommand : GenerateCommand {
...
+					return 2;
+
+				auto p = search.front.front;
+				writefln("%s wasn't found locally, but it's available online:", package_name);
+				writefln("Description: %s", p.description);
+				writefln("Version: %s", p.version_);
+
+				bool answer = m_yes ? true : input("Do you want to fetch %s?".format(package_name));
+				auto dep = Dependency(p.version_);
+				dub.fetch(package_name, dep, dub.defaultPlacementLocation, FetchOptions.none);
```
ZombineDev: Would be nice to be able to select the package repository and version, like with `docker run`, e.g.:
```
dub run ${PACKAGE_REGISTRY}/${PACKAGE_PATH}:${PACKAGE_VERSION}
```

source/dub/commandline.d
wilzbach: Hmm the `dub init` `input` is quite different to this input as (1) a default value is allowed and (2) multiple values are allowed for a single value (all `yes`, `y`, `1`, `YeS`, `Y`, ...) map to `1`.

Though I'm not against unifying them by any means if someone wants to try it _after_ this addition.

source/dub/commandline.d
wilzbach: Hmm in `dub init` `std.stdio.writef` is used too.

source/dub/commandline.d
wilzbach: That duplicates `return super.execute(dub, free_args, app_args);`, but no strong feelings either way. I changed it to the early exit.

source/dub/commandline.d
wilzbach: Nice idea :+1: 
This also removes the need for asking for verification in these cases. I only added support for selecting a specific version with `{PACKAGE_NAME}:{VERSION}` as selecting thee registry isn't really supported by `dub` at all at the moment.

Also note that e.g. `dub fetch` only supports `--version 1.0.4` (though it would be trivial to add support for `dub fetch mypackage:1.0.4`)

wilzbach: So it turns out that the recently added `dub add` command uses `=`, but simply supporting both `=` and `:` will work.

> though it would be trivial to add support for dub fetch mypackage:1.0.4

https://github.com/dlang/dub/pull/1643

source/dub/commandline.d
```diff
@@ -848,12 +849,70 @@ class BuildCommand : GenerateCommand {
...
 		m_generator = "build";
 	}
 
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
+		// single package files don't need to be downloaded, they are on the disk.
+		if (free_args.length < 1 || m_single)
+		    return super.execute(dub, free_args, app_args);
+
+        const package_parts = free_args[0].split(":");
```
wilzbach: This could use `splitPackageName` from https://github.com/dlang/dub/pull/1643

wilzbach:
Rebased, used `splitPackageName` from #1643 and should hopefully be good 2 go now (: 

We missed 2.085.0 by a day, but well at least we can get it in the next release.

wilzbach:
Ping @thewilsonator

MartinNowak:
We already have a `--non-interactive` option for other commands, would be better to consistently stick with it.

MartinNowak:
Well `--yes` and `--non-interactive` are semantically differrent.

PR #1415, opened 2018-03-17
Maven repository package supplier added

andre2007:
The purpose of this pull request is to be a door opener for dub in commercial sector. In build infrastructure it is in many cases not allowed to connect to the public internet (code.dlang.org) for security reasons. Instead the companies runs their own repositories using software like Nexus or Artifactory.
It is quite hard to convince system administrators to also run additional software like dub registry.

To solve this problem, this pull request enables dub to connect to maven repositories to retrieve dub packages. Environment variable ```DUB_REGISTRY``` and console argument ```--registry``` now accepts additional protocols (dub/mvn). Example:

--registry dub+https://code.dlang.org
--registry mvn+http://localhost:8040/artifactory/libs-release/dubpackages

These Maven urls consists of the path to a maven repository + a group id. 
(libs-release is the maven repository, dubpackages is the group id)

In the maven repository you deploy your dub package with group id "dubpackages" and
as artifactId you use dub package name and as version you use dub package version.

Maven creates a metadata xml file which get quite handy for retrieving metadata needed for dub
```
<?xml version="1.0" encoding="UTF-8"?>
<metadata modelVersion="1.1.0">
  <groupId>dubpackages</groupId>
  <artifactId>d-unit</artifactId>
  <version>0.8.2</version>
  <versioning>
    <latest>0.8.2</latest>
    <release>0.8.2</release>
    <versions>
      <version>0.8.2</version>
    </versions>
    <lastUpdated>20180317184845</lastUpdated>
  </versioning>
</metadata>
```
This metadata and the archives (zip)files are stored using canonical paths.
http://localhost:8040/artifactory/libs-release/dubpackages/d-unit/maven-metadata.xml
http://localhost:8040/artifactory/libs-release/dubpackages/d-unit/0.8.2/d-unit-0.8.2.zip

As far as I know Maven does not provide a nice generic way to implement the searchPackages method.
Therefore I left this method empty.

Can you have a look at this pr?

PR #1429, opened 2018-03-28
Use the more generic 0.8 version for Vibe.d in 'dub init' to avoid it getting outdated too fast

wilzbach:
A quick fix that should push the problem of having an outdated Vibe.d version when running `dub init`  a bit more into the future.

At some point a proper `init` template system / bootstrapping system would be great though.

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -111,7 +111,7 @@ void main()
 private void initVibeDPackage(NativePath root_path, ref PackageRecipe p, scope void delegate() pre_write_callback)
 {
 	if ("vibe-d" !in p.buildSettings.dependencies)
-		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.8.2");
+		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.8");
 	p.description = "A simple vibe.d server application.";
 	pre_write_callback();
 


andre2007:
A little bit related to your pull request. I wondered why your pull request went to green because my new test which is based on fetchzip.sh went to red on travis for dmd 2.071.2 and below. By analyzing your travis reports I saw, that test fetchzip.sh is not executed for dmd 2.071.2 and below. Do you know how it is disabled? I need the same disable log for my new test issue1416-maven-repo-pkg-supplier.sh
from pr #1415

wilzbach:
>  Do you know how it is disabled? I need the same disable log for my new test issue1416-maven-repo-pkg-supplier.sh
from pr #1415

Have a look at `test/fetchzip.sh.min_frontend`

s-ludwig:
The problem is that this will also allow 0.9.0, 0.10.0 etc. On the other hand, having `~>0.8.2` isn't really a problem usually. It will still use the latest 0.8.x release, and the number of breaking changes is small enough to make it unlikely to have compatibility issues with the base version in practice.

andre2007:
With pr https://github.com/dlang/dub/pull/1658 the init template system was implemented. Maybe vibe.d could make use of this and the special logic in dub can be removed?

Geod24:
Based on @s-ludwig 's comment that it's not wanted (because of the unintentional upgrade), the fact that we have a new init system as pointed by @andre2007, and this is 2 years old, I'm just going to give this the axe.

PR #1433, opened 2018-04-03
Use full Link-Time Optimization (LTO) for the release binaries

wilzbach:
Without LTO
-----------

```
~/dlang/avgtime/avgtime -q -r1000 ~/dlang/dub-repo/bin/dub.stable describe

------------------------
Total time (ms): 110249
Repetitions    : 1000
Sample mode    : 100 (586 occurrences)
Median time    : 108.899
Avg time       : 110.249
Std dev.       : 6.25044
Minimum        : 100.357
Maximum        : 142.289
95% conf.int.  : [97.9979, 122.499]  e = 12.2506
99% conf.int.  : [94.1484, 126.349]  e = 16.1001
EstimatedAvg95%: [109.861, 110.636]  e = 0.387399
EstimatedAvg99%: [109.739, 110.758]  e = 0.509129
```

With LTO
--------

```
~/dlang/avgtime/avgtime -q -r1000 ~/dlang/dub-repo/bin/dub.lto describe

------------------------
Total time (ms): 97952.6
Repetitions    : 1000
Sample mode    : 93 (117 occurrences)
Median time    : 96.2915
Avg time       : 97.9526
Std dev.       : 5.86976
Minimum        : 89.173
Maximum        : 146.487
95% conf.int.  : [86.4481, 109.457]  e = 11.5045
99% conf.int.  : [82.8331, 113.072]  e = 15.1195
EstimatedAvg95%: [97.5888, 98.3164]  e = 0.363805
EstimatedAvg99%: [97.4745, 98.4307]  e = 0.478121
```

Note that the binary size goes up from 17M to 22M.
(Compressed from 3.8M to 5.2M), but imho space is cheap compared to this huge speed-up.

Learn more about LTO -> http://johanengelen.github.io/ldc/2016/11/10/Link-Time-Optimization-LDC.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,8 @@ jobs:
 stages:
   - name: test
     if: type = pull_request or (type = push and branch = master)
-  - name: deploy
-    if: type = push and tag =~ ^v
+  # Until deployment of the release binaries is fixed, always build them
+  #- name: deploy
+    #if: type = push and tag =~ ^v
   - name: update-latest
     if: type = push and tag =~ ^v
--- a/build.sh
+++ b/build.sh
@@ -52,8 +52,8 @@ else
 	echo Using existing version file.
 fi
 
-# For OSX compatibility >= 10.7
-MACOSX_DEPLOYMENT_TARGET=10.7
+# For OSX compatibility >= 10.8
+MACOSX_DEPLOYMENT_TARGET=10.8
 
 echo Running $DMD...
 $DMD -ofbin/dub -g -O -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
--- a/release.sh
+++ b/release.sh
@@ -1,22 +1,30 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-set -v -e -o pipefail
+set -eux -o pipefail
 
 VERSION=$(git describe --abbrev=0 --tags)
 ARCH="${ARCH:-64}"
-CUSTOM_FLAGS=""
-
+CUSTOM_FLAGS=()
 unameOut="$(uname -s)"
 case "$unameOut" in
     Linux*)
         OS=linux
-        CUSTOM_FLAGS="-L--export-dynamic"
+        CUSTOM_FLAGS+=("-L--export-dynamic")
         ;;
     Darwin*)
         OS=osx
+        CUSTOM_FLAGS+=("-L-dead_strip")
         ;;
     *) echo "Unknown OS: $unameOut"; exit 1
 esac
+
+if [[ $(basename "$DMD") =~ ldmd.* ]] ; then
+    CUSTOM_FLAGS+=("-flto=full")
+    # ld.gold is required on Linux
+    if [ ${OS:-} == "linux" ] ; then
+        CUSTOM_FLAGS+=("-linker=gold")
+    fi
+fi
 
 case "$ARCH" in
     64) ARCH_SUFFIX="x86_64";;
@@ -27,5 +35,5 @@ esac
 archiveName="dub-$VERSION-$OS-$ARCH_SUFFIX.tar.gz"
 
 echo "Building $archiveName"
-DFLAGS="-release -m$ARCH ${CUSTOM_FLAGS}" DMD="$(command -v $DMD)" ./build.sh
+DFLAGS="-release -m$ARCH ${CUSTOM_FLAGS[@]}" DMD="$(command -v $DMD)" ./build.sh
 tar cvfz "bin/$archiveName" -C bin dub


release.sh
```diff
@@ -1,23 +1,27 @@
...
         ;;
     Darwin*)
         OS=osx
+        CUSTOM_FLAGS+=("-L-macosx_version_min" "-L10.7" "-L-lcrt1.o")
         ;;
     *) echo "Unknown OS: $unameOut"; exit 1
 esac
 
+if [[ $(basename "$DMD") =~ ldmd.* ]] ; then
+    CUSTOM_FLAGS+=("-flto=full" "-linker=gold")
```
wilzbach: Ported the configuration from https://github.com/dlang-community/D-Scanner/pull/598 to this script.
(the binaries are here https://github.com/dlang-community/D-Scanner/releases/tag/v0.5.1)
Note: LDC's `-static` linking can't be used because `curl` (it's `dlopen-ed` in std.net.curl and apparently is `std.socket` not very static link-friendly either)

release.sh
```diff
@@ -1,23 +1,27 @@
...
 unameOut="$(uname -s)"
 case "$unameOut" in
     Linux*)
         OS=linux
-        CUSTOM_FLAGS="-L--export-dynamic"
+        CUSTOM_FLAGS+=("-L--export-dynamic")
         ;;
     Darwin*)
         OS=osx
+        CUSTOM_FLAGS+=("-L-macosx_version_min" "-L10.7" "-L-lcrt1.o")
```
wilzbach: From: https://github.com/dlang-community/discussions/issues/9#issuecomment-378725199

jacob-carlborg:
> Note that the binary size goes up from 17M to 22M.

Have a look at [1]. On macOS for D-Scanner the binary size is reduced using LTO and `-L-dead_strip` reduces it even more.

[1] https://github.com/dlang-community/D-Scanner/pull/598#issuecomment-379161594

release.sh
wilzbach: Maybe even 10.9? (as this will come with 2.079)

https://dlang.org/changelog/2.079.0.html#osx-10_9

wilzbach: BTW what for do we need `lcrt1.o`?
I assumed this is a legacy thing too?

build.sh
```diff
@@ -52,8 +52,8 @@ else
 	echo Using existing version file.
 fi
 
-# For OSX compatibility >= 10.7
-MACOSX_DEPLOYMENT_TARGET=10.7
+# For OSX compatibility >= 10.8
+MACOSX_DEPLOYMENT_TARGET=10.8
```
MartinNowak: That's due to a recent bump in dmd, isn't it? Seems unrelated to the PR though, at least it's uncommented why it's here.

.travis.yml
```diff
@@ -107,7 +107,8 @@ jobs:
 stages:
   - name: test
     if: type = pull_request or (type = push and branch = master)
-  - name: deploy
-    if: type = push and tag =~ ^v
+  # Until deployment of the release binaries is fixed, always build them
```
MartinNowak: It don't understand this sentence, also seems unrelated to LTO ;).
Disabling auto-deploy to GH releases seems fine for now, as mentioned earlier I'd prefer to use the official binaries instead of rebuilding from sources. IIUC this PR adds support for LDC building and the GH binaries would prolly use it, while the official binaries still use dmd.
This inconsistency is likely to require additional maintenance and debug efforts down the road. Would be slightly cleaner to switch over dub together with using gdc or ldc to build optimized dmd release binaries.

In fact integrating gdc or ldc with [installer/create_dmd_release](https://github.com/dlang/installer/tree/6763102708fbf6ff59bd13ed6cf761aa05c0e07c/create_dmd_release) just for dub would be a good first step towards changing the optimized compiler for dmd releases.

release.sh
```diff
@@ -1,23 +1,31 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-set -v -e -o pipefail
+set -eux -o pipefail
```
MartinNowak: We should standardize the writing to simplify reviews ;).
How about  `set -xeuo pipefail` and `set -euo pipefail`?

release.sh
```diff
@@ -1,23 +1,31 @@
...
 ARCH="${ARCH:-64}"
-CUSTOM_FLAGS=""
-
+CUSTOM_FLAGS=()
 unameOut="$(uname -s)"
 case "$unameOut" in
     Linux*)
         OS=linux
-        CUSTOM_FLAGS="-L--export-dynamic"
+        CUSTOM_FLAGS+=("-L--export-dynamic")
```
MartinNowak: Generally prefer single ticks for strings w/o interpolation.

build.sh
jacob-carlborg: It causes linker warnings because linked dependencies are built targeting 10.8.

jacob-carlborg: See previous comment: https://github.com/dlang/dub/pull/1433#discussion_r179668984.

MartinNowak:
Also compiler optimizations are mood for an app that mostly waits on remote network requests, and optimizing offline 100ms latencies isn't paying back that much ;).
I guess the dependency resolver is the only hot loop in dub for complex dependency graphs.

jacob-carlborg:
LTO reduces binary size, at least on macOS.

release.sh
jacob-carlborg: Perhaps we should have a version statement that allows to statically link curl. Then we could use the musl C standard library that fully supports static linking.

PR #1437, opened 2018-04-05
Install the newest version of dub to avoid random network failures

wilzbach:
Analogous to https://github.com/vibe-d/vibe.d/pull/2140 and https://github.com/vibe-d/vibe-core/pull/78

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,12 @@ addons:
     packages:
       - libevent-dev
 
+before_install:
+  # Use the dub-updating fork of the installer script until https://github.com/dlang/installer/pull/301 is merged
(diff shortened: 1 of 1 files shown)


s-ludwig:
Still ran into two issues, but maybe those are simply GitHub related:
> Failed to extract zip archive for vibe-d 0.8.3-alpha.4...
> ZipException: no end record

wilzbach:
Hmm yeah I have seen this failure too (e.g. https://github.com/dlang/dub/issues/1336 and I tried to address it with https://github.com/dlang/dub/pull/1339) but it seems like the logic in https://github.com/dlang/dub/pull/1339 still has space for improvement :/

Not so sure what we could do to solve this - maybe even more retries with exponential backup off or falling back to `git clone`?

Issue #1438, opened 2018-04-06
DUB should expose a package's version to the respective package


wilzbach:
Problem: the built program should expose its version to the user (e.g. `--version`).

### Semi-generic workaround:

This will generate a `bin/dubhash.txt` file that can be imported anywhere in the program with `import("dubhash.txt")`.

```json
  "stringImportPaths": ["bin"],
  "preGenerateCommands": [
    "rdmd --eval='auto dir=environment.get(`DUB_PACKAGE_DIR`); dir.buildPath(`bin`).mkdirRecurse; auto gitVer = (`git -C `~dir~`describe --tags`).executeShell; (gitVer.status == 0 ? gitVer.output : dir.dirName.baseName.findSplitAfter(environment.get(`DUB_ROOT_PACKAGE`)~`-`)[1]).ifThrown(`0.0.0`).toFile(dir.buildPath(`bin`, `dubhash.txt`));'",
  ]
```

This workaround will default to git describe --tags (within the packages's directory) and if that fails it will detect its version based on the DUB package path.If that fails too, it will fallback to 0.0.0.

The workaround "works", but obviously it would be nicer if `dub` found a standardized way to expose the package's current version to the package itself. It has this information after all...

WebFreak001:
yes pls add `enum PackageCommitSha = "...";` and `enum PackageVersionTag = "v0.1.0";` and the same for subpackages using `enum Package_<id>_CommitSha` and `Package_<id>_VersionTag`, dub already knows about these and I just needed that in serve-d but instead used the github api to get commit date and built with `__TIMESTAMP__` to find out if a program is outdated

andre2007:
Does this PR somehow helps you?
Environment variable DUB_PACKAGE_VERSION added (https://github.com/dlang/dub/pull/1562)

WebFreak001:
I don't think that is helpful as it's just for the commands dub can run. What would fix this issue imo would be having defines (enums) set to the versions.

wilzbach:
> Does this PR somehow helps you?

It simplifies the logic quite a bit - though it's probably such a common use case that we won't it to be possible without the need for any custom D code?

andre2007:
I fully agree. I have a similar logic in all of my productive applications.

I have 2 things I want to mention.

- I know the version attribute in dub.json is not officially supported but could the solution respects this value if set? (By using Package.version_ attribute). 
[...]

PR #1439, opened 2018-04-06
Auto-generate the version_.d file from DUB

wilzbach:
It's really annoying that we have to `version_.d` file standing showing up in the `git status`.
AFAICT this is because the dub build can't generate a version - now it can :)

See also: https://github.com/dlang/dub/issues/1438


gedaiu:
@wilzbach as I understand from your code, you generate the version_ file after you tag a commit. 
This means that the new version number will not be inside that tagged commit.

wilzbach:
@gedaiu the idea is to _remove_ the version file_ from the source tree.
AFAICT dub was the last build solution that didn't generate the version_ file during/before the build, that's fixed now, which means we can delete the file now, because it will always be regenerated on every build.

I think it's reasonable to assume that dub gets tagged first and then built.
That's the case for the new auto-deployment pipeline for Travis and also the case for the build of the official release binaries:

https://gist.github.com/MartinNowak/a471fe7ddbfeef205cdf04c93a94c6d0#file-release-sh-L86

gedaiu:
@wilzbach I understand, and what you propose might work well with using dub as an executable. My problem is with using dub as a library. I use it with `trial` and I want to be able to know the right dub version. When you add a tag, code.dlang.com will fetch the code with the wrong version.

wilzbach:
> I want to be able to know the right dub version. When you add a tag, code.dlang.com will fetch the code with the wrong version.

I'm not sure I understand.
If you use dub as a library, it will detect its version based on the folder it's stored in, e.g. `~/.dub/packages/dub-1.10.0`
How can this be the wrong version? It will be exactly the version you use in your program...

gedaiu:
@wilzbach I looked again and the script should work.. sorry for the confusion :) :+1:

Issue #1447, opened 2018-04-17
Support for more variables in the build settings


rtbo:
Variables such as `$DUB_ARCH` are only supported in the commands environments.

I'd like to turn 
```json
    "lflags-posix-x86": [ "-L$PACKAGE_DIR/lib/posix-x86" ],
    "lflags-posix-x86_64": [ "-L$PACKAGE_DIR/lib/posix-x86_64" ],
```
into
```json
    "lflags-posix": [ "-L$PACKAGE_DIR/lib/posix-$ARCH" ],
```

PR #1450, opened 2018-04-19
Build settings vars

rtbo:
Fixing issue #1447.

variables:

name | values
--- | ---
`$ARCH` | `"x86"`, `"x86_64"`
`$PLATFORM` | `"linux"`, `"windows"`, ...
`$PLATFORM_POSIX` | `"posix"`, `"windows"`, ...
`$BUILD_TYPE` | `"debug"`, `"release"`, ...

Don't hesitate to ask for more.

I'm not sure whether 3c7a56be6f6e66959d45393110f7ff5d3ab6e8fc is the right thing to do.
Another possibililty is to bring `GeneratorSettings` to `BuildSettingsTemplate.getPlatformSettings` in order to resolve variables here too.


wilzbach:
BTW you might be interested in https://github.com/dlang/dub/pull/1392 too

wilzbach:
Sorry that this got so lost in the queue.
We have since then adopted the dmd/druntime/phobos changelog format. Would you mind adding a short entry, s.t. it will should up in the next changelog?

See: https://github.com/dlang/dub/tree/master/changelog

I don't see anything else blocking this though of course it wouldn't hurt to have another pair of eyes on this.

WebFreak001:
this will definitely require at least a minor version bump, some public APIs break with this. Though I guess these aren't very commonly used APIs.

@wilzbach @s-ludwig should we add automatic tests for the API with common use cases so functions don't break and it's all documented by example? I think it would make a lot of sense.

wilzbach:
> this will definitely require at least a minor version bump, 

Yeah the next release should be 1.11

> Is there anything blocking the merge?

I don't see much. I was waiting for input from Martin or Sönke, but I think we can move forward here.
Maybe
- a PR to the dub-registry, s.t. we ensure that the changes will be documented too?
- Squash all commits

rtbo:
All done :-)

wilzbach:
Thanks!

PR #1459, opened 2018-05-05
Add a "customCachePaths" configuration field

### LINKED ISSUE #838: Please make dub find & use packages installed to system locations
This is a follow-up of issue #811
At time, it is impossible to properly package dub packages in Linux distributions.
In Linux distros, we need to have all stuff required to build a software present in the distribution archives, this includes dub packages which other software depends on.
Unfortunately, dub does not search for matching packages in `/usr/include/d`.
So, ideally dub would allow some kind of `dub -v --allow-registries=local build` command to build using only locally installed packages.
So, I am proposing:
- Install dub packages locally in the file system under a standard path
- The path could e.g. be `/usr/include/d/{package}/{version}/` or `/usr/share/dlang/{package}/{version}/` (the former if the dub package can also be used without dub without changes, since that directory seems to be used by non-dub D software too)
- When resolving dependencies, have dub prefer these locally installed packages. This requires to stat once to see if the expected package name is present, and then one quick search to find the highest matching version. This should be really quick, so it won't hurt performance.
- Build with the packages found locally
- Optional, but ideal: Use the shared/static libraries built from the sources which have been installed into a library path (e.g. `/usr/lib/<triplet>/libfoobar.so`) to link against, instead of recompiling the package that has already been pre-built.

Not having a mechanism like this creates real problems when packaging any dub-using D software for Linux distributions, see https://github.com/gnunn1/terminix/issues/304 for an example.
Would be great to have this fixed, in order to be able to build D code in distributions without crude hacks.

s-ludwig:
Allows defining additional paths that contain packages in subfolders with the pattern "(name)-(version)/(name)/" by defining a "customCachePaths" field in /etc/dub/settings.json or ~/.dub/settings.json.

Fixes #838.


wilzbach:
CC @ximion

source/dub/dub.d
```diff
@@ -188,7 +188,12 @@ class Dub {
 			ps ~= defaultPackageSuppliers();
 
 		m_packageSuppliers = ps;
-		m_packageManager = new PackageManager(m_dirs.localRepository, m_dirs.systemSettings);
+		m_packageManager = new PackageManager(m_dirs.localRepository, m_dirs.systemSettings, true);
```
wilzbach: Could be a flag: `RefreshPackages.Yes`

test/issue838-custom-cache-paths.sh
```diff
@@ -0,0 +1,14 @@
...
+
+mkdir $CURR_DIR/../etc && mkdir $CURR_DIR/../etc/dub || true
+echo "{\"customCachePaths\": [\"$CURR_DIR/issue838-custom-cache-paths/cache\"]}" > $CONFIG_FILE
+
+if ! { $DUB build --root=issue838-custom-cache-paths --skip-registry=all; }; then
+	rm $CONFIG_FILE
+	die $LINENO 'Failed to build package with custom cache path for dependencies.'
+fi
+
+rm $CONFIG_FILE
```
wilzbach: Both `rm`s could be combined with `trap 'rm $CONFIG_FILE'  EXIT`

source/dub/packagemanager.d
```diff
@@ -71,14 +72,30 @@ class PackageManager {
...
+
+		Cache paths have the same structure as the default cache paths, such as
+		".dub/packages/".
+	*/
+	@property void customCachePaths(NativePath[] custom_cache_paths)
+	{
+		import std.algorithm.iteration : map;
+		import std.array : array;
+
+		m_repositories.length = LocalPackageType.max+1;
```
wilzbach: Might be worth pointing out that this removes all previously set custom cache paths (though it gets kinda implied by "set").

wilzbach:
This fails on the CI:

```
ERROR] /home/travis/build/dlang/dub/test/issue838-custom-cache-paths.sh:11 Failed to build package with custom cache path for dependencies.
[ERROR] Script failure.
```

Maybe the `$CURR_DIR/../etc` trick doesn't work?
Though dub does:

```
thisExePath.parentPath ~ "../etc/dub/settings.json
```

source/dub/dub.d
s-ludwig: I'll actually remove that from the diff, as that was left over from an intermediate version. But, I agree that a concrete type is preferable here.

ximion:
Looks good to me!
By using `/usr/include/d/` as package path and making projects using Meson or plain Makefiles install their data into `/usr/include/d/(name)-(version)/(name)/` (and put a dub.json in there as well), we can also easily achieve some basic level of Meson/Make --> dub interoperability[1], which would be really neat! (and make testing the new dub feature a bit easier)

[1]: This would likely mean generating a new dub.json file though, because Meson would expect the project's source code to be directly in `/usr/include/d/(name)-(version)/(name)/`, adding `/usr/include/d/(name)-(version)` as include directory...

WebFreak001:
this should get a changelog entry imo

s-ludwig:
@wilzbach, what's the plan with change logs? Is CHANGELOG.md deprecated now, and shouldn't entries in the form of ".dd" files at least have their PR number prefixed, so that it's possible to find them?

I committed a change to CHANGELOG.md now, because I didn't realize there was an update already, but I'll just reset that.

wilzbach:
> what's the plan with change logs?

I think the general idea is to include it with the release notes on dlang.org


>  Is CHANGELOG.md deprecated now

I think @MartinNowak mentioned that he already wrote a little tool that adds the changelog entries to the changelog.md file

> and shouldn't entries in the form of ".dd" files at least have their PR number prefixed, so that it's possible to find them?

I wasn't sure about this as generally we haven't done this for dmd/druntime/phobos changelog entries. Maybe a "learn more" link at the end of the entry with a link to the PR might work and would be clickable?
(the title can't contain links as it's already one)

> I committed a change to CHANGELOG.md now, because I didn't realize there was an update already, but I'll just reset that.

Sorry, just wanted to speed-up things.

MartinNowak:
Cool stuff thanks for the work.
The term cache is slightly confusing here (also for `~/.dub/packages`), would be nice to find a better name for this concept.
Maybe sth. along the lines of package search paths? Repository would be good as well, but is already used by VCSs.

PR #1460, opened 2018-05-06
Split-up package suppliers into individual modules

wilzbach:
Just a dumb split-up into individual modules with a bit of local, selective imports.


source/dub/packagesuppliers/isupplier.d
```diff
@@ -0,0 +1,54 @@
+module dub.packagesuppliers.isupplier;
+
+public import dub.dependency : Dependency, Version;
+public import dub.internal.vibecompat.core.file : NativePath;
```
s-ludwig: Should be `dub.internal.vibecompat.core.inet.path : NativePath, toNativeString;`. Should be noted that this is the first use of public imports (I used to public import all types necessary for the public API in a module, but stopped doing over time so for no real reason). The `toNativeString` might also be cleaner to import directly in `filesystemsupplier`.

s-ludwig:
I would suggest a modified naming scheme to be consistent with the other dub.* packages:
- `isupplier` -> `packagesupplier` (I'd probably also put the definitions in `utils` there to avoid creating a useless public module (otherwise they should go into `dub.internal`))
- `filesystemsupplier` -> `filesystem` (and so on)

wilzbach:
> I would suggest a modified naming scheme to be consistent with the other dub.* packages:

Didn't know about this naming scheme. Done.

> I'd probably also put the definitions in utils there to avoid creating a useless public module (otherwise they should go into dub.internal)

Well it's undocumented, so shouldn't appear as part of the docs.
Not sure whether using `dub.internal.packagesuppliers.d` is better, `dub.packagesuppliers.utils` is at least the place I would be looking for these methods.

s-ludwig:
And why not put the definitions in packagesupplier.d? They don't create any cyclic dependencies, so that should be fine.

wilzbach:
Sorry, but what are you referring to with definitions?

s-ludwig:
I mean the contents of utils.d (now just `getBestPackage`).

WebFreak001:
uh shouldn't this go through a deprecation phase? this isn't in the `internal` package and I think I even might have used it somewhere in workspace-d.

wilzbach:
> I mean the contents of utils.d (now just getBestPackage).

Ok done.

> uh shouldn't this go through a deprecation phase?

Add the dub.packagesupplier.d back with a deprecation.

build-files.txt
```diff
@@ -7,6 +7,12 @@ source/dub/dub.d
 source/dub/init.d
 source/dub/packagemanager.d
 source/dub/packagesupplier.d
+source/dub/packagesuppliers/package.d
+source/dub/packagesuppliers/fallback.d
+source/dub/packagesuppliers/filesystem.d
+source/dub/packagesuppliers/packagesupplier.d
+source/dub/packagesuppliers/mvn.d
```
s-ludwig: I think this is really the last issue, but I'd really write that out as "maven", so that the contained class name is always the concatenation of module name and "PackageSupplier". There is also no real gain in saving characters here.

wilzbach: Sure, I just used `mvn` because that's the name of the CLI tool. Renamed to "maven".

s-ludwig:
Okay, thanks!

PR #1475, opened 2018-05-19
Fix misleading message when package description is invalid

dcarp:
If the package description file is incomplete or contains errors, the produced error message can be misleading: "No package found in current working directory.", when in fact the package file is there but invalid.


dcarp:
rebase to trigger rebuild

Issue #1478, opened 2018-05-22
Migration of package dir fails in Wine (dub 1.7 -> 1.8/1.9); xcopy returns error code 4. Workaround: delete old package dir


SimonN:
### System information
* Arch Linux 64-bit
* Wine 3.7
* dmd 2.080
* dub.exe 1.9.0 (the one that ships with Windows dmd 2.080. But also dub.exe 1.8.0 and current master 0863be36c4c488fce84d392cf72e368a9016ec13 have this issue)

### Bug Description

After you migrate your dub 1.7.0 (for Windows, installed on Linux via Wine) to 1.9.0 (for Windows, again on Linux), then try to build with `wine dub`, dub will migrate the package directory to `Local Settings\Application Data\dub\packages`. But `xcopy` fails with exit code 4 (see Log 1 at bottom of this post). This prevents even the most trivial dub projects from being built. `wine dub` refuses to compile anything.

I improved the error message by compiling my own dub for Windows, see Log 2. (I will submit a PR with this.)

**Workaround:** Delete the old package cache dir at `.wine/drive_c/users/yourname/Application Data/dub/packages` by hand. Have dub fetch everything anew.

### How to reproduce?

1. ~~Install (through Wine) dub 1.7.0 (or earlier) for Windows. (We're on Linux.)~~
2. ~~Use that dub 1.7.0 to build unrelated projects, thus generate a filled package cache directory at `Application Data\dub\packages`.~~ It's enough to create an empty file at `.wine/drive_c/users/yourname/"Application Data"/dub/packages/a.txt`.
3. Install (through Wine) dub 1.9.0 for Windows.
4. Create a minimal dub-using D project by `dub init`. (I used native Linux dub but shouldn't matter.)
5. Build that trivial dub-using D project with dub through Wine: `wine dub`
6. Observe Log 1 (instead of expected behavior)

Optionally, improve dub's error message to see exactly what command fails, observing Log 2.

### Expected Behavior

Projects build with dub 1.9.0 in Wine even when package dir exists at `.wine/drive_c/users/yourname/Application Data/dub/packages`.

(Error message should be better; if it fails at all, prefer Log 2 over Log 1. I will submit a PR for this better error message, but not for the failing call to `xcopy` in Wine.)

### Logs

Log 1:

```
object.Exception@source\dub\internal\utils.d(201): Command failed with exit code 4
```

Log 2:

```
[...]

PR #1479, opened 2018-05-22
Report failed external commands explicitly

SimonN:
Improve the error on failed external commands: dub will now print the exact command that failed.

This was useful to troubleshoot #1478.

Before:
```
object.Exception@source\dub\internal\utils.d(201): Command failed with exit code 4
```
After:
```
object.Exception@source\dub\internal\utils.d(201): Command ``xcopy /s /e /y C:\users\simon\Application Data\dub\packages C:\users\simon\Local Settings\Application Data\dub\packages > NUL'' failed with exit code 4
```

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -198,7 +198,8 @@ void runCommands(in string[] commands, string[string] env = null)
 		Pid pid;
 		pid = spawnShell(cmd, stdin, childStdout, childStderr, env, config);
 		auto exitcode = pid.wait();
-		enforce(exitcode == 0, "Command failed with exit code "~to!string(exitcode));
+		enforce(exitcode == 0, "Command failed with exit code "
+			~ to!string(exitcode) ~ ": " ~ cmd);
 	}
 }
 


source/dub/internal/utils.d
```diff
@@ -198,7 +198,8 @@ void runCommands(in string[] commands, string[string] env = null)
 		Pid pid;
 		pid = spawnShell(cmd, stdin, childStdout, childStderr, env, config);
 		auto exitcode = pid.wait();
-		enforce(exitcode == 0, "Command failed with exit code "~to!string(exitcode));
+		enforce(exitcode == 0, "Command ``" ~ cmd
+			~ "'' failed with exit code "~to!string(exitcode));
 	}
```
wilzbach: Why two backticks and then two single quotes?
Usually one is enough and typically you end the quotation with the same character, no?

SimonN: Right, (two backticks, then two single quotes) is uncommon style. It's found in old GNU manuals and in AsciiDoc. I've merely tried to be distinct from any quotes within the shell command itself.

Maybe the best is to avoid quoting altogether, and rewrite the error as such?
```
Command failed with exit code 4: xcopy /s /e /y C:\... C:\...
```

I'll re-push with that quoteless style.

I'd also be happy with any other quoting style, feel free to pick one, although dub output generally avoids quotes.

SimonN:
Re-pushed with a writing style that avoids quotes. This may be best because a shell command can contain various quotes itself.

The original error is now reported as:
```
object.Exception@source\dub\internal\utils.d(201): Command failed with exit code 4: xcopy /s /e /y C:\users\simon\Application Data\dub\packages C:\users\simon\Local Settings\Application Data\dub\packages > NUL
```

SimonN:
Thanks for the super-quick review & merge!

PR #1481, opened 2018-05-23
Fix #1478 - Quote paths passed to xcopy

### LINKED ISSUE #1478: Migration of package dir fails in Wine (dub 1.7 -> 1.8/1.9); xcopy returns error code 4. Workaround: delete old package dir

SimonN:
This fixes two lines:
**(1)** migrate dub package dir,
**(2)** copy DDoc files.

I have tested (1) on Wine and on a friend's native Windows 8.1. Both times, the fix works as expected: `xcopy` finds and copies the old package directory even when source or target dir contain spaces.

**Reviewers:** I have not tested (2), neither on Wine nor on Windows. Please test (2)!

---

On Windows, `cd` accepts unquoted paths, but for any calls to `xcopy`,
paths arguments must be double-quoted. This patch surrounds all paths
with double-quotes before passing them to `xcopy`.

This fixes #1478: On Wine (and presumably Windows XP), migrating the dub
package directory from dub 1.7 to dub 1.8/1.9 via `xcopy` failed an
enforce in dub (because `xcopy` quit with nonzero exit code) and thus
prevented even the most simple dub projects from being built.

D WYSIWYG strings (backtick-surrounded strings) allow all characters
unescaped, including double quotes and backslashes.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -311,7 +311,7 @@ class Dub {
             mkdirRecurse(localDirPath);
         }
 
-        runCommand("xcopy /s /e /y " ~ roamingDirPath ~ " " ~ localDirPath ~ " > NUL");
+        runCommand(`xcopy /s /e /y "` ~ roamingDirPath ~ `" "` ~ localDirPath ~ `" > NUL`);
         rmdirRecurse(roamingDirPath);
 	}
 
@@ -1278,7 +1278,7 @@ class Dub {
 
 		if (!run) {
 			// TODO: ddox should copy those files itself
-			version(Windows) runCommand("xcopy /S /D "~tool_path~"public\\* docs\\");
+			version(Windows) runCommand(`xcopy /S /D "`~tool_path~`public\*" docs\`);
 			else runCommand("rsync -ru '"~tool_path~"public/' docs/");
 		}
 	}


SimonN:
What would be a test case for (2) from the first post? I'm willing to test (2) manually, then.

This PR is hard to cover with automated tests -- on most Windows systems, the standard paths don't have any spaces to enquote.

andre2007:
Could you rebase your commit to restart the test pipeline?

SimonN:
Done! :-)

andre2007:
Is the xcopy command for ddox correct?
It has following value
```xcopy /S /D "C:\Users\pc\AppData\Local\dub\packages\ddox-0.16.12\ddox\"public\* docs\```

I would expect
```xcopy /S /D "C:\Users\pc\AppData\Local\dub\packages\ddox-0.16.12\ddox\public\*" docs\```

**Comment by andre2007** [comment]
At the moment it is quite hard to test the ddox xcopy due to bug described here https://github.com/dlang/dub/issues/1512

I got around this bug by passing default architecture to determinePlatform call. This might be the solution for the bug but I have to do more tests here.

**Comment by SimonN** [comment]
Good find. Even if both quoting methods behave correctly, we should use the clearest.

I'll re-check exactly how quoted/unquoted `*` behaves. My guts suggest that `"a b c"/public/*` and `"a b c/public/*"` behave the same, because the `*`is passed as-is to `xcopy` (not resolved like a Unix shell glob to multiple arguments) and both quotings encompass all spaces in the argument.

**Comment by SimonN** [comment]
Both quotation methods indeed behave the same (copy all files in `public/`) and I like yours better, too. I've amended and re-pushed my commit, it's still on top of master.

**Comment by andre2007** [comment]
The failing tests are caused by a timeout, could you rebase to get the tests green?

**Comment by andre2007** [comment]
Thanks. At the moment ddox generation cannot be tested on windows, due to the OPTLINK bug. I created PR https://github.com/dlang/dub/pull/1588 which should solve the issue. Therefore your pr can be manually tested.

**Comment by andre2007** [comment]
Manually validated the changes. Looks good to me.
These are the new generated strings:
```
xcopy /s /e /y "C:\Users\pc\AppData\Roaming\dub\packages" "C:\Users\pc\AppData\Local\dub\packages" > NUL
```
```
xcopy /S /D "C:\Users\pc\AppData\Local\dub\packages\ddox-0.16.12\ddox\public\*" docs\
```

SimonN:
Thanks for the check and the merge!

PR #1488, opened 2018-06-08
Add skipRegistry option to settings.json

andre2007:
In a company scenario you want to specify that only the company internal dub registry is used and not any public registry (security issues...). While it was already possible to define registryUrls in settings.json, disabling the public dub registries (code.dlang.org, ....) wasn't possible. This pull request allows to set the skipRegistry option in settings.json.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -167,6 +167,46 @@ class Dub {
 
 		init();
 
+		if (skip_registry == SkipPackageSuppliers.none)
+			m_packageSuppliers = getPackageSuppliers(additional_package_suppliers);
+		else
+			m_packageSuppliers = getPackageSuppliers(additional_package_suppliers, skip_registry);
+
+		m_packageManager = new PackageManager(m_dirs.localRepository, m_dirs.systemSettings);
+
+		auto ccps = m_config.customCachePaths;
+		if (ccps.length)
+			m_packageManager.customCachePaths = ccps;
+
+		updatePackageSearchPath();
+	}
+
+	unittest
+	{
+		scope (exit) environment.remove("DUB_REGISTRY");
+		auto dub = new Dub(".", null, SkipPackageSuppliers.configured);
+		assert(dub.m_packageSuppliers.length == 0);
+		environment["DUB_REGISTRY"] = "http://example.com/";
+		dub = new Dub(".", null, SkipPackageSuppliers.configured);
+		assert(dub.m_packageSuppliers.length == 1);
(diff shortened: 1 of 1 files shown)


andre2007:
That's true. But I wonder whether adding new constructor would introduce too much complexity. I have another suggestion. I extract the whole package supplier logic from the current constructor into a new function which has a import argument the skipPackageSupplier value and returns the list of determined package suppliers. In the existing constructor I can then decide whether to call this function with the skip package supplier value from the constructor import argument or the default value from settings.json. This would also ease writing a unit test for the new functionality. What do you think?

WebFreak001:
0 cost testable abstraction which makes the code more readable always sounds good

andre2007:
I did the changes and added a unittest. Please have a look.

andre2007:
@WebFreak001 could you have a second look? Thanks

source/dub/dub.d
```diff
@@ -167,6 +167,47 @@ class Dub {
...
+	/** Get the list of package suppliers.
+
+		Params:
+			additional_package_suppliers = A list of package suppliers to try
+				before the suppliers found in the configurations files and the
+				`defaultPackageSuppliers`.
+			skip_registry = Can be used to skip using the configured package
+				suppliers, as well as the default suppliers.
+	*/
+	private PackageSupplier[] getPackageSuppliers(PackageSupplier[] additional_package_suppliers, SkipPackageSuppliers skip_registry)
```
WebFreak001: if this is well documented and unittested I think it would make sense to make this public

WebFreak001:
looks good, can you just make that a public symbol (and the ditto function), it seems like a function that could be useful if used as a library. (For example show suppliers in IDE)

source/dub/dub.d
andre2007: done :)

source/dub/dub.d
```diff
@@ -167,6 +167,47 @@ class Dub {
...
+	}
+
+	unittest
+	{
+		scope (exit) environment.remove("DUB_REGISTRY");
+		auto dub = new Dub(".", null, SkipPackageSuppliers.configured);
+		assert(dub.m_packageSuppliers.length == 0);
+		environment["DUB_REGISTRY"] = "http://example.com/";
+		dub = new Dub(".", null, SkipPackageSuppliers.configured);
+		logInfo("%s", dub.m_packageSuppliers);
```
wilzbach: I would probably remove this to avoid polluting the output when `dub test` is run.

andre2007: fixed :)

PR #1490, opened 2018-06-11
Colored output

gdelazzari:
This PR adds colored text output to Dub, in order to (in my opinion) enhance the user experience and make the output clearer and easier to navigate. See the first post of this forum thread for the motivations regarding this proposal: https://forum.dlang.org/thread/vefplpvwvrcbednsufli@forum.dlang.org

The change has been implemented by substituting the "old" dub.internal.vibecompat.core.log module with a new dub.logging module that provides the same functionality (same function names etc...) but also handles colored text and the concept of a colored "tag" at the beginning of each logging line. Non-TTY output is handled correctly, to avoid ASCII escape sequences when piping stdout/stderr to something else. No functionality of the program has been changed, just the output look. Also, the library colorize (https://github.com/yamadapc/d-colorize) has been imported (by copying the source code in dub/internal) as an helper to write formatted text to the console in a cross-platform manner.

Here are a couple of screenshots of the final result while running clean & build commands.

build-from-scratch
build-error

More screenshots of other commands here https://imgur.com/a/VgIGPBy

Hope this is appreciated and doesn't create issues. As I said, IMHO the output is way easier to navigate and understand now.


source/dub/commandline.d
```diff
@@ -557,12 +561,12 @@ class InitCommand : Command {
...
 				m_templateType = free_args[0];
 				free_args = free_args[1 .. $];
-				logInfo("Deprecated use of init type. Use --type=[vibe.d | deimos | minimal] in future.");
+				logWarn("Deprecated use of init type. Use --type=[vibe.d | deimos | minimal] in future.");
 			}
 		}
+
 		dub.createEmptyPackage(NativePath(dir), free_args, m_templateType, m_format, &depCallback);
 
-		logInfo("Package successfully created in %s", dir.length ? dir : ".");
```
WebFreak001: why remove this log line?

source/dub/commandline.d
```diff
@@ -1239,10 +1243,10 @@ class FetchCommand : FetchRemoveCommand {
...
 			try {
 				dub.fetch(name, Dependency(">=0.0.0"), location, fetchOpts);
-				logInfo(
-					"Please note that you need to use `dub run <pkgname>` " ~
-					"or add it to dependencies of your package to actually use/run it. " ~
-					"dub does not do actual installation of packages outside of its own ecosystem.");
+				logInfo("Finished", Color.green, "%s fetched", name.color(Mode.bold));
+				logInfo("Please note that you need to use `dub run <pkgname>` or add it");
+				logInfo("to dependencies of your package to actually use/run it.");
+				logInfo("Dub does not do actual installation of packages outside of its own ecosystem.");
```
WebFreak001: I don't know if it's a good idea to separate the lines, if the window is small it will wrap each one individually

source/dub/dub.d
```diff
@@ -509,19 +509,20 @@ class Dub {
...
 					continue;
 				}
 				auto sver = m_project.selections.getSelectedVersion(basename);
 				if (!sver.path.empty) continue;
 				if (ver.version_ <= sver.version_) continue;
-				logInfo("Package %s can be upgraded from %s to %s.",
-					basename, sver, ver);
+        logInfo("Upgrade", Color.yellow,
+          "%s can be upgraded %s -> %s",
+					basename.color(Mode.bold), sver, ver);
```
WebFreak001: you mixed in spaces to indent here

source/dub/generators/build.d
```diff
@@ -47,8 +47,12 @@ class BuildGenerator : ProjectGenerator {
 	{
 		scope (exit) cleanupTemporaries();
 
-		logInfo("Performing \"%s\" build using %s for %-(%s, %).",
-			settings.buildType, settings.platform.compilerBinary, settings.platform.architecture);
+    logInfo("Starting", Color.green,
+      "Performing %s build using %s for %-(%s, %)",
+			settings.buildType.color(Color.magenta),
```
WebFreak001: mixed in spaces to indent

source/dub/generators/build.d
```diff
@@ -92,7 +96,9 @@ class BuildGenerator : ProjectGenerator {
 			buildTargetRec(m_project.rootPackage.name);
 
 			if (any_cached) {
-				logInfo("To force a rebuild of up-to-date targets, run again with --force.");
+				logInfo("Finished", Color.green,
+          "To force a rebuild of up-to-date targets, run again with --force"
+        );
```
WebFreak001: mixed in spaces to indent

source/dub/generators/build.d
```diff
@@ -524,7 +530,7 @@ class BuildGenerator : ProjectGenerator {
 				if (!exe_path_string.startsWith(".") && (exe_path_string.length < 2 || exe_path_string[1] != ':'))
 					exe_path_string = ".\\" ~ exe_path_string;
 			}
-			logInfo("Running %s %s", exe_path_string, run_args.join(" "));
+			logInfo("Running", Color.green, "%s %s", exe_path_string, run_args.join(" "));
```
WebFreak001: can you make sure before running the console gets reset?

source/dub/internal/colorize/colors.d
```diff
@@ -0,0 +1,185 @@
...
+ * License: Licensed under the MIT license. See LICENSE for more information
+ * Version: 1.0.2
+ */
+module dub.internal.colorize.colors;
+
+import std.string : format;
+
+private template color_type(int offset)
+{
+  static enum type : int
```
WebFreak001: You used different formatting than what all other code is using, please adjust.

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
...
+module dub.internal.colorize.cwrite;
+
+import std.stdio : File, stdout;
+
+import dub.internal.colorize.winterm;
+
+/// Coloured write.
+void cwrite(T...)(T args) if (!is(T[0] : File))
+{
+  stdout.cwrite(args);
```
WebFreak001: Please respect previous coding & formatting style.

source/dub/internal/vibecompat/core/log.d
```diff
@@ -1,99 +1,14 @@
-/**
-	Central logging facility for vibe.
+/**	+deprecated("Please use dub.logging")
+	This module contained logging functions, but has been replaced by dub.logging
+  which provides colored output and other facilities.
 
-	Copyright: © 2012 rejectedsoftware e.K.
+  Copyright: © 2012 rejectedsoftware e.K., © 2018 Giacomo De Lazzari
```
WebFreak001: You mixed in spaces

source/dub/logging.d
```diff
@@ -0,0 +1,383 @@
...
+  there for a list of possible values.
+*/
+public alias Mode = mode;
+
+/// The tag width in chars, defined as a constant here
+private const int TAG_WIDTH = 12;
+
+/// Possible log levels supported
+enum LogLevel {
+  debug_,
```
WebFreak001: Respect other coding & formatting style

source/dub/packagesuppliers/maven.d
```diff
@@ -11,8 +11,8 @@ import dub.packagesuppliers.packagesupplier;
 class MavenRegistryPackageSupplier : PackageSupplier {
 	import dub.internal.utils : download, HTTPStatusException;
 	import dub.internal.vibecompat.data.json : serializeToJson;
-	import dub.internal.vibecompat.core.log;
 	import dub.internal.vibecompat.inet.url : URL;
+  import dub.logging;
```
WebFreak001: spaces

source/dub/packagesuppliers/registry.d
```diff
@@ -12,9 +12,9 @@ package enum PackagesPath = "packages";
 */
 class RegistryPackageSupplier : PackageSupplier {
 	import dub.internal.utils : download, HTTPStatusException;
-	import dub.internal.vibecompat.core.log;
 	import dub.internal.vibecompat.data.json : parseJson, parseJsonString, serializeToJson;
 	import dub.internal.vibecompat.inet.url : URL;
+  import dub.logging;
```
WebFreak001: spaces

andre2007:
If possible please also check on Windows Powershell. The good old command prompt is deprecated by Microsoft and Powershell is the default console on windows now. While doing a quick check it seems colorize has a background issue. For this code 

```
import colorize : fg, color, cwriteln, cwritefln;

void main()
{
  cwriteln("This is blue".color(fg.blue));

  auto c = "red";
  cwritefln("This is %s".color(c), c);
}
```
it has this output 
image

(it seems it assumes black background on windows (cmd prompt) while PowerShell has  a blue background.

source/dub/commandline.d
gdelazzari: I removed it since there's already a logInfo at the end of dub.createEmptyPackage, the previous output when running "dub init my-project" ended something like

```
Successfully initialized empty project in /home/user/things/my-project
Package successfully created in my-project
```

I found that repetitive so I thought about removing one of the two. Maybe that isn't the one to be removed, but the one in dub.createEmptyPackage?

source/dub/commandline.d
gdelazzari: You are right. I think I made that change because now, since logInfo("...") lines are indented by 12 chars (or whatever is configured as tag width in the logging module), if the terminal automatically wraps them, they look a bit messier since the wrapped part doesn't obviously get indented as well.
I don't have any problem restoring that text to how it was before if, in your opinion, that's a better idea. Let me know.

WebFreak001: maybe the log function could wrap with indentation based on the terminal width, though that would be a much more complex function

source/dub/dub.d
gdelazzari: In response to this and any other "mixed in spaces to indent" etc... 
I'm really sorry, I didn't check my editor before and I'm totally used to spaces to indent. My editor had 2 chars as tab width so I didn't spot any issue until I saw the code here on GitHub. I will fix everything.

gdelazzari:
@andre2007 I just noticed that too on Windows. Unfortunately I almost never use Windows so I didn't think about testing also the PowerShell. I guess the issue is that colorize changes the background even when it's not needed, I'll see if I can change its behaviour.

Also I noticed I'll need to fix some tests that check the terminal output and search for certain strings. That is TODO as well.

@BBasile disabling colors is just a matter of calling a function in the new logging module so I guess I could simply add a new command line option that disables them if needed, I mean it should be trivial

test/feat663-search.sh
```diff
@@ -10,6 +10,6 @@ fi
 if ! OUTPUT=$(${DUB} search dub -v 2>&1); then
     die $LINENO '`dub search dub` failed' "$OUTPUT"
 fi
-if ! grep -q '^dub (.*)\s'<<<"$OUTPUT"; then
-    die $LINENO '`grep -q '"'"'^dub (.*)\s'"'"'` failed' "$OUTPUT"
+if ! grep -q '^\s*dub (.*)\s'<<<"$OUTPUT"; then
+    die $LINENO '`grep -q '"'"'^\s*dub (.*)\s'"'"'` failed' "$OUTPUT"
```
wilzbach: These adaptions shouldn't be necessary as it should be possible to detect that no terminal is available here. No?

source/dub/logging.d
```diff
@@ -0,0 +1,383 @@
...
+  Params:
+    str = The string to color
+    color = The color to apply
+    mode = An optional mode, such as bold/underline/etc...
+*/
+string color(const string str, const Color color, const Mode mode = Mode.init)
+{
+  import dub.internal.colorize;
+
+  if (_printColors == true)
```
wilzbach: This is a weird style, why not `if (_printColors)` ?

source/dub/logging.d
```diff
@@ -0,0 +1,383 @@
...
+
+    if (output.isOpen)
+    {
+      output.cwrite(result, "\n");
+      output.flush();
+    }
+  }
+  catch (Exception e)
+  {
+    debug assert(false, e.msg);
```
wilzbach: This is pretty dangerous as it means that by default all exceptions will be silently caught.

test/feat663-search.sh
gdelazzari: Sure, the output isn't colored, but I added two trailing spaces (like when you do `dub list`) to every line in the output of `dub search`. There's no reason why `dub lists` prints something like 

```
Packages known to dub:
  package-a 0.1.2
  package-b 0.2.4
```

while the results of `dub search` don't appear formatted the same way, IMHO. The regex change is to accomodate for this. Hope you understand.

source/dub/logging.d
gdelazzari: I agree, actually that code comes from dub.internal.vibecompat.core.log (100% copy pasted, I mean the try-catch) and I didn't take the time to find a better solution. But, as the comment that was there before says... what can we do in that case? :)

https://github.com/dlang/dub/blob/master/source/dub/internal/vibecompat/core/log.d#L84

I mean if we fail to print something I don't think we should throw an exception, in my opinion it makes sense for `log` to be `nothrow`, you shouldn't worry about exceptions when logging.

Anyway, I'm all ears for suggestions

source/dub/logging.d
gdelazzari: Right, not sure how that line ended up being that way

source/dub/generators/build.d
gdelazzari: Uhm, that shouldn't be necessary, does it? The ASCII escape sequences are all "terminated" after being used. I mean I can't see a way for that not being the case since colored text is always generated with the `color()` method that always puts an escape sequence at the beginning and one at the end that restores the state (on Posix). On Windows, the `cwrite()` method that is used to print colored text, as soon as the `WinTermEmulation` struct gets destroyed, resets everything as it was before.

Or did you mean something else that I'm not aware of?

gdelazzari:
Okay, I fixed the indentation issue (I'm really sorry for that), fixed the PowerShell background color problem that has been reported, fixed the failing tests and added a common option to disable color output in any case (--no-colors) which appears with `dub --help` and gets its own description.

Let me know if other things are needed in order to merge this, besides the things that have been commented on the code for which I'm waiting directions regarding what to do.

Thanks for the review, I hope now it looks better.

Vild:
Could you add a way to force colors as well?
It is useful when the dub command is run in a buildsystem that output to a buffer before outputing to stdout.

Real life example: If you want colors from gcc when you use ninja you can use: `gcc -fdiagnostics-color=always [...]`.

source/dub/generators/build.d
WebFreak001: hm sure just wanted to make sure this was thought about

source/dub/commandline.d
WebFreak001: ah sure ok

gdelazzari:
I see on Travis that some old versions of dmd (and ldc/gdc) fail to compile because of colors.d in dub/internal/colorize. Is that a problem?
Also the test coverage is now lower, but how could I test the code given it's almost only code that prints to stdout/stderr?

Another thing I wanted to note is that now, even if the same function names were maintained (logInfo, logWarn, logError, etc...) their output has a different look, i.e. they always put a tag at the beginning (actually, there's logInfoNoTag if needed) so there may be some corner cases I didn't manage to test that print output which may not look as pretty as the stuff I've tested and adjusted. But I'm not sure about what to do, should I try every single code path that prints stuff and ensure everything is "pretty" enough, or given they're rare cases I can leave them there and, if someone discovers that something looks really ugly after the PR, we fix it as needed?

source/dub/commandline.d
```diff
@@ -136,6 +140,16 @@ int runDubCommandLine(string[] args)
 		else if (options.quiet) loglevel = LogLevel.warn;
 		else if (options.verror) loglevel = LogLevel.error;
 		setLogLevel(loglevel);
+
+		if (options.noColors && options.forceColors) {
+			logError("Incompatible options: --no-colors and --force-colors");
```
BBasile: i'd rather see `--color={on|off|force}`

- `on`: activated, however detect if DUB is piped
- `off`: deactivated.
- `force`: activated, don't care about piping (or whatever else is the semantic for that).

However i wont get mad if other do't agree.

gdelazzari: I agree the current implementation is not that great. What about `--colors={auto|on|off}`? With `auto` being the default

BBasile: yes that would be nicer.

gdelazzari:
>>but how could I test the code given it's almost only code that prints to stdout/stderr?
>
>You could force the output of colors and then grep for it or use something like hexdump?

I added a test as you said. However, when I wrote that message, I was referring to the code coverage, since the only bot that is still not happy is codecov/patch. What I meant is that I'm not sure how to increase the unittests code coverage, how am I supposed to test stuff that prints to stdout/stderr in `unittest { }` blocks?

source/dub/logging.d
```diff
@@ -0,0 +1,384 @@
...
+
+		if (output.isOpen)
+		{
+			output.cwrite(result, "\n");
+			output.flush();
+		}
+	}
+	catch (Exception e)
+	{
+		debug assert(false, e.msg);
```
wilzbach: Can we remove the `debug` here?

test/colored-output.sh
```diff
@@ -0,0 +1,17 @@
...
+${DUB} build --colors=off --compiler=${DC} 2>&1 | { ! grep -P '\e\[' -c; }
+
+# Test --colors=auto detecting no TTY
+${DUB} build --colors=auto --compiler=${DC} 2>&1 | { ! grep -P '\e\[' -c; }
+
+# Test no --colors= option defaulting to auto
+${DUB} build --compiler=${DC} 2>&1 | { ! grep -P '\e\[' -c; }
+
+# Test --colors=on enabling colors in any case
+${DUB} build --colors=on --compiler=${DC} 2>&1 | grep -P '\e\[' -c
```
wilzbach: FYI: you could use `strings` to generate readable/diffable string code.
E.g. sth. like this should work:

```sh
check()
{
    local out expected
    out=$("$1" 2>&1 | strings | tr -d "\n")
    expected="$2"
    if [ "$out" != "$expected" ] ; then
        echo "Expected: $expected"
        echo "Actual:   $out"
        exit 1
    fi
}
check "${DUB} build --colors=on --compiler=${DC}" "<output log>"
```

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
...
+void cwritef(Char, T...)(in Char[] fmt, T args) if (!is(T[0] : File))
+{
+	stdout.cwritef(fmt, args);
+}
+
+/// Coloured writefln.
+void cwritefln(Char, T...)(in Char[] fmt, T args)
+{
+	stdout.cwritef(fmt ~ "\n", args);
+}
```
wilzbach: Not used.

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
...
+
+/// Coloured writeln.
+void cwriteln(T...)(T args)
+{
+	// Most general instance
+	stdout.cwrite(args, '\n');
+}
```
wilzbach: Not used.

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
...
+void cwrite(T...)(T args) if (!is(T[0] : File))
+{
+	stdout.cwrite(args);
+}
+
+/// Coloured writef.
+void cwritef(Char, T...)(in Char[] fmt, T args) if (!is(T[0] : File))
+{
+	stdout.cwritef(fmt, args);
+}
```
wilzbach: Not used.

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
...
+	stdout.cwrite(args, '\n');
+}
+
+/// Coloured writef to a File.
+void cwritef(Char, A...)(File f, in Char[] fmt, A args)
+{
+	import std.string : format;
+	auto s = format(fmt, args);
+	f.cwrite(s);
+}
```
wilzbach: Not used.

source/dub/internal/colorize/cwrite.d
```diff
@@ -0,0 +1,74 @@
+/**
+ * Authors: ponce
+ * Date: July 28, 2014
+ * License: Licensed under the MIT license. See LICENSE for more information
+ * Version: 1.0.2
+ */
+module dub.internal.colorize.cwrite;
```
wilzbach: This entire module is only used once by `winterm`.
I think it would be best to move only the relevant code to `winterm` and delete the rest.

source/dub/internal/colorize/winterm.d
```diff
@@ -0,0 +1,191 @@
+/**
+ * Authors: ponce
+ * Date: July 28, 2014
+ * License: Licensed under the MIT license. See LICENSE for more information
+ * Version: 1.0.2
+ */
+module dub.internal.colorize.winterm;
```
wilzbach: Compared to what [DMD does](https://github.com/dlang/dmd/blob/master/src/dmd/console.d), this looks overly complicated.
If all this logic is really required for Windows, it would be great to have some explanation for it (and future readers).

source/dub/logging.d
```diff
@@ -0,0 +1,384 @@
...
+		else
+			_printColors = false;
+	}
+	else version (Posix)
+	{
+		import core.sys.posix.unistd;
+
+		if (!isatty(STDERR_FILENO) || !isatty(STDOUT_FILENO))
+			_printColors = false;
+	}
```
wilzbach: I think it would be best to have a special `inTerminal` method, s.t. this can be simplified to: `setLoggingColorsEnabled(!inTerminal());`

source/dub/logging.d
```diff
@@ -0,0 +1,384 @@
...
+	auto boldTag = false;
+	if (level >= LogLevel.warn)
+		boldTag = true;
+
+	try
+	{
+		string result = format(fmt, args);
+
+		if (hasTag)
+			result = tag.rightJustify(TAG_WIDTH, ' ').color(tagColor, boldTag ? Mode.bold : Mode.init) ~ " " ~ result;
```
wilzbach: Looks like at least two allocations could be avoided here:
- `formattedWrite` instead of `format`
- direct writes instead of `~`
(`color` + `rightJustify` could be concatenated ranges too).

source/dub/internal/colorize/colors.d
```diff
@@ -0,0 +1,185 @@
...
+ */
+
+string color(
+	const string str,
+	const fg c=fg.init,
+	const bg b=bg.init,
+	const mode m=mode.init
+) pure
+{
+	return format("\033[%d;%d;%dm%s\033[0m", m, c, b, str);
```
wilzbach: This could be a range and thus cut-down on a lot of allocations.

source/dub/internal/colorize/colors.d
```diff
@@ -0,0 +1,185 @@
...
+}
+
+alias colorHelper!bg background;
+alias colorHelper!fg foreground;
+alias colorHelper!mode style;
+
+alias background color;
+alias foreground color;
+alias style color;
+alias colorHelper color;
```
wilzbach: This `alias` style is deprecated. Use `alias background = colorHelper!bg`.

wilzbach:
BTW I'm not a huge fan of the explosion of `logInfoNoTag` and friends.
How about adding a simple `Tag` struct that wraps a string and its respective tag?
Then it should be possible to figure out everything at CT and we don't need so many overloads nor do we have this separation of the color and tag:

```d
logInfo("Finished".tag(Color.green), "%s fetched", name.color(Mode.bold));
logInfo("%s fetched", name.color(Mode.bold)); // no tag
```

source/dub/internal/colorize/cwrite.d
gdelazzari: I didn't get this one, do you mean that this module uses `winterm` only once? (i.e. not the other way around)? How would you propose to adjust the dub.internal.colorize subpackage? At this point it wouldn't even make a lot of sense to keep it under there, since some stuff will be removed (some sub modules too, right?)

Should I simply make the changes you proposed and leave everything there (but then it would seem like I "imported" d-colorize but changed the library from top to bottom, which would be a bit strange) or make a new module somewhere else? And how should that be called/where should it be?

source/dub/internal/colorize/winterm.d
gdelazzari: I agree the approach of d-colorize is indeed more complicated, also on Windows the performance will be a bit lower, since this module is doing this sort of emulation we see here. Maybe at this point it would make more sense to just bring in DMD's `console` module and ditch this mess? :smile:

wilzbach: Dmd's Code is well-tested and only a quarter of d-colorize, so this means a lot less maintenance cost. So I would be in favor if you don't mind throwing away a bit of your work.

source/dub/internal/colorize/cwrite.d
wilzbach: Oh I meant that winterm only calls one method from this module, so there's a lot of dead code that could simply be removed?

> Should I simply make the changes you proposed and leave everything there (but then it would seem like I "imported" d-colorize but changed the library from top to bottom, which would be a bit strange) 

Ah I didn't know that you simply copied the library in. I don't think that modifying this code is a problem, especially because I doubt that we will ever need to merge new features from this library. So I don't see a problem with removing dead code and making it more light-weight if you are interested in going in this direction.

source/dub/internal/colorize/winterm.d
gdelazzari: No problem, I'll definitely switch to DMD's `console` module, it's a better decision indeed. Also, that shouldn't be a lot of work to do. Sorry for not doing that in the first place, I'm a bit new to D and its community and I didn't even think about checking DMD's code before, the first thing I did was search for a dub package that did the job, that's how I ended up with d-colorize. I'll make the changes in the next days/weeks, I have some exams pretty soon unfortunately.

wilzbach:
> Can we merge it in a different branch so others (me too) can improve it further?

As a maintainer you can just push to this PR. If you want automation for this, I use https://github.com/wilzbach/git-tools

gdelazzari:
I have picked this up again and started swapping the colorize-d library I brought in with DMD's console module (https://github.com/dlang/dmd/blob/master/src/dmd/console.d). However I noticed that there's one catch: while DMD's module is lighter, there's no easy way to use it to color parts of the text in the middle of a log string.

Let me show you an example: with the current implementation we can do stuff like this
```d
logInfo("Starting", Color.green,
    "Performing %s build using %s for %-(%s, %)",
    settings.buildType.color(Color.magenta),
    settings.platform.compilerBinary,
    settings.platform.architecture
);
```
which, IMHO, looks nice and allows to easily add colored parts wherever one wants. This is thanks to the Posix-to-Win32 translation that colorize-d does (now I see why they went with that approach): by making the `color()` method just insert ASCII escape sequences in the middle of the string and then

- if on Posix, just spitting it out
- if on Windows, interpreting the escape sequences while writing to stdout/stderr/your-stream-of-choice and calling the Win32 API to set the color accordingly

DMD's console module is pretty barebone and basically provides a `setColor()` method you can call before printing something (and then you can use `resetColor()` to cleanup. While I can use such an interface for handling the colored tag, coloring text in the middle of strings is really hard to do in a "clean" way. In fact, to write something like "hello this is red text" with the "red" word colored, the code would be something like:
```d
write("hello this is ");
console.setColor(Color.red);
write("red");
console.resetColor();
writeln(" text");
```
While you can do something like this standalone, fitting it with the logging methods sounds pretty hard without having to do weird things such as
```d
logInfoNoNewline(Tag("Starting", Color.green), "Performing ");
console.setColor(Color.magenta);
logInfoNoNewline(Tag.no(), "%s", settings.buildType);
console.resetColor();
logInfo(Tag.no(), " build using %s for %-(%s, %)",
    settings.platform.compilerBinary,
    settings.platform.architecture
);
```
which is as bad as it can possibly get. This leads me to think that the approach colorize-d is taking is actually not bad.

So, in conclusion, I would prefer for simplicity reasons to leave untouched the coloring side of things.

I'll still make some changes suggested by @wilzbach regarding the "explosion of `logInfoNoTag` and friends", which I totally agree upon.

Let me know what you think about this, I would really like to see this merged.

ohdatboi:
> which, IMHO, looks nice and allows to easily add colored parts wherever one wants. This is thanks to the Posix-to-Win32 translation that colorize-d does (now I see why they went with that approach): by making the color() method just insert ASCII escape sequences in the middle of the string and then

This is an unnecessary overhead because windows [supports](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) virtual terminal sequences. Setup for this approach is [primitive](https://gitlab.com/ohboi/silly/blob/931bded68f2d2dd03db2bf55795760344caf85fa/silly.d#L219) and it allows you to use same escape sequences, removing `version(Windows) {} else version(Posix) {}` everywhere.

Two things to consider though:
1. Looks like it is supported only in Windows 10. Not a big deal anyway, dmd itself can be used only on Windows 7+ and Windows 10 is the only Windows version which is supported by Microsoft (Not counting extended support period).
2. No idea if it works on OS X. I can't see a reason why it wouldn't, but I haven't tried.

ohdatboi:
I tried only with PowerShell but if I understand Windows correctly, this feature is a feature of Windows Console itself meaning that it can be used with cmd.exe/PowerShell or any other shell.

> Also, is it possible to detect if the functionality is available or not?

Yes, it is:

```
// Set output mode to handle virtual terminal sequences
    HANDLE hOut = GetStdHandle(STD_OUTPUT_HANDLE);
    if (hOut == INVALID_HANDLE_VALUE)
    {
        return GetLastError();
    }

    DWORD dwMode = 0;
    if (!GetConsoleMode(hOut, &dwMode))
    {
        return GetLastError();
    }

    dwMode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
    if (!SetConsoleMode(hOut, dwMode))
    {
        return GetLastError();
    }
```

In this sample `SetConsoleMode` would fail if `ENABLE_VIRTUAL_TERMINAL_PROCESSING` cannot be set.

gdelazzari:
Nice, I'll take a look the next days and I'll try to implement this method. It makes sense that Microsoft is finally allowing the same sequences since Windows was the only OS not supporting them, this solves a lot of problems

gdelazzari:
I just tried on a Windows 10 VM and everything is working beautifully well. I'll go with this approach if no one is against supporting colors only on Windows 10 and beyond (?).

andre2007:
Targeting Windows 10 seems quite reasonable in my opinion.

What should not happen: users below Windows 10 sees garbage by default. Either the colouring should be off by default or only on by default when you can detect the virtual terminal sequences are supported.

Nice work, please continue.

source/dub/internal/colorize/winterm.d
p0nce: `d-colorize` does it this way in order to have color codes inside the strings...

> now I see why they went with that approach

Indeed, that's why I went with this approach, after too much time lost with colorizing console output like DMD does.

Basile-z:
what's up @gdelazzari ?

wilzbach:
Another ping @gdelazzari. Any chance to push this through the finish line?

WebFreak001:
moved to #2268

PR #1492, opened 2018-06-13
Support of snapshots in maven supplier

andre2007:
While dub now supports maven repositories in general, snapshots aren't supported currently. Snapshots enables the developer to push a developer version (e.g. 1.0.0) multiple times before the productive version 1.0.0 is released. Every time a developer version is deployed to Maven it is stored with a timestamp and the build number is increased:

Content of maven repository folder dubpackages/package1489
- 1.0.5-SNAPSHOT/
  - maven-metadata.xml
  - package1489-1.0.5-20090719.104640-2.zip
- maven-metadata.xml

Content of maven-metadata.xml in root folder:
```
<?xml version="1.0" encoding="UTF-8"?>
<metadata modelVersion="1.1.0">
  <groupId>dubpackages</groupId>
  <artifactId>package1489</artifactId>
  <versioning>
    <latest>1.0.5-SNAPSHOT</latest>
    <versions>
      <version>1.0.5-SNAPSHOT</version>
    </versions>
    <lastUpdated>20180317184845</lastUpdated>
  </versioning>
</metadata>
```

Content of maven-metadata.xml in subfolder:
```
<?xml version="1.0" encoding="UTF-8"?>
<metadata>
  <groupId>dubpackages</groupId>
  <artifactId>package1489</artifactId>
  <version>2.1.2-SNAPSHOT</version>
  <versioning>
    <snapshot>
      <timestamp>20090719.104640</timestamp>
      <buildNumber>2</buildNumber>
    </snapshot>
    <lastUpdated>20090719104644</lastUpdated>
  </versioning>
</metadata>
```

An application which want to use the snapshot version of this dependency would add -SNAPSHOT to the version:

```
"dependencies": {
		"package1489": ">=1.0.0-SNAPSHOT"
	}
```

My initial goal was, that adding -SNAPSHOT is not necessary and snapshot handling is done hidden from the end-users. But there is a major drawback. In a repository you could have both, a 1.0.0 release and a 1.0.0 snapshot. Only with the -SNAPSHOT suffix it is clear, that the user want to use the snapshot versions. In specific build setups snapshots should even never be used.

My implementation is in a early state but I want to use this pull request to also discuss some open points I am currently struggle with:

-> In interface method "getVersions" for snapshot versions the value "1.0.0-SNAPSHOT-20090719.104640-2" (baseVersion-SNAPSHOT-timestamp-buildNumber) is returned as version.
Timestamp contains a dot. I am not 100% sure whether this conforms to SemVer.

-> As dependency version you have to specify greaterEqual  ">=1.0.0-SNAPSHOT". It does not work with "1.0.0-SNAPSHOT".

-> There is some functionality implemented in dub called "prerelease". As far as I can see only the upgrade command knows this argument (prerelease). While this functionality goes somehow in a similar direction it seems it isn't fully supported by all commands.

-> If you have a dub package with 20 snapshot versions, then for every snapshot version a xml has to be retrieved (The xml containing the snapshot timestamp and build number). This could become a performance penalty.


WebFreak001:
hm does it really need to be called snapshot? In dub you can already create these prerelease versions with semver: `1.0.0-alpha1`, `1.0.0-alpha2`, `1.0.0-beta1`, `1.0.0-beta2`, `1.0.0-beta3`, `1.0.0-beta4`. Like vibe.d does.

Or well I guess the snapshot part could also be there, but when depending on it you would the specific version instead of SNAPSHOT and dub would resolve it to SNAPSHOT on the server maybe?

How does depending on snapshot packages look in Java with maven?

andre2007:
The concept of artefacts in Maven mandates the suffix -SNAPSHOT. You upload your artefact to Maven with the boolean flag "is snapshot" and Maven automatically adds the suffix -SNAPSHOT. Also if you want to use these pre development versions within your Java application you have to add this suffix to The version.<dependency>
        <groupId>org.springframework</groupId>
        <artifactId>spring</artifactId>
        <version>1.8-SNAPSHOT”</version>
    </dependency>

andre2007:
I did some more tests. The version e.g. "1.0.0" in the header line means there is an artifact on maven registry which is only available with version "1.0.0". Version "1.0.0 and 1.0.0-SNAPSHOT" means the artifact is available with version "1.0.0" and with version "1.0.0-SNAPSHOT"

The versions in the first column are the versions which should be fetched.

| Version to be fetched  |  1.0.0 | 1.0.0-SNAPSHOT  | 1.0.0-SNAPSHOT and 1.0.0-TS-COMMIT_ID | 1.0.0 and 1.0.0-SNAPSHOT |
|--------|-----------|-------|------------------|-----------------------|
| 1.0.0   | Success    |  Error    | Error  | Success |
| 1.0.0-SNAPSHOT  | Error      | Success   | Success     | Success |
| 1.0.0-TS-COMMIT_ID    | -   | -      | Success  | - |

The import part to notice: If you specify a SNAPSHOT version only SNAPSHOT version is found but not the release version. If you enter a version without SNAPSHOT only the release version is found but never the SNAPSHOT one.

If you enter a SNAPSHOT version, after each 24 hours the maven building process will check whether a new SNAPSHOT version is available on registry. Of course you can on any time force the check using the maven upgrade command. I assume the timestamp in the SNAPSHOT version is used for this check.

andre2007:
Closing this pull request until there is a need for this feature and also a clear picture what and how to implement it.

PR #1506, opened 2018-07-05
Bump stdx-allocator version to 2.77.2

wilzbach:
This should fix the testsuite for 2.081


wilzbach:
Tagged as v1.11.0-alpha.1, s.t. the Project Tester uses this version.

PR #1509, opened 2018-07-07
remove regular upgrade check

MartinNowak:
- avoids "build" failures due to connectivity issues or pending dependency conflicts
- `dub build --nodeps` now only suppresses resolution of missing dependencies
- add an explicit `dub upgrade --dry-run` to check for new packages

### DIFF
--- a/changelog/upgrade_check.dd
+++ b/changelog/upgrade_check.dd
@@ -1,0 +1,8 @@
+The regular upgrade check has been removed
+
+Previously dub would regularly (once a day) check for possible package upgrades before building a packages.
+This lead to unexpected build failures, e.g. when internet connectivity was down or dependency resolution failed, and caused unnecessary delays.
+
+The build flag `--nodeps` now only suppresses resolution of missing dependencies.
+
+The new upgrade flag `--dry-run` was added to explicitly check for upgradable packages without actually upgrading anything.
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -525,21 +525,12 @@ class Dub {
 		}
 
 		Dependency[string] versions;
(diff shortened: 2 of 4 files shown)


MartinNowak:
I'm tired of this nonsense, it seems responsible for a lot of unnecessary dub usage problems.
Right now dpl-docs no longer builds because the upgrade check runs into a dependency resolution error, but it has a working resolution in it's repo and could build just fine.

Some more reasons.
- I've never updated anything because dub told me so, and always did it manually when I had time and was updating my package anyhow.
- Don't think I know any other package manager that would eagerly notify me of possible updates.
- People are generally overloaded enough with notifications.

source/dub/dub.d
```diff
@@ -1381,8 +1372,9 @@ enum UpgradeOptions
 	preRelease = 1<<2, /// inclde pre-release versions in upgrade
 	forceRemove = 1<<3, /// Deprecated, does nothing.
 	select = 1<<4, /// Update the dub.selections.json file with the upgraded versions
-	printUpgradesOnly = 1<<5, /// Instead of downloading new packages, just print a message to notify the user of their existence
-	useCachedResult = 1<<6, /// Use cached information stored with the package to determine upgrades
+	dryRun = 1<<5, /// Instead of downloading new packages, just print a message to notify the user of their existence
+	/*deprecated*/ printUpgradesOnly = dryRun, /// deprecated, use dryRun instead
+	/*deprecated*/ useCachedResult = 1<<6, /// deprecated, has no effect
```
wilzbach: FYI: DMD 2.082 finally supports deprecated on enums.

changelog/upgrade_check.dd
```diff
@@ -0,0 +1,8 @@
+The regular upgrade check has been removed
```
wilzbach: I capitalized the first word of the title.

PR #1511, opened 2018-07-07
Fix multiple protocols in URL issue

andre2007:
URL had issues with multiple protocols (dub+https://code.dlang.org)
For  URL("dub+https://code.dlang.org") attribute localURI returned https:////code.dlang.org.
This pr fixes that issue.


WebFreak001:
Hm I dislike that these few protocols are hardcoded into there, why doesn't every URL get that `//` part or rather why isn't there a boolean flag for that? @s-ludwig

source/dub/internal/vibecompat/inet/url.d
```diff
@@ -66,45 +66,40 @@ struct URL {
...
-						enforce(pi < m_host.length-1, "Empty port in URL.");
-						m_port = to!ushort(m_host[pi+1..$]);
-						m_host = m_host[0 .. pi];
-					}
-
-					enforce(!requires_host || m_schema == "file" || m_host.length > 0,
-							"Empty server name in URL.");
-					str = str[si .. $];
+			auto schema_parts = m_schema.split("+");
+			if (schema_parts.any!(p => p.canFind("http", "https", "ftp", "spdy", "file", "sftp")))
```
wilzbach: Shouldn't this be just `.back`?

source/dub/internal/vibecompat/inet/url.d
```diff
@@ -215,16 +210,10 @@ struct URL {
...
-			case "file":
-			case "http":
-			case "https":
-			case "ftp":
-			case "spdy":
-			case "sftp":
-				dst.put("//");
-				break;
+		auto schema_parts = schema.split("+");
+		if (schema_parts.any!(p => p.canFind("http", "https", "ftp", "spdy", "file", "sftp")))
```
wilzbach: Shouldn't this be just `.back`?

source/dub/internal/vibecompat/inet/url.d
wilzbach: Also as the list of valid protocol is repeated, maybe a good idea to define them as an AliasSeq globally?

andre2007: I do not really know AliasSeq but what is the benefit of AliasSeq compared to a static const string array? As far as I see AliasSeq allows a list of values of different types but that seems not be necessary here.

source/dub/internal/vibecompat/inet/url.d
andre2007: I do not understand how the range .back function is related to this source code part. Could you explain what you mean?

wilzbach:
```diff
diff --git a/source/dub/internal/vibecompat/inet/url.d b/source/dub/internal/vibecompat/inet/url.d
index 54c3699..d2c346a 100644
--- a/source/dub/internal/vibecompat/inet/url.d
+++ b/source/dub/internal/vibecompat/inet/url.d
@@ -211,7 +211,7 @@ struct URL {
 		dst.put(schema);
 		dst.put(":");
 		auto schema_parts = schema.split("+");
-		if (schema_parts.any!(p => p.canFind("http", "https", "ftp", "spdy", "file", "sftp")))
+		if (schema_parts.back.canFind("http", "https", "ftp", "spdy", "file", "sftp"))
 		{
 			dst.put("//");
 		}
```

source/dub/internal/vibecompat/inet/url.d
wilzbach: It auto-expands:

```diff
diff --git a/source/dub/internal/vibecompat/inet/url.d b/source/dub/internal/vibecompat/inet/url.d
index 54c3699..a40f691 100644
--- a/source/dub/internal/vibecompat/inet/url.d
+++ b/source/dub/internal/vibecompat/inet/url.d
@@ -67,7 +67,9 @@ struct URL {
 			bool requires_host = false;
 
 			auto schema_parts = m_schema.split("+");
-			if (schema_parts.any!(p => p.canFind("http", "https", "ftp", "spdy", "file", "sftp")))
+			import std.meta : AliasSeq;
+			alias schemes = AliasSeq!("http", "https", "ftp", "spdy", "file", "sftp");
+			if (schema_parts.any!(p => p.canFind(schemes)))
 			{
 				// proto://server/path style
 				enforce(str.startsWith("//"), "URL must start with proto://...");
```

source/dub/internal/vibecompat/inet/url.d
WebFreak001: back could RangeError if schema_parts is empty

andre2007: @wilzbach I included your suggestions. Thanks.
@WebFreak001 Range error is checked using "empty" function

source/dub/internal/vibecompat/inet/url.d
```diff
@@ -66,45 +68,40 @@ struct URL {
...
-						enforce(pi < m_host.length-1, "Empty port in URL.");
-						m_port = to!ushort(m_host[pi+1..$]);
-						m_host = m_host[0 .. pi];
-					}
-
-					enforce(!requires_host || m_schema == "file" || m_host.length > 0,
-							"Empty server name in URL.");
-					str = str[si .. $];
+			auto schema_parts = m_schema.split("+");
+			if (schema_parts.empty == false && schema_parts.back.canFind(m_schemes))
```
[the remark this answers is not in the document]
andre2007: fixed

wilzbach:
Added the "72h no objection -> merge" label.

Issue #1484, opened 2018-05-30
Endless loop with conflicting packages - "The dependency graph blah blah blah"


SenorCMasMas:
### System information
**Windows 10**
    * DUB version 1.8.0, built on Mar  3 2018
    * DMD32 D Compiler v2.080.0
**Ubuntu 16.04**
    * DUB version 1.8.0, built on Mar  3 2018
    * DMD64 D Compiler v2.079.1
**Debian9**
    * DUB version 1.7.1, built on Jan 21 2018
    * DMD64 D Compiler v2.079.1

### Bug Description
Dub goes into an endless loop when I specify two specific packages and attempt to build.
I have tested against the three platforms specified above.

I was able to cache the sdlang package locally and hack the unit-threaded version to get around this.

### How to reproduce?
dub init
+dependency "sdlang-d" version="~ >0.10.2" //<- note no space after tilde
+dependency "mysql-native" version="~ >2.2.2" //<- note no space after tilde
dub build

Note: tilde greaterthan (above) caused strike-through so I added a space.

### Expected Behavior
Download packages or fail with something other than "The dependency graph is likely hitting a pathological case"

### Logs
I can provide more log information if required.
*** From the middle of the process ***
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
Package mysql-native contains invalid dependency unit-threaded (no version candidates) (ci=0)
Package sdlang-d contains invalid dependency unit-threaded (no version candidates) (ci=13)
[...]

Issue #1345, opened 2018-01-20
dub gets into infinite loop


carun:
### System information

- **dub version**: master, 1.7.0, 1.6.0
- **OS Platform and distribution**: Ubuntu 17.10 64 bit, Ubuntu 16.04 64 bit
- **compiler version** (e.g. dmd-2.077.1)

### Bug Description
DUB gets into an infinite while loop for pretty much any operation) dub clean --all-packages, dub build, dub upgrade, etc).

### How to reproduce?
A repo to reproduce: https://github.com/carun/dub-error

### Expected Behavior
dub should not get into indefinite loop

### Logs
Strace has been uploaded here: https://drive.google.com/open?id=1GgLIw1dCCpY0STsfwmlaivAgXDLERO96

carun:
Adding the code here, just in case the repo to reproduce gets deleted in future

* D code
```d
import std.stdio;
import darg;

private struct Options {
    @Option("help", "h")
    @Help("Prints this help.")
    OptionFlag help;
}

// Generate the usage and help string at compile time.
private immutable usage = usageString!Options("test-client");
private immutable help = helpString!Options;

void main(string[] args) {
    __gshared Options options;
    try {
        options = parseArgs!Options(args[1 .. $]);
    }
    catch (ArgParseError e) {
        writeln(e.msg);
        writeln(usage);
        return;
    }
    catch (ArgParseHelp e) {
        // Help was requested
        writeln(usage);
        write(help);
        return;
    }

    return get(options);
}

private void get(in ref Options options) {
    immutable url = "http://www.google.com";
    import std.net.curl : get;
    const res = get(url);
    import std.json : parseJSON;
    writeln(res.parseJSON.toPrettyString);
}
```

* dub.json
```json
{
    "name": "test-client",
    "authors": [
        "Arun Chandrasekaran"
    ],
    "dependencies" : {
        "darg": "~>0.0.4",
        "dproto": "~>2.1.3",
        "painlessjson": "~>1.3.8",
        "requests": "~>0.6.0"
    },
}
```

carun:
Got this error message today after I left to run for 10 mins:

```
The dependency resolution process is taking too long. The dependency graph is likely hitting a pathological case in the resolution algorithm. Please file a bug report at https://github.com/dlang/dub/issues and mention the package recipe that reproduces this error.
```

rjframe:
[...]

Issue #1001, opened 2016-11-22
dub suddenly can't satisfy root package reference


drug007:
## Attention: didn't tested on other machines ##
I did a reduced case [here](https://github.com/drug007/dub-reduced-case)
Test command:
```
cd rdpl
dub build :inspector
```
failed with
```
Root package rdpl:core reference gfm:math ~>6.1.4 cannot be satisfied.
```
but if in 'rdpl/inspector/dub.sdl' you comment out line 6
```
//dependency "rdpl:core" version="~>0.0.3"
```
and rebuild then dub starts satisfying the reference above. You can now uncomment line 6 and it works as expected.

Again, if you remove rdpl/.dub folder then error appears again.

SingingBush:
I've just encountered this with the latest version of dub while trying to get an older version of ddbc into my project.

andre2007:
Issue still reproducible with dub 1.11.0

PR #1513, opened 2018-07-10
Reimplement dependency resolution as a recursive process

### LINKED ISSUE #1300: Dependency deadlock depending on order in json
Happens on linux & windows (Pure-D/serve-d#14)
dub 1.6.0

### Bug Description

There is a dependency deadlock that I failed to reproduce with local packages, but I found a reproducible way using dub registry packages. When specifying the dependencies `libdparse *` and `dscanner ~>0.4.1` (in that order) dub hangs up in the upgrade method, effectively breaking an entire project. It is weird that the order matters here because json objects should be treated as having random order.

### How to reproduce?

`dub init .`

```
{
	"name": "foo",
	"dependencies": {
		"libdparse": "*",
		"dscanner": "~>0.4.1"
	}
}
```

When swapping the 2 dependencies it seems to work, upgrading dscanner to `0.4.2-alpha.1` works too, but they only changed `"libdparse" : "~>0.7.2-alpha.1"` to `"libdparse" : "~>0.7.2-alpha.3"` in the dub.json. libdparse didn't have any dub.json changes between these 2 versions.

### Expected Behavior

It should resolve libdparse to the latest version because the root project doesn't care which version it has, only dscanner specifies a version. Swapping the dependencies should not affect behavior of dub.

### Logs

```
Collecting dependencies for b
  Version selection for dependency libdparse (libdparse) of b is missing.
  Missing dependency libdparse >=0.0.0 of b
  Version selection for dependency dscanner (dscanner) of b is missing.
  Missing dependency dscanner ~>0.4.1 of b
Upgrading project in /tmp/b
Search for versions of libdparse (1 package suppliers)
Downloading metadata for libdparse
Getting from http://code.dlang.org/packages/libdparse.json
Getting http://code.dlang.org/packages/libdparse.json...
Return for libdparse: [0.7.1, 0.7.0, 0.6.0, 0.5.0, 0.4.3, 0.4.2, 0.4.1, 0.4.0, 0.3.0, 0.2.1, 0.2.0, 0.1.1, 0.1.0, 0.7.2-alpha.3, 0.7.2-alpha.2, 0.7.2-alpha.1, 0.7.1-beta.9, 0.7.1-beta.8, 0.7.1-beta.7, 0.7.1-beta.6, 0.7.1-beta.5, 0.7.1-beta.4, 0.7.1-beta.3, 0.7.1-beta.2, 0.7.1-beta.1, 0.7.0-beta.8, 0.7.0-beta.7, 0.7.0-beta.6, 0.7.0-beta.5, 0.7.0-beta.4, 0.7.0-beta.3, 0.7.0-beta.2, 0.7.0-beta.1, 0.7.0-alpha9, 0.7.0-alpha8, 0.7.0-alpha7, 0.7.0-alpha6, 0.7.0-alpha5, 0.7.0-alpha4, 0.7.0-alpha3, 0.7.0-alpha2, 0.7.0-alpha12, 0.7.0-alpha11, 0.7.0-alpha10, 0.7.0-alpha1, 0.7.0-alpha.15, 0.7.0-alpha.14, 0.7.0-alpha.13, ~master, ~issue-158]
Search for versions of experimental_allocator (1 package suppliers)
Downloading metadata for experimental_allocator
Getting from http://code.dlang.org/packages/experimental_allocator.json
Getting http://code.dlang.org/packages/experimental_allocator.json...
Return for experimental_allocator: [2.70.0-b1, ~master]
Search for versions of dscanner (1 package suppliers)
Downloading metadata for dscanner
Getting from http://code.dlang.org/packages/dscanner.json
Getting http://code.dlang.org/packages/dscanner.json...
Return for dscanner: [0.4.1, 0.4.0, 0.3.0, 0.2.0, 0.4.2-alpha.1, 0.4.0-beta.6, 0.4.0-beta.5, 0.4.0-beta.4, 0.4.0-beta.3, 0.4.0-beta.2, 0.4.0-beta.1, 0.4.0-alpha6, 0.4.0-alpha5, 0.4.0-alpha4, 0.4.0-alpha3, 0.4.0-alpha2, 0.4.0-alpha1, 0.4.0-alpha.8, 0.4.0-alpha.7, ~master, ~stringcache-bc, ~revert-459-workaround-17506, ~phobos]
Search for versions of dsymbol (1 package suppliers)
Downloading metadata for dsymbol
Getting from http://code.dlang.org/packages/dsymbol.json
Getting http://code.dlang.org/packages/dsymbol.json...
Return for dsymbol: [0.2.9, 0.2.8, 0.2.7, 0.2.6, 0.2.5, 0.2.4, 0.2.3, 0.2.2, 0.2.1, 0.2.0, 0.1.2, 0.1.1, 0.1.0, 0.2.1-alpha.2, 0.2.1-alpha.1, 0.2.0-beta.2, 0.2.0-beta.1, 0.2.0-alpha7, 0.2.0-alpha6, 0.2.0-alpha5, 0.2.0-alpha4, 0.2.0-alpha3, 0.2.0-alpha2, 0.2.0-alpha1, 0.2.0-alpha.10, 0.2.0-alpha.9, 0.2.0-alpha.8, ~master, ~tests-1, ~temp-issue34-ci, ~issue_16, ~dcd-323, ~auto-bools]
Search for versions of emsi_containers (1 package suppliers)
Downloading metadata for emsi_containers
Getting from http://code.dlang.org/packages/emsi_containers.json
Getting http://code.dlang.org/packages/emsi_containers.json...
Return for emsi_containers: [0.5.3, 0.5.2, 0.5.1, 0.5.0, 0.4.3, 0.4.2, 0.4.1, 0.4.0, 0.3.0, ~master]
Search for versions of inifiled (1 package suppliers)
Downloading metadata for inifiled
Getting from http://code.dlang.org/packages/inifiled.json
Getting http://code.dlang.org/packages/inifiled.json...
Return for inifiled: [1.0.2, 1.0.1, 1.0.0, 0.0.6, 0.0.5, 0.0.4, 0.0.3, 0.0.2, 0.0.1, ~master]
Search for versions of libddoc (1 package suppliers)
Downloading metadata for libddoc
Getting from http://code.dlang.org/packages/libddoc.json
Getting http://code.dlang.org/packages/libddoc.json...
Return for libddoc: [0.2.0, 0.1.0, ~master, ~missing_parameters_fix, ~loop_fix, ~issue-26, ~fix_section_splitter, ~fix-example-code]
Configurations used for dependency resolution:
  inifiled (5, required): [1.0.2, 1.0.1, 1.0.0, 0.0.6, 0.0.5, 0.0.4, 0.0.3, 0.0.2, 0.0.1, ~master]
  experimental_allocator (1, required): [2.70.0-b1, ~master]
  libdparse (0, required): [0.7.1, 0.7.0, 0.6.0, 0.5.0, 0.4.3, 0.4.2, 0.4.1, 0.4.0, 0.3.0, 0.2.1, 0.2.0, 0.1.1, 0.1.0, 0.7.2-alpha.3, 0.7.2-alpha.2, 0.7.2-alpha.1, 0.7.1-beta.9, 0.7.1-beta.8, 0.7.1-beta.7, 0.7.1-beta.6, 0.7.1-beta.5, 0.7.1-beta.4, 0.7.1-beta.3, 0.7.1-beta.2, 0.7.1-beta.1, 0.7.0-beta.8, 0.7.0-beta.7, 0.7.0-beta.6, 0.7.0-beta.5, 0.7.0-beta.4, 0.7.0-beta.3, 0.7.0-beta.2, 0.7.0-beta.1, 0.7.0-alpha9, 0.7.0-alpha8, 0.7.0-alpha7, 0.7.0-alpha6, 0.7.0-alpha5, 0.7.0-alpha4, 0.7.0-alpha3, 0.7.0-alpha2, 0.7.0-alpha12, 0.7.0-alpha11, 0.7.0-alpha10, 0.7.0-alpha1, 0.7.0-alpha.15, 0.7.0-alpha.14, 0.7.0-alpha.13, ~master, ~issue-158]
  dsymbol (3, required): [0.2.9, 0.2.8, 0.2.7, 0.2.6, 0.2.5, 0.2.4, 0.2.3, 0.2.2, 0.2.1, 0.2.0, 0.1.2, 0.1.1, 0.1.0, 0.2.1-alpha.2, 0.2.1-alpha.1, 0.2.0-beta.2, 0.2.0-beta.1, 0.2.0-alpha7, 0.2.0-alpha6, 0.2.0-alpha5, 0.2.0-alpha4, 0.2.0-alpha3, 0.2.0-alpha2, 0.2.0-alpha1, 0.2.0-alpha.10, 0.2.0-alpha.9, 0.2.0-alpha.8, ~master, ~tests-1, ~temp-issue34-ci, ~issue_16, ~dcd-323, ~auto-bools]
  libddoc (6, required): [0.2.0, 0.1.0, ~master, ~missing_parameters_fix, ~loop_fix, ~issue-26, ~fix_section_splitter, ~fix-example-code]
  emsi_containers (4, required): [0.5.3, 0.5.2, 0.5.1, 0.5.0, 0.4.3, 0.4.2, 0.4.1, 0.4.0, 0.3.0, ~master]
  dscanner (2, required): [0.4.1, 0.4.2-alpha.1]
Dependency dscanner -> libdparse >=0.7.2-alpha.1 <0.8.0 mismatches with selected version 0.7.1 (ci=2)
Interation (ci=2) libdparse 0.7.1, experimental_allocator 2.70.0-b1, dscanner 0.4.1 (C), dsymbol 0.2.9 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> libdparse >=0.7.2-alpha.3 <0.8.0 mismatches with selected version 0.7.1 (ci=2)
Interation (ci=2) libdparse 0.7.1, experimental_allocator 2.70.0-b1, dscanner 0.4.2-alpha.1 (C), dsymbol 0.2.9 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> libdparse >=0.7.2-alpha.1 <0.8.0 mismatches with selected version 0.7.1 (ci=2)
Interation (ci=2) libdparse 0.7.1, experimental_allocator ~master, dscanner 0.4.1 (C), dsymbol 0.2.9 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> libdparse >=0.7.2-alpha.3 <0.8.0 mismatches with selected version 0.7.1 (ci=2)
Interation (ci=2) libdparse 0.7.1, experimental_allocator ~master, dscanner 0.4.2-alpha.1 (C), dsymbol 0.2.9 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dsymbol -> libdparse ~>0.7.1 mismatches with selected version 0.7.0 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.9 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dsymbol -> libdparse >=0.7.1-beta.7 <0.8.0 mismatches with selected version 0.7.0 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.8 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dsymbol -> libdparse >=0.7.1-beta.7 <0.8.0 mismatches with selected version 0.7.0 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.7 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dsymbol -> libdparse >=0.7.1-beta.7 <0.8.0 mismatches with selected version 0.7.0 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.6 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.5 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.5 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.4 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.4 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.3 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.3 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.2 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.2 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.1 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.1 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.2.0 (ci=3)
Interation (ci=3) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.2.0 (C), emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.5.3 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.5.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.5.2 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.5.2 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.5.1 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.5.1 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.5.0 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.5.0 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.4.3 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.4.3 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.4.2 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.4.2 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.4.1 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.4.1 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)
Dependency dscanner -> dsymbol ~>0.2.6 mismatches with selected version 0.1.2 (ci=3)
Dependency dsymbol -> emsi_containers ~>0.3.0 mismatches with selected version 0.4.0 (ci=4)
Interation (ci=4) libdparse 0.7.0, experimental_allocator 2.70.0-b1, dscanner 0.4.1, dsymbol 0.1.2, emsi_containers 0.4.0 (C), inifiled 1.0.2 (C), libddoc 0.2.0 (C)

...
```

### LINKED ISSUE #1508: Backtracking for dependency resolver
Both bundler and carge seem to have good experience with backtracking to speed up dependency resolution and print humanly understandable conflicts.
https://github.com/CocoaPods/Molinillo/blob/b97b2f96440925b93150de20709c6403c5ee22f0/ARCHITECTURE.md#backtracking
https://github.com/rust-lang/cargo/pull/5187

s-ludwig:
The new approach minimizes the dependency version graph while descending, making it *much* less likely to run into pathological cases. It also is a lot easier to understand and reason about than the previous iterative approach.

The downside is dynamic and stack memory use, where the latter could be optimized using a COW approach. If stack overflow should turn out to become an issue in large dependency graphs, the algorithm could be converted to use a heap based stack.

Fixes #1300.
Fixes #1508.

Looks related, but appears to be fixed already: #1484
#1001 appears to have been fixed already (gives a proper error message)
#1345 also appears to have been fixed, this time externally by fixing the faulty dependencies

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -96,6 +96,10 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		// returned explicitly
 		context.result.remove(rootbase);
 
+		logDiagnostic("Dependency resolution result:");
+		foreach (d; context.result.keys.sort())
+			logDiagnostic("  %s: %s", d, context.result[d]);
+
 		return context.result;
 	}
 
@@ -225,7 +229,7 @@ class DependencyResolver(CONFIGS, CONFIG) {
 					// try the configuration on a cloned context
 					auto subcontext = context.clone;
 					constrain(TreeNode(dep.pack, c.config), subcontext, max_iterations);
-					constrainDependencies(TreeNode(dep.pack, c.config), dependencies, depidx+1, subcontext, max_iterations);
+					constrainDependencies(n, dependencies, depidx+1, subcontext, max_iterations);
 					// if a branch succeeded, replace the current context
 					// with the one from the branch and return
 					context = subcontext;
@@ -237,8 +241,12 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		}
 
 		// ignore unsatisfiable optional dependencies
-		if (dep.depType != DependencyType.required)
+		if (dep.depType != DependencyType.required) {
+			auto subcontext = context.clone;
+			constrainDependencies(n, dependencies, depidx+1, subcontext, max_iterations);
+			context = subcontext;
 			return;
+		}
 
 		// report the first error encountered to the user
 		if (first_err) throw first_err;
@@ -481,5 +489,18 @@ unittest {
 		} catch (DependencyLoadException e) {
 			assert(e.msg == "Failed to find any versions for package b, referenced by a 0");
 		}
+	}
+
+	// regression: unresolvable optional dependency skips the remaining dependencies
+	with (TestResolver) {
+		auto res = new TestResolver([
+			"a:0": [
+				TreeNodes("b", ics([ic(2)]), DependencyType.optional),
+				TreeNodes("c", ics([ic(1)]))
+			],
+			"b:1": [],
+			"c:1": []
+		]);
+		assert(res.resolve(TreeNode("a", ic(0))) == ["c":ic(1)]);
 	}
 }
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -449,4 +449,37 @@ unittest {
 		]);
 		assert(res.resolve(TreeNode("a", ic(0))) == ["b":ic(2)], to!string(res.resolve(TreeNode("a", ic(0)))));
 	}
+
+	// check error message for multiple conflicting dependencies
+	with (TestResolver) {
+		auto res = new TestResolver([
+			"a:0": [TreeNodes("b", ics([ic(1)])), TreeNodes("c", ics([ic(1)]))],
+			"b:1": [TreeNodes("d", ics([ic(1)]))],
+			"c:1": [TreeNodes("d", ics([ic(2)]))],
+			"d:1": [],
+			"d:2": []
+		]);
+		try {
+			res.resolve(TreeNode("a", ic(0)));
+			assert(false, "Expected resolve to throw.");
+		} catch (ResolveException e) {
+			assert(e.msg ==
+				"Unresolvable dependencies to package d:"
+				~ "\n  b 1 depends on d [1]"
+				~ "\n  c 1 depends on d [2]");
+		}
+	}
+
+	// check error message for invalid dependency
+	with (TestResolver) {
+		auto res = new TestResolver([
+			"a:0": [TreeNodes("b", ics([ic(1)]))]
+		]);
+		try {
+			res.resolve(TreeNode("a", ic(0)));
+			assert(false, "Expected resolve to throw.");
+		} catch (DependencyLoadException e) {
+			assert(e.msg == "Failed to find any versions for package b, referenced by a 0");
+		}
+	}
 }
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -1,7 +1,7 @@
 /**
 	Dependency configuration/version resolution algorithm.
 
-	Copyright: © 2014 rejectedsoftware e.K.
+	Copyright: © 2014-2018 Sönke Ludwig
 	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
 	Authors: Sönke Ludwig
 */
@@ -18,7 +18,23 @@ import std.typecons : Nullable;
 import std.string : format, indexOf, lastIndexOf;
 
 
+/** Resolves dependency graph with multiple configurations per package.
+
+	The term "configuration" can mean any kind of alternative dependency
+	configuration of a package. In particular, it can mean different
+	versions of a package.
+
+	`CONFIG` is an abstract type denoting a single configuration of a certain
+	package, whereas `CONFIGS` denotes a set of configurations. The
+	representation of both can be freely chosen, so that `CONFIGS` for example
+	can be defined in terms of a version range.
+*/
 class DependencyResolver(CONFIGS, CONFIG) {
+	/** Encapsulates a list of outgoing edges in the dependency graph.
+
+		A value of this type represents a single dependency with multiple
+		possible configurations for the target package.
+	*/
 	static struct TreeNodes {
 		string pack;
 		CONFIGS configs;
@@ -37,6 +53,10 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		}
 	}
 
+	/** A single node in the dependency graph.
+
+		Nodes are a combination of a package and a single package configuration.
+	*/
 	static struct TreeNode {
 		string pack;
 		CONFIG config;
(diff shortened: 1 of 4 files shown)


WebFreak001:
Are there also changes in performance by this?

test/expected-issue1037-output
```diff
@@ -1,4 +1,3 @@
-Root package issue1037-better-dependency-messages reference gitcompatibledubpackage 1.0.1 cannot be satisfied.
-Packages causing the conflict:
-	issue1037-better-dependency-messages depends on 1.0.1
-	b depends on ~>1.0.2
+Conflicting dependencies to gitcompatibledubpackage:
+  b >=0.0.0 @/home/runner/dub/test/issue1037-better-dependency-messages/b depends on gitcompatibledubpackage ~>1.0.2
+  issue1037-better-dependency-messages ~master depends on gitcompatibledubpackage 1.0.1
```
WebFreak001: love the new details in this message

source/dub/dependencyresolver.d
```diff
@@ -353,6 +294,12 @@ private string basePackage(string p)
 	return p[0 .. idx];
 }
 
+private string subPackage(string p)
+{
+	auto idx = indexOf(p, ':');
+	if (idx < 0) return "";
+	return p[idx+1 .. $];
```
WebFreak001: you could make this a one-liner using findSplit: `return p.findSplit(":")[2];`.

imo it's actually a lit bit more readable (and it would be even more readable if findSplit returned a tuple with start, match and end members) but it's not really required. Actually also think the performance might be a little bit worse.

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
 
-		// find all possible configurations of each possible dependency
-		size_t[string] package_indices;
-		string[size_t] package_names;
-		PackageConfigs[] configs;
-		bool[string] maybe_optional_deps;
-		bool[TreeNode] visited;
+		// Leave the possibility to opt-out from the loop limit
+		import std.process : environment;
+		bool no_loop_limit = environment.get("DUB_NO_RESOLVE_LIMIT") !is null;
```
WebFreak001: this new flag should be documented somewhere

source/dub/dependencyresolver.d
wilzbach: > it's actually a lit bit more readable (and it would be even more readable if findSplit returned a tuple with start, match and end members)

Yeah I tried this once

https://github.com/dlang/phobos/pull/5968

tl;dr: it was reverted due to a regression in CTFE usage of tuples.

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
 
-				// eliminate configurations from which we know that they can't satisfy
-				// the uniquely defined root dependencies (==version or ~branch style dependencies)
-				if (parent_unique) config.allConfigs = config.allConfigs.filter!(c => matches(ch.configs, c)).array;
+	private static struct PackageConfigs
+	{
+		import std.bitmanip : BitArray;
 
-				configs[pidx] = config;
+		bool[string] visited; // key = sub package or empty string
```
WebFreak001: while this is being rewritten, maybe we could consider making this a normal array (or some container like RedBlackTree) and add helper functions with readable names such as `isVisited(string name)` and `visit(string name)` that check if it is in the container and add it respectively.

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
+		CONFIG[] configs;
```
WebFreak001: Every time I look at this code again I get confused what CONFIG is supposed to be, being in PackageConfigs. Can we add a little bit of documentation here and to PackageConfigs describing what they are used for?

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
-					config = configs[*pi];
-				} else {
-					if (basepack == root_base_pack) config.allConfigs = [root.config];
-					else config.allConfigs = getAllConfigs(basepack);
-					configs ~= config;
-					package_indices[basepack] = pidx;
-					package_names[pidx] = basepack;
+		// build up the dependency graph, eliminating as many configurations/
+		// versions as possible
+		PackageConfigs[string] packs;
```
WebFreak001: should we maybe call this `cache` or `visited` instead to make more clear what packs are stored in this variable?

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
+		auto base = n.pack.basePackage;
+		auto sub = n.pack.subPackage;
+		if (base in packs && sub in packs[base].visited) return;
+		packs[base].visited[sub] = true;
+
+		auto dependencies = getChildren(n);
+
+		foreach (dep; dependencies) {
+			// lazily load all dependency configurations
+			auto depbase = basePackage(dep.pack);
```
WebFreak001: why aren't we using UFCS here like in the other places?

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
+				if (di.included[i]) {
+					if (!matches(dep.configs, c))
+						di.included[i] = false;
+					else any_config = true;
 				}
+
+			if (!any_config && dep.depType == DependencyType.required) {
+				if (di.configs.length)
+					throw new ResolveException(n, dep);
+				else throw new DependencyLoadException(n, dep);
```
WebFreak001: this formatting seems a little weird (first throw in a new line then in the same line)

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
-				// the uniquely defined root dependencies (==version or ~branch style dependencies)
-				if (parent_unique) config.allConfigs = config.allConfigs.filter!(c => matches(ch.configs, c)).array;
+	private static struct PackageConfigs
+	{
+		import std.bitmanip : BitArray;
 
-				configs[pidx] = config;
+		bool[string] visited; // key = sub package or empty string
+		CONFIG[] configs;
+		BitArray included;
```
WebFreak001: also the correlation between this length and configs.length could maybe be better represented by a struct containing both CONFIG and bool included, otherwise in the future someone might manipulate one array without knowing to change the other and somewhere else in the code a range error will occur.

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
-		Nullable!ConflictError first_error;
-		size_t loop_counter = 0;
-
-		// Leave the possibility to opt-out from the loop limit
-		import std.process : environment;
-		bool no_loop_limit = environment.get("DUB_NO_RESOLVE_LIMIT") !is null;
+		try constrainDependencies(n, dependencies, 0, packs, loop_counter);
+		catch (ResolveException e) {
+			// add any relevant information for this stack frame
+			e.chain(n, dependencies);
```
WebFreak001: +1, I like this way of adding dependencies to the error

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
-	protected abstract bool matches(CONFIGS configs, CONFIG config);
+		// ignore unsatisfyable optional dependencies
+		if (dep.depType != DependencyType.required)
+			return;
+
+		// report the first error encountered to the user
+		if (first_err) throw first_err;
+
+		// should have thrown in constrainRec before reaching this
+		assert(false, "Got no configuration for dependency!?");
```
WebFreak001: just in case this actually gets triggered, including which package triggered this would probably be a good idea.

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
 
-				configs[pidx] = config;
+		bool[string] visited; // key = sub package or empty string
+		CONFIG[] configs;
+		BitArray included;
+	}
 
-				foreach (v; config.allConfigs)
-					findConfigsRec(TreeNode(ch.pack, v), parent_unique && config.allConfigs.length == 1);
+	private void constrainRec(TreeNode n, ref PackageConfigs[string] packs, ref long loop_counter)
```
WebFreak001: Why is there constrainRec and constrainDependencies? I would have guessed constrainRec stands for constrainRecursive by the name but that doesn't really make sense with constrainDependencies being there

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
+		assert (--loop_counter > 0,
+			"The dependency resolution process is taking too long. The"
+			~ " dependency graph is likely hitting a pathological case in"
+			~ " the resolution algorithm. Please file a bug report at"
+			~ " https://github.com/dlang/dub/issues and mention the package"
+			~ " recipe that reproduces this error.");
+
+		auto dep = &dependencies[depidx];
+		auto depbase = dep.pack.basePackage;
+		auto dp = &packs[depbase];
```
WebFreak001: these variable names tho

source/dub/dependencyresolver.d
```diff
@@ -54,236 +54,149 @@ class DependencyResolver(CONFIGS, CONFIG) {
...
-
-			// print out current iteration state
-			logDebug("Interation (ci=%s) %s", conflict_index, {
-				import std.array : join;
-				auto cs = new string[configs.length];
-				foreach (p, i; package_indices) {
-					if (configs[i].allConfigs.length)
-						cs[i] = p~" "~configs[i].allConfigs[config_indices[i]].to!string~(i >= 0 && i >= conflict_index ? " (C)" : "");
-					else cs[i] = p ~ " [no config]";
+	private void constrainDependencies(TreeNode n, TreeNodes[] dependencies, size_t depidx, ref PackageConfigs[string] packs, ref long loop_counter)
```
WebFreak001: is the n argument even used in this function?

WebFreak001:
travis is failing because of the absolute paths in the test file, maybe we should use relative paths relative to the package root directory instead?

andre2007:
By chance, does this pull request has any influence on optional dependencies? (https://github.com/dlang/dub/pull/1148)

source/dub/dependencyresolver.d
s-ludwig: It's not new, but yeah, it should still be documented.

source/dub/dependencyresolver.d
s-ludwig: That would cause a range error for package names without a sub package part, wouldn't it?

WebFreak001: no findSplit returns a tuple, it is always exactly 3 long (pre, match, post) no matter the input

source/dub/dependencyresolver.d
s-ludwig: `constrainDependencies` calls back into `constrainRec`, but also calls itself recursively. Probably makes sense to just drop the `Rec` suffix.

source/dub/dependencyresolver.d
```diff
@@ -18,7 +18,23 @@ import std.typecons : Nullable;
...
+	configuration of a package. In particular, it can mean different
+	versions of a package.
+
+	`CONFIG` is an abstract type denoting a single configuration of a certain
+	package, whereas `CONFIGS` denotes a set of configurations. The
+	representation of both can be freely chosen, so that `CONFIGS` for example
+	can be defined in terms of a version range.
+*/
 class DependencyResolver(CONFIGS, CONFIG) {
+	/** Encasulates a list of outgoing edges in the dependency graph.
```
John-Colvin: `s/Encasulates/Encapsulates/`

s-ludwig:
@WebFreak001
> Are there also changes in performance by this?

It definitely improves performance for complex graphs (which is the whole point of this of course). For simple graphs it is probably a bit slower due to the heavy memory allocation and copying, but it doesn't matter because we are talking about milliseconds or less then. Most of the time is currently spent in superfluous queries to the registry anyway, since the cache mechanism stopped working at some point.

source/dub/dependencyresolver.d
s-ludwig: Now it's used in the assertion message at the bottom of the function ;-)

s-ludwig:
@andre2007 
> By chance, does this pull request has any influence on optional dependencies? (#1148)

It shouldn't affect that issue. A proper solution for that requires on a higher level approach, one that comes after the configurations have been determined and the project is about to be built.

s-ludwig:
> travis is failing because of the absolute paths in the test file, maybe we should use relative paths relative to the package root directory instead?

I just fixed the test for now, as outputting relative paths would be more complicated (relative to the CWD would be easier than relative to the root directory).

source/dub/dependencyresolver.d
s-ludwig: Ah, I actually missed the "split" part.

s-ludwig:
> Do we have test cases that we can include that were previously breaking dependency resolution but don't anymore?

The problem is that the known test cases depend on a rather large number of public packages. Capturing them in a local test case would be quite involved and just sourcing them from the registry is error prone and slows down the CI tests unnecessarily.

WebFreak001:
but if we do have one public test case where it happens, can we at least try locally if the issue happened before and if it is fixed?

s-ludwig:
> but if we do have one public test case where it happens,  can we at least try locally if the issue happened before and if it is  fixed?

#1300 does get fixed by this and is the only public test case that I found that still fails on master.

s-ludwig:
I just noticed that not all sources end up in the conflict error message, because the stack doesn't generally encompass the whole dependency graph, so an additional map will be required to fix that.

s-ludwig:
You are right, those were definitely easy enough to add. I'll rebase/clean up once the CI is green again.

s-ludwig:
Not sure what those CI failures are about. Travis looks fine and I can't spot the failure for Jenkins.

wilzbach:
> Not sure what those CI failures are about. Travis looks fine and I can't spot the failure for Jenkins.

We're currently migrating off from Jenkins to Buildkite due to these spurious failures. You can ignore them safely though we should probably

andre2007:
I tried this PR. While it seems to solve an issue for me, I found another issue. Maybe this issue is completely unrelated to this pr or maybe it is somehow affected.

I have a small app with dependency to vibe-d. I have to execute "dub" 4 times until all dependencies are resolved right. The first 3 times there are following errors:
 
First try:
core.exception.AssertError@source\dub\project.d(156): Non-optional dependency libasync of vibe-d:core not found in dependency tree!?.

Second try:
core.exception.AssertError@source\dub\project.d(156): Non-optional dependency memutils of libasync not found in dependency tree!?.

Third try:
core.exception.AssertError@source\dub\project.d(156): Non-optional dependency taggedalgebraic of eventcore not found in dependency tree!?.

My dub packages folder is completely empty while executing dub the first time. My assumption was, if a dependency 1 (vibe.d) depends on 2 (vibe-d:core) which has a dependency to 3 (libasync) then this will fail if dependency 2 and 3 aren't downloaded yet.
But this assumption is wrong. I left in dub packages folder "libasync-0.8.3" and removed all other packages and still the error is thrown.
core.exception.AssertError@source\dub\project.d(156): Non-optional dependency libasync of vibe-d:core not found in dependency tree!?.

MartinNowak:
> We're currently migrating off from Jenkins to Buildkite due to these spurious failures.

No, we're migrating from Jenkins cause it's using too many server-side resources.
Most "spurious" failures are caused by our test runners/caches/cleanup and are not the fault of Jenkins.

MartinNowak:
Will we get this merged until next week for 2.082 @s-ludwig @wilzbach?
It's still missing a changelog entry.

s-ludwig:
Still need to look into the issue @andre2007 mentioned, but I'll try to do it this week.

andre2007:
@s-ludwig I found the root cause for my issue. It is not related to your pr.
While the Dub Registry provider returns the list of the dependencies of a package in method fetchPackageRecipe, the Maven provider does not return the list of dependencies. (Dub registry provider has this information directly). Maven provider doesn't have this information at hand.
Maven provider would need to download the package archive, extract dub.json /dub.sdl and return the dependencies.
I am currently think about what is the best solution. 
Sorry for the noise.

gedaiu:
It would be really nice to have this in the next release.

I tested this branch on some of my projects and it worked good :) If it's needed, I can test it with more projects.

s-ludwig:
Okay, good to hear, was just going to look into it myself. I guess someone can pull the trigger then.

WebFreak001:
no changelog entry?

s-ludwig:
Of course... forgot about that.

s-ludwig:
Added now.

WebFreak001:
tests pass, reviews agree, feel like it's time to merge

Issue #1436, opened 2018-04-04
Build type: syntax


andre2007:
In my dub.json I added a new build type "syntax" which has as build option ["syntaxOnly"].

This build type helps for fast iteration while developing. In this iteration mode I do not need any compilation but only want to check my syntax.

``` dub build -b syntax```

I think it makes sense to add this build type to dub in general.

andre2007:
Merged

PR #1514, opened 2018-07-10
Add build type syntax

andre2007:
The build type "syntax" should be highly valuable for IDEs like IntelliJ (D plugin).
At the moment the syntax check in IntelliJ D plugin compiles on every code change a new binary leading to several issues (performance, file locked errors).

Using build type "syntax" (will cause -o-) these issues will be gone.

### DIFF
--- a/changelog/buildTypeSyntax.dd
+++ b/changelog/buildTypeSyntax.dd
@@ -1,0 +1,3 @@
+DUB supports build type "syntax" 
+
+With this release new build type "syntax" is added. This build type is useful for IDEs to check the syntax of D coding without generating binaries.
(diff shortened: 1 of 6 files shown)


ohdatboi:
Looks good to me. After this gets merged, don't forget to add description of this build type to the registry.

Also, referencing your issue here #1436 so it can be closed afterwards.

wilzbach:
> don't forget to add description of this build type to the registry.

The dmd policy is to merge only after the spec PR has been submitted as the updates have frequently been forgotten in the past.
Maybe a good idea here too?

andre2007:
Change log is added. As wilzbach pointed out, it makes sense to wait for PR merge before adapting the registry documentation. After merge I will create a PR for registry.

wilzbach:
> As wilzbach pointed out,

I was actually saying the opposite ;-)

> to merge only __after__ the spec PR has been submitted

But I don't feel strongly about this here.

andre2007:
You are right, I am sorry. I will create a PR for registry too

andre2007:
Documentation added
https://github.com/dlang/dub-registry/pull/364

PR #1516, opened 2018-07-18
issue1158-stdin-for-single-files

andre2007:
This pull request allows running single source code files from stdin.

Example:
curl -s http://localhost:8080/singledub.d | dub -

The - (dash) is the marker for reading stdin.


source/dub/internal/utils.d
```diff
@@ -40,8 +40,14 @@ NativePath getTempDir()
 NativePath getTempFile(string prefix, string extension = null)
 {
 	import std.uuid : randomUUID;
+    import std.array: replace;
```
wilzbach: I think the existing file uses tabs whereas you use whitespace here. We should probably just convert all files to whitespace at some point...

test/issue1158-stdin-for-single-files/stdin.d
```diff
@@ -0,0 +1,7 @@
+/+ dub.sdl:
+	name "hello"
```
wilzbach: The name is optional since 1.6

source/dub/internal/utils.d
```diff
@@ -40,8 +40,14 @@ NativePath getTempDir()
...
 
-	auto path = getTempDir() ~ (prefix ~ "-" ~ randomUUID.toString() ~ extension);
+	string fileName = prefix ~ "-" ~ randomUUID.toString() ~ extension;
+
+	if (extension !is null && extension == ".d")
+		fileName = fileName.replace("-", "_");
```
wilzbach: Maybe directly use _ above?

wilzbach:
Maybe a short changelog entry so that people know about this?

test/issue1158-stdin-for-single-files/stdin.d
andre2007: By removing name attribute, the file not longer compiles. I tried different formats:
```
/+ dub.sdl:
+/
```

```
/+ dub.sdl
+/
```

source/dub/internal/utils.d
andre2007: randomUUID also returns a string containing dashes. Therefore my idea was to keep the dash and replaces all dashes in case of a d module file.

source/dub/internal/utils.d
andre2007: fixed

andre2007:
Change log added

test/issue1158-stdin-for-single-files/stdin.d
wilzbach: Ah sorry. Apparently one items is needed then :/

source/dub/commandline.d
```diff
@@ -97,6 +97,16 @@ int runDubCommandLine(string[] args)
 		environment["TEMP"] = environment["TEMP"].replace("/", "\\");
 	}
 
+	// special stdin syntax
+	if (args.length >= 2 && args[1] == "-")
+	{
+		import dub.internal.utils: getTempFile;
+
+		auto path = getTempFile("app", ".d");
+		std.file.write(path.toNativeString(), stdin.byLineCopy.join);
```
wilzbach: FYI if you don't want the the FQSN (std.file), there exists `toFile` just for this reason:

`stdin.byChunk(4096).joiner.toFile(path.toNativeString())`

Also use `byChunk` instead of `byLineCopy` to avoid unnecessary allocation + copying and as a bonus you don't loose the line endings.
Furthermore with `joiner` you avoid allocations :)

source/dub/commandline.d
```diff
@@ -97,6 +97,16 @@ int runDubCommandLine(string[] args)
...
+		args = args[0]~[path.toNativeString()]~args[2..$];
```
wilzbach: No need for [] here.
Also mind the style of the existing file (and [DStyle](https://dlang.org/dstyle.html)): in other words spaces around `~`.

source/dub/commandline.d
andre2007: great! looks much better now.

source/dub/commandline.d
andre2007: fixed

andre2007: Square brackets are needed.
Args without []: ["C:\\Users\\home\\git\\dub\\bin\\dub.exeC:\\Users\home\\Desktop\\abc.d"]
Args with []: ["C:\\Users\home\\git\\dub\\bin\\dub.exe", "C:\\Users\home\\Desktop\\abc.d"]

source/dub/commandline.d
andre2007: DMD 2.071.2 and lower does not support toFile ):

andre2007:
@wilzbach Unfortunately I had to switch back to std.file.write as toFile isn't available in the lower DMD/Phobos releases. The write function also had some issue with the joiner result, I had to add .array.

Will we stay on 2.068.2 as lowest version to be supported or is there a plan to remove some of the oldest versions?

wilzbach:
The latest GDC release has been the blocking reason for an upgrade.
The newest GDC release is at 2.076 (master is at 2.082).

I can make that upgrade tomorrow.

andre2007:
@wilzbach did you already had the chance to make the upgrade (GDC was blocking)?

wilzbach:
Not yet, but done now https://github.com/dlang/dub/pull/1520 (might require a few tries of experimentation to get it right tough, was a bit more complex than I remembered).

So if you don't want to depend/wait on this, feel free to revert here for now ;-)
(I don't mind including the "new" `toFile` and other changes in my PR - actually a good check that the upgrade worked.)

andre2007:
Yes I really like your proposed syntax (toFile). I will revert to your proposal when the merge was done.

andre2007:
It seems there is a general error in semaphoreci. The gdc build fails with:
-bash: ./semaphore-ci.sh: No such file or directory

Also the dmd build fails, here no clear error description can be found.

https://semaphoreci.com/wilzbach/dub/branches/pull-request-1516/builds/7

wilzbach:
Yes, we need to merge https://github.com/dlang/dub/pull/1520 first (I changed the scripts while I was setting up the newer `gdc` compiler.)
I have been waiting on the approval from Martin & Sönke on this one, which I got over the weekend, so we can finally move ahead with the upgrade.

andre2007:
@wilzbach I changed the coding to your proposal (stdin.byChunk(4096).joiner.toFile(path.toNativeString());) and it works fine. Except some dub packages throwing errors. But I assume this is a general issue.

PR #1518, opened 2018-07-24
Better build reuse

John-Colvin:
The fact that builds contain the buildType string means that every dependency has to be recompiled for all different build types, even when they cause no change in compiler options. It doesn't appear to be necessary as all that matters is that the compiler options are the same.

Seeing as VibeCustomMain doesn't actually do anything any more, I've stopped it being added.

The combination of these two changes means that - assuming no other differences with versions or optional dependencies - the same build can be used as a dependency of a unittest build as a normal build.


MartinNowak:
See #1589 for an alternative @John-Colvin

PR #1520, opened 2018-07-26
Use GDC from Semaphore and upgrade the minimally required host compiler

wilzbach:
Now that GDC on the latest Ubuntu 18.04 is built on top 2.076 (master is at 2.082 btw), we can upgrade it and thus finally increase the minimall required D version to build Dub (for DMD it's 2.079 btw, so we could increase a lot here, for now I wasn't so drastic though).

I used SemaphoreCI, because gdcdownloads.org still doesn't provide binaries for the newest GDC release.

See also: https://github.com/dlang/dmd/pull/8431

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,13 @@ jobs:
     - d: gdc
   include:
     - stage: test
-      d: dmd-2.078.0
+      d: dmd-2.081.1
+      env: [FRONTEND=2.081]
+    - d: dmd-2.080.1
+      env: [FRONTEND=2.080]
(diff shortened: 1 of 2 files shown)


wilzbach:
From @MartinNowak:

> Sounds OK to me.
> I would apply the same requirement as for dmd since dub is now part of
the official release builds. It must compile with all stable releases of
dmd, ldc, and gdc so that we can switch the bootstrap compiler and also
so that ldc/gdc don't have to deal with eagerly adopted features in
basic D tools.

Issue #1180, opened 2017-06-30
cache local ist broken


andre2007:
--cache local is not working. 

```
dub init vibetest -t vibe.d
cd vibetest
dub build --cache local
```

C:\Users\user\Desktop\vibetest>dub build --cache local
Fetching libevent 2.0.2+2.0.16 (getting selected version)...
Fetching diet-ng 1.2.1 (getting selected version)...
Fetching openssl 1.1.5+1.0.1g (getting selected version)...
Fetching memutils 0.4.9 (getting selected version)...
Fetching vibe-d 0.7.31 (getting selected version)...
Fetching libasync 0.8.3 (getting selected version)...
Non-optional dependency vibe-d of vibetest not found in dependency tree!?.

atilaneves:
I ran into this today. It works if you `dub fetch $PACKAGE --cache=local` first. The issue is that the behaviour is different from the default - then it does a dub fetch for you. It should do the same with `--cache=local`.

andre2007:
@wilzbach @WebFreak001 
With --cache=local the dependeny is stored in the package root directory but not added to dub.json nor found by the dependency collector. The dependency collector looks only in the user and system path. I wonder whether this is by purpose? 
Should the dependency collector also look into the package path?

andre2007:
Pull request https://github.com/dlang/dub/pull/1568

Geod24:
PR merged.

PR #1526, opened 2018-08-15
Late dependencies resolver

andre2007:
Update: I updated the title to a more appropriate one. Although the solution described here is currently needed only for the maven package supplier, the concept is also useful for other package supplier which might come in future (git package supplier...). 

Problem description: 
A project has dependency to package **A**. Package **A** has dependency to package **B**.
Maven repository is set as package supplier. Executing dub upgrade for the project will only fetch package **A** but not package **B**.

Root cause: 
In contrast to dub registry, maven repository does not have the knowledge about dependencies. Dub registry package supplier retrieves the dependency information while calling method "fetchPackageRecipe". Dub relies on the fact, that package recipe will have info about the dependencies.
But that is not the case for maven repositories.

Solution: 
Method upgrade will call itself recursively until all dependencies are resolved. To avoid an endless recursive call, upgrade will only be called if at least 1 package was retrieved (bool dependendPackageFetched).


source/dub/dub.d
```diff
@@ -557,6 +557,8 @@ class Dub {
 			return;
 		}
 
+		bool dependendPackageFetched;
+
 		foreach (p; versions.byKey) {
 			auto ver = versions[p]; // Workaround for DMD 2.070.0 AA issue (crashes in aaApply2 if iterating by key+value)
```
wilzbach: FWIW we don't support DMD 2.070.0 anymore (:

source/dub/dub.d
```diff
@@ -557,6 +557,8 @@ class Dub {
 			return;
 		}
 
+		bool dependendPackageFetched;
```
wilzbach: It's __dependent__, but I think sth. like `hasDependenciesFetched` or `hasDependentPackagesFetched` is probably even better?

andre2007:
This will take some time. Today ```Project``` contains only information that a dependency is missing but not the list of missing dependencies which could be used for ```packages_to_upgrade```.
First step is to enhance ```Project``` with a new property ```string[] missingDependencies() {return m_missingDependencies;}```.
This array could be filled in method ```reinit```

andre2007:
@wilzbach I did the changes proposed by you. Could you have another look?

andre2007:
@wilzbach ping :)

source/dub/project.d
```diff
@@ -116,6 +117,9 @@ class Project {
 	*/
 	bool hasAllDependencies() const { return m_hasAllDependencies; }
```
WebFreak001: can we change this to `return missingDependencies.length == 0;` and also make the `fetch` function remove from that instead of having that `hasDependentPackagesFetched` boolean in dub.d? I think it would be a lot more fail-safe for future changes on this

andre2007: Thanks for the feedback. The first part I understand, I will change it. The consequence is, the attribute m_hasAllDependencies can be removed at all.

The second part I am not sure. The whole purpose of hasDependentPackagesFetched is to avoid an endless loop due to the recursive function call. In case there are missing packages but they cannot be retrieved, without the Boolean flag there will be an endless loop. The Boolean flag guarantees that only if at least 1 package were successful fetched then the recursive function call will not lead to an endless loop.

As far as I understand your proposed change, endless loops could occur.

WebFreak001: fetch doesn't call fetch recursively right? After the fetch call it should then check if there are still dependencies unfetched

andre2007: I am not sure whether the proposed solution is getting quite complex to be prepared for future requirements which might come. The fetch is called within a loop. There might be further entries in "versions", which then be handled after the recursive call correctly or maybe incorrectly. Triggering recursive call to function upgrade might add side effects I cannot judge at the moment.

I have to little knowledge about the dub logic here, what consequences it has.
Are you still convinced, adding the recursive function call within the loop after the fetch is the right solution?
If yes, then I will go on with you proposal. If not, I propose to stay with the boolean flag solution until there are further requirements.

andre2007: @WebFreak001 m_hasAllDependencies  is removed as proposed

andre2007:
This pr might solve also #1180 but at the moment, it causes an endless loop 
for this:
```
dub init vibetest -t vibe.d
cd vibetest
dub build --cache local
```

I have to check what is going on and how to solve it.

andre2007:
Update:
if you use --cache=local the dependencies are stored correctly but they they are ignored by the "collecting dependencies" logic (https://github.com/dlang/dub/blob/master/source/dub/project.d#L318)

This is another bug not related to this pr. Before this pr the bug behavior was the error message "Non-optional dependency vibe-d of vibetest not found in dependency tree!?."
With this pr the bug behavior is an infinite loop. I check whether I can correct this additional bug as part of this pr or an independent pr.

source/dub/project.d
```diff
@@ -403,6 +410,7 @@ class Project {
 			}
 		}
 		collectDependenciesRec(m_rootPackage);
+		m_missingDependencies = m_missingDependencies.sort().array;
```
WebFreak001: `m_missingDependencies.sort();` will already sort the source array, you don't have to allocate a new one with `.array`

source/dub/dub.d
```diff
@@ -592,8 +592,17 @@ class Dub {
 			}
 		}
 
+		string[] missingDependenciesBeforeReinit = m_project.missingDependencies;
 		m_project.reinit();
 
+		if (!m_project.hasAllDependencies) {
+
+			string[] resolvedDependencies = setDifference(missingDependenciesBeforeReinit, m_project.missingDependencies).array;
```
WebFreak001: all you do is check `.empty` here, so why not store an auto with the setDifference?

```suggestion
			auto resolvedDependencies = setDifference(missingDependenciesBeforeReinit, m_project.missingDependencies);
```

source/dub/project.d
WebFreak001:
```suggestion
		m_missingDependencies.sort();
```

WebFreak001:
I really like the new logic, seems really easily understandable now

source/dub/project.d
andre2007: fixed

source/dub/dub.d
andre2007: fixed

andre2007:
@WebFreak001 Thank you a lot for your reviews.

buildkite/dub fails due to Memory Exceptions in vibe.d - this seems not related to my pr.
core.exception.InvalidMemoryOperationError@src/core/exception.d(706): Invalid memory operation

wilzbach:
Yep it's due to https://github.com/vibe-d/eventcore/pull/86

andre2007:
Anything left I can do to get this merged?

andre2007:
Ping :)

WebFreak001:
looking good imo, would merge if no objections

andre2007:
Ping :)

Issue #1531, opened 2018-08-18
support for "minDubVersion" recipe entry


rtbo:
Not all recipes may work with old dub versions. see #1450 or #1453.
Specifying the minimum required dub version in the json/sdl file allows meaningful error message instead of a build error the user would be unable to understand.

dhasenan:
More ideally, dub would use semantic versioning for things like this, and dub.json would allow you to rely on a range of dub versions. Then dub would automatically download a compatible version of itself, just like regular dependencies, and invoke that.

That's a much bigger change, but we can at least lay the groundwork by using `"dubVersion": "~>1.11.0"` instead of `"minDubVersion": "1.11.0"`.

dhasenan:
Related: https://github.com/dlang/dub/issues/1021

Issue #1021, opened 2016-12-17
Include D, and dub itself in build dependencies for reproducibility


biocyberman:
Hi I am a Java `maven` and Scala `sbt` user. One thing I like about both tools is that it allows me to build with a specific Java/Scala version, and in case of sbt, a specific build tool version. This is very helpful for reproducibility. I think the same idea can be implemented for `dub`. For example:

- When I find an interesting project P on github that have not been updated for 2 years, but I want to try and develop it. The delopers of P run and tested it with `dub` and `D` versions at that time, and everything was fine. But now I try with newer version of dub and D, many things broken, even though I still can download other dependencies with correct versions. The solution is to have dub definition file (i.e. dub.json) to include `dub` and `D` versions at that time. The build with be based on those versions. And the build guarantied to be successful for practically longer period. 

This is also very helpful for production uses, in which developers are not pressed to update the project whenever there are new `D` and `dub` versions. 

That's my 2c of idea to make dub and D more appealing for adaption.

p0nce:
Idea: DUB version could be something written in the generated `dub.selections.json` perhaps (an artifact that has to versionned if looking for reproducibility).

wilzbach:
> This is also very helpful for production uses, in which developers are not pressed to update the project whenever there are new D

See also: https://github.com/dlang/dub/issues/863

biocyberman:
@wilzbach Indeed, #863 issue is asking about something quite close to what this one is discussing. And to rephrase for clarity:

In essence, **A dub-backed project should be re-runnable from source**, in terms of time: after, say, 5 years; in terms of versions: with specified versions of dub, d, and direct dependencies. Dub should be able to pull everything by itself, not requiring developers to recreate the environment manually.

p0nce:
A minimum front-end version, while a thing that does exist in all D project (sometimes unbeknownst to the author), would not help recompile projects.

An old project would fail to compile anyway because of:
- SemVer not respected
- added language warnings/errors
- things being deprecated then phased out (5 years is larger than the deprecation period), often for good reasons
- regressions

So yeah the only way to build for sure would have to have the **exact** same setup as the original (deps and compiler).

If you want to have a project build in 5 years, you would have to version `dub.selections.json`.
That seems like DVM should be improved instead so it parses some dub input (or output).
I mean, does `cargo` handle Rust version? Nope, it comes with the compiler.

[...]

PR #1532, opened 2018-08-18
implement "minDubVersion" field

rtbo:
see #1531


WebFreak001:
let's speed this up a little, I think this is a fairly straightforward change that wouldn't get a lot of negative feedback and formally everything is there: changelog, tests, code

test/issue1531-min-dub-version/source/app.d
```diff
@@ -0,0 +1,3 @@
+module app;
+
+void main() {}
```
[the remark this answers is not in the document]
rtbo: `run_unittest.sh` also iterates the directories under test.
i.e. the package is built and build passes or fail depending on the `minDubVersion` specified. (checked out a one that passes)
```sh
$ DUB=... test/run_unittest.sh
Testing summary:
[INFO]  Building /home/remi/dev/dlang/dub/test/issue1531-min-dub-version/...
```
Now I guess I can have a script to test for both success and failure.

source/dub/package_.d
```diff
@@ -619,6 +620,34 @@ class Package {
...
+			static if (dubVersion[0] == 'v') {
+				enum dv = dubVersion[1 .. $];
+			}
+			else {
+				enum dv = dubVersion;
+			}
+			static assert(isValidVersion(dv));
+
+			enforce(compareVersions(dv, mdv) >= 0,
+				"dub-"~dv~" does not comply with minDubVersion specification: "~mdv);
```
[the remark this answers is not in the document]
rtbo: did this:
```
dub-1.10.0 does not comply with minDubVersion specification: 99.0.0.
Please consider upgrading your dub installation.
```

ohdatboi:
Discussing this in #1021 it got me thinking about `minDubVersion`. Maybe this is not the best name, since in the (bright) future we would have DMD following semver which will make it possible to add the same version-checking feature for dub and we would want to add `minDmdVersion`, `minLdcVersion`, `minGdcVersion`.

Why don't we use some new namespace for that, for example:
```json
{
  "name": "test",
  ...
  "devDependencies": {
    "dmd": "~>2.81.0",
    "dub": "~>1.10.0"
  }
}
```

It could be `requirements` or something like that. Actually, since LDC already follows semver and GDC too (well, when it gets merged into GCC), it could be implemented too.

@wilzbach, @WebFreak001 what's your opinion on that? Otherwise, this PR is fully complete  with test and changelog.

WebFreak001:
I like the idea but something else than devDependencies would be good. Node already uses it, making it even harder to differentiate between the files if they are called package.json and because they are not actual dependencies but only for checking. `requirements` would be good as it would also not imply that all dmd, ldc and gdc are required to be installed for the project to build if the project specifies versions for all 3 of them

test/issue1531-min-dub-version.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+
+DIR1531=${CURR_DIR}/issue1531-min-dub-version
+
+rm -rf $DIR1531/.dub/
+rm -rf $DIR1531/test-application*
+
+sed -i 's/^minDubVersion ".*"$/minDubVersion "1.0"/' ${DIR1531}/dub.sdl
```
wilzbach: It's not a good idea to mutate files that are checked in the source tree by the testsuite, better put the `dub.sdl` into a temporary directory.
Also since the last version dub supports passing files in from stdin, so you could use `--single` plus pass the file in via stdin which would eliminate the need for a temporary directory entirely.

test/issue1531-min-dub-version.sh
```diff
@@ -0,0 +1,19 @@
...
+${DUB} run --root ${DIR1531} || die "Did not pass minDubVersion \"1.0\""
+
+sed -i 's/^minDubVersion ".*"$/minDubVersion "99.0"/' $DIR1531/dub.sdl
+! ${DUB} run --root ${DIR1531} || die "Did pass minDubVersion \"99.0\"!"
+
```
wilzbach: Maybe a good idea to add one test that works, e.g. with `2.068` or similar?

rtbo: What do you mean by this? I already test for success and for failure.

WebFreak001:
Seems to be good now, will merge if there are no more objections

ZombineDev:
I think this change was aiming in the right direction, but it has one significant drawback - it doesn't integrate through the established dependency system.
dmd and dub are build dependencies just like any other package yours depends on, because without all the right versions of all of the above you can't build your package.
Not being able to specify `"dub": "~>1.0"` means that we won't be able to make a breaking change to dub and release version dub `2.0`, since any package with `"minDubVersion": "1.10"` will happily accept `2.0`.

ZombineDev:
I mean essentially what @ohdatboi was proposing, but either putting dub, dmd, ldc, etc. next to the other dependencies, or having a separate `toolchainDependencies` section.

MartinNowak:
Compatibility is a reasonable argument for `~>` style dependencies.
How about coming up with a PR that implements this instead of minDubVersion?

rtbo:
I agree with this. We should have this done before the first `v1.12` beta is out (in a week or so?), or otherwise have `minDubVersion` reverted.

MartinNowak:
Kudos for taking action on this, was a bit sad to see this leaking into the release. Very glad to see #1571 :).

wilzbach:
@MartinNowak this shouldn't be part of the real release anymore as it's reverted: https://github.com/dlang/dub/pull/1581

rtbo:
This has leaked in [changelog](https://dlang.org/changelog/2.083.0.html#minDubVersion) and beta1unfortunately.

wilzbach:
Yes, and that's not ideal, but it won't be part of the official release and that's what really matters.

PR #1541, opened 2018-08-24
Support targetType 'wasm'

lempiji:
WebAssembly is supported with LDC 1.11.0.
If wasm is added to the targetType, I think that development becomes easier.

```json
  "targetType": "wasm"
```

I wrote such a code. How about that?

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -171,9 +171,9 @@ private:
 	static bool pathMatch(string path, string pattern)
 	{
 		import std.functional : memoize;
-		
+
 		alias nativePath = memoize!((string stringPath) => NativePath(stringPath));
-			
+
 		return nativePath(path) == nativePath(pattern) || globMatch(path, pattern);
 	}
 
@@ -249,7 +249,8 @@ enum TargetType {
 	sourceLibrary,
 	dynamicLibrary,
 	staticLibrary,
-	object
+	object,
+	wasm
 }
 
 enum BuildRequirement {
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -167,6 +167,8 @@ class DMDCompiler : Compiler {
 				if (platform.platform.canFind("windows"))
 					return settings.targetName ~ ".obj";
 				else return settings.targetName ~ ".o";
+			case TargetType.wasm:
+				assert(false, "dub not support targetType: wasm");
 		}
 	}
 
@@ -189,6 +191,8 @@ class DMDCompiler : Compiler {
 			case TargetType.object:
 				settings.addDFlags("-c");
 				break;
+			case TargetType.wasm:
+				assert(false, "dub not support targetType: wasm");
 		}
 
 		if (tpath is null)
--- a/source/dub/compilers/gdc.d
+++ b/source/dub/compilers/gdc.d
@@ -161,6 +161,8 @@ class GDCCompiler : Compiler {
 				if (platform.platform.canFind("windows"))
 					return settings.targetName ~ ".obj";
 				else return settings.targetName ~ ".o";
+			case TargetType.wasm:
+				assert(false, "gdc not support targetType: wasm");
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


ohdatboi:
I don't think that it should be implemented as `targetType`.  Target type defines what will be produced: an executable or a library.

It'd be better to add `--arch=wasm32` option instead. After all, D programs are cross-platform and `--arch` is meant to force a different architecture (in this example, wasm).

WebFreak001:
@ZombineDev imo vendor doesn't make any sense in there. What could it possibly add that sys and abi not already describe?

ZombineDev:
@WebFreak001 the idea is to adapt a standard, well-known format and not invent our own.

WebFreak001:
@ZombineDev can we use something else than these triples then? Well if dub had a way to list all possible triple combinations it would be ok but otherwise it's just even worse than when using clang then. I mean otherwise even for your example triple combinations it would be basically impossible to find them out without looking them up somewhere online. That's like such a huge pain with these triples which is why I don't like them.

s-ludwig:
@WebFreak001 I think thís is a bit beside the point. If you want to cross compile to such a platform, you *need* to know the triple, no matter what the format is. On the other hand, for native compilation the existing architecture flag is going to stay, so the additional complexity won't show up.

Of course, we could define our own identifiers that map to such triples to simplify that, but then we could also just put up a documentation page with common triples. Also, the list would grow to a point where it will be just as hard to find out the right identifier for a given target system.

WebFreak001:
hm ok I can see advantages in regards with LDC compatibility. On DMD I guess it could be somewhat interpreted for some settings.

But if we have these triples how is dub going to decide on the build steps then? Especially linking will require different tools and different arguments potentially for every triple.

I think there should definitely be some documentation page with common triples and some custom identifiers mapping to some.

Also about the suggested triples: `x86_64-pc-windows-gnu` what would happen if I didn't specify pc here? (as you also didn't specify it for the android ones) There would need to be basically default values for everything based on other parameters. Also the range of valid parameters would basically be depending on all other parameters. Like with windows other valid systems could be xbox one or hololens but they clearly don't apply on other systems.

Also I think having `unknown` (or rather the undefined behaviour the name implies) would be bad design as it could change any release.

ZombineDev:
1. Yes for DMD we'll interpret the target triple initially, until we add `-mtriple` support to dmd as well.
2. The advantage is for compatability with C  toolchain, which LDC and GDC already support.
3. > But if we have these triples how is dub going to decide on the build steps then? Especially linking will require different tools and different arguments potentially for every triple

Good question. It depends on the compiler used. LDC's config file supports specifying details like this per target triple in its configuration file. So in the most basic case Dub will call LDC and LDC will handle the rest. AFAIU, for GDC you will have a cross-compiler for each target platform installed in a different folder. 
On Dub's side we may want to support specifying per target triple toolchain options (assembler, linker, C compiler, D compiler, system libs to link, etc.) to support more advanced use cases.

3. > I think there should definitely be some documentation page with common triples and some custom identifiers mapping to some.

Yes, definitely.

4. > Also about the suggested triples

Just to clarify: I didn't invent any of those (except for the `i686-pc-windows-dmc`?) triples. They're what's already established in the real world today.

> x86_64-pc-windows-gnu what would happen if I didn't specify pc here? (as you also didn't specify it for the android ones) There would need to be basically default values for everything based on other parameters. Also the range of valid parameters would basically be depending on all other parameters. Like with windows other valid systems could be xbox one or hololens but they clearly don't apply on other systems.

For vendor: as mentioned before most of the time you probably wouldn't need to specify it as the compiler will infer the right thing from the OS field. But it's there for those who might need it. Not specifying the vendor or typing unknown explicitly should always be safe as it means the compiler will either determine the right thing automatically, or make as few assumptions as possible when the OS is not specified - useful for platform bring up and/or bare metal development. 
As for the set of valid combination of values, it's up for the compiler to decide / support.

Geod24:
Superseeded by #1755 ?

kinke:
Yes;

> dub build --arch=wasm32-unknown-unknown-wasm

that works now with master.

lempiji:
I can build to the wasm, so I close this PR.
It seems that it may still fail with Windows, so I will report it separately.

PR #1548, opened 2018-08-29
Implement separate linking for LDC

s-ludwig:
Also enables separate linking for GDC when building with `--build-mode=separate`.

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -212,7 +212,19 @@ class LDCCompiler : Compiler {
 
 	void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback)
 	{
-		assert(false, "Separate linking not implemented for LDC");
+		import std.string;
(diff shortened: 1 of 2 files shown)


source/dub/compilers/ldc.d
```diff
@@ -224,4 +236,17 @@ class LDCCompiler : Compiler {
...
+	private static bool isLinkerDFlag(string arg)
+	{
+		switch (arg) {
+			default:
+				if (arg.startsWith("-defaultlib=")) return true;
+				if (arg.startsWith("-march=")) return true;
+				if (arg.startsWith("-fsanitize=")) return true;
+				return false;
+			case "-g", "-gc", "-shared", "-lib":
+				return true;
```
wilzbach: Shouldn't this be before the `default` case?

JabbaDa:
Where are we with this? I need this to build gdkd on the Pi. Is there anything I can do to help?

Geod24:
Superseeded (or rather, included) by #1755

PR #1549, opened 2018-08-29
Adding support for $DUB variable

rtbo:
Adding support for `$DUB` variable in build scripts.
```sdl
preBuildCommands "$DUB run --single somescript.d"
```
Resolves to `thisExePath`, unless `dub` was built as a library. In such case, it searches for the first dub application in `$PATH`.
This is useful if dub is not in the `$PATH` (really seldom i guess), or if several versions of dub are installed (that's more for devs than users).
$DUB env var is also set within build command processes.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -430,6 +430,65 @@ string getDUBVersion()
 	return verstr;
 }
 
+
+/**
+	Get current executable's path if running as DUB executable,
+	or find a DUB executable if DUB is used as a library.
+	For the latter, the following locations are checked in order:
+	$(UL
+		$(LI current working directory)
+		$(LI same directory as `compilerBinary` (if supplied))
+		$(LI all components of the `$PATH` variable)
+	)
+	Params:
+		compilerBinary = optional path to a D compiler executable, used to locate DUB executable
+	Returns:
+		The path to a valid DUB executable
+	Throws:
+		an Exception if no valid DUB executable is found
+*/
+public string getDUBExePath(in string compilerBinary=null)
+{
+	version(DubApplication) {
+		import std.file : thisExePath;
+		return thisExePath();
+	}
+	else {
+		// this must be dub as a library
+		import std.algorithm : filter, map, splitter;
+		import std.array : array;
+		import std.file : exists, getcwd;
+		import std.path : chainPath, dirName;
+		import std.range : chain, only, take;
+		import std.process : environment;
+
+		version(Windows) {
+			enum exeName = "dub.exe";
+			enum pathSep = ';';
+		}
+		else {
+			enum exeName = "dub";
+			enum pathSep = ':';
+		}
+
+		auto dubLocs = only(
+			getcwd().chainPath(exeName),
+			compilerBinary.dirName.chainPath(exeName),
+		)
+		.take(compilerBinary.length ? 2 : 1)
+		.chain(
+			environment.get("PATH", "")
+				.splitter(pathSep)
+				.map!(p => p.chainPath(exeName))
+		)
+		.filter!exists;
+
+		enforce(!dubLocs.empty, "Could not find DUB executable");
+		return dubLocs.front.array;
+	}
+}
+
+
 version(DubUseCurl) {
 	void setupHTTPClient(ref HTTP conn, uint timeout)
 	{
--- a/build-gdc.sh
+++ b/build-gdc.sh
@@ -17,7 +17,7 @@ echo "module dub.version_;" > source/dub/version_.d
 echo "enum dubVersion = \"$GITVER\";" >> source/dub/version_.d
 
 echo Running $GDC...
-$GDC -obin/dub -lcurl -w -fversion=DubUseCurl -Isource $* $LIBS @build-files.txt
+$GDC -obin/dub -lcurl -w -fversion=DubUseCurl -fversion=DubApplication -Isource $* $LIBS @build-files.txt
 echo DUB has been built as bin/dub.
 echo
 echo You may want to run
--- a/build.cmd
+++ b/build.cmd
@@ -7,7 +7,7 @@
 @echo enum dubVersion = "%GITVER%"; >> source\dub\version_.d
 
 @echo Executing %DC%...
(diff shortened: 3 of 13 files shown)


source/dub/internal/utils.d
```diff
@@ -342,6 +342,39 @@ string getDUBVersion()
...
+			enum exeName = "dub";
+		}
+
+		foreach (p; pathSplitter(environment["PATH"])) {
+			const exe = buildPath(p, exeName);
+			if (exists(exe) && isFile(exe)) {
+				return exe;
+			}
+		}
+		throw new Exception("Could not find dub application in $PATH");
```
WebFreak001: if the user (or whatever invokes dub) specified a compiler path, maybe you could search in the folder where the compiler lives for a dub executable. For example when it's a portable D install with a portable editor install which has an extension using dub as a library and invoking with the portable compiler

source/dub/internal/utils.d
```diff
@@ -342,6 +342,39 @@ string getDUBVersion()
...
+		import std.path : buildPath, pathSplitter;
+		import std.process : environment;
+
+		version(Windows) {
+			enum exeName = "dub.exe";
+		}
+		else {
+			enum exeName = "dub";
+		}
+
```
wilzbach: Check in the CWD first?

source/dub/internal/utils.d
```diff
@@ -342,6 +342,54 @@ string getDUBVersion()
 	return verstr;
 }
 
+/// Returns: current executable's path if running as dub executable,
+/// or as dub library, look for dub executable in the following locations
+///   - current working directory
+///   - same directory as compilerBinary (if supplied)
+///   - all components of the $PATH variable
+/// Throws: an Exception if nothing succeded.
```
wilzbach: Please use ddoc

source/dub/internal/utils.d
```diff
@@ -342,6 +342,54 @@ string getDUBVersion()
...
+		if (exists(cwdDub) && isFile(cwdDub)) {
+			return cwdDub;
+		}
+
+		if (compilerBinary) {
+			const exe = buildPath(dirName(compilerBinary), exeName);
+			if (exists(exe) && isFile(exe)) {
+				return exe;
+			}
+		}
```
wilzbach: Have a look at https://github.com/dlang/dmd/blob/74281dc320329953b44a20f4aa2dfaacc967cc64/src/dmd/frontend.d#L121 for an idea on how this could be simplified with ranges

source/dub/internal/utils.d
```diff
@@ -342,6 +342,54 @@ string getDUBVersion()
...
+
+		foreach (p; pathSplitter(environment["PATH"])) {
+			const exe = buildPath(p, exeName);
+			if (exists(exe) && isFile(exe)) {
+				return exe;
```
wilzbach: This is repeated three times.

rtbo:
Please tell when I'm good to squash

source/dub/internal/utils.d
```diff
@@ -342,6 +342,63 @@ string getDUBVersion()
...
+			enum pathSep = ';';
+		}
+		else {
+			enum exeName = "dub";
+			enum pathSep = ':';
+		}
+
+		auto dubLocs = only(
+			getcwd().buildPath(exeName),
+			compilerBinary ? compilerBinary.dirName.buildPath(exeName) : null,
```
WebFreak001: you should be doing `compilerBinary.length ? ...` here instead so if an empty string is passed it won't try to find the exe there.

```suggestion
			compilerBinary.length ? compilerBinary.dirName.buildPath(exeName) : null,
```

source/dub/internal/utils.d
```diff
@@ -342,6 +342,63 @@ string getDUBVersion()
...
+		)
+		.chain(
+			environment.get("PATH", "")
+				.splitter(pathSep)
+				.map!(p => p.buildPath(exeName))
+		)
+		.filter!exists;
+
+		enforce(!dubLocs.empty, "Could not find DUB executable");
+		return dubLocs.front;
```
WebFreak001: can we also use `chainPath` on all of these because they don't allocate? :p

WebFreak001:
I think this has been sitting long enough

rtbo:
rebased on master and squashed

WebFreak001:
hm issue616-describe-vs-generate-commands seems to keep erroring. Should we fix that first or merge now? @wilzbach

wilzbach:
I am only on my phone, so can't check whether master is erroring because of this too.
If it's not, then this needs to be fixed here first.

rtbo:
master passes on both semaphore and travis.
I'll have this checked.

WebFreak001:
can you merge in master again?

wilzbach:
> can you merge in master again?

Master will be merged in when you restart the builds on Travis which is what I have done just now, so yeah I think we need to fix Travis here before we can merge this :/

wilzbach:
> can you merge in master again?

Also we don't merge in master, but rebase ;-)

WebFreak001:
looking good, is passing. Nice catch with the environment variable name and escaped values.

thewilsonator:
Please squash again.

WebFreak001:
`dub add` unittest makes this fail with old dmd because of missing/changed feature in a dependency (mir) in the test.

imo we should drop these pretty old dmd versions in support, how many people actually build and update dub from source for years without upgrading dmd?

wilzbach:
IIRC We can't. We need to support the latest GDC version though @ibuclaw might not require this anymore now that the GDC master is up to date with the frontend.

andre2007:
Fix for travis ci: https://github.com/dlang/dub/pull/1598

andre2007:
By restarting the checks, the pull request should go to green.
I wonder whether the dlang bot can be enhanced: by writing a comment with the content "restart tests" the bot could restart the tests.
Would be quite nice :)

wilzbach:
https://github.com/dlang/dlang-bot - it's open source. The biggest problem is that we use so many different CIs.

rtbo:
travis failure seems unrelated

thewilsonator:
Seems so.

PR #1552, opened 2018-09-04
Fix #1551: Escaping a variable ($$) is ignored

### LINKED ISSUE #1551: Regression: Dub v1.11.0 single mode is broken
It breaks SemaphoreCI for DMD, effectively stalling all PRs ( https://semaphoreci.com/dlang/dmd-2/branches/pull-request-8575/builds/69 )

```
# In dmd/test/dub_package/
% dub --version
DUB version 1.10.0, built on Jul  4 2018
% dub --single frontend.d
Invalid source/import path: /Users/geod24/projects/dmd/generated/dub
Running pre-generate commands for dmd:frontend...
dmd:frontend ~argv0-slice: building configuration "library"...
frontend ~master: building configuration "application"...
Linking...
Running ./frontend
core.exception.AssertError@../../src/dmd/frontend.d(208): No valid config found.
----------------
??:? _d_assert_msg [0xd2b1b5e]
../../src/dmd/frontend.d:208 std.algorithm.iteration.MapResult!(std.path.buildNormalizedPath, std.algorithm.iteration.UniqResult!(std.functional.binaryFun!("a == b", "a", "b").binaryFun, std.range.SortedRange!(immutable(char)[][], "a < b").SortedRange).UniqResult).MapResult dmd.frontend.findImportPaths() [0xd1a8208]
frontend.d:14 _Dmain [0xd0e7609]
Program exited with code 1
dub --single frontend.d  6.78s user 1.36s system 93% cpu 8.731 total
```

```
% ~/projects/dub/bin/dub --version
DUB version 1.11.0, built on Sep  4 2018
% ~/projects/dub/bin/dub --single frontend.d
Invalid variable: DUB_PACKAGE_DIR
```

Geod24:
```
This was broken by the refactoring in #1392 (commit 14c00c49a9e756e93653b5e83e22fa05923cde67)
```

### DIFF
--- a/test/issue1551-var-escaping/dub.json
+++ b/test/issue1551-var-escaping/dub.json
@@ -1,0 +1,7 @@
+{
+	"name": "issue1551-var-escaping",
+    "preGenerateCommands": [
+        "echo $${DUB_PACKAGE_DIR}",
+        "echo $$DUB_PACKAGE_DIR"
+    ]
+}
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1185,8 +1185,10 @@ private string processVars(Project, Package)(string var, in Project project, in 
 {
 	import std.regex : regex, replaceAll;
 
-	auto varRE = regex(`\$([\w_]+)|\$\{([\w_]+)\}`);
-	var = var.replaceAll!(m => getVariable(m[1].length ? m[1] : m[2], project, pack, gsettings))(varRE);
+	auto varRE = regex(`\$([\w_]+)|\$\{([\w_]+)\}|(\$\$[\w_]+|\$\$\{[\w_]+\})`);
+	var = var.replaceAll!(
+		m => m[3].length ? m[3][1..$] : (getVariable(m[1].length ? m[1] : m[2], project, pack, gsettings))
+		)(varRE);
 	if (is_path) {
 		auto p = NativePath(var);
 		if (!p.absolute) {
@@ -1310,6 +1312,8 @@ unittest
 	assert(processVars("Hello $PACKAGE_DIR"~dirSeparator~"foobar", proj, pack, gsettings, !isPath) == "Hello "~(pack.path ~ "foobar").toNativeString);
 	// test with isPath
 	assert(processVars("local", proj, pack, gsettings, isPath) == (pack.path ~ "local").toNativeString);
+	assert(processVars("foo/$$ESCAPED", proj, pack, gsettings, isPath) == (pack.path ~ "foo/$ESCAPED").toNativeString);
+	assert(processVars("$$ESCAPED", proj, pack, gsettings, !isPath) == "$ESCAPED");
 	// test other env variables
 	import std.process : environment;
 	environment["MY_ENV_VAR"] = "blablabla";
@@ -1317,6 +1321,7 @@ unittest
 	assert(processVars("${MY_ENV_VAR}suffix", proj, pack, gsettings, !isPath) == "blablablasuffix");
 	assert(processVars("$MY_ENV_VAR-suffix", proj, pack, gsettings, !isPath) == "blablabla-suffix");
 	assert(processVars("$MY_ENV_VAR:suffix", proj, pack, gsettings, !isPath) == "blablabla:suffix");
+	assert(processVars("$MY_ENV_VAR$MY_ENV_VAR", proj, pack, gsettings, !isPath) == "blablablablablabla");
 	environment.remove("MY_ENV_VAR");
 }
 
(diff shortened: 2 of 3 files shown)


Geod24:
Fixed the unittest failures (and added some in the process).

Geod24:
@ZombineDev Pointed out that this solution was not working for `$VAR$VAR` and suggested a much better alternative, which I have adopted here.

ZombineDev:
The main problem is that the expected behavior is not thoroughly documented. I testes 3 solutions of which one of the first two is probably correct, but I'm not sure which: https://run.dlang.io/is/4pg7bz

@s-ludwig @MartinNowak what is the expected behavior, given input like this:
```D
const input = "$ $$ $$$ $$$$ $$$$$ $$$$$$ asd$$var1$var2$var3${var4}${var5}$${var6}$$var7$${var8} a$var9 b$$var10 c$$$var11 d$$$$var12 e$$$$$var13 f$$$$$$var14 g$$$$$$$var15 h$$$$$$$$var16 $$&$!$$?$$$$var18";
```
?

Geod24:
In doubt, I think we should just revert back to the old behavior.

ZombineDev:
> In doubt, I think we should just revert back to the old behavior.

That's not an option as commit 14c00c49a9e756e93653b5e83e22fa05923cde67 is not a pure refactoring - it implements the support for the `${var}` syntax and reverting the code to the previous version would break users of the new syntax.

I will ask on forums for feedback on the expected behavior.

s-ludwig:
I'd say that we should do what sh does, since this is already heavily inspired by their syntax.

ZombineDev:
https://forum.dlang.org/post/exerdovrnivrsloetxtg@forum.dlang.org

ZombineDev:
> I'd say that we should do what sh does, since this is already heavily inspired by their syntax.

AFAIK, in most shell languages, `$` is escaped as `\$`. I thought that Make is the original inspiration, as its the only tool I know that uses this syntax (`$$`).

From http://www.tldp.org/LDP/abs/html/internalvariables.html: 
> $$
>    * Process ID (PID) of the script itself

wilzbach:
> https://forum.dlang.org/post/exerdovrnivrsloetxtg@forum.dlang.org

I personally like Martin's approach though as we already broke this and apparently almost no one used this, it really makes sense to adopt the more shell-like `\$` escape technique.

ZombineDev:
@wilzbach note that, because the backslash also needs to be escaped in JSON/SDL, so the actual syntax would be `\\$`.

ZombineDev:
@wilzbach / @s-ludwig are you ok with force merging? ("force" because AppVeyor is broken.)

Geod24:
And LDC v1.12.0 was released with this broken dub, and SemaphoreCI for DMD is broken again.

test/issue1551-var-escaping/dub.json
```diff
@@ -0,0 +1,7 @@
+{
+	"name": "issue1551-var-escaping",
+    "preGenerateCommands": [
+        "echo $${DUB_PACKAGE_DIR}",
+        "echo $$DUB_PACKAGE_DIR"
```
wilzbach: Not a real test.

Geod24: Good point, that looks like a leftover. But this test was written before the unittests were put in place.

Geod24:
Can we have a point release @wilzbach ?

wilzbach:
Single point releases here wouldn't help you for dmd as there only the DMD bundle gets downloaded.
For point releases, please contact @MartinNowak 

I think we missed 2.082.1 closely (https://forum.dlang.org/post/ppokpt$23ql$1@digitalmars.com).
Though there should be a beta for 2.083 soon: https://dlang.org/changelog/release-schedule.html

Geod24:
Sorry, meant patch not point (this went to stable).
I know it gets bundled, the next step would have been to talk to the LDC folks and Martin for that, so we can remove the hacks in DMD.

wilzbach:
I'm aware, but the beta for 2.083 has already been released: https://forum.dlang.org/post/pq7941$mfg$1@digitalmars.com
So in two weeks you can remove the hack at DMD.

jacob-carlborg:
@wilzbach Dub shipped with 2.083.0-beta.1  does not seem to include this fix [1].

[1] https://travis-ci.com/jacob-carlborg/dlp/jobs/154389016

wilzbach:
Yes, it was merged a few hours after the beta tagging, but it's on stable and thus will be part of the upcoming release.

jacob-carlborg:
> Yes, it was merged a few hours after the beta tagging, but it's on stable and thus will be part of the upcoming release.

Ok. It's a bit annoying that `- dmd` (to get the latest DMD) cannot be used in Travis CI.

wilzbach:
This should now be part of "beta" as the second beta has been released today and GitHub also shows the tag on the commit:

image

PR #1559, opened 2018-09-09
Fix detection for other msvc triples

thewilsonator:
e.g. x86-pc-win32-msvc

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -150,7 +150,7 @@ class LDCCompiler : Compiler {
 			~platform.compilerBinary~" -version\" failed with exit code "
 			~result.status.to!string()~".");
 
-		bool generates_coff = result.output.splitLines.find!(l => l.strip.toLower.startsWith("default target:")).front.canFind("-windows-msvc");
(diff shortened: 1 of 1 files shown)


source/dub/compilers/ldc.d
```diff
@@ -150,7 +150,7 @@ class LDCCompiler : Compiler {
 			~platform.compilerBinary~" -version\" failed with exit code "
 			~result.status.to!string()~".");
 
-		bool generates_coff = result.output.splitLines.find!(l => l.strip.toLower.startsWith("default target:")).front.canFind("-windows-msvc");
+		bool generates_coff = result.output.splitLines.find!(l => l.strip.toLower.startsWith("default target:")).front.canFind("msvc");
```
ZombineDev: Pedantically speaking, the `cygwin` and `mingw` ABIs also produce COFF object files, but
1. They're not well supported in neither LDC or GDC
2. I intend to solve this with `TargetTriple("...").objectFormat == ObjectFormat.COFF` in the future.

ZombineDev:
(Merged as the CI failures were all due to network timeouts.)

source/dub/compilers/ldc.d
ibuclaw: GDC doesn't support MSVC at all.  Compiler backend is support, no one is actively building or checking the runtime library against mingw or cygwin though.

PR #1560, opened 2018-09-09
Cache the COFF generation status for LDC.

s-ludwig:
This can speed up the up-to-date check considerably when using LDC2 on Windows.


wilzbach:
Maybe at some point we could cache such global state over multiple invocations of dub, though that's a bit tricky as we probably would need to track the timestamp of the compiler binary and other environment variables ...

PR #1562, opened 2018-09-14
Environment variable DUB_PACKAGE_VERSION added

andre2007:
The new environment variable DUB_PACKAGE_VERSION can be used in shell / batch scripts called from "preGenerateCommands". The version of the package can be written to module file.
This would simplify e.g. this logic https://github.com/dlang/dub/blob/master/build.cmd

Is a unit test needed for the new environment variable?

### DIFF
--- a/changelog/env_var_package_version.dd
+++ b/changelog/env_var_package_version.dd
@@ -1,0 +1,3 @@
+Environment variable DUB_PACKAGE_VERSION added
+
+DUB now supports the environment variable DUB_PACKAGE_VERSION containing the version of the package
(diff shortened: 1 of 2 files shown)


andre2007:
@ZombineDev done:)

wilzbach:
This will be a lot nice than e.g. this hack here: https://github.com/dlang-community/D-Scanner/blob/master/dub.json#L27

WebFreak001:
well it will sadly not be enough to fix it recompiling every time just yet, but useful nevertheless.

andre2007:
Documents update https://github.com/dlang/dub-docs/pull/7

PING :)

PR #1566, opened 2018-09-18
Refactor retry download

andre2007:
The logic for download retry was duplicated. This pull requests moves the retry logic into the internal
utils module. In addition the old logic had the flaw, in case of CurlException no retry was done.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -325,6 +325,94 @@ ubyte[] download(URL url, uint timeout = 8)
 	return download(url.toString(), timeout);
 }
 
+/**
+	Downloads a file from the specified URL with retry logic.
+
+	Downloads a file from the specified URL with up to n tries on failure
+	Throws: `Exception` if the download failed or `HTTPStatusException` after the nth retry or
+	on "unrecoverable failures" such as 404 not found
+	Otherwise might throw anything else that `download` throws.
+	See_Also: download
+**/
+void retryDownload(URL url, NativePath filename, size_t retryCount = 3)
+{
+	foreach(i; 0..retryCount) {
+		version(DubUseCurl) {
+			try {
+				download(url, filename);
+				return;
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) throw e;
+				else {
+					logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+					if (i == retryCount - 1) throw e;
+					else continue;
+				}
+			}
+			catch(CurlException e) {
+				logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+				continue;
+			}
+		}
+		else
+		{
+			try {
+				download(url, filename);
+				return;
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) throw e;
+				else {
+					logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+					if (i == retryCount - 1) throw e;
+					else continue;
+				}
+			}
+		}
+	}
+	throw new Exception("Failed to download %s".format(url));
+}
+
+///ditto
+ubyte[] retryDownload(URL url, size_t retryCount = 3)
+{
+	foreach(i; 0..retryCount) {
+		version(DubUseCurl) {
+			try {
+				return download(url);
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) throw e;
+				else {
+					logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+					if (i == retryCount - 1) throw e;
+					else continue;
+				}
+			}
+			catch(CurlException e) {
+				logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+				continue;
+			}
+		}
+		else
+		{
+			try {
+				return download(url);
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) throw e;
+				else {
+					logDebug("Failed to download %s (Attempt %s of %s)", url, i + 1, retryCount);
+					if (i == retryCount - 1) throw e;
+					else continue;
+				}
+			}
+		}
+	}
+	throw new Exception("Failed to download %s".format(url));
+}
+
 /// Returns the current DUB version in semantic version format
 string getDUBVersion()
 {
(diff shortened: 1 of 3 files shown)


source/dub/packagesuppliers/maven.d
```diff
@@ -92,27 +88,16 @@ class MavenRegistryPackageSupplier : PackageSupplier {
...
+
+		try
+			xmlData = cast(string)retryDownload(url);
+		catch(HTTPStatusException e) {
+			if (e.status == 404) {
+				logDebug("Maven metadata %s not found at %s (404): %s", packageId, description, e.msg);
+				return Json(null);
 			}
+			else throw e;
 		}
```
WebFreak001: this now behaves differently. Before it would have just tried to parse an empty xml after 3 fails and put that in the cache, but now it throws after 3 fails. Though this seems like it was a bug before so I think this is a good change here

source/dub/packagesuppliers/registry.d
```diff
@@ -92,28 +87,18 @@ class RegistryPackageSupplier : PackageSupplier {
...
+			jsonData = cast(string)retryDownload(url);
+		catch(HTTPStatusException e) {
+			if (e.status == 404) {
+				logDebug("Package %s not found at %s (404): %s", packageId, description, e.msg);
+				return Json(null);
 			}
+			else throw e;
 		}
+
 		Json json = parseJsonString(jsonData, url.toString());
```
WebFreak001: same as above

source/dub/internal/utils.d
```diff
@@ -325,6 +325,83 @@ ubyte[] download(URL url, uint timeout = 8)
 	return download(url.toString(), timeout);
 }
 
+/**
+	Downloads a file from the specified URL with retry logic.
```
WebFreak001: I think the documentation could be improved here. Something like
```
Downloads a file from the specified URL with up to 3 trys on failure
Throws: `Exception` if the download failed or `HTTPStatusException` on "unrecoverable failures" such as 404 not found. Otherwise might throw anything else that `download` throws.
See_Also: download
```

andre2007:
@WebFreak001 thanks for your feedback, I added the description.

I also corrected the 2 fetchPackage functions. In case  a HTTPStatusException with status code != 404 was caught, no new Exception was thrown.

source/dub/internal/utils.d
andre2007: done

source/dub/packagesuppliers/registry.d
```diff
@@ -57,19 +57,16 @@ class RegistryPackageSupplier : PackageSupplier {
...
-					continue;
-				}
-			}
+		try {
+			retryDownload(url, path);
+			return;
+		}
+		catch(HTTPStatusException e) {
+			if (e.status == 404) throw e;
+			else logDebug("Failed to download package %s from %s", packageId, url);
```
WebFreak001: this code can't trigger, if a HTTPStatusException is ever thrown it's only a 404 one (because you catch the rest). Only at the end you throw a normal Exception. Maybe rethrow the last HTTPStatusException on failure after 3 retries instead?

andre2007: HTTPStatusException is now rethrown
The throw Exception on last line of retryDownload I kept in, because of the CurlException

source/dub/internal/utils.d
```diff
@@ -325,6 +325,92 @@ ubyte[] download(URL url, uint timeout = 8)
...
+			}
+			catch(HTTPStatusException e) {
+				if (e.status == 404) throw e;
+				else {
+					logDebug("Failed to download %s (Attempt %s of 3)", url, i + 1);
+					if (i == 2) throw e;
+					else continue;
+				}
+			}
+			catch(CurlException e) {
```
wilzbach: Why not putting this only in a `version` block?

source/dub/internal/utils.d
```diff
@@ -325,6 +325,92 @@ ubyte[] download(URL url, uint timeout = 8)
...
```
wilzbach: Why not putting this only in a `version` block?

source/dub/internal/utils.d
```diff
@@ -325,6 +325,92 @@ ubyte[] download(URL url, uint timeout = 8)
...
+				else {
+					logDebug("Failed to download %s (Attempt %s of 3)", url, i + 1);
+					if (i == 2) throw e;
+					else continue;
+				}
+			}
+		}
+	}
+	throw new Exception("Failed to download %s".format(url));
+}
```
wilzbach: One empty line between functions

source/dub/internal/utils.d
```diff
@@ -325,6 +325,92 @@ ubyte[] download(URL url, uint timeout = 8)
...
+	Downloads a file from the specified URL with retry logic.
+
+	Downloads a file from the specified URL with up to 3 tries on failure
+	Throws: `Exception` if the download failed or `HTTPStatusException` on "unrecoverable failures" such as 404 not found.
+	Otherwise might throw anything else that `download` throws.
+	See_Also: download
+**/
+void retryDownload(URL url, NativePath filename)
+{
+	foreach(i; 0..3) {
```
wilzbach: Probably a good idea to make n configurable if we already generalize this function?

source/dub/internal/utils.d
andre2007: I am not sure whether I understood your question correctly.
Did you asked why I didn't put the the catch(CurlException) only
In a version scope?
I tried this but got syntax errors. It seems it isn't allowed.

source/dub/internal/utils.d
andre2007: done

source/dub/internal/utils.d
andre2007: done

source/dub/internal/utils.d
```diff
@@ -325,6 +325,93 @@ ubyte[] download(URL url, uint timeout = 8)
 	return download(url.toString(), timeout);
 }
 
+/**
+	Downloads a file from the specified URL with retry logic.
+
+	Downloads a file from the specified URL with up to n tries on failure
+	Throws: `Exception` if the download failed or `HTTPStatusException` on "unrecoverable failures" such as 404 not found.
```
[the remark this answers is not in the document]
andre2007: done

andre2007:
Ping :)

PR #1568, opened 2018-09-24
Fix 1180: local cache broken

andre2007:
If you use --cache=local

e.g.
```
dub init vibetest -t vibe.d
cd vibetest
dub build --cache=local
```

the dependencies are stored in the package folder but ignored by dub. That was the reason why for
the example the error message "Non-optional dependency vibe-d of vibetest not found in dependency tree!?." is raised.

This pull request propose to store the packages for --cache=local in folder ```<dub-package-folder>/.dub/packages``` instead of  ```<dub-package-folder>/```

This new location is also added the the package folders, Dub is aware of.

Benefits
* no conflicts with sub packages
* dependencies are cached in a structured manner and the noise in the dub package folder is removed
* as .dub folder is already in gitignore, the cached dependencies are also not tracked by git
* consistent with user and system cache


andre2007:
Also a questions, should the dependencies which are retrieved with cache=local should be stored in a sub folder "packages" of the dub package folder?

At the moment they are stored directly in dub package folder. 

By storing them in a sub folder the developer has a better overview and it is more consistent with the user and system cache.

source/dub/packagemanager.d
```diff
@@ -37,9 +37,10 @@ class PackageManager {
 		bool m_disableDefaultSearchPaths = false;
 	}
 
-	this(NativePath user_path, NativePath system_path, bool refresh_packages = true)
```
WebFreak001: Please add a new constructor, this argument is breaking the API.

test/issue1180-local-cache-broken/dub.json
```diff
@@ -0,0 +1,7 @@
+{
+	"name": "test",
+	"dependencies": {
+		"local-cache-lib": "~master"
```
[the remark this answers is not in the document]
andre2007: It is hard to answer here, because I do not know whether --cache=local is fully implemented and works as designed or a half implemented feature. There is no documentation.

Yes, as the depencenies(fetched with cache local) stored in the package folder, this could conflict with sub modules. My idea is to enhance the cache=local functionality and store the depencenies in a subfolder "packages". 

But maybe there is a reason why cache=local works as it works. Is the idea that the developer adds the path by hand to dub json?

andre2007:
I understand your last answer as response to my proposal?
In my opinion the developers expects that cache=local works similar to the user/system cache. The dependencies are retrieved, stored and automatically found by dub. If the intention is that the developers need to add these depencies by hand to dub.json maybe we could document it.

WebFreak001:
Well the cache is not something supposed to be added or edited manually (for permanent solutions) which this PR would imply and allow people to do.

Reasons are that for example if vibe would start to use this as way to depend on some local package then anyone could just create a malicious package on the dub registry with the same name as the now depended package, running some prebuild command messing with your local machine.

Also the other way around if the local packages are higher in precedence it could happen that you add a subpackage with the same name as a dub package you would like to depend on (for example you make a opengl subpackage for your package)

andre2007:
I need some hours to reason about. For the moment I am not sure what steps needs to be done.
* store packages in $dub_package_folder (current implementation)
* store packages in $dub_package_folder/packages
* store packages in $dub_package_folder/.dub/packages 
* remove cache=local
* other solution?

source/dub/packagemanager.d
andre2007: done

andre2007:
I implemented option 3. Please see pr description for the benefits. Test is adapted.

source/dub/dub.d
```diff
@@ -804,7 +804,7 @@ class Dub {
 
 		NativePath placement;
 		final switch (location) {
-			case PlacementLocation.local: placement = m_rootPath; break;
+			case PlacementLocation.local: placement = m_rootPath ~ ".dub/packages/"; break;
```
wilzbach: Shouldn't we use buildPath for Windows support?

source/dub/packagemanager.d
```diff
@@ -39,9 +39,20 @@ class PackageManager {
...
+			Repository(user_path ~ "packages/"),
+			Repository(system_path ~ "packages/")];
+
+		if (refresh_packages) refresh(true);
+	}
+
+    this(NativePath package_path, NativePath user_path, NativePath system_path, bool refresh_packages = true)
+	{
+		m_repositories = [
+			Repository(package_path~ ".dub/packages/"),
```
wilzbach: buildPath

test/issue674-concurrent-dub.sh
```diff
@@ -15,5 +15,4 @@ sleep 0.5
 cd ${TMPDIR} && $DUB fetch --cache=local bloom &
 pid2=$!
 wait $pid1
-wait $pid2
-[ -d ${TMPDIR}/bloom* ]
```
wilzbach: Isn't this nullifying the test? Why was this necessary?

andre2007: Fixed

source/dub/dub.d
andre2007: It is working fine on windows, as placement is of type NativePath

source/dub/packagemanager.d
andre2007: It is working fine on windows, as placement is of type NativePath

andre2007:
The local cache makes most sense for me on CI servers like Jenkins.
For my projects I want always an empty package list. The easiest way is to use cache=local

Also after the build, I need all of my project files and all dependencies source code files for further processing. For example I want to execute WhiteSource on them which is a tool for IP scanning. This is also quite easy with cache=local.

wilzbach:
I always hacked around this with `HOME=$(pwd) dub...`, but yeah I can now see from what point you are coming from. Thanks!

wilzbach:
Though I would love to hear a third option on this before merging, @s-ludwig or @martinnowak maybe?

Geod24:
Another reason the user would want to have those packages locally is for interaction with tools, such as IDE, or simply `docker`, as `docker build` calling `dub build` would ALWAYS re-download the packages by default.

@andre2007 : Can you rebase on master ?

andre2007:
@Geod24 it is rebased now.

Issue #1504, opened 2018-07-05
stringImportPaths + $ENV_VAR always yield an error message, but works


p0nce:
### System information

- **dub version**: DUB version 1.9.0, built on Jun  7 2018
- **OS Platform and distribution**: Windows 10
- **compiler version** DMD32 D Compiler v2.080.1

### Bug Description

It seems you can't use environment variable within stringImportPaths without having error messages.
Those error messages don't correspond to a real bug.

(We use this to convert C++ headers to D at compile-time.)

### How to reproduce?

Minimal `dub.json`

```json
{
    "name": "test-thing",
    "stringImportPaths": ["$MY_VAR"],
    "targetType": "none"
}
```
Set the `MY_VAR` environment variable to an existing, valid directory.

Output is:
```
C:\Users\ponce\Desktop\dub-test-case>dub
Invalid source/import path: C:\Users\ponce\Desktop\dub-test-case\$MY_VAR
Main package must have a binary target type, not none. Cannot build.
```

Same thing with `importPaths` and `sourcePaths`.

### Expected Behavior

I would expect the same output, but without the error message.

### Logs

C:\Users\ponce\Desktop\dub-test-case>dub -v
Using dub registry url 'https://code.dlang.org/'
Refreshing local packages (refresh existing: true)...
Looking for local package map at C:\ProgramData\dub\packages\local-packages.json
Looking for local package map at C:\Users\ponce\AppData\Local\dub\packages\local-packages.json
Note: Failed to determine version of package test-thing at .. Assuming ~master.
Refreshing local packages (refresh existing: false)...
Looking for local package map at C:\ProgramData\dub\packages\local-packages.json
Looking for local package map at C:\Users\ponce\AppData\Local\dub\packages\local-packages.json
Refreshing local packages (refresh existing: false)...
Looking for local package map at C:\ProgramData\dub\packages\local-packages.json
Looking for local package map at C:\Users\ponce\AppData\Local\dub\packages\local-packages.json
Checking for upgrades.
Using cached upgrade results...
Generating using build
Invalid source/import path: C:\Users\ponce\Desktop\dub-test-case\$MY_VAR
Configuring dependent test-thing, deps:
Main package must have a binary target type, not none. Cannot build.

0xEAB:
I can reproduce this "bug".

In case someone using Windows wants to quickly check it out himself, I wrote the following handy test script:
```batch
REM     Reproduction script for dlang/dub issue #1504

cd %~dp0

[...]

PR #1569, opened 2018-10-02
Fix Issue #1504

p0nce:
This also checks environment variable $ENV_VAR or ${ENV_VAR} in `importPaths`, `stringImportPaths`, etc.

### DIFF
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -20,6 +20,7 @@ import std.array : join, split;
 import std.exception : enforce;
 import std.file;
 import std.range;
+import std.process : environment;
 
 
 /**
@@ -179,6 +180,9 @@ struct BuildSettingsTemplate {
 		{
 			auto files = appender!(string[]);
 
+			import dub.project : buildSettingsVars;
+			auto envVars = environment.toAA();
+
 			foreach (suffix, paths; paths_map) {
 				if (!platform.matchesSpecification(suffix))
 					continue;
@@ -188,9 +192,8 @@ struct BuildSettingsTemplate {
 					auto path = NativePath(spath);
 					if (!path.absolute) path = base_path ~ path;
 					if (!existsFile(path) || !isDir(path.toNativeString())) {
-						import dub.project : buildSettingsVars;
 						import std.algorithm : any, find;
-						const hasVar = buildSettingsVars.any!((string var) {
+						const hasVar = chain(buildSettingsVars, envVars.byKey).any!((string var) {
 							return spath.find("$"~var).length > 0 || spath.find("${"~var~"}").length > 0;
 						});
 						if (!hasVar)
--- a/test/issue1504-envvar-in-path.sh
+++ b/test/issue1504-envvar-in-path.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -e
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+
+rm -rf ${CURR_DIR}/issue1504-envvar-in-path/.dub
+rm -rf ${CURR_DIR}/issue1504-envvar-in-path/test
+rm -rf ${CURR_DIR}/output-1504.txt
+
+
+export MY_VARIABLE=teststrings 
+# pragma(msg) outputs to stderr
+${DUB} build --force --root ${CURR_DIR}/issue1504-envvar-in-path 2> ${CURR_DIR}/output-1504.txt
+
+grep "env_variables_work" < ${CURR_DIR}/output-1504.txt
+
+# Don't manage to make it work
+#grep "Invalid source" < ${CURR_DIR}/output-1504.txt && true
+
--- a/test/issue1504-envvar-in-path/dub.json
+++ b/test/issue1504-envvar-in-path/dub.json
@@ -1,0 +1,5 @@
+{
+    "name": "test",
+    "stringImportPaths": ["$MY_VARIABLE"]
+}
+
--- a/test/issue1504-envvar-in-path/source/app.d
+++ b/test/issue1504-envvar-in-path/source/app.d
@@ -1,0 +1,5 @@
+pragma(msg, import("message.txt"));
+
+void main()
+{    
+}
--- a/test/issue1504-envvar-in-path/teststrings/message.txt
+++ b/test/issue1504-envvar-in-path/teststrings/message.txt
@@ -1,0 +1,1 @@
+env_variables_work


WebFreak001:
One bug this introduces is that if AA.byKey isn't implemented as input range (with save method) then the modified code path will only work properly once (because `chain` only implements save if all ranges implement save).

As AA is implementation dependent this is a possibility, but considering dmd and ldc implement byKey as input range I think this is ok. Thoughts? @wilzbach @s-ludwig

p0nce:
The PDF spec says nothing but I've found this:
https://dlang.org/spec/hash-map.html

It says:
> `.byKey()`  Returns a **forward range** suitable for use as a ForeachAggregate to a ForeachStatement         which will iterate over the keys of the associative array.

And a forward range has `save()` https://dlang.org/phobos/std_range_primitives.html#isForwardRange so by virtue of `chain` implementation:
https://github.com/dlang/phobos/blob/master/std/range/package.d#L972

The result should have `save()`.



Anyway, we use it in a range-based `foreach` and that property doesn't seem useful if we are to believe:
https://dlang.org/spec/statement.html#foreach_over_tuples

Indeed foreach doesn't rollback.

source/dub/recipe/packagerecipe.d
```diff
@@ -179,6 +180,9 @@ struct BuildSettingsTemplate {
 		{
 			auto files = appender!(string[]);
 
+			import dub.project : buildSettingsVars;
+			auto allVars = chain(buildSettingsVars, environment.toAA.byKey);
```
wilzbach: Could be const

wilzbach: And maybe a more explicit name like `buildEnviromentVars`?

wilzbach:
byKey is a forward range, see e.g. https://github.com/dlang/druntime/blob/master/src/rt/aaA.d

@WebFreak001 it works because of by-value copies of struct ranges.

p0nce:
What is the next step to get it approved?

p0nce:
>  then the modified code path will only work properly once 

Oh my, I see the bug now.

p0nce:
This should be a lot safer because the forward ranges are regenerated.

WebFreak001:
yeah I think that won't impact performance or GC at all but makes it safe with any D runtime implementation. I agree with adding a test for this so it won't regress

p0nce:
You want to add a test for what exactly?

WebFreak001:
the feature, when somebody thinks this code is not needed while refactoring :)

The test could just be

```
// app.d
pragma(msg, import("message.txt"));
```

```
// teststrings/message.txt
env_variables_work
```

```json
{
    "name": "env-test",
    "stringImportPaths": ["$MY_VARIABLE"]
}
```

```sh
env MY_VARIABLE=teststrings dub build 2>&1 > output.txt
grep env_variables_work output.txt || exit 1
grep "Invalid source" output.txt && exit 1 || true
```

p0nce:
Sorry but I don't find any information about 
- how to run tests (on which OS etc)
- how to add a test

p0nce:
So it seems you need to run `./test/run-unittest.sh` but this doesn't run under WSL, is it meant to only work on Linux?

p0nce:
```
ponce@vps37353:~/dub$ ./build.sh
Generating version file...
Running dmd...
./build.sh: line 59:  3571 Killed                  $DMD -ofbin/dub -g -O -w -version=DubUseCurl -Isource $* $LIBS @build-files.txt
```
Unfortunately I don't have access to a Linux machine with enough RAM to build DUB.

p0nce:
Hello, why exactly is this PR pending ?
- it is marked "accepted"
- it fixes an annoying bug
- there is no one who knows how to add a test AND willing to add the test AND that can run the test suite

I feel like this is really unfortunate, it has become much more difficult to contribute to DUB than in the past. To me this PR is more in the "no brainer" category.

andre2007:
You can run ./test/run-unittest.sh from Windows without wsl. Just open git bash which come with git client installation. Before running the script you have to say where your dub executable is located:
export DUB=c:/path/to/dub.exe

You can pass as first argument a test number to the bash script to execute that test only.

andre2007:
In folder "test" Create a new issue folder with a naming schema similar to the existing ones and create the first 3 files as proposed by @WebFreak001. The fourth file is a bash test file which you create directly in the test folder similar to the existing ones.
Do not forget to add the executable flag while adding the fourth file to git.

p0nce:
Rebased, squashed, added a test.

p0nce:
Whoops, fixed :)

andre2007:
Maybe the issue is due to the missing .nobuild file.
This avoid an automatic build which might cause as side effect this failing test.

p0nce:
OK should be fixed, rebased, resquashed
What happened is that I had put `--root issue1504-envvar-in-path` instead of `--root ${CURR_DIR}/issue1504-envvar-in-path` because locally I was testing from the wrong directory.

andre2007:
Maybe the test still fail due to the missing .nobuild file.
See here how it looks like
https://github.com/dlang/dub/tree/master/test/5-convert

My assumption is, the automatic dub test logic builds the test application long time before your shell script is executed. Of course your environment variable is not set here. The dub build command in your shell script does not trigger a new build.

andre2007:
You added the .no_test file which has a different meaning than the .nobuild file

p0nce:
No I haven't checked the return value yet.

andre2007:
You can simplify your test script by having a change directory command in line 5
```
cd ${CURR_DIR}/issue1504-envvar-in-path
```
then all occurrences of ${CURR_DIR} will become superfluous

andre2007:
As far as I understand this is an enhancement, please include a changelog like e.g.
https://github.com/dlang/dub/blob/master/changelog/env_var_package_version.dd

p0nce:
This is not en enhancement but a bug fix.
Straight from http://code.dlang.org/package-format?lang=sdl#environment-variables

" Any variable not matching a predefined name will be taken from the program environment."

It worked, but with this erroneous message.

p0nce:
Another attempt.

andre2007:
The interactive error is a known error, @wilzbach is working on

wilzbach:
> The interactive error is a known error, @wilzbach is working on

Yes, it's actually ready: https://github.com/dlang/dub/pull/1579

andre2007:
I assume you will be asked to reduce your commits to 1 commit.

p0nce:
Squashed.

ZombineDev:
Now that #1579 is merged @p0nce can you please rebase again?

p0nce:
Commito ergo sum.

test/issue1504-envvar-in-path.sh
```diff
@@ -0,0 +1,19 @@
...
+rm -rf ${CURR_DIR}/issue1504-envvar-in-path/test
+rm -rf ${CURR_DIR}/output-1504.txt
+
+
+export MY_VARIABLE=teststrings 
+# pragma(msg) outputs to stderr
+${DUB} build --force --root ${CURR_DIR}/issue1504-envvar-in-path 2> ${CURR_DIR}/output-1504.txt
+
+grep "env_variables_work" < ${CURR_DIR}/output-1504.txt
+
```
wilzbach: The temporary files should have been removed here :/

p0nce: They are removed before the test, like in the other test about variables.

PR #1570, opened 2018-10-05
fix loading package with relative path

tchaloupka:
I had a trouble to build subpackage with dependency with relative path in our pretty complex project structure.
It ened up with a really strange error with `Non-optional dependency xx of xx not found in dependency tree` but with `--vverbose` I figured there is an error much earlier.
Figured that it happens when dub tries to upgrade dependencies but not in `Project.reinit` so I just used same relative path handling here.
And it fixes my problem.


tchaloupka:
CI errors are unrelated.

ZombineDev:
Looks good, but can you also add a test for the bug fix? Take a look at [`test/`](https://github.com/dlang/dub/tree/master/test) dir for inspiration.

tchaloupka:
Sorry for the delay.
This fixes #1577 I've added for reference.

s-ludwig:
Can you also add a dub invocation that builds "foo" instead of "foo:sub". To me it seems that both the relative path resolution here and the one in `Project.reinit`, is wrong and needs to be based on the path of the package that has the dependency instead of the root package.

tchaloupka:
Building of "foo" worked for me even before the change problem was with submodule where the generated relative path used for upgrade was invalid.
All is ok when the package is already fetched.

s-ludwig:
Looking closer at it, the odd thing is that there is a relative path in that place at all, all `Dependency` values that are used as tree nodes should be resolved using `.mapToPath` already.

PR #1575, opened 2018-10-16
AppVeyor CI: Temporary disable DMD nightly testing

ZombineDev:
Until the dmd nightlies server is back online


andre2007:
It fails due to a timeout issue.

There is an approved pull request, which retries on timeout issues;)
https://github.com/dlang/dub/pull/1566

PR #1579, opened 2018-10-19
GitHub's API + UI don't return 0.9.20 anymore

wilzbach:
The 0.9.20 tag is still there, see e.g.

```
git ls-remote --tags https://github.com/dlang/dub
```

But not on the web UI nor API, e.g.

https://github.com/dlang/dub/tags?after=v0.9.22-beta.2

This is the pragmatic solution to fixing the test suite.

### DIFF
--- a/test/interactive-remove.sh
+++ b/test/interactive-remove.sh
@@ -2,30 +2,30 @@
 
 . $(dirname "${BASH_SOURCE[0]}")/common.sh
 
-$DUB fetch dub --version=0.9.20 && [ -d $HOME/.dub/packages/dub-0.9.20/dub ]
-$DUB fetch dub --version=0.9.21 && [ -d $HOME/.dub/packages/dub-0.9.21/dub ]
+$DUB fetch dub --version=1.9.0 && [ -d $HOME/.dub/packages/dub-1.9.0/dub ]
+$DUB fetch dub --version=1.10.0 && [ -d $HOME/.dub/packages/dub-1.10.0/dub ]
 if $DUB remove dub --non-interactive 2>/dev/null; then
     die $LINENO 'Non-interactive remove should fail'
 fi
-echo 1 | $DUB remove dub | tr -d '\n' | grep --ignore-case 'select.*0\.9\.20.*0\.9\.21.*'
-if [ -d $HOME/.dub/packages/dub-0.9.20/dub ]; then
-    die $LINENO 'Failed to remove dub-0.9.20'
+echo 1 | $DUB remove dub | tr -d '\n' | grep --ignore-case 'select.*1\.9\.0.*1\.10\.0.*'
+if [ -d $HOME/.dub/packages/dub-1.9.0/dub ]; then
+    die $LINENO 'Failed to remove dub-1.9.0'
 fi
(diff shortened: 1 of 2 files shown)


test/interactive-remove.sh
```diff
@@ -2,30 +2,30 @@
...
-$DUB fetch dub --version=0.9.21 && [ -d $HOME/.dub/packages/dub-0.9.21/dub ]
+$DUB fetch dub --version=1.9.0 && [ -d $HOME/.dub/packages/dub-1.9.0/dub ]
+$DUB fetch dub --version=1.10.0 && [ -d $HOME/.dub/packages/dub-1.10.0/dub ]
 if $DUB remove dub --non-interactive 2>/dev/null; then
     die $LINENO 'Non-interactive remove should fail'
 fi
-echo 1 | $DUB remove dub | tr -d '\n' | grep --ignore-case 'select.*0\.9\.20.*0\.9\.21.*'
-if [ -d $HOME/.dub/packages/dub-0.9.20/dub ]; then
-    die $LINENO 'Failed to remove dub-0.9.20'
+echo 1 | $DUB remove dub | tr -d '\n' | grep --ignore-case 'select.*1\.9\.0.*1\.10\.0.*'
```
wilzbach: Hmm this doesn't seem to work with DMD 2.077:

```
Select version of 'dub' to remove from location 'user':1) 1.1.02) 1.9.03) 1.10.04) all versions> Removing dub in /home/travis/.dub/packages/dub-1.1.0/dub/Removed package: 'dub'Removed dub, version 1.1.0.
[ERROR] /home/travis/build/dlang/dub/test/interactive-remove.sh:12 Failed to remove dub-1.9.0
```

wilzbach: Ouch this is probably due to COVERAGE mode and dub's testsuite being rerun multiple times.

wilzbach:
Ok so this is passing Travis now
Though hard-cleaning the package cache feels like an ugly hack, but I don't really have time to look into which test scripts don't clean themselves up at the moment.

wilzbach:
As no one seems to be having an opinion and the testsuite is broken, I will merge this tomorrow baring no objections until then.

MartinNowak:
Well it's there now, was this just a temporary github issue?
https://github.com/dlang/dub/releases/tag/v0.9.20

MartinNowak:
Maybe it was removed from the dub registry due to having an invalid package recipe. It's definitely no longer in the db. Or the registry got confused by the github hickup.

It's a bit confusing that https://buildkite.com/dlang/ci/builds/511#4c02ce8c-f1a7-4899-bc6d-bcf91ba4d9ee passed and https://buildkite.com/dlang/phobos/builds/582#3ddd20c7-ec9f-4455-838e-061bb4d3a14b/106-716 subsequently failed.

The culprit seems to be that no distinction is made between building dub (as in dub the tool) where we want to use the dev version, and dub (as in dub the project with testsuite) where we want to use the latest stable release.
There is no point in rerunning dub's test-suite with the dub development version, that task is the duty of dlang/dub itself.
https://github.com/dlang/ci/blob/73ebbe8a6ff8caf919b4de8516b471dac2bdd363/buildkite/build_project.sh#L52

PR #1582, opened 2018-10-22
Pre/post run commands added

andre2007:
After the run of an executable there are artefacts created like docs.json (-Xdocs.json). This artefacts should be further processed. postGenerateCommands and postBuildCommands runs before the exectutable is executed.

postRunCommands helps in this scenario. preRunCommands is added for symmetry.

### DIFF
--- a/changelog/pre-post-run-commands.dd
+++ b/changelog/pre-post-run-commands.dd
@@ -1,0 +1,6 @@
+Pre/Post run commands added
+
+DUB now supports commands preRunCommands which are executed before the target run
+and postRunCommands which are executed after the target run.
+Environment variable DUB_TARGET_EXIT_STATUS contains the target executable call status
+and is available in postRunCommands.
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -529,17 +529,40 @@ class BuildGenerator : ProjectGenerator {
 				if (!exe_path_string.startsWith(".") && (exe_path_string.length < 2 || exe_path_string[1] != ':'))
 					exe_path_string = ".\\" ~ exe_path_string;
 			}
+			runPreRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
 			logInfo("Running %s %s", exe_path_string, run_args.join(" "));
 			if (settings.runCallback) {
 				auto res = execute(exe_path_string ~ run_args);
 				settings.runCallback(res.status, res.output);
+				settings.targetExitStatus = res.status;
+				runPostRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
 			} else {
 				auto prg_pid = spawnProcess(exe_path_string ~ run_args);
 				auto result = prg_pid.wait();
+				settings.targetExitStatus = result;
+				runPostRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
 				enforce(result == 0, "Program exited with code "~to!string(result));
 			}
 		} else
 			enforce(false, "Target is a library. Skipping execution.");
+	}
+
+	private void runPreRunCommands(in Package pack, in Project proj, in GeneratorSettings settings,
+		in BuildSettings buildsettings)
+	{
+		if (buildsettings.preRunCommands.length) {
+			logInfo("Running pre-run commands...");
+			runBuildCommands(buildsettings.preRunCommands, pack, proj, settings, buildsettings);
+		}
+	}
+
+	private void runPostRunCommands(in Package pack, in Project proj, in GeneratorSettings settings,
+		in BuildSettings buildsettings)
+	{
+		if (buildsettings.postRunCommands.length) {
+			logInfo("Running post-run commands...");
+			runBuildCommands(buildsettings.postRunCommands, pack, proj, settings, buildsettings);
+		}
 	}
 
 	private void cleanupTemporaries()
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -439,6 +439,7 @@ struct GeneratorSettings {
 	string buildType;
 	BuildSettings buildSettings;
 	BuildMode buildMode = BuildMode.separate;
+	int targetExitStatus;
 
 	bool combined; // compile all in one go instead of each dependency separately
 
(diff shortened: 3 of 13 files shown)


andre2007:
Docs update https://github.com/dlang/dub-docs/pull/10

andre2007:
Ping

test/issue1396-pre-post-run-commands/dub.sdl
```diff
@@ -0,0 +1,3 @@
+name "test"
+preRunCommands "echo \"pre-run\n\" >> test.txt"
```
wilzbach: Not sure why you need the newline

test/issue1396-pre-post-run-commands.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd ${CURR_DIR}/issue1396-pre-post-run-commands
+rm -rf .dub
```
wilzbach: Isn't this dangerous when the testsuite is running in parallel (though that's not really supported as of now)

test/issue1396-pre-post-run-commands/dub.sdl
andre2007: My impression was, the newline is needed for the grep command. But that is not true. I removed them.

test/issue1396-pre-post-run-commands.sh
andre2007: Depending how the test suite parallel runner works this fails or not. I am in doubt to add here a more complex logic which might be needed or not.

test/issue1396-pre-post-run-commands/test.txt
```diff
@@ -0,0 +1,4 @@
+pre-run
+
+post-run
+
```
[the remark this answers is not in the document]
andre2007: removed

source/dub/generators/build.d
```diff
@@ -510,6 +510,11 @@ class BuildGenerator : ProjectGenerator {
 	private void runTarget(NativePath exe_file_path, in BuildSettings buildsettings, string[] run_args, GeneratorSettings settings)
 	{
 		if (buildsettings.targetType == TargetType.executable) {
+			if ( buildsettings.preRunCommands.length) {
```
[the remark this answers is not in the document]
andre2007: fixed

source/dub/generators/build.d
```diff
@@ -538,6 +543,11 @@ class BuildGenerator : ProjectGenerator {
 				auto result = prg_pid.wait();
 				enforce(result == 0, "Program exited with code "~to!string(result));
 			}
+
+			if (buildsettings.postRunCommands.length) {
+				logInfo("Running post-run commands...");
+				runBuildCommands(buildsettings.postRunCommands, m_project.rootPackage, m_project, settings, buildsettings);
```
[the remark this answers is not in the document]
andre2007: At the moment, there is an existing  enforce call, therefore the post run commands are only called if return code is 0. 
Should post run commands be called even if return code is not equal 0?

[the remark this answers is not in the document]
andre2007: Environment variable DUB_TARGET_STATUS added, containing the status of the executable call.
Also the pre / post run commands are no longer executed for callback (dustmite).

Please have a look

changelog/pre-post-run-commands.dd
```diff
@@ -0,0 +1,6 @@
+Pre/Post run commands added
+
+DUB now supports commands preRunCommands which are executed before the target run
+and postRunCommands which are executed after the target run.
+Environment variable DUB_TARGET_STATUS contains the target executable call status
```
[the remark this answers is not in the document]
andre2007: fixed

source/dub/generators/build.d
```diff
@@ -529,13 +529,23 @@ class BuildGenerator : ProjectGenerator {
...
+					runBuildCommands(buildsettings.preRunCommands, m_project.rootPackage, m_project, settings, buildsettings);
+				}
+				logInfo("Running %s %s", exe_path_string, run_args.join(" "));
 				auto prg_pid = spawnProcess(exe_path_string ~ run_args);
 				auto result = prg_pid.wait();
+				settings.targetStatus = result;
+				if (buildsettings.postRunCommands.length) {
+					logInfo("Running post-run commands...");
+					runBuildCommands(buildsettings.postRunCommands, m_project.rootPackage, m_project, settings, buildsettings);
+				}
```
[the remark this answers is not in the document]
andre2007: fixed

source/dub/generators/generator.d
```diff
@@ -439,6 +439,7 @@ struct GeneratorSettings {
 	string buildType;
 	BuildSettings buildSettings;
 	BuildMode buildMode = BuildMode.separate;
+	int targetStatus;
```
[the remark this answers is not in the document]
andre2007: fixed

source/dub/generators/generator.d
```diff
@@ -439,6 +439,7 @@ struct GeneratorSettings {
 	string buildType;
 	BuildSettings buildSettings;
 	BuildMode buildMode = BuildMode.separate;
+	int targetExitStatus;
```
WebFreak001: is GeneratorSettings really the proper place to put this? All the other environment variables are always set after all and targetExitStatus is only set once the program has run once, so it wouldn't make sense to include it in a pre-run call for example.

Also why is this called "target" exit status?

source/dub/generators/build.d
```diff
@@ -529,19 +529,41 @@ class BuildGenerator : ProjectGenerator {
 				if (!exe_path_string.startsWith(".") && (exe_path_string.length < 2 || exe_path_string[1] != ':'))
 					exe_path_string = ".\\" ~ exe_path_string;
 			}
+			runPreRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
 			logInfo("Running %s %s", exe_path_string, run_args.join(" "));
 			if (settings.runCallback) {
 				auto res = execute(exe_path_string ~ run_args);
 				settings.runCallback(res.status, res.output);
+				runPostRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
```
WebFreak001: shouldn't targetExitStatus also be set here before the run callback?

andre2007: you are right, fixed

source/dub/generators/generator.d
andre2007: It seemed for me reasonable to name it "target" as e.g. the existing method also is called "runTarget". 
You are right, it is a compromise. I didn't wanted to add too much new complexity just to get the targetExitStatus as environment variable.

WebFreak001:
oh yeah sorry, I think it's fine to merge now. I still don't like that the environment variable is in every run, but I think this makes the code the simplest and nobody will notice anyway.

Issue #790, opened 2016-03-15
dub fetch <package> --save


wilzbach:
It would be quite convenient to be able to modify the package.json and `sdl` file and add or update the new package. This is a solution to #789 (multiple package formats).

Expected behavior

1) run `dub fetch <package>` (-> abortion on errors)
2a) When a package is newly installed, add it to the users dependencies
2b) When a package is already installed, but newly "saved" - update its version in the package file.

Ideally - as I expect this command to be posted on the registry - this command should get a nice alias (or even better: without `--save`) like `dub save <package>`.

MartinNowak:
> which is a bit difficult to implement nicely without destroying user formatting in dub.(json/sdl)

Should be easy for dub.sdl b/c it can be appended to the file, for dub.json we'd need to only manipulate the dependencies array (plus detecting leading whitespace).

greenify:
>  for dub.json we'd need to only manipulate the dependencies array

I think json is trivial as there are no comments to preserve. That's why every package manager (npm, bower, ...) just deserializes the file, adds the dependencies and then serializes it back. The best we can do is to detect the pretty-print format (tabs, 2 spaces, 4 spaces, etc.) and use the same option for the newly saved file.

s-ludwig:
Appending for dub.sdl *works* of course, but won't lead to clean results in many cases. Searching for the last top-level "dependency" tag would be okay, though. But you also need to be aware of semicolon separated nodes or backslash line continuations.

JSON, even if it doesn't have comments, has quite some formatting opportunities, see for example <https://github.com/rejectedsoftware/vibe.d/blob/master/dub.json>. Simply deserializing is not an option, IMO. But of course something like what Martin proposed is doable. It's just a little involved and will at least require some modifications to the JSON parser.

In all honesty, I'm still undecided about the net value of this feature. On one hand it kind of makes sense. On the other hand it crosses the border of messing with source files for providing trivial and rather rarely used functionality.

greenify:
> In all honesty, I'm still undecided about the net value of this feature.

Well it's kind of known from other package manager (`bower`, `npm`, `godep`, ...) and it allows a handy command that can be shown on the dub registry and allow easy copy & pasted!

> On the other hand it crosses the border of messing with source files

If there are people out there which really don't like this at all - they can still use `fetch`. It's just an additional, very useful feature.

> and rather rarely used functionality.

[...]

PR #1587, opened 2018-11-04
Create "add" command to add a dependency

### LINKED ISSUE #1574: [enhancement] `add` command
It would be nice if there was a dub command like `dub add <package>` that will automatically add the latest version of `<package>` as a dependency. There really should be a way to add dependencies without editing raw JSON/SDL.

Flying-Toast:
Usage: `dub add vibe-d` will add `dependency "vibe-d" version="~>X.Y.Z"`(X.Y.Z=latest version) to the dub.sdl file, or the equivalent to dub.json.
This command is a simpler alternative to manually editing the dub.json/dub.sdl file. It is also no longer necessary to check what the latest version of a package is before adding the dependency.

fixes #1574

### DIFF
--- a/changelog/addcommand.dd
+++ b/changelog/addcommand.dd
@@ -1,0 +1,11 @@
+Add Command
+
+The `add` command adds a dependency to the dub.json/dub.sdl file.
+
+Running `dub add vibe-d` is equivalent to manually adding the following line to dub.sdl (X.Y.Z represents the latest version):
+dependency "vibe-d" version="~>X.Y.Z"
+
+Or the following in dub.json:
+"dependencies": {
+	"vibe-d": "~>X.Y.Z"
+}
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1147,6 +1147,7 @@ class AddCommand : Command {
 	override int execute(Dub dub, string[] free_args, string[] app_args)
 	{
 		import dub.recipe.io : readPackageRecipe, writePackageRecipe;
+		import dub.internal.vibecompat.core.file : existsFile;
 		enforceUsage(free_args.length != 0, "Expected one or more arguments.");
 		enforceUsage(app_args.length == 0, "Unexpected application arguments.");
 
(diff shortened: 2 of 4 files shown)


Flying-Toast:
If this is merged, we should probably also change the "To use this package, put the following dependency into your project's dependencies section:" part of the dub registry to use the `dub add` command. (https://github.com/dlang/dub-registry/blob/master/views/view_package.dt#L67)

Flying-Toast:
@wilzbach Is there something wrong with the Travis CI build? It looks like it failed because of a problem with the compiler, not the actual build.

andre2007:
Could you please add an automatic test for the new command?

The version file should not be included within the pull request.

Another question, by executing the new add command, does this change the order of the existing attributes within dub.json?

Flying-Toast:
@andre2007 std.json does not preserve the order of attributes. I can't think of a method that would save the order, other than manually editing the raw string of the file.

Flying-Toast:
@andre2007 what is the naming convention for the test scripts?

andre2007:
I use issueNNN-short-descriptive-name.sh
Example issue1416-maven-repo-pkg-supplier.sh

Flying-Toast:
@andre2007 IMO the order of the attributes doesn't matter. By using the `add` command you won't really ever need to edit the package file, so the changes won't be noticed.

andre2007:
I think if you work in a team and you have to review a dub.json on github which is completely different because there was 1 new dependency added, that might be annoying. 
Does it make sense to change the internal json implementation to store the attributes sorted by name?

By chance do you know how npm handles this issue?

Flying-Toast:
It looks like npm is just using javascript's built in `JSON.stringify` to write the package json (https://github.com/npm/stringify-package). I guess we'll just have to change dub's JSON implementation to preserve order.

ZombineDev:
I think the way to go is not to preserve the order, but to always sort by name.

Flying-Toast:
@ZombineDev good point.
I am not sure when/if I will be able to implement JSON sorting. Could anyone else help with this?

andre2007:
I think a good starting point is here
https://github.com/dlang/dub/blob/master/source/dub/internal/vibecompat/data/json.d#L1782

As proposal: ```pretty == true``` also means sorting the object keys. What do you think?

Flying-Toast:
#1593 is done.

thewilsonator:
@Flying-Toast sorting PR merged. Is this ready to be merged?

Flying-Toast:
@thewilsonator yes.

WebFreak001:
I think this functionality is great, but maybe we could do something smarter than just modifying the data and writing it back. Even if it is less elegant and a bit worse to maintain in code, I think keeping the users original file formatting and comments should be a high priority for the `dub add` command.

In serve-d I got [code](https://github.com/Pure-D/serve-d/blob/master/source/served/extension.d#L2528) doing the same, but keeping all the user formatting in tact. Even though it works in fewer cases, it pretty much guarantees that the user doesn't lose any of their formatting.

0xEAB:
so using `dub add` will blow up all formatting? :neutral_face: 

my opinion on this: https://github.com/dlang/dub/issues/790#issuecomment-419663811
> This should be **clearly** stated and documented, otherwise someone will complain soon that this command destroyed their formatting.

PR #1589, opened 2018-11-06
Add experimental feature to improve build cache efficiency

MartinNowak:
Using version identifiers to configure certain features can often lead to
unnecessary rebuilds of dependencies that don't use those version
identifiers. In order to improve the build cache efficiency, dub gained a new
experimental `--filter-versions` switch.

When `--filter-versions` is passed to any build, test, or generate command, dub
will grep for all the version identifiers packages actually use and only apply
those during building. This allows for example to reuse a cached build for a
library between two applications using different version identifiers when the
library isn't using any of those itself.

Reduces time for vibe-d's examples test suite by about 50% for me.

### DIFF
--- a/changelog/version-identifier-filter.dd
+++ b/changelog/version-identifier-filter.dd
@@ -1,0 +1,17 @@
+Added experimental build settings to filter applicable version identifiers
+
+Those settings allow to filter version identifiers and debug version identifiers
+used during building a package. This can improve the build cache efficiency for
+libraries that are used by multiple applications with different version
+identifiers.
+
+dub.json:
+```json
+"-versionFilters": ["Have_vibe_d"]
+"-versionFilters-posix": ["UseUnixSockets", "UseMMap"]
+```
+dub.sdl:
+```sdl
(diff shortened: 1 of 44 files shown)


jacob-carlborg:
Grepping feels like a hack. I guess you don't want to use DMD as a library?

MartinNowak:
Not in it's current state. Yes grepping substitutes a parser (which couldn't find all version identifiers either), but version declarations are also very simple constructs (see https://dlang.org/spec/grammar.html#VersionSpecification) so grep is a suitable tool.

jacob-carlborg:
> which couldn't find all version identifiers either

The Dub package provides semantic analyze, but that might take longer than just using grep.

thewilsonator:
> Yes grepping substitutes a parser

That won't handle `version` or `debug` as part of a comment, but I don't think that it will be a particularly common occurrence. This is experimental anyway, LGTM.

andre2007:
Can this be merged?

thewilsonator:
Whoops, thanks!

PR #1590, opened 2018-11-06
Fix temp copy files

andre2007:
Executing app.d
```
/+ dub.sdl:

name "app"
dependency "vibe-d" version=">=0.8.0-beta.5"
+/
void main(){}
```
with command ```dub app.d``` which is similiar to ```dub run --temp-build --single app.d```
fails on windows because the OpenSSL dlls are copied to the location of app.d instead of
the location of the executable.

The actual location of the executable ```getTempDir() ~ format(".dub/build/%s-%s/%s/", packageName, pack.version_, build_id)``` is only known to BuildGenerator but not to ProjectGenerator. This is an issue as
copying the files is triggered from ProjectGenerator. I solved this issue by introducing protected variable ```m_tempTargetExecutablePath``` within ProjectGenerator. BuildGenerator will fill this variable.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -34,7 +34,6 @@ class BuildGenerator : ProjectGenerator {
 	private {
 		PackageManager m_packageMan;
 		NativePath[] m_temporaryFiles;
-		NativePath m_targetExecutablePath;
 	}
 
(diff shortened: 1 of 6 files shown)


andre2007:
Fixed issue 1136

Issue #1522, opened 2018-08-05
DUB should not automatically fall-back to non-HTTPs repository


mbierlee:
In module `dub.dub`, DUB's registry resolution is instructed to fall-back to a non-HTTPS alternative in case something is wrong with the HTTPs default. A man in the middle can invalidate the HTTPs certificate and make DUB fall back to the non-HTTPs version, after which they can inject their own code in your projects.

I suggest to only allow fall-back to a non-HTTPs alternative when the user explicitly indicated they allow this (through a command-line flag, dub config, etc.).

andre2007:
Was fixed.

PR #1592, opened 2018-11-08
improve package supplier fallback logic

MartinNowak:
- scope (failure) did swallow fatal errors
- improve logging of failures and successes
- rethrow failure of default registry instead of from the last one
- remove failing package suppliers, so we don't keep beating dead horses

Note that some retry logic is already done interally by package suppliers,
so a failing one is unlikely to recover on the next usage.

### DIFF
--- a/source/dub/packagesuppliers/fallback.d
+++ b/source/dub/packagesuppliers/fallback.d
@@ -5,19 +5,26 @@ import std.typecons : AutoImplement;
 
 package abstract class AbstractFallbackPackageSupplier : PackageSupplier
 {
-	protected PackageSupplier[] m_suppliers;
+	protected import core.time : minutes;
+	protected import std.datetime : Clock, SysTime;
+
+	static struct Pair { PackageSupplier ps; SysTime failTime; }
+	protected Pair[] m_suppliers;
 
 	this(PackageSupplier[] suppliers)
 	{
 		assert(suppliers.length);
-		m_suppliers = suppliers;
+		m_suppliers.length = suppliers.length;
+		foreach (i, ps; suppliers)
+			m_suppliers[i].ps = ps;
 	}
 
 	override @property string description()
 	{
 		import std.algorithm.iteration : map;
 		import std.format : format;
-		return format("%s (fallbacks %-(%s, %))", m_suppliers[0].description, m_suppliers[1 .. $].map!(x => x.description));
+		return format("%s (fallbacks %-(%s, %))", m_suppliers[0].ps.description,
+			m_suppliers[1 .. $].map!(pair => pair.ps.description));
 	}
 
 	// Workaround https://issues.dlang.org/show_bug.cgi?id=2525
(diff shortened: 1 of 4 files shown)


source/dub/packagesuppliers/fallback.d
```diff
@@ -40,19 +39,33 @@ private template fallback(T, alias func)
...
+
+		Exception firstEx;
+		try
+			return m_suppliers[0].%1$s(args);
+		catch (Exception e)
+		{
+			logDiagnostic("Package supplier %%s failed with '%%s', trying fallbacks.",
+				m_suppliers[0].description, e.msg);
+			m_suppliers = m_suppliers[1 .. $];
+			firstEx = e;
```
wilzbach: `-dip1008` won't allow this, but lots of stuff in Phobos does similar stuff which is why it will take a very long time until that switch gets ever activated ...

source/dub/packagesuppliers/fallback.d
```diff
@@ -40,19 +39,33 @@ private template fallback(T, alias func)
...
+		import dub.internal.vibecompat.core.log : logDiagnostic;
+
+		Exception firstEx;
+		try
+			return m_suppliers[0].%1$s(args);
+		catch (Exception e)
+		{
+			logDiagnostic("Package supplier %%s failed with '%%s', trying fallbacks.",
+				m_suppliers[0].description, e.msg);
+			m_suppliers = m_suppliers[1 .. $];
```
wilzbach: Not sure whether this is a good idea when dub is used as a library and potentially runs for days, but I think only DLS does so and I'm just sending a quick ping to @LaurentTreguier to make him aware of this change.

source/dub/packagesuppliers/fallback.d
```diff
@@ -40,19 +39,33 @@ private template fallback(T, alias func)
...
 	import std.format : format;
 	enum fallback = q{
-		import std.range : back, dropBackOne;
-		import dub.internal.vibecompat.core.log : logDebug;
-		scope (failure)
+		import dub.internal.vibecompat.core.log : logDiagnostic;
+
+		Exception firstEx;
+		try
+			return m_suppliers[0].%1$s(args);
```
wilzbach: Won't this trigger and RangeError if all suppliers failed?

andre2007:
There was also the issue https://github.com/dlang/dub/issues/1522 created: Falling back to non https should not be the default. Maybe you can address this security related issue in the pr too?

source/dub/packagesuppliers/fallback.d
MartinNowak: I think ignoring that use-case (i.e. [`collectExceptions`](https://dlang.org/library/std/exception/collect_exception.html)) is reason enough for DIP1008 to never work until it's accompanied by a `.dup`/`.clone` mechanism.

source/dub/packagesuppliers/fallback.d
MartinNowak: Interesting point! When all suppliers failed the exception will be rethrown and quit the program.

andre2007:
LGTM

MartinNowak:
We have a major issue with the search api in dub atm. hence the failures (see https://github.com/dlang/dub-registry/pull/397 for a to be deployed fix).

PR #1593, opened 2018-11-09
sort JSON

Flying-Toast:
see #1587


Flying-Toast:
This is ready to be reviewed.
Is it possible to get dlang-bot to remove the "WIP" tag now?

andre2007:
Could you add an unit test to the module to check the sorting is working as expected?
You can run the test using command ```dub test ```

Flying-Toast:
It turns out there was already a unittest for `Json.toPrettyString`: 
```
assert(a.toPrettyString() ==
`{
        "a": [],
         "b": [
		1,
		{}
	]
}` || a.toPrettyString() ==
`{
	"b": [
		1,
		{}
	],
	"a": []
}`);
```

The existing unittest checked for both possible orders of the JSON, so I just changed it to only check for the sorted one.

ZombineDev:
I think most people would want only the entries of the dependency section to be sorted, but not the rest. If we sort the entries of every object in the json, we'll end up with the `name` and `description` properties being placed after the `dependency` section.
If I had to formalize the desired behavior I would say: "keep the order of entries in non-managed sections intact while sorting the entries of managed sections (e.g. by `dub add`)". I understand that may not a small task to implement, so perhaps we should try to see if it's possible to sort only specific sections. For example: `toPrettyString(string[] keysOfObjectsToSort = [ "dependencies" ])`.

Flying-Toast:
The issue with only sorting the dependencies section is that the other sections would still get shuffled around and, like andre said, 
> If you work in a team and you have to review a dub.json on github which is completely different because there was 1 new dependency added, that might be annoying.

What if instead of sorting the JSON alphabetically, it was sorted into a custom order, something like `name`, `authors`, `description`, `dependencies`? That way, the attributes would be in a logical order, and they would also be consistent when the json file is modified.

Flying-Toast:
@ZombineDev I added an array of [all the possible JSON properties in dub.json](https://code.dlang.org/package-format?lang=json) that determines the sorting order. This allows the order to be preserved while maintaining a logical order.

andre2007:
There are 2 new attributes in JSON since this morning :)
preRunCommands / postRunCommands

https://github.com/dlang/dub/pull/1582

source/dub/internal/vibecompat/data/json.d
```diff
@@ -1779,24 +1779,76 @@ void writeJsonString(R, bool pretty = false)(ref R dst, in Json json, size_t lev
...
+					"versions",
+					"debugVersions",
+					"importPaths",
+					"stringImportPaths",
+					"preGenerateCommands",
+					"postGenerateCommands",
+					"preBuildCommands",
+					"postBuildCommands",
+					"dflags",
+					"lflags"
```
thewilsonator: Is there any particular reason for this order? e.g. why not alphabetic? could take advantage of its sortedness then.

Flying-Toast: No, there is not a specific reason for this order. It's just the order that I copied it down. 

Initially, I made it [sort alphabetically](https://github.com/dlang/dub/pull/1593/commits/a9e5dc1561aca03647e0cf434a26f9148d5f01a3), but changed it because @ZombineDev said it would make more sense if `name` came before `dependencies`.

I still think alphabetical order would be a better choice though. That way, the code wouldn't have to be updated each time there is a new JSON property is added (like happened this morning, in #1582). What are your ideas on this?

andre2007: I wonder whether it makes sense to sort the fields by reading the fields of structure PackageDescription. https://github.com/andre2007/dub/blob/0719452de25c93e8c590b01e08ec5b938f58a582/source/dub/description.d#L65

This avoids the problem of having multiple places to be adapted when new fields are introduced.

Flying-Toast: @andre2007 I don't think that would work, because the order of the fields in a struct have no semantic value. After parsing the fields from `PackageDescription`, we would still have to sort them in some way.

@ZombineDev does it really even matter what order the elements are in, as long as they are consistent?

Flying-Toast: I've reverted to the alphabetical order sorting, because that seems like the most efficient way to sort the fields.

thewilsonator: Thanks, I'm fine with whatever order really, as long as there is reason for it and if its not immediately obvious what the order is there is some documentation saying what it is and why.

PR #1595, opened 2018-11-14
fix unix executables gitignore

Flying-Toast:
I recently merged #1594, but it turns out that putting `{project-name}` in the gitignore also ignores the directory `source/{project-name}`. I added a leading `/` to only ignore the executable in the root project directory, and nothing from the `source/` directory.


thewilsonator:
Ooh, good catch.

PR #1600, opened 2018-11-27
Fix wrong cast

RazvanN7:
This is currently blocking [1]. The cast to BuildOption is useless since parent.options is already a BuildOption object. What happened before the patch in [1]: parent.options.values (the alias this member)
was being  wrongfully cast to BuildOptions (and opCast from std.typecons.Bitflags was called); now the opCast is no longer called directly on the alias this member.

[1] https://github.com/dlang/dmd/pull/9017

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -396,7 +396,7 @@ class ProjectGenerator
 	{
 		child.addVersions(parent.versions);
 		child.addDebugVersions(parent.debugVersions);
-		child.addOptions(BuildOptions(cast(BuildOptions)parent.options & inheritedBuildOptions));
+		child.addOptions(BuildOptions(parent.options & inheritedBuildOptions));
 	}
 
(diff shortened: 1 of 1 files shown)


source/dub/version_.d
```diff
@@ -1,2 +1,2 @@
 module dub.version_;
-enum dubVersion = "v1.12.0";
```
Geod24: Don't commit this

thewilsonator: I was just about to ask about that.

RazvanN7: I don't know how that got changed. I fixed it now. Thanks!

Geod24: Probably `git commit -am` ;)
This is updated during the build.

wilzbach:
I'm guessing that this should have been in stable, so that you don't need to wait too long for this to become usable in buildkite?
Though actually as we release DMD with dub, we should probably be fine with always checking out master.

PR #1602, opened 2018-11-29
when running `dub build :<subpackage>`, explicitly load subpackage in current folder.

FeepingCreature:
To reproduce the problem that prompted this:

```
cd $HOME
rm -rf .dub
dub fetch unit-threaded --version 0.7.52
dub fetch unit-threaded --version 0.7.53
cd .dub/packages/unit-threaded-0.7.52/unit-threaded
dub build :runner
```
Observe that what is built is unit-threaded-0.7.**53**:runner!

This is because dub handles :runner by building "unit-threaded:runner", but it uses the normal lookup logic to find what "unit-threaded:runner" is. Normally the current path would be in `m_temporaryPackages`, but since the current path is already in the search path, it is not added again. As a result, `unit-threaded-0.7.53` is found first.

The attached commit explicitly looks up subpackages specified as ":<package>" in the project loaded from the cwd.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -697,11 +697,16 @@ abstract class PackageBuildCommand : Command {
 			return true;
 		}
 
+		bool from_cwd = package_name.length == 0 || package_name.startsWith(":");
 		// load package in root_path to enable searching for sub packages
-		if (loadCwdPackage(dub, package_name.length == 0)) {
+		if (loadCwdPackage(dub, from_cwd)) {
 			if (package_name.startsWith(":"))
-				package_name = dub.projectName ~ package_name;
-			if (!package_name.length) return true;
+			{
+				auto pack = dub.packageManager.getSubPackage(dub.project.rootPackage, package_name[1 .. $], false);
+				dub.loadPackage(pack);
+				return true;
+			}
+			if (from_cwd) return true;
 		}
 
 		enforce(package_name.length, "No valid root package found - aborting.");


thewilsonator:
Thanks, added `72h -> merge to let others express their opinions. Ping me if I forget.

FeepingCreature:
@thewilsonator ping!

wilzbach:
A test should have been added here to ensure it actually fixes the problem and that it doesn't regress in the future.

FeepingCreature:
It's hard to test because it specifically only happens when you run dub in `$HOME/.dub`.

andre2007:
Is it possible to set the environment variable $HOME in the test?
Then it would be easy to add a test, by settings $HOME to a specified folder.

wilzbach:
> Is it possible to set the environment variable $HOME in the test?

Yes.

> Then it would be easy to add a test, by settings $HOME to a specified folder.

Yes.

andre2007:
@FeepingCreature Could you add a small test? There is a chance the problem will occur in future by an accidental change.

FeepingCreature:
I'm honestly not sure how, which may be partially because I don't know what reproduces it, because I don't understand how dub import paths are ordered. I know it happened in my case because I had two versions of a library in .dub/packages, and I was trying to build one of them from within .dub/packages, but I don't even know if it always selects the newer version, and coworkers have indicated that it sometimes didn't happen for them. So I really don't have a clean repo to offer.

PR #1605, opened 2018-12-02
Shebang without .d extension

andre2007:
Dub single-file packages can now be called without .d extension.
In addition to ```dub app.d --param``` you can call ```dub app --param```.


changelog/shebang-without-d-extension.dd
```diff
@@ -0,0 +1,4 @@
+Shebang without .d extension
+
+Dub single-file packages can now be called without .d extension.
+In addition to `dub app.d --param` you can call `dub app --param`.
```
wilzbach: This is confusing because the file still needs to be named app.d

andre2007: You are right but you gave me an inspiration. We could allow single file packages without d extension by using same approach like stdin: copy the file ./app to a temp file .../app.d.

I have to check whether it is technically possible and if it works with posix shebang line.

wilzbach: Why not pipe the file directly to stdin?

andre2007: I am not sure whether I understand you 100%. The stdin feature of dub works by storing stdin as temp-file.d. It does not leverage dmd stdin feature. 

If I understand you correct, you propose to pipe file ./app to stdin which will then write the file to a temp-file.d.  I cannot see a benefit. Just copying file ./app to a temp-file.d seems much more straightforward.

wilzbach: IIRC DMD doesn't create temporary files for files from stdin, but if you want to implement this just go for what is the easiest for you.

andre2007: I did the changes, please retest. Now it is possible to:
- have files without .d extension
- files with .d extension but call them without

My assumption is, single buildkite error is unrelated
https://buildkite.com/dlang/dub/builds/315#2695a8fc-ff2d-4ae6-bd6f-99f45148c88c

wilzbach: Yes -> https://github.com/dlang/ci/issues/360

andre2007:
Ping :)

PR #1608, opened 2018-12-06
call setDifference with SortedRange

burner:
in preperation for

https://issues.dlang.org/show_bug.cgi?id=10706
https://github.com/dlang/phobos/pull/6795

the passed string[] are sorted so assumeSorted is ok.


burner:
thank you

PR #1616, opened 2018-12-10
Filesystem package supplier updated

andre2007:
There was a filesystem supplier available but not registered. Also some minor additions were missing to get it working with json/sdl package descriptors.


andre2007:
Ping :)

source/dub/internal/utils.d
```diff
@@ -130,6 +130,39 @@ Json jsonFromFile(NativePath file, bool silent_fail = false) {
...
+	import std.zip : ZipArchive, ArchiveMember;
+	import dub.package_ : packageInfoFiles;
+
+	auto f = openFile(zip, FileMode.read);
+	ubyte[] b = new ubyte[cast(size_t)f.size];
+	f.rawRead(b);
+	f.close();
+	auto archive = new ZipArchive(b);
+	alias PSegment = typeof (NativePath.init.head);
+	outer: foreach (ArchiveMember am; archive.directory) {
```
thewilsonator: Why is the outer label doing there>

andre2007: Good catch:)
It is superfluous and was needed in a previous implementation, I will remove it.

andre2007: Done

PR #1618, opened 2018-12-13
extraDependencyFiles attribute added

andre2007:
Using attribute `dependencyFiles` a list of files can be specified
causing rebuild on change.

Fixes https://github.com/dlang/dub/issues/1117

PR #1621, opened 2018-12-16
[WIP] Add --force-no-deps option to build command

### LINKED ISSUE #606: --force-all to rebuild all dependencies
Doing `dub build [project] --force` forces a rebuild of the root project (but not dependencies). That's good, but it'd be useful to also have a `dub build [project] --force-all` to automatically force rebuild of all dependencies in addition to the root project.

### LINKED ISSUE #1620: Force rebuild without dependencies
The `--force` option for `dub build` forces dependencies to be rebuilt, along with the actual package. There are some cases when it is useful to be able to force rebuild *only* the package, for example with [import expressions](https://dlang.org/spec/expression.html#import_expressions):
```
import std.stdio;

void main() {
    writeln( import("foo.txt") );
}
```

If the contents of `foo.txt` are updated and no changes to the package's actual source are made, dub won't trigger a rebuild. You have to use `dub build -f`, but then that also rebuilds dependencies, which is unnecessary and takes much longer.

Flying-Toast:
The `--force` option for `dub build` forces dependencies to be rebuilt, along with the actual package. There are some cases when it is useful to be able to force rebuild *only* the package, for example with [import expressions](https://dlang.org/spec/expression.html#import_expressions):
```
import std.stdio;

void main() {
    writeln( import("foo.txt") );
}
```

If the contents of `foo.txt` are updated and no changes to the package's actual source are made, dub won't trigger a rebuild. You have to use `dub build -f`, but then that also rebuilds dependencies, which is unnecessary.

This PR adds a `--force-no-deps` (shorthand `-F`) option to the `build` command.

fixes #1620
closes #606

### DIFF
--- a/changelog/forceNoDeps.dd
+++ b/changelog/forceNoDeps.dd
@@ -1,5 +1,5 @@
 build --force-no-deps
 
-The `-F`/`--force-no-deps` option to the `build` command forces a rebuild of the current package, excluding dependencies.
+The `--force-no-deps` option for the `build` command forces a rebuild of the current package, excluding dependencies.
 
 `--force-no-deps` does the same thing as `--force`, except that `--force-no-deps` does not rebuild dependencies.
--- a/changelog/forceNoDeps.dd
+++ b/changelog/forceNoDeps.dd
@@ -1,0 +1,5 @@
+build --force-no-deps
+
+The `-F`/`--force-no-deps` option to the `build` command forces a rebuild of the current package, excluding dependencies.
+
+`--force-no-deps` does the same thing as `--force`, except that `--force-no-deps` does not rebuild dependencies.


Flying-Toast:
I cant figure out where the logic for `--force` is. Where should I implement `--force-no-deps`?

ZombineDev:
If the consensus is to add more fine grained  
rebuild control (assuming improvement like https://github.com/dlang/dub/pull/1618 can't solve the underlying problem), I would prefer a more general approch along the lines of:

```
# Rebuild target dep1 and dep2
# (dependencies of the root project)
dub build --target=dep1 --force

s-ludwig:
Actually, it *should* trigger a rebuild in the example above. The path that contains "foo.txt" needs to be registered with `stringImportPaths` in the package recipe, though.

Issue #1619, opened 2018-12-15
`--build-mode=singleFile` rebuilds already built files


JohanEngelen:
Using `dub build --build-mode=singleFile` starts compiling files one by one. If halfway the build is cancelled, then the compilation output object files are in the `./dub/build/...` subdir, as expected. However, if now `dub build --build-mode=singleFile` is run again, it starts building from scratch, recompiling to the object files that were already built. Expected: compared date of object files already in built cache dir. (can people reproduce this? or does it depend on my filesystem settings?)

( Possibly related #1121  ? )

MartinNowak:
It's slightly more complex than that, the relevant code is at [build.d](source/dub/generators/build.d).
Atm. only the dates of input files and targets are compared (e.g. `.d` sources and `.a` libs), you'd need to integrate a multi-level date checking à la make to partially rebuild only some files.
Furthermore this would either need to rebuild all objects if some input source changes or needed to track all imported dependencies for fine-grained partial rebuilds.
Given that singleFile compilation is inherently slow (due to repeated imports+semantic) and the effort necessary to only implement a make-like approach, it's debatable whether this would be a good use of resources.

What's the use-case for interrupting and the continuing builds?
Could you achieve your goal with the cmake generator?
Would it make sense to write a generator for [Ninja](https://ninja-build.org/) and use that for building?

JohanEngelen:
"Single file compilation is inherently slow" is only true for certain cases. In other cases, single file compilation is much slower. On a 96 core machine...
Anyway, it was just a note. I also think it isn't very effective use of time to spend improving dub as a build tool for large projects. Better use cmake etc.

MartinNowak:
Better generator integration with ninja could be worthwhile for large projects (or using it via cmake's ninja build mode).

MartinNowak:
See https://github.com/dlang/dub/pull/1625 for a PoC ninja generator, might be particularly interesting for dual-socket build servers ;).

PR #1625, opened 2018-12-23
[PoC] add ninja generator

MartinNowak:
- based on parallel single file compilation
- writes object files to local .dub/ninja/<package> folder
  (use `ninja -t clean` to cleanup)
- different build types require regeneration of .ninja files
  (see https://ninja-build.org/manual.html#_philosophical_overview)
- requires dlang/dmd#9122 for non-segfaulting `-deps`

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -120,7 +120,7 @@ int runDubCommandLine(string[] args)
 			args[1] = args[1].setExtension(".d");
 		}
 	}
-	
+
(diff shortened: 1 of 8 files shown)


MartinNowak:
FYI @JohanEngelen, regarding the discussion at https://github.com/dlang/dub/issues/1619#issuecomment-449569259. I don't plan to finish or polish this, but it might be a useful starting point for quicker incremental rebuilds in large projects.

PR #1626, opened 2018-12-24
fixup #1587 - cleanup and version-spec support for dub add

MartinNowak:
- support version specifications in new add command
- improve changelog entry for add command
- reuse existing code to find and load package recipe
- share code for dependency addition between init and add cmd


Flying-Toast:
@MartinNowak see #1630

PR #1630, opened 2018-12-27
Fix broken link in changelog

Flying-Toast:
The [changelog for the AddCommand](https://dlang.org/changelog/2.084.0.html#addcommand) has a broken link.

The link text and link location must have been mixed up, because the link shows as
> It is also possible to add multiple packages at once and explicitly add a simple [http://code.dlang.org/package-format?lang=json#version-specs](https://dlang.org/changelog/version%20specification) for some of them. 

(link href is `version specification`)

Instead of
> It is also possible to add multiple packages at once and explicitly add a simple [version specification](https://code.dlang.org/package-format?lang=json#version-specs) for some of them. 

(link href is `https://code.dlang.org/package-format?lang=json#version-specs`)

### DIFF
--- a/changelog/addcommand.dd
+++ b/changelog/addcommand.dd
@@ -18,7 +18,7 @@ dependency "vibe-d" version="~>X.Y.Z"
 ------
 
 It is also possible to add multiple packages at once and explicitly add a
-simple $(LINK2 version specification,http://code.dlang.org/package-format?lang=json#version-specs)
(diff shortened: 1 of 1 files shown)


wilzbach:
You need to change the base branch to stable. I'm on my phone, but you can run:

```
git rebase --onto upstream/stable upstream/master
```

And then force-push.

wilzbach:
CI failures were obviously unrelated.

PR #1471, opened 2018-05-16
Build {32,64}-bit Windows binaries on Travis

wilzbach:
Similar to how we build the binaries at Dscanner, DCD and dfmt, e.g.

https://github.com/dlang-community/DCD/blob/master/release-windows.sh
https://github.com/dlang-community/DCD/releases/tag/v0.9.7

I guess generating the setup.exe is a bit more tricky (@s-ludwig: do you have a script for it?), but this should already bring us a lot further down the road.

PR #1638, opened 2019-01-04
Build win{32,64} binaries with LDC

wilzbach:
Cross-compilation to Windows got super-easy with recent LDCs.

Supersedes https://github.com/dlang/dub/pull/1471


ZombineDev:
BTW, Travis now has native Windows support. Why not use it?

PR #1641, opened 2019-01-20
dub.project: Improve variable expansion logic

CyberShadow:
See https://github.com/dlang/dmd/pull/9275#issuecomment-455861862

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1194,20 +1194,100 @@ private void processVars(ref Appender!(string[]) dst, in Project project, in Pac
 	foreach (var; vars) dst.put(processVars(var, project, pack, gsettings, are_paths));
 }
 
-private string processVars(Project, Package)(string var, in Project project, in Package pack,in GeneratorSettings gsettings, bool is_path)
+private string processVars(Project, Package)(string var, in Project project, in Package pack, in GeneratorSettings gsettings, bool is_path)
 {
-	import std.regex : regex, replaceAll;
-
-	auto varRE = regex(`\$([\w_]+)|\$\{([\w_]+)\}|(\$\$[\w_]+|\$\$\{[\w_]+\})`);
-	var = var.replaceAll!(
-		m => m[3].length ? m[3][1..$] : (getVariable(m[1].length ? m[1] : m[2], project, pack, gsettings))
-		)(varRE);
+	var = var.expandVars!(varName => getVariable(varName, project, pack, gsettings));
 	if (is_path) {
 		auto p = NativePath(var);
 		if (!p.absolute) {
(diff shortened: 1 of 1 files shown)


source/dub/project.d
```diff
@@ -1210,6 +1205,91 @@ private string processVars(Project, Package)(string var, in Project project, in
...
+	assertThrown(expandVars!expandVar("$"));
+	assertThrown(expandVars!expandVar("${}"));
+	assertThrown(expandVars!expandVar("$|"));
+	assertThrown(expandVars!expandVar("x$"));
+	assertThrown(expandVars!expandVar("$X"));
+	assertThrown(expandVars!expandVar("${"));
+	assertThrown(expandVars!expandVar("${X"));
+
+	// https://github.com/dlang/dmd/pull/9275
+	assert(expandVars!expandVar("$${DUB_EXE:-dub}") == "${DUB_EXE:-dub}");
```
jacob-carlborg: Should we test for other Bash variable syntaxes: https://www.tldp.org/LDP/abs/html/parameter-substitution.html ?

CyberShadow: Well, the point of this test is to explicitly check that Dub doesn't do anything crazy here, and it ignores anything following `$$` and passes it through verbatim, whatever the syntax might be. The previous implementation expected very specific syntax to be used even when escaping Dub processing, which was presumptuous and led to wrong behavior. This test (and any other tests we might add in the same vein) are supplemental / illustrative to the behavior tests above, and don't actually add to the test coverage. An implementation of `expandVars` which somehow passes all of these tests but would behave incorrectly with other kinds of Bash variable substitution syntax would be very weird indeed.

jacob-carlborg: Ok, good enough for me.

ZombineDev:
> Instead of looking for $ (minding escapes), it uses a regular expression to match what it thinks shell variable use looks like.

@CyberShadow Yes, it is working as intended - it's just that we have a different definition of what's intended. I asked for input on this matter on the newsgroup, but apparently no one cared enough to respond at the time, so we went with the current behavior.

* https://forum.dlang.org/post/exerdovrnivrsloetxtg@forum.dlang.org
* https://github.com/dlang/dub/pull/1392/files#diff-fc1737ce75bc9102f5d5b54be258a268R1188
* https://github.com/dlang/dub/pull/1552

CyberShadow:
@ZombineDev Sorry, what do you mean? Of the unit tests I added, how would you consider an implementation that failed them as "working as intended"?

wilzbach:
I'm going to merge this now as it still passes all the tests of the referenced https://github.com/dlang/dub/pull/1392 which is in master now.

PR #1643, opened 2019-01-31
Add dub fetch <package>@<version> shortcut

wilzbach:
Follow-up to https://github.com/dlang/dub/pull/1428#discussion_r252577400

FWIW the recently added `dub add` uses `=`, but as both `=` and `:` can't be part
of a valid version we can just support both as valid splitting symbols to avoid any potential confusion.

edit: now changed to `@` instead of `:`

### DIFF
--- a/changelog/dub-fetch-shortcut.dd
+++ b/changelog/dub-fetch-shortcut.dd
@@ -1,0 +1,9 @@
+`dub fetch` now supports `<package>@<version>` as a shortcut
+
+`dub fetch <package>@<version>` is a shortcut for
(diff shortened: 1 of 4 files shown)


s-ludwig:
`:` is used as the sub package delimiter, so that may lead to confusion (while I think the set of valid package names is distinct from the set of valid version names). I'd suggest to use the more commonly found `@` instead.

wilzbach:
> I'd suggest to use the more commonly found `@` instead.

Fair point. Changed to `@`.

PR #1648, opened 2019-02-11
issue 1645

vporton:
successfully tested

### DIFF
--- a/test/issue1645-dflags-build.sh
+++ b/test/issue1645-dflags-build.sh
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
+set -e
+
 # If DFLAGS are not processed, dub for library would fail
 DFLAGS="-w" $DUB --root="$CURR_DIR"/1-staticLib-simple --build=plain
 if DFLAGS="-asfdsf" $DUB --root="$CURR_DIR"/1-staticLib-simple --build=plain 2>/dev/null; then
-  echo "Should not accept this DFLAGS"
+  echo "Should not accept this DFLAGS";
+  false # fail
 fi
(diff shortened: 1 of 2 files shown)


test/issue1645-dflags-build.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# If DFLAGS are not processed, dub for library would fail
+DFLAGS="-w" $DUB --root="$CURR_DIR"/1-staticLib-simple --build=plain
+if DFLAGS="-asfdsf" $DUB --root="$CURR_DIR"/1-staticLib-simple --build=plain 2>/dev/null; then
+  echo "Should not accept this DFLAGS"
+fi
```
wilzbach: This isn't testing anything. It will always pass. You need to use at least `set -e`

vporton: Fixed.

wilzbach:
@vporton this still fails on the CIs.

```
[INFO] Running /home/travis/build/dlang/dub/test/issue1645-dflags-build.sh...
Performing "plain" build using dmd for x86_64.
staticlib-simple ~master: building configuration "library"...
Target is a library. Skipping execution.
[ERROR] Script failure.
```

joseph-wakeling-frequenz:
Have I understood right that the intended (and effective) behaviour of this change is that command-line set `DFLAGS` will be taken _in addition_ to anything set by the `--build` flag?  And that one must explicitly specify `--build="$DFLAGS"` to use just the environment-set DFLAGS?

If so (i) I like this, because the logical use of `DFLAGS='...' dub build ...` is to add extra custom flags, but (ii) it might be worth tweaking the docs for `dub build` to explain this.

Geod24:
Fixed in #2169

PR #1650, opened 2019-02-14
Central build cache

s-ludwig:
Add support for storing cached build artefacts in a central directory instead of in each package's ".dub" directory.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1214,6 +1214,11 @@ class GenerateCommand : PackageBuildCommand {
 		if (!gensettings.config.length)
 			gensettings.config = m_defaultConfig;
 		gensettings.runArgs = app_args;
+		gensettings.force = m_force;
+		gensettings.rdmd = m_rdmd;
+		gensettings.tempBuild = m_tempBuild;
+		gensettings.parallelBuild = m_parallel;
+		gensettings.buildCacheDirectory = dub.buildCacheDirectory;
 
 		logDiagnostic("Generating using %s", m_generator);
 		dub.generateProject(m_generator, gensettings);
@@ -1432,6 +1437,7 @@ class TestCommand : PackageBuildCommand {
 		settings.compiler = getCompiler(this.baseSettings.platform.compilerBinary);
 		settings.run = true;
 		settings.runArgs = app_args;
+		settings.buildCacheDirectory = dub.buildCacheDirectory;
 
 		dub.testProject(settings, this.baseSettings.config, NativePath(m_mainFile));
 		return 0;
@@ -2386,6 +2392,7 @@ class DustmiteCommand : PackageBuildCommand {
 			gensettings.compileCallback = check(m_compilerStatusCode, m_compilerRegex);
 			gensettings.linkCallback = check(m_linkerStatusCode, m_linkerRegex);
 			gensettings.runCallback = check(m_programStatusCode, m_programRegex);
+			gensettings.buildCacheDirectory = dub.buildCacheDirectory;
 			try dub.generateProject("build", gensettings);
 			catch (DustmiteMismatchException) {
 				logInfoNoTag("Dustmite test doesn't match.");
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1214,10 +1214,6 @@ class GenerateCommand : PackageBuildCommand {
 		if (!gensettings.config.length)
 			gensettings.config = m_defaultConfig;
 		gensettings.runArgs = app_args;
-		gensettings.force = m_force;
-		gensettings.rdmd = m_rdmd;
-		gensettings.tempBuild = m_tempBuild;
-		gensettings.parallelBuild = m_parallel;
 		gensettings.buildCacheDirectory = dub.buildCacheDirectory;
 
 		logDiagnostic("Generating using %s", m_generator);
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -417,6 +417,14 @@ class Dub {
 	@property const(string[string]) defaultPreRunEnvironments() const { return this.m_config.defaultPreRunEnvironments; }
 	@property const(string[string]) defaultPostRunEnvironments() const { return this.m_config.defaultPostRunEnvironments; }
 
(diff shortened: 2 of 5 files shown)


wilzbach:
I added auto-merge to it as its a very straight-forward change and the master/stable branch-off will likely happen tomorrow.
If there are any controversies with this, we can always temporarily revert this and include it in 2.086

source/dub/generators/build.d
```diff
@@ -248,6 +244,18 @@ class BuildGenerator : ProjectGenerator {
...
+	private NativePath getBuildCacheDirectory(GeneratorSettings settings, in Package pack, string build_id)
+	{
+		NativePath base_path;
+		if (settings.buildCacheDirectory != NativePath.init)
+			base_path = settings.buildCacheDirectory;
+		else if (settings.tempBuild)
+			base_path = getTempDir() ~ format("dub/builds");
+		else return pack.path ~ format(".dub/build/%s/", build_id);
+
+		return base_path ~ format("%s/%s/%s/%s/", pack.basePackage.name, pack.version_, build_id);
```
wilzbach: Let's better use `buildPath` for all of these formats here.
This will also avoid the error on Windows:

```
std.format.FormatException@C:\dmd2\windows\bin\..\..\src\phobos\std\format.d(486): Orphan format specifier: %s
----------------
0x00007FF6305A94E3 in std.exception.bailOut!(std.format.FormatException).bailOut at C:\dmd2\windows\bin\..\..\src\phobos\std\exception.d(515)
0x00007FF6305A93E9 in std.exception.enforce!(std.format.FormatException).enforce!bool.enforce at C:\dmd2\windows\bin\..\..\src\phobos\std\exception.d(436)
0x00007FF63062B242 in std.format.formattedWrite!(std.array.Appender!string, char, string, dub.dependency.Version, string).formattedWrite at C:\dmd2\windows\bin\..\..\src\phobos\std\format.d(487)
0x00007FF6306E39AA in std.format.format!(char, string, dub.dependency.Version, string).format at C:\dmd2\windows\bin\..\..\src\phobos\std\format.d(6385)
0x00007FF63083BE36 in dub.generators.build.BuildGenerator.getBuildCacheDirectory at C:\projects\dub\source\dub\generators\build.d(256)
```

s-ludwig: I used `NativePath`'s concatenation operator now. It's actually strange that I didn't get the orphan format specifier error, since I wrote this on a Windows machine.

s-ludwig:
I didn't want to do this in the first step, but I think we should probably eventually make this behavior the default, allowing to opt-out by setting the configuration field to an empty/null entry. Especially for single-file packages it is much nicer than polluting the parent directory or performing an uncached temp-build.

ZombineDev:
Agreed, let's make it the default.

Geod24:
This is a great feature. The behavior of `dub` is not compatible with e.g. Docker and path dependency.
In our use case, we have submodules as path dependencies, and we had to add [a few special cases to our Dockerfile](https://github.com/bpfkorea/agora/blob/21bef9a33b9377c080064f9bed0cf82e7f9d0976/.dockerignore#L13-L21) to prevent long build (otherwise we were sending > 1 GB of data to the container, instead of ~50 mbs).

source/dub/dub.d
```diff
@@ -417,6 +417,14 @@ class Dub {
...
 	@property const(string[string]) defaultPostRunEnvironments() const { return this.m_config.defaultPostRunEnvironments; }
 
+	/** An optional directory that is used for storing cached build artefacts.
+
+		This is used by `BuildGenerator` to store intermediate object files
+		for reuse in incremental builds. The value of this proprerty may be
+		an empty path, in which case no directory is configured.
+	*/
+	@property NativePath buildCacheDirectory() const { return m_config.buildCacheDirectory; }
+
```
Geod24, line 427: Instead of adding piece-meal getters, which I think doesn't scale, we could just return the `m_config`?

source/dub/dub.d
```diff
@@ -1884,6 +1893,7 @@ private struct UserConfiguration {
 
 	@Optional string[] registryUrls;
 	@Optional NativePath[] customCachePaths;
+	@Optional NativePath buildCacheDirectory;
```
Geod24, line 1896: I'd like to re-think a bit the configuration, and have something a bit more segmented.
In this case, we could have:
```JSON
{
    "build": {
        "cache": "path/to/cache"
    }
}
```

source/dub/commandline.d
```diff
@@ -1214,6 +1214,11 @@ class GenerateCommand : PackageBuildCommand {
 		if (!gensettings.config.length)
 			gensettings.config = m_defaultConfig;
 		gensettings.runArgs = app_args;
+		gensettings.force = m_force;
+		gensettings.rdmd = m_rdmd;
+		gensettings.tempBuild = m_tempBuild;
+		gensettings.parallelBuild = m_parallel;
```
Geod24, line 1220: Those are written directly to the `baseSettings` now.

source/dub/dub.d
s-ludwig: Hm, not sure about that. `UserConfiguration` (unless I'm mistaken) represents the format of the settings file. Tying that to the public API and at the same time considering changing the file format makes it sound like a leaky abstraction that we should probably avoid.

Geod24: It is indeed. But I don't see why we are exposing `buildCacheDirectory` and other settings. It seems that every time someone touches the configuration, they just cargo-cult-ishly copy the existing approach.

Geod24:
In general, I'd like us to be able to express the default behavior using the configuration. Meaning that there could be a configuration the user can enter that allows them to do exactly what no configuration would do. It's IMO important to avoid building special case, and it usually leads to cleaner code as you can then set a default value as field initializer and things just work.

In this case, I think we should take into account environment / package variable. The user should IMO be able to use `DUB_PACKAGE_DIR` and `DUB_ROOT_PACKAGE_DIR` to accomplish what the default is doing.

Geod24:
Closing in favor of https://github.com/dlang/dub/pull/2542

PR #1571, opened 2018-10-07
toolchainRequirements recipe support

rtbo:
following discussion in #1531 and #1532 

Suppresses `minDubVersion` field and add support for parsing new recipe entry `toolchainRequirements`:
SDL:
```sdl
toolchainRequirements dub="~>1.11" dmd="~>2.082" ldc=... gdc=...
```
JSON:
```json
{
    "toolchainRequirements": {
        "dub": "~>1.11",
        "dmd": "~>2.082"
    }
}
```
Currently only actually checks for the dub version compatibility (therefore only replaces the `minDubVersion` field).
`toolchainRequirements` terminology is of course not set in stone. I have no preference at all on the name we give to this recipe section.

PR #1655, opened 2019-02-17
Fix syntax inconsistency in toolchainRequirements

s-ludwig:
Removes the additional '|' syntax for toolchain requirements from #1571.

- '|' seems confusing, given that it is used as a logical AND
- Dependency already supports specifying an arbitrary range and this is the only meaningful use for '|'
- Keeps the possibility open to use '|' as a logical OR in a later version
- Rewords the change log entry and adds a link to the DUB documentation

Another thing that seems a bit off is the fact that the dependency specifications for tool chain requirements are stored as strings whereas the usual ones are stored as `Dependency`. Changing this later on will require a new major version, so this should better be right from the start.

### DIFF
--- a/changelog/toolchain_requirements.dd
+++ b/changelog/toolchain_requirements.dd
@@ -28,7 +28,7 @@ $(UL
 )
 
 Each can contain a
-$(LINK2 https://dub.pm/package-format-sdl.html#version-specs version specification),
+$(LINK2 https://dub.pm/package-format-sdl.html#version-specs, version specification),
 where DMD-like versions are supported in addition to SemVer versions. For
 compilers, instead of a version specification, the keyword `no` can also be used
 to indicate that the compiler should not be used for this package.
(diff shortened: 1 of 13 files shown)


s-ludwig:
Changed the internal representation to `Dependency`. This also removes a bunch of public API functions to keep the technical dept under control. I think that the two remaining methods added to `Compiler` should also be removed, as they can be realized from the outside by looking at `Compiler.name` instead.

s-ludwig:
Should be okay now. The public API changes are now limited to the new `ToolchainRequirements` struct.

The last thing that I'm wondering is whether the build should really fail if a requirement is not met. If people now start adding an upper bound on the version (e.g. `"frontend": ">=2.068 <=2.084"`), those packages will be broken by every new compiler release, by default.

wilzbach:
(added a missing comma for LINK2 in the changelog entry)

PR #1658, opened 2019-02-25
Custom init type

andre2007:
Dub packages which have an "init" sub package can be used in command

```
dub init -n myPackage --format sdl -t custom-dub-init-dubpackage -- --foo=bar
```

Folder myPackage will be created containing a template dub.sdl.
Package ```custom-dub-init-dubpackage``` will be fetched and sub package ```init```
will be built and executed. The execution work directory is the new package directory with
the template dub.sdl.
Free args ```--foo=bar``` will be passed to the "init" executable.

The init executable has the task to create skeleton source code.


changelog/custom-dub-init-type.dd
```diff
@@ -0,0 +1,8 @@
+Support of custom dub init type
+
+Command dub init now supports custom dub packages for argument -t.
+`dub init -n myPackage --format sdl -t custom-dub-init-dubpackage -- --foo=bar`
```
wilzbach:
```suggestion
$(CONSOLE
> dub init -n myPackage --format sdl -t custom-dub-init-dubpackage -- --foo=bar
)
```

changelog/custom-dub-init-type.dd
```diff
@@ -0,0 +1,8 @@
+Support of custom dub init type
+
+Command dub init now supports custom dub packages for argument -t.
```
wilzbach:
```suggestion
Command dub init now supports custom dub packages for argument `-t`.
```

wilzbach:
Added 72h no objection -> merge.

s-ludwig:
Unfortunately I missed this before the merge. Do I get it right that this requires the package to create the generated package programmatically? In that case, what would be the anticipated advantages that outweigh the additional complexity required in each project that provides such an initialization? Additional complexity meaning in contrast to an approach like the one outlined in https://github.com/dlang/dub/pull/600#issuecomment-143585527.

andre2007:
@s-ludwig 
Unfortunately I do not understand your last comment.

In the referenced comment you proposed that instead of specifying 
```-t custom-dub-init```, there should be some logic which automatically
find out, that ```custom-dub-init``` has an init logic and execute it implicit.

```dub init -n myPackage custom-dub-init-dubpackage package2 package3```

If you meant that with you recent comment, I see this as enhancement on top
of pull request 1658. This pull request enables to specify the init package explicitly.
Adding the implicit call is now rather easy to implement.

s-ludwig:
Okay, no, the critical difference I meant was that the approach in this PR requires procedural generation of the created package, while I was proposing to take the `:init` package as a template for the newly created one. So the `:init` package would be copied to the destination directory recursively and the package recipe would then be adjusted to match the information given to `dub init`.

While this would of course be less flexible, I think that it should suffice for 99% of the cases and would be trivial to implement for package authors, while at the same time serving as a simple usage example/test. I don't mind a procedural approach for the few cases where that is needed, but I think the default should be something with a very low implementation barrier for package authors.

andre2007:
@s-ludwig I fully agree with you. As suggestion, the new logic will only be executed if there is a sub package ```init-exec```. 
Therefore, if there is a package ```init``` then the logic you described will be executed. 
What do you think?

PR #1666, opened 2019-03-05
Upgrade dependencies

s-ludwig:
This is the result of a "dub upgrade". In particular, it fixes a deprecation warning triggered by an old diet-ng version that currently blocks https://github.com/dlang/dmd/pull/9393. See also https://github.com/rejectedsoftware/diet-ng/pull/62.

### DIFF
--- a/dub.selections.json
+++ b/dub.selections.json
@@ -1,18 +1,19 @@
 {
 	"fileVersion": 1,
 	"versions": {
-		"botan": "1.12.9",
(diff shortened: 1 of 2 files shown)


s-ludwig:
Changed to keep the taggedalgebraic dependency at 0.10.x, because 0.11.x is not compatible with the required old compiler versions (maybe we should also revisit the current compiler support list again).

PR #1668, opened 2019-03-09
Support of init-exec sub package

andre2007:
Quote from @s-ludwig (https://github.com/dlang/dub/pull/1658)

> Okay, no, the critical difference I meant was that the approach in this PR requires procedural generation of the created package, while I was proposing to take the :init package as a template for the newly created one. So the :init package would be copied to the destination directory recursively and the package recipe would then be adjusted to match the information given to dub init.
> 
> While this would of course be less flexible, I think that it should suffice for 99% of the cases and would be trivial to implement for package authors, while at the same time serving as a simple usage example/test. I don't mind a procedural approach for the few cases where that is needed, but I think the default should be something with a very low implementation barrier for package authors.

With this pr, the existing procedural generation is no longer executed for sub package ```init```
but for sub package ```init-exec```. The copy logic for ```init``` as propose by @s-ludwig still
has to be implemented in another pull request.


ZombineDev:
```
[INFO] Running /home/runner/dub/test/issue1651-custom-dub-init-type.sh...
custom-dub-init-dubpackage is not present, getting and storing it user wide
Fetching custom-dub-init-dubpackage 1.0.1...
Storing package 'custom-dub-init-dubpackage', version requirements: >=0.0.0
Sub package "custom-dub-init-dubpackage:init-exec" doesn't exist.
[ERROR] /home/runner/dub/test/issue1651-custom-dub-init-type.sh:7 command failed
[ERROR] Script failure.
```

andre2007:
@ZombineDev thanks. In the zip file I forgot to adapt the name in 1 place. Hopefully I have this evening 5 free minutes to correct it.

andre2007:
@ZombineDev done. I think the failing travis build is unrelated.

wilzbach:
> @ZombineDev done. I think the failing travis build is unrelated.

Yup.

andre2007:
Merge? :)

PR #1661, opened 2019-02-28
Windows: Don't use Optlink by default

wilzbach:
There's __a lot__ of evidence that the Windows experience of D is frustrating as almost every big Dub package doesn't compile with Optlink and the user manually needs to figure out to switch to COFF.

This PR proposes to use COFF by default on Windows.

Evidence collection:
--------------------------

- https://github.com/vibe-d/vibe.d/issues/2243
- https://forum.dlang.org/thread/bobqtoesasjlwvdiisih@forum.dlang.org
- https://forum.dlang.org/post/wpxsfavlgqtfqplwizme@forum.dlang.org
- https://forum.dlang.org/post/hoyaqzytksiuvdlsitaw@forum.dlang.org
- https://forum.dlang.org/post/uaprrobsefqhaqahcthu@forum.dlang.org
- https://forum.dlang.org/post/xbhsbhepqcnfbcveuduw@forum.dlang.org
- https://forum.dlang.org/post/xbzqqhzjrjxndzgqogyk@forum.dlang.org
- https://issues.dlang.org/show_bug.cgi?id=17508
- https://github.com/vibe-d/vibe.d/issues/2122
- https://github.com/vibe-d/vibe.d/issues/1910
- https://github.com/vibe-d/vibe.d/issues/1771
- https://github.com/vibe-d/vibe.d/issues/1439
- https://forum.dlang.org/post/owoazhliszwqcqvsnifl@forum.dlang.org
- https://forum.dlang.org/post/abxmmjwpwmelxyicuzip@forum.dlang.org
- https://forum.dlang.org/post/zjhvufwmratwxcpywjsg@forum.dlang.org
- https://forum.dlang.org/post/ybufckdvasbnehwbdnec@forum.dlang.org
- https://forum.dlang.org/post/gemkzmxsdyilziloorqv@forum.dlang.org
- https://forum.dlang.org/post/yawsuwxbzkdlboniyyio@forum.dlang.org
- https://forum.dlang.org/post/ryzmhycnzowbyurvnlsu@forum.dlang.org
...

TODO:
- [x] changelog entry
- [x] Test on Windows

I don't have a Windows box at the moment, so if one of the many devs who actually run into this issue would step to test this, this would be awesome.

PR #1674, opened 2019-03-22
Fix LDC versions for the deployment stage

wilzbach:
https://github.com/dlang/dub/pull/1661#issuecomment-475217413

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
     - d: ldc-1.3.0
       env: [FRONTEND=2.073]
     - stage: deploy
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           api_key: $GH_REPO_TOKEN
           on:
             tags: true
-    - d: ldc
+    - d: ldc-1.13.0
```
wilzbach: Looks like this started to fail with LDC 1.14: https://github.com/dlang/dub/pull/1661#issuecomment-475217413

PR #1366, opened 2018-02-09
use single api request to fill metadata cache for all dependencies

MartinNowak:
- avoids various RTTs and possible failures

Requires https://github.com/dlang/dub-registry/pull/278

PR #1676, opened 2019-04-07
Use new minimized API endpoint for package queries

wilzbach:
Branch-off from https://github.com/dlang/dub/pull/1366 as the `/infos` endpoint
isn't available yet, but using the minimized JSON already yields a huge benefit.


wilzbach:
CC @thewilsonator. This is a very trivial and good optimization for DUB's fetching process. Any chance we can get this in soonish (i.e. before 2.086)?

PR #1677, opened 2019-04-08
Fetch dependencies in parallel

wilzbach:
A very low hanging fruit to speed up dub's build time (especially for e.g. CI systems where the dependencies need to be fetched every time).

 Example | `master` | `parallel`
-------------|-------------|------
Vibe.d      |  26s       | 10s
DLS         |  61s       | 28s

(Before all test `~/.dub` was purged.)

For `dls` I used this:

```
time ~/dlang/dub-repo/bin/dub build --yes dls --verbose
```

(so the actual build time is included in there too)

It would be nice to get this in with the 2.086 release (scheduled to branch off on the ~14/15.4).


wilzbach:
Argh, it's apparently not that easy. Sometimes it's segfaulting :/

```
/home/travis/build/dlang/dub/test/fetchzip.sh: line 10:  5273 Segmentation fault      (core dumped) "$DUB" build --single "$DIR"/test_registry.d
[ERROR] :10 command failed
```

source/dub/dub.d
```diff
@@ -556,7 +556,10 @@ class Dub {
 			return;
 		}
 
-		foreach (p; versions.byKey) {
+		import std.parallelism : TaskPool, totalCPUs;
+		// Avoid being blocked by a slow download (even on a single-threaded machines)
+        auto taskPool = new TaskPool(totalCPUs.clamp(3, 6));
```
WebFreak001: You mixed tabs and spaces here.

Also I think this could be `scope` to automatically destroy the TaskPool object at the end.

Should the download speed even be regulated by CPU count at all? I think moving out the parsing/whatever out of the parallel loop would allow us to pick some well working static count of download threads. Then it can submit all downloaded packages for work to some other task pool which will run the more difficult operations on exactly the right count of CPU tasks so systems with many cores can process more packages at once.

andre2007:
It would be really great if we can get this PR merged. The performance gain for DUB registry is already great but for Maven repositories it is even more. Maven provider like Nexus need 8-12 seconds while retrieving the meatadata pom file (The pom file is not stored on Nexus but calculated on the fly). 
Therefore retrieving dependencies from Maven sequentially is a huge pain at the moment. This PR will mitigate the pain:)

Issue #1671, opened 2019-03-13
Don't parse compiler version if toolchainRequirements is not used


jacob-carlborg:
Compiling a project using LDC nightly build I got the follow error: `Invalid SemVer format: 1.15.0git-e3e96b8` [1]. `1.15.0git-e3e96b8` is the version of LDC. The `dub.json` file does not contain `toolchainRequirements`.

[1] https://travis-ci.com/jacob-carlborg/dvm/jobs/184455053#L222

wilzbach:
This is related to https://github.com/dlang/dub/pull/1571 as now the compiler version is parsed on every target generation. In particular, for `ldc` `parseDependency` is called which tries to parse `1.15.0git-e3e96b8` as a SemVer dependency and thus subsequently fails.

See also:

https://github.com/dlang/dub/blob/c707c032591d4506d7087bc160819bd8f7ee0bb9/source/dub/generators/build.d#L57

https://github.com/dlang/dub/blob/c707c032591d4506d7087bc160819bd8f7ee0bb9/source/dub/recipe/packagerecipe.d#L396-L404

PR #1679, opened 2019-04-19
Use -dumpfullversion to determine GDC version

belka-ew:
Parsing `gdc -v` output is not reliable. On my machine it looks like:

```
Target: x86_64-slackware-linux
Configured with: ../gcc-8.2.0/configure --prefix=/opt/gdc --libdir=/opt/gdc/lib64 --mandir=/usr/man --infodir=/usr/info --enable-shared --enable-bootstrap --enable-languages=c,c++,d --enable-threads=posix --enable-checking=release --with-system-zlib --enable-libstdcxx-dual-abi --disable-libunwind-exceptions --enable-__cxa_atexit --enable-libssp --enable-lto --disable-install-libiberty --with-gnu-ld --verbose --with-arch-directory=amd64 --program-suffix=-8 --with-pkgversion='GDC 2.081.2' --disable-multilib --target=x86_64-slackware-linux --build=x86_64-slackware-linux --host=x86_64-slackware-linux
Thread model: posix
gcc version 8.2.0 (GDC 2.081.2) 
COLLECT_GCC_OPTIONS='-v' '-o' 'a.out' '-shared-libgcc' '-mtune=generic' '-march=x86-64'
 /opt/gdc/libexec/gcc/x86_64-slackware-linux/8.2.0/cc1d ../test.d -quiet -dumpbase test.d -mtune=generic -march=x86-64 -auxbase test -version -v -o /tmp/ccFfoCDi.s
GNU D (GDC 2.081.2) version 8.2.0 (x86_64-slackware-linux)
	compiled by GNU C version 8.2.0, GMP version 6.1.1, MPFR version 3.1.4, MPC version 1.0.3, isl version none
GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
GNU D (GDC 2.081.2) version 8.2.0 (x86_64-slackware-linux)
	compiled by GNU C version 8.2.0, GMP version 6.1.1, MPFR version 3.1.4, MPC version 1.0.3, isl version none
GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
binary    /opt/gdc/libexec/gcc/x86_64-slackware-linux/8.2.0/cc1d
version   v2.081.2
predefs   GNU D_Version2 LittleEndian GNU_DWARF2_Exceptions GNU_StackGrowsDown GNU_InlineAsm D_LP64 assert all X86_64 D_HardFloat Posix linux CRuntime_Glibc
```

The regular expressions prefers "GDC 2.081.2" and fails with "Invalid SemVer".

GCC/GDC has `-dumpfullversion` which outputs the compiler version as SemVer.

Two flags, `-dumpversion` and `-dumpfullversion`, are required for compatibility with elder GCC (`-dumpfullversion` is available since GCC 7, for GCC 6 and earlier `-dumpversion` does the same).

clang has `-dumpversion` as well, but it doesn't seem to work with ldc.

By the way https://github.com/dlang/dub/issues/1671 can be probably closed, it is fixed on the ldc side: https://github.com/ldc-developers/ldc/issues/3025

### DIFF
--- a/source/dub/compilers/gdc.d
+++ b/source/dub/compilers/gdc.d
@@ -54,31 +54,15 @@ class GDCCompiler : Compiler {
 
 	@property string name() const { return "gdc"; }
 
-	enum gdcVersionRe1 = `GDC\s+(\d+\.\d+\.\d+[A-Za-z0-9.+-]*)`;
-	enum gdcVersionRe2 = `^GNU D \(.*?\) version (\d+\.\d+\.\d+[A-Za-z0-9.+-]*)`;
-	enum gdcVersionRe3 = `^gcc version (\d+\.\d+\.\d+[A-Za-z0-9.+-]*)`;
+	string determineVersion(string compiler_binary, string verboseOutput)
+	{
+		const result = execute([
+			compiler_binary,
+			"-dumpfullversion",
+			"-dumpversion"
+		]);
 
-	unittest {
-		import std.algorithm : equal, map;
-		import std.range : only;
-		import std.regex : matchFirst, regex;
-		auto probe = `
-Target: x86_64-pc-linux-gnu
-Thread model: posix
-gcc version 8.2.1 20180831 (GDC 8.2.1 based on D v2.068.2 built with ISL 0.20 for Arch Linux)
-GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
-GNU D (GDC 8.2.1 based on D v2.068.2 built with ISL 0.20 for Arch Linux) version 8.2.1 20180831 (x86_64-pc-linux-gnu)
-GGC heuristics: --param ggc-min-expand=100 --param ggc-min-heapsize=131072
-binary    /usr/lib/gcc/x86_64-pc-linux-gnu/8.2.1/cc1d
-version   v2.068.2
-predefs   GNU D_Version2 LittleEndian GNU_DWARF2_Exceptions GNU_StackGrowsDown GNU_InlineAsm D_LP64 D_PIC assert all X86_64 D_HardFloat Posix linux CRuntime_Glibc
-`;
-		auto vers = only(gdcVersionRe1, gdcVersionRe2, gdcVersionRe3)
-			.map!(re => matchFirst(probe, regex(re, "m")))
-			.filter!(c => c && c.length > 1)
-			.map!(c => c[1]);
-
-		assert(vers.equal([ "8.2.1", "8.2.1", "8.2.1" ]));
+		return result.status == 0 ? result.output : null;
 	}
 
 	BuildPlatform determinePlatform(ref BuildSettings settings, string compiler_binary, string arch_override)
@@ -97,8 +81,7 @@ predefs   GNU D_Version2 LittleEndian GNU_DWARF2_Exceptions GNU_StackGrowsDown G
 		return probePlatform(
 			compiler_binary,
 			arch_flags ~ ["-S", "-v"],
-			arch_override,
-			[ gdcVersionRe1, gdcVersionRe2, gdcVersionRe3 ]
+			arch_override
 		);
 	}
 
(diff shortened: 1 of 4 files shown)


source/dub/compilers/gdc.d
```diff
@@ -54,31 +54,16 @@ class GDCCompiler : Compiler {
...
-
-		assert(vers.equal([ "8.2.1", "8.2.1", "8.2.1" ]));
+	string determineVersion(string compiler_binary, string verboseOutput)
+	{
+		const args = [
+			compiler_binary,
+			"-dumpfullversion",
+			"-dumpversion"
+		];
+		auto result = executeShell(escapeShellCommand(args));
```
wilzbach: No need to create a shell here.

source/dub/compilers/gdc.d
```diff
@@ -54,31 +54,16 @@ class GDCCompiler : Compiler {
...
-			.filter!(c => c && c.length > 1)
-			.map!(c => c[1]);
-
-		assert(vers.equal([ "8.2.1", "8.2.1", "8.2.1" ]));
+	string determineVersion(string compiler_binary, string verboseOutput)
+	{
+		const args = [
+			compiler_binary,
+			"-dumpfullversion",
+			"-dumpversion"
```
wilzbach: Does this work with older GDC versions too?

wilzbach:
> By the way #1671 can be probably closed, it is fixed on the ldc side: 

I think LDC only fixed their nightly tags, no?

source/dub/compilers/gdc.d
belka-ew: Surprisingly no, all GDC versions just ignore even invalid flags if `-dumpfullversion` or `-dumpversion` is present in the command line.

`-dumpversion` returns only the major version since gcc-7.

See here an output example: https://github.com/smanders/externpro/issues/160#issuecomment-355154694

Here is gcc-5 and gdc-8 output:

> $ gcc -dumpfullversion -dumpversion
> 5.5.0
> $ gdc -dumpfullversion -dumpversion
> 8.2.0

belka-ew:
> I think LDC only fixed their nightly tags, no?

I thought, only nightly tags are a problem. Otherwise all compilers use SemVer and dmd is handled separately anyway.

source/dub/compilers/gdc.d
belka-ew: Or to be more precise `-dumpversion` outputs the version used in the directory structure, in the directories like "/opt/gdc/include/c++/**8.2.0**/algorithm" and it can be "/opt/gdc/include/c++/**8**/algorithm" if `--with-gcc-major-version-only` is given at build time.

source/dub/compilers/gdc.d
belka-ew: Replaced with `execute`.

Geod24:
From the example provided [here](https://github.com/dlang/dub/pull/1679#discussion_r277734174) it looks like the version provided is GCC's, not the version of the frontend. I don't think we want to expose this version in DUB...

belka-ew:
> From the example provided here it looks like the version provided is GCC's, not the version of the frontend ?

But it is exactly what `Compiler.determineVersion` does, for LDC as well. It returns the LDC version, e.g. **1.11.0**, not the frontend version.

```
	unittest {
		import std.regex : matchFirst, regex;
		auto probe = `
binary    /usr/bin/ldc2
version   1.11.0 (DMD v2.081.2, LLVM 6.0.1)
config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
`;
		auto re = regex(ldcVersionRe, "m");
		auto c = matchFirst(probe, re);
		assert(c && c.length > 1 && c[1] == "1.11.0");
	}
```

So it is consistent with what it is doing now.

wilzbach:
Yup, getting the right frontend version is important for e.g. dub's platform compatibility feature which has been part of the last release.
On the long run though I think this should be united with dmd's `-Xf-` JSON interface which allows us to get rid of this entire platform probing.

PR #1687, opened 2019-04-30
Fix #1686: Running single-file package as pre generate command always rebuilds

### LINKED ISSUE #1686: Running single-file package as pre generate command always rebuilds
If a single-file Dub package is used as a pre generate command the project containing the pre generate command will always be rebuilt. This happens because the path to the single-file package is not correctly added to the package. This will later lead to the current directory, `.`, being added as a dependency instead of the path to single-file package.

This happens to the DMD Dub package.

jacob-carlborg:
If a single-file Dub package is used as a pre generate command the project containing the pre generate command will always be rebuilt. This happens because the path to the single-file package is not correctly added to the package. This will later lead to the current directory, `.`, being added as a dependency instead of the path to single-file package.

I added a new overload to `loadSingleFilePackage` that takes both the path and the file content. This allows to easier unit test the function.


source/dub/dub.d
```diff
@@ -458,12 +468,25 @@ class Dub {
 			recipe.buildSettings.targetType = TargetType.executable;
 
 		auto pack = new Package(recipe, path.parentPath, null, "~master");
-		loadPackage(pack);
+		pack.m_infoFile = path;
```
jacob-carlborg: This is the actual fix.

wilzbach:
Have a look at the CIs:

```
Running ./bin/dub-test-library-nonet 
object.Exception@../.dub/packages/vibe-core-1.6.0/vibe-core/source/vibe/core/path.d(481): Path has no parent path
----------------
```

source/dub/dub.d
```diff
@@ -458,12 +468,25 @@ class Dub {
...
+		enum filePath = NativePath("foo.d");
+
+		enum content = q{
+			/+ dub.sdl:
+				name "foo"
+			+/
+		};
+
+		auto dub = new Dub(".", null, SkipPackageSuppliers.all);
+		const pack = dub.loadSingleFilePackage(filePath, content);
```
wilzbach: This fails when tested with Vibe.d:

```
Running ./bin/dub-test-library-nonet 
object.Exception@../.dub/packages/vibe-core-1.6.0/vibe-core/source/vibe/core/path.d(481): Path has no parent path
----------------
../.dub/packages/vibe-core-1.6.0/vibe-core/source/vibe/core/path.d:482 const @nogc @safe vibe.core.path.GenericPath!(vibe.core.path.PosixPathFormat).GenericPath vibe.core.path.GenericPath!(vibe.core.path.PosixPathFormat).GenericPath.parentPath() [0xb3168a]
source/dub/dub.d:470 dub.package_.Package dub.dub.Dub.loadSingleFilePackage(vibe.core.path.GenericPath!(vibe.core.path.PosixPathFormat).GenericPath, immutable(char)[]) [0xa6e163]
source/dub/dub.d:487 void dub.dub.Dub.__unittest_L476_C2() [0xa6e2d3]
```

The problem is that parentPath requires the path to have a parent. An easy fix would be to use `foo/bar` as dummy test path as `loadSingleFilePackage` can only be called with an absolute path.

jacob-carlborg:
@wilzbach the tests are passing now.

jacob-carlborg:
I'm not familiar with the Dub code base, but I don't understand why the root path is passed in and then the path to the recipe file is set after contracting the `Package`. Why not pass the full path to the recipe file and extract get the root path from that.

andre2007:
@jacob-carlborg I am not able to reproduce the issue you mentioned. Could you please add a test, maybe similiar like https://github.com/dlang/dub/pull/1811/files#diff-89f43ad1c5b2a1679f0a0bf54a215d4c ?

Imperatorn:
What remains to be done for this to be merged?

source/dub/dub.d
```diff
@@ -411,15 +411,25 @@ class Dub {
...
 		import std.file : mkdirRecurse, readText;
 		import std.path : baseName, stripExtension;
 
-		path = makeAbsolute(path);
-
-		string file_content = readText(path.toNativeString());
-
 		if (file_content.startsWith("#!")) {
 			auto idx = file_content.indexOf('\n');
 			enforce(idx > 0, "The source fine doesn't contain anything but a shebang line.");
```
Imperatorn, line 435: file

Imperatorn (commented):
Some typos and a question

PR #1696, opened 2019-05-05
fix #576 - Clean up tempory files on SIGINT

### LINKED ISSUE #576: Temporary files not cleaned up
dub creates tons of `dub_platform_probe*` and `dub-build*` files in `/tmp` (openSUSE, probably other platforms too). It should delete them again when they are no longer needed.

Samathy:
Default handler means the module dtor doesnt get called.
Calling exit() explicitly ensures cleanup happens.

Suggestions on testing this welcome. I'm sure you can send a signal to a process in a script and test for files existing, but maybe there is a better way?


source/dub/internal/utils.d
```diff
@@ -95,6 +97,18 @@ auto lockFile(string path, Duration timeout)
 	}
 }
 
+static this()
```
thewilsonator: This gets called once per thread, I'm not sure if dub is multithreaded, but either way this only needs to be done once with `shared static this()`.

Samathy: :+1: Updated.

Samathy:
Fab, 

I'll give tests some thought.

thewilsonator:
@Samathy ping. any ideas?

Samathy:
@thewilsonator I havent had any ideas on how to test this, maybe its worth making an issue to make sure we remember to make some tests at some point.

wilzbach:
> I'm sure you can send a signal to a process in a script

Why not do that then? ;-)

Samathy:
@wilzbach Aha, yes, sorry :woman_facepalming: I'll do that.

source/dub/internal/utils.d
```diff
@@ -95,6 +97,18 @@ auto lockFile(string path, Duration timeout)
...
 
+shared static this()
+{
+    signal(SIGINT, &handleSigInt);
+}
+
+extern(C) @nogc @system nothrow 
+void handleSigInt(int sig)
+{
+    exit(sig);
```
kubo39: exit() is not an async-signal-safe function,  `_Exit` or `_exit` would be nice.

PR #1700, opened 2019-05-09
fix #1372 - Ignore files in hidden directories.

### LINKED ISSUE #1372: Dub doesn't ignore non-empty hidden files
### System information
- **dub version**:DUB version 1.6.0, built on Jan 23 2018
- **OS Platform and distribution**: arm v7 (raspberry pi 3)
- **compiler version**
LDC - the LLVM D compiler (1.7.0):
  based on DMD v2.077.1 and LLVM 5.0.1
  built with LDC - the LLVM D compiler (1.6.0)
  Default target: arm-unknown-linux-gnueabihf
  Host CPU: cortex-a53
  http://dlang.org - http://wiki.dlang.org/LDC

  Registered Targets:
    arm     - ARM
    armeb   - ARM (big endian)
    thumb   - Thumb
    thumbeb - Thumb (big endian)

### When mounting a drive from OS X using afp, OS X automatically creates .AppleDouble folders within each mounted folder. The .AppleDouble folders contain files with the same names as the ones existing in the original folder. For example:
```
pi@raspberrypi:~/ $ ls -al source/
drwxr-xr-x  4 pi pi 4096 Feb 11 16:58 .
drwxr-xr-x 10 pi pi 4096 Feb 11 16:58 ..
-rw-r--r--  1 pi pi 1889 Feb 11 14:06 app.d
drwxr-xr-x  2 pi pi 4096 Feb 11 16:58 .AppleDouble
-rw-r--r--  1 pi pi 6148 Feb 11 14:17 .DS_Store

pi@raspberrypi:~/ $ ls -al source/.AppleDouble/
drwxr-xr-x 2 pi pi 4096 Feb 11 16:58 .
drwxr-xr-x 4 pi pi 4096 Feb 11 16:58 ..
-rw-rw-rw- 1 pi pi  741 Feb 11 16:58 app.d
-rw-rw-rw- 1 pi pi  741 Feb 11 16:58 .DS_Store
-rw-rw-rw- 1 pi pi  741 Feb 11 16:58 .Parent
```

when running "dub build", I get the following error for one of the project submodules:
`crate/source/crate/.AppleDouble/base.d: Error: module base odd length of UTF-16 char source 741`

If I create an empty hidden file that ends in .d, everything is fine, but once I have some data in that file, a compiler error is thrown.

### How to reproduce?

1. create a non empty hidden file in your project
2. run dub build
### Expected Behavior

- Hidden files and folders should be ignored by dub even if they are not empty. 

Or

- One should be able to exclude folders using wildcards. Going back to the OSX example, I'd like to be able to exclude "\*AppleDouble/\*.d" (I tried to set this in _dub.json_ using _excludedSourceFiles_ and it had no effect. maybe I don't know they right format...)

Samathy:
I can't see a reason why dub would need to detect files in hidden folders, so we should be okay to ignore them outright.

The modified file `packagerecipe.d` is using tabs, I have followed this convention.

### DIFF
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -247,8 +247,11 @@ struct BuildSettingsTemplate {
 					}
 
 					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
-						import std.path : baseName;
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						if (baseName(d.name)[0] == '.' || d.isDir ||
+							pathSplitter(d).canFind!(name => name[0] == '.'))
+							continue;
 						auto src = NativePath(d.name).relativeTo(base_path);
 						files ~= src.toNativeString();
 					}
--- a/test/issue1372-ignore-files-in-hidden-dirs.sh
+++ b/test/issue1372-ignore-files-in-hidden-dirs.sh
@@ -2,4 +2,15 @@
 
 cd ${CURR_DIR}/issue1372-ignore-files-in-hidden-dirs/
 
+echo "Compile and ignore hidden directories"
+rm dub.json
+ln -s dub_json_no_hidden.json dub.json
 ${DUB} build --force
+
+rm dub.json
+
+echo "Compile and explcitly include file in hidden directories"
+ln -s dub_json_hidden.json dub.json
+${DUB} build --force
+
+rm dub.json
--- a/test/issue1372-ignore-files-in-hidden-dirs.sh
+++ b/test/issue1372-ignore-files-in-hidden-dirs.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd ${CURR_DIR}/issue1372-ignore-files-in-hidden-dirs/
+
+${DUB} build --force
--- a/test/issue1372-ignore-files-in-hidden-dirs/dub.json
+++ b/test/issue1372-ignore-files-in-hidden-dirs/dub.json
@@ -1,9 +1,0 @@
-{
-	"authors": [
-		"--"
-	],
-	"copyright": "Copyright © 2019, --",
-	"description": "--",
-	"license": "--",
-	"name": "issue1372-ignore-files-in-hidden-dirs"
-}
--- a/test/issue1372-ignore-files-in-hidden-dirs/dub.json
+++ b/test/issue1372-ignore-files-in-hidden-dirs/dub.json
@@ -1,0 +1,9 @@
+{
+	"authors": [
+		"--"
+	],
+	"copyright": "Copyright © 2019, --",
+	"description": "--",
+	"license": "--",
+	"name": "issue1372-ignore-files-in-hidden-dirs"
+}
--- a/test/issue1372-ignore-files-in-hidden-dirs/dub_json_hidden.json
+++ b/test/issue1372-ignore-files-in-hidden-dirs/dub_json_hidden.json
@@ -1,0 +1,5 @@
+{
+	"name": "issue1372-ignore-files-in-hidden-dirs",
+    "sourceFiles":["source/.compileMe/hello.d"]
+    
+}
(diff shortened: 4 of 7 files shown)


Samathy:
SemaphoreCI is failing due to something to do with `gitcompatabledubpackage` :woman_shrugging: 

```
Fetching gitcompatibledubpackage 1.0.2 (getting selected version)...
Configuration 'library' of package gitcompatibledubpackage contains no source files. Please add {"targetType": "none"} to its package description to avoid building it.
Performing "debug" build using gdc for x86_64.
b ~master: building configuration "application"...
source/app.d:1:8: error: module file is in file 'gitcompatibledubpackage/subdir/file.d' which cannot be read
 import gitcompatibledubpackage.subdir.file;
        ^
import path[0] = /usr/lib/gcc/x86_64-linux-gnu/8/include/d
import path[1] = /home/runner/dub/test/issue990-download-optional-selected/source
gdc failed with exit code 1.
[ERROR] /home/runner/dub/test/issue990-download-optional-selected.sh:7 command failed
[ERROR] Script failure.

```

Samathy:
Hm, lots of things are complaining they have no source files. 
Sounds like my doing, investigations will progress.

andre2007:
Is it maybe related to how the gitcompatibledubpackage is structured 
https://github.com/dlang-community/gitcompatibledubpackage/blob/master/dub.sdl#L18

There is no source folder but the sources are contained in "."

wilzbach:
The problem is that the following is a valid path:

```
ls ./././
```

It returns the CWD ;-)
So for hidden directories you need to make sure that the folder name isn't `.` or `..`

See e.g. this related discussion: https://unix.stackexchange.com/questions/252165/how-to-show-only-hidden-directories-and-then-find-hidden-files-separately

source/dub/recipe/packagerecipe.d
```diff
@@ -247,8 +247,11 @@ struct BuildSettingsTemplate {
 					}
 
 					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
-						import std.path : baseName;
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						if (baseName(d.name)[0] == '.' || d.isDir ||
+							pathSplitter(d).canFind!(name => name[0] == '.'))
```
wilzbach: Maybe sth. like:

```suggestion
							pathSplitter(d).canFind!(name => name.startsWith('.') && name != ".." && name.length >= 2)
```

Though it probably is better to make this a separate function, s.t. we can add more tester easier.

Geod24:
Ping @Samathy , do you want to pursue this further ?

Samathy:
Thanks for the poke @Geod24 - I'll have a look at this over this weekend ( 04, 05th ) and try to get it finished, or report back otherwise.

But by all means, if someone else has motivation and time to take it over, then they should.

Apologies for not getting on with it, as it often does, other life aspects took priority ( apparently for a long time oops ).

schveiguy:
I just ran into this yesterday, I recently set up an AFP server so I can edit files from my Mac but build on the remote machine. I have an even more obscure bug, I run dub test and it says this:

```
Generating test runner configuration 'foo-test-library' for 'library' (library).
Invalid UTF-8 sequence (at index 1)
```

With no indication of where the error is. I tracked it down using gdb to it calling `readText` on one of these AppleDouble files, and the lack of stack trace or error handling is disturbing. It doesn't even print the file it's about to read.

schveiguy:
As I need this fixed soon, I'll try to make an effort to complete this.

Samathy:
@schveiguy Yeah, as much as I'd love to finish this, I rarely have enough energy. Please do go ahead.

schveiguy:
@Samathy thanks. Quick question, I see that you have in your test a an attempt to compile `source/.compileMe/hello.d`, but you didn't add such a source file to the PR. Is there a reason? I think I've fixed the concern above, and tested it in my project that has `.AppleDouble` files, and it seems to be working.

But I want to address the concerns above. Did you just forget to push those files? If so, that's fine, I'll add them. If not, then please explain what the test is trying to do.

schveiguy:
@Samathy nevermind, I think I'm good.

schveiguy:
#1916 is my update for this. You can close this now. Thanks @Samathy for the head start!

PR #1702, opened 2019-05-10
Use std.digest instead of the deprecated std.digest.digest.

jmdavis:


wilzbach:
Bumped the CI to 2.076.0 as this is the minimum requirement of GDC and still covers 10 versions.

PR #1703, opened 2019-05-11
Before opening a dub file, it checks if it is has contents

### LINKED ISSUE #1649: `dub convert -s -f json > dub.json` fails
On linux, `cd` to DUB repository and execute

```bash
dub convert -s -f json > dub.json
```

to get an empty json and the message

> rawRead must take a non-empty buffer

This is minor issue since to convert to a file we can just do

```bash
dub convert -f json
```

Zevenberge:
Fixes #1649

When enumerating possible dub files, it checks if the file has any contents.

### DIFF
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -131,9 +131,14 @@ class Package {
 	*/
 	static NativePath findPackageFile(NativePath directory)
 	{
+		static bool hasContents(NativePath path)
+		{
+			import std.file : getSize;
+			return getSize(path.toNativeString()) > 0;
+		}
 		foreach (file; packageInfoFiles) {
 			auto filename = directory ~ file.filename;
-			if (existsFile(filename)) return filename;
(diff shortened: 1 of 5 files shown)


andre2007:
I am unsure, an empty json file is invalid but is an empty dub.sdl invalid too?
As far as I know the name attribute isn't required anymore and will be defaulted if missing.

Zevenberge:
Fair point. Currently I get the same error as mentioned in the issue if I use Dub 1.14 on an empty dub.sdl file.

andre2007:
@wilzbach I remember a forum post of you, is the name attribute not required anymore in general or only for "single dub package file"?

wilzbach:
> @wilzbach I remember a forum post of you, is the name attribute not required anymore in general or only for "single dub package file"?

This is only for single dub package files where we can infer the project name to be the file name.

Geod24:
Closing as the issue was marked as WONTFIX.

Issue #996, opened 2016-11-17
Override dependency's dependency version


JesseKPhillips:
This is similar to #409 but for the desired package version rather than the library name.

It would be nice to be able to bump a dependency's version of a dependency, such that the dependency can still require a newer dependency.

Example:
https://github.com/forbjok/jsonserialized

This project wraps std_data_json with a simpler API. However std_data_json has been updated so newer compilers don't throw deprecations. Now the jsonserialized package needs its meta data updated and a new version tagged.

Instead I would like to say my project uses jsonserialized and wants std_data_json at version 1.18, later jsonserialized can require 1.19 which my project will use, but until then let me override this version.

https://github.com/forbjok/jsonserialized/pull/3
https://github.com/forbjok/jsonserialized/issues/2

s-ludwig:
You *can* override version selections in the root package by setting dub.selections.json accordingly (DUB will output a warning that the version choice conflicts with the dependency specification in jsonserialized). However, adding a way in the package *recipe* to override dependency choices seems odd to me, since overriding the version would just be a very special case.

The need to override things could also apply to any other recipe property or even to the source code of the dependency itself. And if the author doesn't fix the package in a timely manner, it's always possible to use a private fork.

I realize that this is less convenient than just putting something into the package recipe of the dependent package, but my gut feeling is that keeping a clean boundary like this is much healthier for the package eco system.

WebFreak001:
closing this as dub.selections.json is the current way how to do this for applications.

PR #1705, opened 2019-05-11
make selected nonmatching dependencies abort build, fix #1389

### LINKED ISSUE #1389: Dub should error on inconsistent dub.selection.json file
I highly recommend that Dub should produce an error if the `dub.selection.json` file is not in sync with the `dub.sdl`/`dub.json` file. Currently I think that `dub build` will automatically update the `dub.selection.json` file, causing change files in a Git repository. See this discussion for the confusion it creates [1].

[1] https://dlang.slack.com/archives/C59UPP30D/p1519047496000385

WebFreak001:
We had this as discussion at dconf. Changing the dependency version inside dub.json/sdl should fail the build if the fetched dependency is unmatching. This will break previous "hacks" where dub.selections.json was manually edited to override the version.

dub.json should be considered as what the developer wanted and therefore should not be overwritten with just a warning message, but instead abort the current build.

Some people said this would break their workflow because they used dub.selections.json to force different invalid versions for different configurations (because the dub.selections.json doesn't include different configuration dependencies) - I think this could be fixed with adding some dependency overrides inside the dub.json/dub.sdl like suggested in #996, until some better versioning approach for configurations is available in dub.

However for the normal use-case this would be an improvement because it will prevent outdated dependency issues in builds (which would just have a warning) so automated build tasks, scripts and maybe just any user compiling with a lot of output or ignoring the warnings don't assume everything was downloaded and compiled as specified in the dub.json file, but still get an executable which is silently malfunctioning in the worst case.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -681,7 +681,7 @@ abstract class PackageBuildCommand : Command {
 		auto validity = dub.project.validate();
 
 		if (!validity.matchingSelections) {
-			logInfo("-> Run \"dub upgrade\" to select specified package versions.");
+			logInfo("-> Run \"dub upgrade\" to select specified package versions in dub.selections.json");
 		}
 
 		enforce(!validity.hasFatalIssues, "Aborting due to fatal issues in project configuration.");
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -678,7 +678,13 @@ abstract class PackageBuildCommand : Command {
 			}
 		}
 
-		dub.project.validate();
+		auto validity = dub.project.validate();
+
+		if (!validity.matchingSelections) {
+			logInfo("-> Run \"dub upgrade\" to select specified package versions.");
+		}
+
+		enforce(!validity.hasFatalIssues, "Aborting due to fatal issues in project configuration.");
 
 		foreach (sc; m_overrideConfigs) {
 			auto idx = sc.indexOf('/');
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -230,10 +230,15 @@ class Project {
 
 		This will emit warnings to `stderr` if any discouraged names or
 		dependency patterns are found.
+
+		Returns: 
 	*/
-	void validate()
+	ValidationResult validate()
 	{
-		// some basic package lint
+		ValidationResult validity;
+
+		// some basic package name lint
+		validity.validName = true;
 		m_rootPackage.warnOnSpecialCompilerFlags();
 		string nameSuggestion() {
 			string ret;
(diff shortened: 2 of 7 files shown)


source/dub/project.d
```diff
@@ -315,6 +332,34 @@ class Project {
...
+		/// Returns: true if building should abort because of potential inconsistencies or issues in the build process.
+		bool hasFatalIssues() const
+		{
+			return !matchingSelections;
+		}
+
+		T opCast(T : bool)() const
+		{
+			return !hasFatalIssues;
+		}
```
wilzbach: Let's not be fancy here.

source/dub/commandline.d
```diff
@@ -678,7 +678,13 @@ abstract class PackageBuildCommand : Command {
...
 		}
 
-		dub.project.validate();
+		auto validity = dub.project.validate();
+
+		if (!validity.matchingSelections) {
+			logInfo("-> Run \"dub upgrade\" to select specified package versions.");
+		}
+
+		enforce(!validity.hasFatalIssues, "Aborting due to fatal issues in project configuration.");
```
wilzbach:
```suggestion
		enforce(!validity.hasFatalIssues, "Dub.selection.json validation failed.  project configuration.");
```

Though we probably can make the message even better.

source/dub/project.d
```diff
@@ -230,10 +230,15 @@ class Project {
...
+		Returns: 
 	*/
-	void validate()
+	ValidationResult validate()
 	{
-		// some basic package lint
+		ValidationResult validity;
+
+		// some basic package name lint
+		validity.validName = true;
```
wilzbach: Struct Initialization here with all true in the beginning here?

WebFreak001: I wanted the init state to be invalid, but I think I can make a fancy tupleof foreach here

source/dub/commandline.d
WebFreak001: the validity check isn't just dub.selections.json. For now it is the only reason how a build should be able to fail after validation, but it's designed so that the project validation can be any issue. I can add a reference to that file in the message above though.

WebFreak001:
@wilzbach the test already gets run (check the run unittest source and the output log)

John-Colvin:
Is this about the difference between the version in the dub.selections.json of the root package with the version spec in the dub.sdl of the root package?

Or is it about the difference between the version in the dub.selections.json of the root package and the inferred version spec from the whole tree?

WebFreak001:
this just changes the behavior where it would have previously told you to `dub upgrade`, so it affects the whole tree of packages when dub.selections.json isn't matching some dependency

FeepingCreature:
This PR synergizes with my submodule PR, which independently makes the same change. This is obviously the right behavior, and I am very much in favor of it.

WebFreak001:
I had forgotten about this PR. Now that I look over it again I still use my workflow with completely ignoring dub.selections.json and always running `dub upgrade` when changing dependencies or cloning something.

I would still be for merging this and I will rebase this later, but maybe we should also ask on the forums what the community would think of changing this.

s-ludwig:
I'd like to keep the old behavior as an opt-in command line switch, because otherwise this would be really disruptive for people who rely on the current relaxed behavior. But with regards to the default behavior, I agree it makes sense to handle this in a strict way.

Thinking about it, skipping the validation in case `--no-deps` is provided is probably already enough.

WebFreak001:
deciding against this now with the direction we are going. dub.selections.json is supposed to be made better instead of trying to work around it.

Issue #1474, opened 2018-05-17
examples/generated-source does not work. Regression in preGenerateCommands.


veelo:
### System information

- **dub version**: dub 1.9.0
- **OS Platform and distribution**: Windows 10, and Travis-CI (Linux)
- **compiler version** dmd-2.080.0

### Bug Description
`examples/generated-sources` does not work.

### How to reproduce?
```
git clone https://github.com/dlang/dub.git
cd dub\examples\generated-sources
dub build
```
Produces a link error. A second invocation of `dub` succeeds.

### Expected Behavior
Success upon initial invocation.

### Logs
```
.dub\build\application-debug-windows-x86-dmd_2080-33BFC7CE2FACCDF4772619001CE7CA74\generated-sources.obj(generated-sources)
 Error 42: Symbol Undefined __D4test3funFZi
Error: linker exited with status 1
```

Maybe add this to the `test` directory? (Make sure the generated file is deleted before the test is run.) This is a regression, see issue #144.

veelo:
For reference: forum post and failing use case https://forum.dlang.org/post/zhndrynwhtoryxiykxvy@forum.dlang.org

veelo:
If all sources are generated, the error is different:
`Configuration 'application' of package [...] contains no source files. Please add {"targetType": "none"} to its package description to avoid building it.
Main package must have a binary target type, not none. Cannot build.`
When this issue is fixed, the above case should probably also be tested.

wilzbach:
A PoC fix for this https://github.com/dlang/dub/pull/1708

andre2007:
@wilzbach my assumption is, the use case is wrong. The pre generate commands are executed "always". Generating source code using pre generate commands causing therefore a painful behavior as in the referenced issue from @MikeWey.
The correct use case is in my opinion to use the pre build commands for generating source code. Therefore the build settings needs also be refreshed here too.

veelo:
> @wilzbach my assumption is, the use case is wrong. The pre generate commands are executed "always". Generating source code using pre generate commands causing therefore a painful behavior as in the referenced issue from @MikeWey.
> The correct use case is in my opinion to use the pre build commands for generating source code. Therefore the build settings needs also be refreshed here too.

[...]

PR #1708, opened 2019-05-11
[PoC] Regenerate buildSettings after preGenerateCommands

wilzbach:
Fixes https://github.com/dlang/dub/issues/1474

This fixes the problem, but obviously we should be smarter 

TODO:
- [ ] add test case (or add the the examples to the CI)
- [x] check whether there actually was a preGenerateCommand run and only do the regeneration then
- [ ] try to find a better solution than this

FWIW it looks like even with the simple hello world example the buildSettings regeneration is already regenerated 4 times before.

### DIFF
--- a/examples/generated-sources/dub.json
+++ b/examples/generated-sources/dub.json
@@ -2,6 +2,6 @@
 	"name": "generated-sources",
 	"description": "Example of using pre generate commands to generate source code.",
 	"preGenerateCommands": [
-		"echo int fun() { return 42; } > source/test.d"
+		"echo 'int fun() { return 42; }' > source/test.d"
 	]
 }
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -100,6 +100,11 @@ class ProjectGenerator
 			targets[pack.name] = TargetInfo(pack, [pack], config, buildSettings);
 
 			prepareGeneration(pack, m_project, settings, buildSettings);
+			// Dub might have generated new files, so we need to regenerate at
+			// least buildSettings.sourceFiles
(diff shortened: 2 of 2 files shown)


examples/generated-sources/dub.json
```diff
@@ -2,6 +2,6 @@
 	"name": "generated-sources",
 	"description": "Example of using pre generate commands to generate source code.",
 	"preGenerateCommands": [
-		"echo int fun() { return 42; } > source/test.d"
+		"echo 'int fun() { return 42; }' > source/test.d"
```
andre2007: Why are the ticks added? I just tested it and the file `test.d` is created as
```d
'int fun() { return 42; }' 
```

andre2007:
@wilzbach 
I tested the PR and get a linker error the first time:

C:\Users\pc\git\dub\examples\generated-sources>dub build
Running pre-generate commands for generated-sources...
generated-sources ~master: building configuration "application"...
Linking...
lld-link: error: undefined symbol: _D4test3funFZi
>>> referenced by C:\Users\pc\git\dub\examples\generated-sources\source\app.d:6
>>>               .dub\build\application-debug-windows-x86_64-dmd_2089-F8E1F3627CAEA609227D49EB76DB892B\generated-sources.obj:(_Dmain)
Error: linker exited with status 1
dmd failed with exit code 1.

it works only the second time.

examples/generated-sources/dub.json
veelo: I can confirm that the ticks cause problems, at least on Windows.

Geod24:
Fixed in #2035

PR #1711, opened 2019-05-15
Build DUB with -preview=dip1000

wilzbach:
Not that much code here would be annotated with `@safe`, but it will check future additions/changes.

See also: https://github.com/dlang/phobos/pull/6931

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ jobs:
     - d: gdc
   include:
     - stage: test
-      d: dmd-2.086.0
-      env: [FRONTEND=2.086]
+      d: dmd-2.087.0
+      env: [FRONTEND=2.087, DFLAGS=-preview=dip1000]
     - d: dmd-2.081.1
       env: [FRONTEND=2.081]
     - d: dmd-2.080.1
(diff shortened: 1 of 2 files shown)


wilzbach:
Hmm, this won't work with all released DMD compilers yet.
So I guess I have to wait until the next release and then detect the version in `build.sh` :/

dub.sdl
```diff
@@ -4,6 +4,7 @@ authors "Sönke Ludwig" "Martin Nowak" "Matthias Dondorff" "Sebastian Wilzbach"
 	"more than 80 contributors total"
 copyright "Copyright © 2012-2016 rejectedsoftware e.K., Copyright © 2012-2014 Matthias Dondorff"
 license "MIT"
+dflags "-preview=dip1000"
```
ZombineDev: Can you move this change to the CI config for now where we know when we're using dmd-nightly?

s-ludwig: Seconded, this will otherwise also appear in projects that use DUB as a library.

wilzbach: Done.

ZombineDev:
It fails to compile on Travis CI with dmd v2.087.

Geod24:
Done in https://github.com/dlang/dub/pull/1791

PR #1721, opened 2019-06-04
Add a sponsorship button

wilzbach:
Analogous to dlang/dmd#9962


thewilsonator:
Semaphore failure is unrelated but required.

PR #1725, opened 2019-06-09
always run preBuildCommands

### LINKED ISSUE #1724: preBuildCommands should always run
I will use preBuildCommands to generate some source code.

Domain:
fix #1724


andre2007:
Could you please provide more info on your use case. From my point of view, with the info you gave, it looks like your problem will be solved by using ```preGenerateCommands``` instead of ```preBuildCommands```.

Domain:
Looks like I misunderstand how preGenerateCommands works.

Issue #1749, opened 2019-08-07
Compiling for WebAssembly uses host file extensions


skoppe:
### System information
- **dub version**: (e.g. dub 1.15.0)
- **OS Platform and distribution**: cross-platform
- **compiler version** (e.g. ldc2-1.16.0 2.086.1)

### Bug Description
Dub isn't aware of the webassembly target and uses host file extensions for static libraries and the final binary. On windows this means the static libraries are named `.lib`, while ldc expect wasm libraries to have the `.a` extensions, and the final binary is named `.exe`. On Linux the final binary has no `.wasm` extension.

### How to reproduce?
- install ldc2 >= 1.15.0
- ensure `ldc2 --version` shows `wasm32` target (I use `dlang2/ldc-ubuntu:1.16.0` docker image)
- start empty dub project
- replace app.d with this:
```
extern(C) void _start() {}
```
- and dub.sdl with this:
```
name "wasm-file-extension"
description "A minimal D application."
authors "Sebastiaan Koppe"
copyright "Copyright © 2019, Sebastiaan Koppe"
license "proprietary"
buildRequirements "allowWarnings"
dflags "-mtriple=wasm32-unknown-unknown-wasm" "-betterC" "-fvisibility=hidden"
configuration "library" {
targetType "library"
}
configuration "executable" {
targetType "executable"
}
```

- and compile with:

`dub build --build=release --config=executable`
`dub build --build=release --config=library`

### Expected Behavior

On windows the library config should result in an `wasm-file-extension.a` file. And the executable config should result in a `wasm-file-extension.wasm` file.

On linux the executable config should result in a `wasm-file-extension.wasm` file.

andre2007:
I think there are different ways to solve this. Either by a new build options "wasm" which would be a rather small fix, or by having a complete solution as described here
https://github.com/dlang/dub/issues/228

@wilzbach 
Maybe this could be tackled by AOC as part of the multi language support
(https://forum.dlang.org/thread/fhtfxkidkeaetdkquatn@forum.dlang.org).

skoppe:
I don't know about a new build option, I feel it doesn't fit the bill. I have to look at the code to really propose something, but a new entry for $ARCH seems to be in the right direction.

andre2007:
@skoppe please see this pull request regarding the triple for wasm
https://github.com/dlang/dub/pull/1541

skoppe:
[...]

Issue #1701, opened 2019-05-10
Profiling disabled for LDC2?


denizzzka:
### System information

- **dub version**: DUB version 1.15.0, built on May  4 2019
- **OS Platform and distribution**: Linux Debian
- **compiler version**: latest ldc2

### Bug Description
dub build -profile is not emits cli command for ldc2 to generate profiling code

### How to reproduce?
```
dub run --compiler=/home/denizzz/ldc2-1.16.0-beta1-linux-x86_64/bin/ldc2 --build=profile
```

### Expected Behavior
Some profiling data files generated after dub run exit

kinke:
Resolved by #1755.

Issue #1523, opened 2018-08-09
Cross compile wrong binary name and extension


rracariu:
### System information
- **dub version**: dub 1.8.1
- **OS Platform and distribution**: Linux
- **compiler version** : ldc 1.10.0

### Bug Description
When cross-compiling from a Linux host to a Windows target the library name is following the host naming convention. For example, a library project `foo` when compiled will generate a library named `libfoo.a`. This is wrong, because if you use that project as a dependency then linking will fail as the linker will be using the library that has a wrong extension. The error in this case is:
```
Error: unrecognized file extension a
```
### How to reproduce?
Make sure you have the VC and Platform libraries from a Visual Studio setup, then use 
```
DFLAGS="-mtriple=x86_64-pc-windows-msvc -link-internally" dub --compiler=ldc2
```
### Expected Behavior
A way to specify the target platform, similar to `--arch`, an option `--platform`, to force dub to use a specific platform. For example:
```
DFLAGS="-mtriple=x86_64-pc-windows-msvc -link-internally" dub --compiler=ldc2 --platform=windows
```

gedaiu:
I think dub should know how to add the compiler flags if you just provide the `platform` argument. This feature will not work with dmd...

kinke:
Resolved by #1755; the switch is `--arch`.

Issue #1431, opened 2018-03-30
How to pass flags to the compiler in separate linking step


JohanEngelen:
Issue discussed here:
https://forum.dlang.org/post/ctsitjonhqinnuyilgds@forum.dlang.org

When compilation and linking are separate:
- compilation step: `dflags` is passed to the compiler
- linking step: `lflags` is prefixed with `-L` and then passed to the compiler

How to pass a flag to the compiler during linking _without_ the `-L` prefix?
This is needed for example for `-flto=...` and `-fsanitize=...` with LDC.

(note that separate linking is not implemented for LDC (https://github.com/dlang/dub/issues/809), but it is possible when `ldmd2` is used as is done in the forum thread)

kinke:
Resolved by #1755 (for `ldc2`, not `ldmd2`): https://github.com/kinke/dub/blob/cfb281e289e7f72bfdd0f3e04407470d27cb8c4b/source/dub/compilers/ldc.d#L270-L288

PR #1755, opened 2019-08-17
Improve user experience with ldc2, incl. cross-compilation

kinke:


kinke:
@skoppe: Could you give this a try and see if it'd be a viable solution for #1749?

kinke:
Other related issues: #809, #1431, #1523, #1701.

source/dub/platform.d
```diff
@@ -42,6 +42,7 @@ enum string platformCheck = q{
 	version(Android) ret ~= "android";
 	version(Cygwin) ret ~= "cygwin";
 	version(MinGW) ret ~= "mingw";
+	version(WebAssembly) ret ~= "wasm";
```
skoppe: This is good for completeness sake, but it won't do much, since it only ever runs when compiling dub itself to WebAssembly.

skoppe:
> @skoppe: Could you give this a try and see if it'd be a viable solution for #1749?

So I just compile with dub `-arch=wasm32-unknown-unknown-wasm`?

source/dub/platform.d
kinke: Nope, this is also mixed into a platform probe file. The compiler compiles that one, using dub's `--arch` as LDC `-mtriple`, so that the BuildPlatform will feature a `wasm` platform.

kinke:
Yep, and remove the `-mtriple` option in the dub config files.

skoppe:
I can see that dub passes the triple to ldc, but it also looks the dflags in the dub.sdl are stripped, including the `betterC` I need to pass compilation.

kinke:
I don't see how previous dflags wouldn't be applied anymore (after the amended fix wrt. a `-betterC` dflag making it into link cmdlines too when linking separately). dub configs + resulting cmdlines would be helpful. [My attempts at building a spasm dummy project fail due to the `optional` package dependency not compiling for wasm (missing `wchar_t` alias etc).]

How is that supposed to work anyway? I mean, each (transitive) dependency of spasm would ideally be compiled with `-betterC -fvisibility=hidden`. Is there a way to enforce some base flags for the whole package tree?

skoppe:
> How is that supposed to work anyway? I mean, each (transitive) dependency of spasm would ideally be compiled with `-betterC -fvisibility=hidden` (`-mtriple` already is via `--arch`). Is there a way to enforce some base flags for the whole package tree?

No there is not. That is a short coming of dub. Properties are only propagated upwards.

With regards to optional, I made a pull request for it to be betterC compatible (and merged). That has worked in the past. But with this pr I get similar errors to you.

I looked at the compiler invocations and with the current dub, the optional package isn't actually build for wasm, no triple, no betterC.

With this pr the wasm triple gets propagates downwards and causes issues. Well, rightfully so. I made a pr for the optional package to fix its configuration see [pull#38](https://github.com/aliak00/optional/pull/38)

The reason it works is that there are only templates in the package.

kinke:
> That is a short coming of dub. Properties are only propagated upwards.

Oh wow, a serious shortcoming indeed. One could apparently work something out with the `DFLAGS` env variable, but I would have expected dub support for some global `--dflags`.

> But with this pr I get similar errors to you.

This is off-topic, but DMD only supports either `version (Windows)` or `version (Posix)`, while LDC predefines none of those 2 for Android, WebAssembly and unknown OS. Predefining `Posix` (currently disallowed in cmdline) for WebAssembly would get around these compile errors.

skoppe:
> This is off-topic, but DMD only supports either `version (Windows)` or `version (Posix)`, while LDC predefines none of those 2 for Android, WebAssembly and unknown OS. Predefining `Posix` (currently disallowed in cmdline) for WebAssembly would get around these compile errors.

also off-topic:

I have a wip [PR](https://github.com/dlang/druntime/pull/2738) in druntime where I define all those things for WebAssembly. Actually I took the WASI headers so it isnt strictly wasm. My reasoning is that is if you are using libc/druntime in wasm, you are actually either targeting WASI or emscripten. And if you don't, you simply dont call them and they will never end up in the binary.

Regardless, I don't think setting the Posix version for WebAssembly is the right thing to do. I remember trying it briefly just to get things compiling without betterC, but I abandoned that in favor of wasi.

skoppe:
> > @skoppe: Could you give this a try and see if it'd be a viable solution for #1749?
> 
> So I just compile with dub `-arch=wasm32-unknown-unknown-wasm`?

If I fix the above mentioned issues with optional (and mir-core as well), I can confirm that dub now sets the file extensions correctly.

Then I get this error:

`lld: error: unknown argument: --no-as-needed`

with

```
LDC - the LLVM D compiler (1.17.0-beta1):
  based on DMD v2.087.1 and LLVM 8.0.1
  built with LDC - the LLVM D compiler (1.17.0-beta1)
```

kinke:
> I have a wip PR in druntime where I define all those things for WebAssembly.

Ah nice, yeah that's a proper solution. I assume it cannot be tested for regressions though?

kinke:
> Then I get this error

Thx, should be fixed now.

skoppe:
> Oh wow, a serious shortcoming indeed. One could apparently work something out with the `DFLAGS` env variable, but I would have expected dub support for some global `--dflags`.

I have hit that issue before with the dip1000/dip25 compiler flag. The issue was that a dependency had a template which I used in a project where dip1000/dip25 was active. It caused mangling issues because the template called a regular function in the dependency, which had different function attributes inferred on it (IIRC it was the return attribute). The workaround was to use the `sourceLibrary` targetType.

The obvious solution would be to propagate all dips (or other future compiler feature flags) to all dependencies. But I don't think that will work, as it is perfectly valid for dependencies to not use various compiler feature flags. Although with something as determent as an arch, it seems more than reasonable to propagate it.

skoppe:
> Thx, should be fixed now.

Works!

kinke:
> Although with something as determent as an arch, it seems more than reasonable to propagate it.

Yep, I think there's a class of flags which make perfect sense as base flags, incl. `-betterC`, `-linker`, `-L`, `-Xcc`, `-fvisibility`, `-flto` (yes, that should suffice for LTO across all packages) etc. I'd leave it up to the user to choose sane ones.

kinke:
I didn't find any more problematic `version (Windows/Posix/linux/OSX)` occurrences, so I assume cross-compiling with `ldc2` and `dub --arch=...` to be fully working now, after setting up `ldc2.conf`. I haven't tested anything yet though. ;)

kinke:
Linux VM:

1) Download & extracted LDC v1.17-beta1.
2) Copy libs dir from windows-x64 package to `<LDC dir>/lib-win64` and append a section in `etc/ldc2.conf`. (see [Wiki](https://wiki.dlang.org/Cross-compiling_with_LDC))
[2.5) Build this dub branch and copy executable to LDC `bin` dir.]
3) `dub init cctest -t vibe.d && cd cctest`
4) `dub build --arch=x86_64-pc-windows-msvc`
5) `cp *.{exe,dll} <Windows dir>`
6) Run `cctest.exe` on Windows (requires Visual C++ runtime installation) and open the page.

Almost too smooth. :]

Geod24:
@kinke : Since you're on an LDC-fixing-spree, I'll shamelessly self plug: https://github.com/dlang/dub/issues/1758

kinke:
@Geod24: Should be fixed by the 1st commit.

Geod24:
Great! Could you add the test then ?

kinke:
A bit overkill IMO, but I wouldn't mind - if only I had an idea how to write suited tests. Is there a way to just emit the command lines without invoking anything and without even checking that the specified compiler can be found? Something like `dub build --compiler=ldc2 -v --dry-run`, so that one can test the issued command lines?

andre2007:
@kinke the dub argument for just printing what would be done is ```--annotate```.

Geod24:
Ping @wilzbach @thewilsonator , it would be great to have this in the next release.

MartinNowak:
A new feature right before the release candidate?
This change could use a test case and a changelog entry for the cross-compilation.

Geod24:
> A new feature right before the release candidate?

A lot of bug fixes, that only affect LDC, and were PRed by an LDC maintainer.
Basic things like coverage and profiling didn't even work with LDC and didn't give any feedback about not working.

jondegenhardt:
> Yep, I think there's a class of flags which make perfect sense as base flags, incl. `-betterC`, `-linker`, `-L`, `-Xcc`, `-fvisibility`, `-flto` (yes, that should suffice for LTO across all packages) etc. I'd leave it up to the user to choose sane ones.

@kinke - Building with LTO such that it included dependent packages would be very useful. Is this what you have in mind?

kinke:
Yes; should be doable today via `DFLAGS='-flto=thin' dub ...`.

skoppe:
> Yep, I think there's a class of flags which make perfect sense as base flags, incl. `-betterC`, `-linker`, `-L`, `-Xcc`, `-fvisibility`, `-flto` (yes, that should suffice for LTO across all packages) etc. I'd leave it up to the user to choose sane ones.

I would also include `-conf` to that list. I am currently porting druntime to wasm and my setup includes having a base package that defines a custom `ldc.conf`. It works well for dependent packages (they inherit the conf), but falls short when using dependencies.

jondegenhardt:
> Yes; should be doable today via `DFLAGS='-flto=thin' dub ...`.

Hm... I thought there was more to it. I'm partly thinking of having dependent packages also pull in phobos/druntime via `-defaultlib=phobos2-ldc-lto,druntime-ldc-lto`, but maybe that'd happen automatically as well.

I'll look into this further and see if it basically takes care of itself or there are other details to be worked through.

PR #1763, opened 2019-08-31
Replace `hash_t` and `sizediff_t` with `size_t` and `ptrdiff_t`

JinShil:
In support of https://github.com/dlang/druntime/pull/2777

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -378,7 +378,7 @@ struct Dependency {
 	}
 
 	/// ditto
-	hash_t toHash()
+	size_t toHash()
 	const nothrow @trusted  {
 		try {
 			size_t hash = 0;
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -40,7 +40,7 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		CONFIGS configs;
 		DependencyType depType = DependencyType.required;
 
-		hash_t toHash() const nothrow @trusted {
+		size_t toHash() const nothrow @trusted {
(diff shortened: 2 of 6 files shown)


joseph-wakeling-frequenz:
The motivating druntime change appears to have been closed without being merged.  Should this still be considered a valid change?

wilzbach:
Why not? Those are duplicate aliases. Even if they don't get deprecated right now, I still think that this was a positive change.

Geod24:
Yeah, I agree with @wilzbach . Those aliases have been "For backward compatibility" for almost a decade.

joseph-wakeling-frequenz:
Because those two names are conceptually different?  It's not intuitive at all why a `toHash` method should return a `size_t`.  Even if `hash_t` is the same in practice, there's a matter of documenting intent.

I don't object per se to the idea of dropping those older aliases, I just think that there should be a proper and well documented (and communicated!) process and rationale for doing so.  And it might be a good idea to actually get the deprecations merged into druntime before starting to spread changes through downstream codebases.

I came across this change somewhat at random when trying to track down other questions about recent changes in dub.  Thought process:

* oh, what, that's unexpected, is there some upcoming change in druntime I don't know about?

* wait, what, all these widely used aliases are being deprecated but there's no changelog, no rationale for the decision, not even a mention of why this is happening now?

* ... OK, so it didn't get merged, so why are changes happening downstream if that justification didn't get merged?

... all of which would have been easily avoided if somebody had bothered to just document the reasons and motivations properly.

Even if aliases have been "for backward compatibility" for a decade (is that documented anywhere on dlang.org?), that doesn't mean that they should be arbitrarily done away with without rationale.  Document the reasons for doing it NOW.

PR #1764, opened 2019-08-31
Single-file build.sh

kettes-leulhetsz:
This is needed for dub to self-host in environments with limited RAM, like Termux/Android. `commandline.d` requires 1 Gb alone. For comparison, the OOM killer kicks in at 1.5 Gb on my system, and takes the entire Termux+editor session with it.

I don't believe there's value in full optimization or debug info here, so I went with `-Os` for a reasonable balance with build time.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -31,7 +31,19 @@ fi
 MACOSX_DEPLOYMENT_TARGET=10.8
 
 echo Running $DMD...
-$DMD -ofbin/dub -g -O -w -version=DubUseCurl -version=DubApplication -Isource $* @build-files.txt
+
+declare -ax objects # array
+while read f; do
+	echo Compiling $f ...
+	object="bin/$f.o"
+	objects+=($object)
+	# compile one file
+	$DMD -of$object -c $f -Os -w -version=DubUseCurl -version=DubApplication -Isource $*
+done < build-files.txt
+
+# link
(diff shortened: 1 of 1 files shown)


John-Colvin:
Will it? Has anyone measured?

John-Colvin:
I agree that it should be a flag

kettes-leulhetsz:
I don't think people care about the build time of build.sh. It's a bootstrap script, ideally ran exactly once per install. For everything else, there's `dub build dub`.

If they did, we'd have at least a Makefile for parallel edit/compile/test cycles.

I'm looking at https://github.com/dcarp/cmake-d, it's mostly usable but doesn't seem to support `-version` flags yet. Is there interest in a full build system?

build.sh
```diff
@@ -31,7 +31,19 @@ fi
...
 echo Running $DMD...
-$DMD -ofbin/dub -g -O -w -version=DubUseCurl -version=DubApplication -Isource $* @build-files.txt
+
+declare -ax objects # array
+while read f; do
+	echo Compiling $f ...
+	object="bin/$f.o"
+	objects+=($object)
+	# compile one file
+	$DMD -of$object -c $f -Os -w -version=DubUseCurl -version=DubApplication -Isource $*
```
Vild: Changing `-O` to `-Os` will break compiling with dmd as it does not have that flag.

lesderid:
Have you tried with just `-lowmem`? It has the benefit that it can be passed through the DFLAGS environment variable, so it wouldn't need changes to the build file.

PR #1765, opened 2019-09-02
support dependencies as git url with exact commit

belka-ew:
Revival of [[WIP]  Support dependencies as git url (including with exact commit hash)](https://github.com/dlang/dub/pull/1403).
The PR introduces new version object type. Dependencies example:

```json
{
	"name": "dub-git",

	"dependencies": {
		"tanya": {
			"repository": "https://github.com/caraus-ecms/tanya",
			"version": "73535568b79a0b124bc1653002637a830ce0fcb8"
		}
	}
}
```

Also `~branch` as version works.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -34,7 +34,6 @@ struct PackageDependency {
 	Dependency spec;
 }
 
-
 /**
 	Represents a dependency specification.
 
@@ -56,6 +55,7 @@ struct Dependency {
 		NativePath m_path;
 		bool m_optional = false;
 		bool m_default = false;
+		Repository m_repository;
 	}
 
 	/// A Dependency, which matches every valid version.
@@ -93,10 +93,27 @@ struct Dependency {
 		m_path = path;
 	}
 
+	this(Repository repository, string spec) {
+		this.versionSpec = spec;
+		this.repository = repository;
+	}
+
 	/// If set, overrides any version based dependency selection.
 	@property void path(NativePath value) { m_path = value; }
 	/// ditto
 	@property NativePath path() const { return m_path; }
+
+	/// If set, overrides any version based dependency selection.
+	@property void repository(Repository value)
+	{
+		m_repository = value;
+	}
+
+	/// ditto
+	@property Repository repository() const
+	{
+		return m_repository;
+	}
 
 	/// Determines if the dependency is required or optional.
 	@property bool optional() const { return m_optional; }
@@ -110,6 +127,8 @@ struct Dependency {
 
 	/// Returns true $(I iff) the version range only matches a specific version.
 	@property bool isExactVersion() const { return m_versA == m_versB; }
+
+	@property bool isGit() const { return !repository.empty; }
 
 	/// Returns the exact version matched by the version range.
 	@property Version version_() const {
@@ -156,7 +175,7 @@ struct Dependency {
 			ves = ves[2..$];
 			m_versA = Version(expandVersion(ves));
 			m_versB = Version(bumpVersion(ves) ~ "-0");
-		} else if (ves[0] == Version.branchPrefix) {
+		} else if (ves[0] == Version.branchPrefix || ves.isHash) {
 			m_inclusiveA = true;
 			m_inclusiveB = true;
 			m_versA = m_versB = Version(ves);
@@ -207,7 +226,6 @@ struct Dependency {
 		string r;
 
 		if (this == invalid) return "invalid";
-
 		if (m_versA == m_versB && m_inclusiveA && m_inclusiveB) {
 			// Special "==" case
 			if (m_versA == Version.masterBranch) return "~master";
@@ -258,7 +276,12 @@ struct Dependency {
 	*/
 	string toString()()
 	const {
-		auto ret = versionSpec;
+		string ret;
+
+		if (!repository.empty) {
+			ret ~= "#";
+		}
+		ret ~= versionSpec;
 		if (optional) {
 			if (default_) ret ~= " (optional, default)";
 			else ret ~= " (optional)";
@@ -288,6 +311,7 @@ struct Dependency {
 			json = Json.emptyObject;
 			json["version"] = this.versionSpec;
 			if (!path.empty) json["path"] = path.toString();
+			if (!repository.empty) json["repository"] = repository.remote;
 			if (optional) json["optional"] = true;
 			if (default_) json["default"] = true;
 		}
@@ -316,6 +340,10 @@ struct Dependency {
 
 				dep = Dependency.any;
 				dep.path = NativePath(verspec["path"].get!string);
+			} else if (auto repository = "repository" in verspec) {
+				enforce("version" in verspec, "No version field specified!");
+				dep = Dependency(Repository(repository.get!string),
+						verspec["version"].get!string);
 			} else {
 				enforce("version" in verspec, "No version field specified!");
 				auto ver = verspec["version"].get!string;
@@ -397,6 +425,7 @@ struct Dependency {
 		A specification is valid if it can match at least one version.
 	*/
 	bool valid() const {
+		if (isGit) return true;
 		return m_versA <= m_versB && doCmp(m_inclusiveA && m_inclusiveB, m_versA, m_versB);
 	}
 
@@ -426,6 +455,7 @@ struct Dependency {
 	/// ditto
 	bool matches(ref const(Version) v) const {
 		if (this.matchesAny) return true;
+		if (this.isGit) return true;
 		//logDebug(" try match: %s with: %s", v, this);
 		// Master only matches master
 		if(m_versA.isBranch) {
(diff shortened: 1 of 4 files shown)


andre2007:
@belka-ew great that you are working in this topic. I am not really sure how the exact syntax could look like. Did you considered:
```
"dependencies": {
		"tanya": {
			"url": "git+https://github.com/caraus-ecms/tanya#73535568b79a0b124bc1653002637a830ce0fcb8"
		}
	}
```

(I am not really an expert on this topic, don't know wheter this git syntax makes sense, but it is somehow consistent to the maven syntax used in Dub)

lesderid:
Does this work with local git repository paths too?

andre2007:
@lesderid I am not sure wheter the question was meant for @belka-ew or for me. Yes, the protocol would be git+file (here some details from npm https://docs.npmjs.com/files/package.json#git-urls-as-dependencies)

lesderid:
> @lesderid I am not sure wheter the question was meant for @belka-ew or for me. Yes, the protocol would be git+file (here some details from npm https://docs.npmjs.com/files/package.json#git-urls-as-dependencies)

Sorry, was meant for @belka-ew

belka-ew:
> @belka-ew great that you are working in this topic. I am not really sure how the exact syntax could look like. Did you considered:
> 
> ```
> "dependencies": {
> 		"tanya": {
> 			"url": "git+https://github.com/caraus-ecms/tanya#73535568b79a0b124bc1653002637a830ce0fcb8"
> 		}
> 	}
> ```
> 
> (I am not really an expert on this topic, don't know wheter this git syntax makes sense, but it is somehow consistent to the maven syntax used in Dub)

This is doable, but I don't like the syntax. Dub already supports objects in version specification for paths `"mypackage": { "path": "..." }`, so why not reuse it. It is more readable and easier to remember than magic characters like `#`; different things have different entries (`repository` for the url and `version` for the git reference). 

npm keeps everything in one string and introduces a magic notation for different elements `#v1.0.27`, `#semver:^5.0` and so on.

belka-ew:
> Does this work with local git repository paths too?

Do you mean a bare git repository locally? The code just does `git clone`, so yes, whatever `git clone` accepts, should work here.

lesderid:
> Do you mean a bare git repository locally? The code just does `git clone`, so yes, whatever `git clone` accepts, should work here.

The code is using `URL` for the repository, so you'd have to use `file://`, which only works with absolute paths.

joseph-wakeling-frequenz:
@belka-ew thanks for reviving this idea.  A few thoughts and queries:

* (in response to others' feedback) let's not mess around with inventing magic syntaxes: git's own URL syntax for remotes should be matched exactly (with all possible options supported)

* (immediately messing around with syntax) if we go with `"repository" : "..."` are we going to run into issues if we want to add support for more repository types in the future (e.g. mercurial, bzr/brz, fossil, SVN, ...)?  Would it be better to have `"git" : "{URL}"` syntax?

  * note, there's not really a blocker here, as in future we could always add a `"vcs"` key to specify the type of repo (`"vcs" : "git"`, `"vcs" : "bzr"`, etc.) but it would be nice to think a little about this ahead of time

* care should be taken to test this with repo providers other than GitHub

  - ideally I'd suggest integration tests with at least: GitHub, GitLab, BitBucket, Launchpad (I'm sure others can suggest extra options)

* relative/local URLs like `../my-other-project/` should be supported

  - this may be important for access to resources in some CI systems, see e.g. GitLab's requirements for referring to URLs of private projects: https://docs.gitlab.com/ee/ci/git_submodules.html (they're talking about submodules but the same URL constraints may apply)

* IIUC version can currently be specified by either commit hash or branch, what about tag?

belka-ew:
I made the PR work with local paths as well. Now it is really the case, that everything `git clone` accepts should work, because I save the path in a string and leave git to decide what is a correct path/address and what is not.

@joseph-wakeling-frequenz, I’m not sure about the tag support. The tags just reference specific commits, so they can be replaced with the hash in 100% of cases. The problem aren’t the tags themselves, but I want really to reuse `“version”: “...”` to have more compact dependency definition and because it already has support for branches (probably the most common case for this kind of dependency specification). The version in dub is currently checked against some conditions: whether it is SemVer, a branch or hash. If I would support any git references (like tags), I have to accept any string as version and have to remove all these checks - and it would affect the current logic for normal dependencies.

belka-ew:
Can I get some official feedback here? What has to be done to get it merged?

belka-ew:
ping

belka-ew:
@thewilsonator, @andre2007, @lesderid, @joseph-wakeling-frequenz thanks.

wilzbach:
I don't have time to dive into this, but
- no changelog
- no tests
- no human reviews
- real code problems like that this should have been a PackageSupplier instead of patching itself through the entire codebase

@thewilsonator !!!. I'm now going to revert as we have a clear policy that we don't merge things that don't e.g. tests.

source/dub/dependency.d
```diff
@@ -93,11 +93,28 @@ struct Dependency {
 		m_path = path;
 	}
 
+	this(Repository repository, string spec) {
```
joseph-wakeling-frequenz: Missing docs for this constructor overload.

source/dub/dependency.d
```diff
@@ -111,6 +128,8 @@ struct Dependency {
 	/// Returns true $(I iff) the version range only matches a specific version.
 	@property bool isExactVersion() const { return m_versA == m_versB; }
 
+	@property bool isGit() const { return !repository.empty; }
```
joseph-wakeling-frequenz: Missing docs on this new property.

belka-ew:
It is not a PackageSuppliert, git repositories are single packages like paths. Things like `PackageSupplier.getVersions()` don't make sense for git.

source/dub/dependency.d
```diff
@@ -258,7 +276,12 @@ struct Dependency {
 	*/
 	string toString()()
 	const {
-		auto ret = versionSpec;
+		string ret;
+
+		if (!repository.empty) {
+			ret ~= "#";
+		}
+		ret ~= versionSpec;
```
joseph-wakeling-frequenz: Not an expert on all the options here, but this formulation looks questionable: is it legit for the git version spec to be followed by extra stuff (which will happen if `optional` is true, or if `!path.empty`)?

Also, why the `#` symbol?

source/dub/dependency.d
```diff
@@ -397,6 +425,7 @@ struct Dependency {
 		A specification is valid if it can match at least one version.
 	*/
 	bool valid() const {
+		if (isGit) return true;
```
joseph-wakeling-frequenz: I suggest being consistent in usage: `this.isGit` (as elsewhere), not just `isGit`

source/dub/dependency.d
```diff
@@ -706,7 +774,7 @@ struct Version {
 		Note that branches are always considered pre-release versions.
 	*/
 	@property bool isPreRelease() const {
-		if (isBranch) return true;
+		if (isBranch || isGit) return true;
```
joseph-wakeling-frequenz: Why should `isGit` necessarily imply a pre-release?  What if the git version spec points to a commit with a release tag?

source/dub/dependency.d
```diff
@@ -805,4 +881,20 @@ unittest {
 	assertThrown(a == b, "Failed: UNKNOWN == UNKNOWN");
 
 	assert(Version("1.0.0+a") == Version("1.0.0+b"));
+
+	assert(Version("73535568b79a0b124bc1653002637a830ce0fcb8").isGit);
+}
+
+private bool isHash(string hash) @nogc nothrow pure @safe
```
joseph-wakeling-frequenz: Missing function documentation

source/dub/project.d
```diff
@@ -1542,6 +1550,17 @@ final class SelectedVersions {
 		m_dirty = true;
 	}
 
+	void selectVersion(string package_id, Repository repository, string spec)
```
joseph-wakeling-frequenz: missing method documentation

source/dub/project.d
```diff
@@ -1604,16 +1623,23 @@ final class SelectedVersions {
 
 	static Json dependencyToJson(Dependency d)
```
joseph-wakeling-frequenz: missing method docs

wilzbach:
@belka-ew I hope I didn't discourage you too much from this. My motivation to revert was solely based on the fact that I think it was merged prematurely (no reviews, no tests). It was not a vote against this feature.

> Things like PackageSupplier.getVersions() don't make sense for git.

Why not? You can list all tags of a git repository, e.g.

```
git ls-remote --tags https://github.com/dlang/dub | sed -n 's|.*refs/tags/\(v\?[0-9]*\.[0-9]*\.[0-9]*$\)|\1|p' | sort --version-sort
```

belka-ew:
> Why not? You can list all tags of a git repository, e.g.

Tags aren't very useful since in the most cases they are versions that can be obtained from the registry. The most useful use case is probably branches, if I want to test a library before tagging it and without waiting for a registry update. But also specific commits are possible, and the amount of commits can be huge.

joseph-wakeling-frequenz:
> Tags aren't very useful since in the most cases they are versions that can be obtained from the registry.

I think you're missing a convenience use-case here.  If I'm specifying the version of a dependency to use, it's much nicer if I can use human semantic versioning, rather than Git's hash-based commit IDs.

DUB config files need to be human-readable first.

PR #1774, opened 2019-09-23
Command dub lint added

andre2007:
Dub supports now command `lint`, which will execute D-Scanner on the dub package.
By default `dub lint` will execute style check. Import paths to dependent dub
packages will be passed to D-Scanner.


andre2007:
@Hackerpilot Could you have a look here?

andre2007:
Ping:)

PR #1776, opened 2019-09-24
Use timeout value for low_speed_time

andre2007:
Maven repositories (Nexus) could be really slow. While connect timeout of 16 seconds works fine, the low speed value of 5 is too low and causing failing builds.

This PR proposes to use the parameter ```timeout``` for low_speed_time instead of the fixed value 5. 
(Per default timeout is 8 and for maven repository it is 16)


andre2007:
@thewilsonator thank you!

PR #1779, opened 2019-10-01
Handle unorthodox linker flags escaping for DMD

kinke:
People use stuff like this in their dub.json (ariovistus/pyd#120):
```
"lflags-windows-x86_64-dmd": [ "\\\"/LIBPATH:$USERPROFILE\\AppData\\Local\\Programs\\Python\\Python27\\libs\\\"" ]
```

This ugliness doesn't work for LDC and is only necessary because DMD's `-L` escaping rule is everything but intuitive.

Let dub handle this for the user (i.e., supporting simple `"/LIBPATH:C:\\My dir"`) while maintaining backwards compatibility for existing config files - **edit**: assuming they use a single flag per array item, i.e., nothing like `[ "-ldArg1 -ldArg2" ]`.

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -295,7 +295,27 @@ config    /etc/dmd.conf
 
 	string[] lflagsToDFlags(in string[] lflags) const
 	{
-        return map!(f => "-L"~f)(lflags.filter!(f => f != "")()).array();
+		static string translate(string lflag)
+		{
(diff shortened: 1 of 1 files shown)


Geod24:
This could use a changelog entry

kinke:
Oh, turns out this is actually a DMD bug.

kinke:
Fixed in upcoming DMD 2.089.

PR #1780, opened 2019-10-08
Add flag --submodules to add git submodules in the project folder as available packages.

FeepingCreature:
Also, adjust git version determination to support submodule folders with the -C flag.
(Submodules are folders that have distinct git state (tags etc.) but do not contain a `.git` folder, since they share the `.git` folder of the parent project.)

### DIFF
--- a/source/dub/git.d
+++ b/source/dub/git.d
@@ -22,7 +22,12 @@ import std.string;
 public void addGitSubmodules(PackageManager packageManager, NativePath rootPath) {
 	import std.process : execute;
 
-	const submoduleInfo = execute(["git", "--git-dir=" ~ (rootPath ~ ".git").toNativeString, "submodule", "status"]);
+	auto rootScmPath = rootPath ~ ".git";
+	const submoduleInfo = execute([
+		"git",
+		"-C", rootPath.toNativeString,
+		"--git-dir=" ~ (rootScmPath.relativeTo(rootPath)).toNativeString,
+		"submodule", "status"]);
 
 	enforce(submoduleInfo.status == 0,
 		format("git submodule status exited with error code %s: %s", submoduleInfo.status, submoduleInfo.output));
--- a/source/dub/git.d
+++ b/source/dub/git.d
@@ -1,0 +1,42 @@
+// functionality to supply packages from git submodules
+module dub.git;
+
+import dub.dependency;
+import dub.internal.vibecompat.core.file;
+import dub.package_;
+import dub.packagemanager;
+
+import std.algorithm;
+import std.ascii : newline;
+import std.exception : enforce;
+import std.range;
+import std.string;
+
+/** Adds the git submodules checked out in the root path as direct packages.
+	Package version is derived from the submodule's tag by `getOrLoadPackage`.
+
+	Params:
+		packageManager = Package manager to track the added packages.
+		rootPath = the root path of the git repository to check for submodules.
+ */
+public void addGitSubmodules(PackageManager packageManager, NativePath rootPath) {
+	import std.process : execute;
+
+	const submoduleInfo = execute(["git", "--git-dir=" ~ (rootPath ~ ".git").toNativeString, "submodule", "status"]);
+
+	enforce(submoduleInfo.status == 0,
+		format("git submodule status exited with error code %s: %s", submoduleInfo.status, submoduleInfo.output));
+
+	foreach (line; submoduleInfo.output.lines) {
+		const parts = line.split(" ").map!strip.filter!(a => !a.empty).array;
+		const subPath = rootPath ~ parts[1];
+		const packageFile = Package.findPackageFile(subPath);
+
+		if (packageFile != NativePath.init) {
+			const scmPath = rootPath ~ NativePath(".git/modules/" ~ parts[1]);
+			packageManager.getOrLoadPackage(subPath, packageFile, false, scmPath);
+		}
+	}
+}
+
+private alias lines = text => text.split(newline).map!strip.filter!(a => !a.empty);
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -75,20 +75,24 @@ class Package {
 			root = The directory in which the package resides (if any).
 			parent = Reference to the parent package, if the new package is a
 				sub package.
+			scm_path = The directory in which the VCS (Git) stores its state.
+				Different than root/.git for submodules.
 			version_override = Optional version to associate to the package
 				instead of the one declared in the package recipe, or the one
-				determined by invoking the VCS (GIT currently).
+				determined by invoking the VCS (Git currently).
 	*/
-	this(Json json_recipe, NativePath root = NativePath(), Package parent = null, string version_override = "")
+	this(Json json_recipe, NativePath root = NativePath(), Package parent = null,
+		 NativePath scm_path = NativePath(), string version_override = "")
 	{
 		import dub.recipe.json;
 
 		PackageRecipe recipe;
 		parseJson(recipe, json_recipe, parent ? parent.name : null);
-		this(recipe, root, parent, version_override);
+		this(recipe, root, parent, version_override, scm_path);
 	}
 	/// ditto
-	this(PackageRecipe recipe, NativePath root = NativePath(), Package parent = null, string version_override = "")
+	this(PackageRecipe recipe, NativePath root = NativePath(), Package parent = null,
+		 string version_override = "", NativePath scm_path = NativePath())
 	{
 		// save the original recipe
 		m_rawRecipe = recipe.clone;
@@ -98,15 +102,15 @@ class Package {
 
 		// try to run git to determine the version of the package if no explicit version was given
 		if (recipe.version_.length == 0 && !parent) {
-			try recipe.version_ = determineVersionFromSCM(root);
+			try recipe.version_ = determineVersionFromSCM(root, scm_path);
 			catch (Exception e) logDebug("Failed to determine version by SCM: %s", e.msg);
 
 			if (recipe.version_.length == 0) {
-				logDiagnostic("Note: Failed to determine version of package %s at %s. Assuming ~master.", recipe.name, this.path.toNativeString());
+				logDiagnostic("Note: Failed to determine version of package %s at %s. Assuming ~master.", recipe.name, root.toNativeString());
 				// TODO: Assume unknown version here?
 				// recipe.version_ = Version.unknown.toString();
 				recipe.version_ = Version.masterBranch.toString();
-			} else logDiagnostic("Determined package version using GIT: %s %s", recipe.name, recipe.version_);
+			} else logDiagnostic("Determined package version using Git: %s %s", recipe.name, recipe.version_);
 		}
 
 		m_parentPackage = parent;
@@ -146,11 +150,15 @@ class Package {
 				empty, the `root` directory will be searched for a recipe file.
 			parent = Reference to the parent package, if the new package is a
 				sub package.
+			scm_path = The directory in which the VCS (Git) stores its state.
+				Different than root/.git for submodules!
 			version_override = Optional version to associate to the package
 				instead of the one declared in the package recipe, or the one
-				determined by invoking the VCS (GIT currently).
+				determined by invoking the VCS (Git currently).
 	*/
-	static Package load(NativePath root, NativePath recipe_file = NativePath.init, Package parent = null, string version_override = "")
+	static Package load(NativePath root,
+		NativePath recipe_file = NativePath.init, Package parent = null,
+		string version_override = "", NativePath scm_path = NativePath.init)
 	{
 		import dub.recipe.io;
 
@@ -163,7 +171,7 @@ class Package {
 
 		auto recipe = readPackageRecipe(recipe_file, parent ? parent.name : null);
 
-		auto ret = new Package(recipe, root, parent, version_override);
+		auto ret = new Package(recipe, root, parent, version_override, scm_path);
 		ret.m_infoFile = recipe_file;
 		return ret;
 	}
@@ -742,21 +750,24 @@ class Package {
 	}
 }
 
-private string determineVersionFromSCM(NativePath path)
+private string determineVersionFromSCM(NativePath path, NativePath scm_path)
 {
+	if (scm_path.empty) {
+		scm_path = path ~ ".git";
+	}
 	// On Windows, which is slow at running external processes,
 	// cache the version numbers that are determined using
-	// GIT to speed up the initialization phase.
+	// Git to speed up the initialization phase.
 	version (Windows) {
 		import std.file : exists, readText;
 
-		// quickly determine head commit without invoking GIT
+		// quickly determine head commit without invoking Git
 		string head_commit;
-		auto hpath = (path ~ ".git/HEAD").toNativeString();
+		auto hpath = (scm_path ~ "HEAD").toNativeString();
 		if (exists(hpath)) {
 			auto head_ref = readText(hpath).strip();
 			if (head_ref.startsWith("ref: ")) {
-				auto rpath = (path ~ (".git/"~head_ref[5 .. $])).toNativeString();
+				auto rpath = (scm_path ~ head_ref[5 .. $]).toNativeString();
 				if (exists(rpath))
 					head_commit = readText(rpath).strip();
 			}
@@ -775,7 +786,7 @@ private string determineVersionFromSCM(NativePath path)
 	}
 
 	// if no cache file or the HEAD commit changed, perform full detection
-	auto ret = determineVersionWithGIT(path);
+	auto ret = determineVersionWithGit(path, scm_path);
 
 	version (Windows) {
 		// update version cache file
@@ -788,14 +799,13 @@ private string determineVersionFromSCM(NativePath path)
 	return ret;
 }
 
-// determines the version of a package that is stored in a GIT working copy
+// determines the version of a package that is stored in a Git working copy
 // by invoking the "git" executable
-private string determineVersionWithGIT(NativePath path)
+private string determineVersionWithGit(NativePath path, NativePath git_dir)
 {
 	import std.process;
 	import dub.semver;
 
-	auto git_dir = path ~ ".git";
 	if (!existsFile(git_dir) || !isDir(git_dir.toNativeString)) return null;
 	auto git_dir_param = "--git-dir=" ~ git_dir.toNativeString();
 
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -803,11 +803,15 @@ private string determineVersionFromSCM(NativePath path, NativePath scm_path)
 // by invoking the "git" executable
 private string determineVersionWithGit(NativePath path, NativePath git_dir)
 {
-	import std.process;
 	import dub.semver;
+	import std.process;
+	import std.typecons : tuple;
 
 	if (!existsFile(git_dir) || !isDir(git_dir.toNativeString)) return null;
-	auto git_dir_param = "--git-dir=" ~ git_dir.toNativeString();
+	auto git_dir_params = tuple(
+		"-C", path.toNativeString(),
+		"--git-dir=" ~ git_dir.relativeTo(path).toNativeString(),
+	).expand;
 
 	static string exec(scope string[] params...) {
 		auto ret = executeShell(escapeShellCommand(params));
@@ -816,7 +820,7 @@ private string determineVersionWithGit(NativePath path, NativePath git_dir)
 		return null;
 	}
 
-	auto tag = exec("git", git_dir_param, "describe", "--long", "--tags");
+	auto tag = exec("git", git_dir_params, "describe", "--long", "--tags");
 	if (tag !is null) {
 		auto parts = tag.split("-");
 		auto commit = parts[$-1];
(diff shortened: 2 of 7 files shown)


FeepingCreature:
Note: I am not sure how the change in git version detection plays out in practice. I fixed it for submodules, but I don't have a testcase for the original usecase that motivated it.

FeepingCreature:
Hm, CI doesn't seem to link in the new file..?

FeepingCreature:
Not sure how. Do I actually add a submodule in the test folder?

FeepingCreature:
I think the reason it's failing is it's now using the dub git version for the version of the test folders. Hm.

FeepingCreature:
Okay, now it uses the `--git-dir` flag again, but I've changed it so it can pass along an override folder for the git dir so it doesn't always assume it's `root/.git`. This is because with submodules, the git folder is `root/.git/modules/[submodule]/`.

FeepingCreature:
@wilzbach Okay, how do I test this? Only way I can think of is initialize a git repo, initialize a second git repo, make a commit, make a tag, add the second git repo as a submodule to the first, and see if dub --submodules picks it up. Does that sound good? It seems kind of overkill...

FeepingCreature:
Added a test that does just that.

Seems to pass..?

FeepingCreature:
Tests are green now.

rikkimax:
Is there any reason why this needs a flag that is manually run?
Can this not be done automatically on build if the file ``.gitmodules`` is found?

FeepingCreature:
It can absolutely be done automatically, but I didn't want to opt the users in to a new feature without requiring some affirmative action. My intention was to set `alias dub=dub --submodules` in our `.bashrc`.

FeepingCreature:
Ping @wilzbach

FeepingCreature:
Suggestions for a later PR:

Should dub consider all tags in the submodule as available versions?

Should dub upgrade change the submodule's checked out tag?

Geod24:
Is this really necessary ? From what I understand (the description is rather short), this will only affect the user's machine, which means anyone checking out the project will need to run this command, right ?

We are using submodules for our project, and the best way we could come up with was just to use path dependency, because users of `git submodules` use dub solely as a build tool, not for dependency management (although a mixed scheme is possible, I doubt it's desirable).

So what problem is this solving that path dependencies is not solving ?

FeepingCreature:
We specifically want to use dub as a dependency management tool. That is, we often have submodules that have recursive submodules that have dependencies that can clash with our own dependencies.

Geod24:
So do you mix DUB dependencies and submodules ?

> The alternative would be to give up submodules, and they're just too convenient.

I suggest the opposite: Go all in on the submodules.

This is what we do:
- [`dub.json`](https://github.com/bpfkorea/agora/blob/fbabf723fe8837bedde7af4f21eb52d935841e09/dub.json#L66-L76)
- [`dub.selections.json`](https://github.com/bpfkorea/agora/blob/fbabf723fe8837bedde7af4f21eb52d935841e09/dub.selections.json)

Once you work past the initial pain (that is, realize you need some dummy dependencies because `dub` tries to download everything, even unused stuff, e.g. https://github.com/bpfkorea/agora/blob/fbabf723fe8837bedde7af4f21eb52d935841e09/submodules/botan/dub.json), it's fairy land and I never had a dependency problem since I started using it.

FeepingCreature:
> I suggest the opposite: Go all in on the submodules.

The problem is recursive submodules.

Ie. A -> X1, A -> B -> X1, and enforcing that X1 and X2 are the same version.

> So do you mix DUB dependencies and submodules ?

Right now, yes. Of course, right now we build with an internal buildsystem that uses dub to build each dependency individually, so there's no version enforcement at all.

Geod24:
> The problem is recursive submodules.

So do you do recursive checkout of submodules ? What we do (and something we took from Sociomantic) is that we never, ever do recursive submodules. If App depends on Lib2 and Lib1, and Lib2 depends on Lib1, we just put them in `submodules` and let our build system use this.

> Ie. A -> X1, A -> B -> X1, and enforcing that X1 and X2 are semver compatible.

That doesn't make too much sense for me: DUB doesn't have to *select* a version, since the version are already selected (and fixed). At least when it comes to submodules.

> Right now, yes.

IMO that's your real problem. If you use both systems, they have to be aware of each other, which is a pain, and will never be perfect.

FeepingCreature:
Ping!

Geod24:
> This is why we want to transition to dub only - and hence, this PR.

Right, but as I said, I don't understand why this switch. If `dub` learns to recognize git submodules' version, it should be enough for the use case you describe, without needing to add them as packages (instead using path dependency for this).

FeepingCreature:
Git submodules added as paths still need the ability to specify a range of versions, ie. combining `path=` and `version=`. But then I think this would imply that the dependency at this path has the version X, rather than that the dependency at this path *should* have the version X.

That aside, adding version requirements for path dependencies seems like it would do the same thing.

FeepingCreature:
ping

thewilsonator:
Sorry, no can do at the moment. Github broke the UI and I'm travelling and so can really upgrade browsers.

cc @ZombineDev ?

FeepingCreature:
Thanks for looking anyway.

source/dub/package_.d
```diff
@@ -75,20 +75,24 @@ class Package {
 			root = The directory in which the package resides (if any).
 			parent = Reference to the parent package, if the new package is a
 				sub package.
+			scm_path = The directory in which the VCS (Git) stores its state.
+				Different than root/.git for submodules!
```
wilzbach:
```suggestion
				Different than root/.git for submodules.
```

source/dub/git.d
```diff
@@ -0,0 +1,40 @@
...
+		const packageFile = Package.findPackageFile(subPath);
+
+		if (!packageFile.empty) {
+			const scmPath = rootPath ~ NativePath(".git/modules/" ~ parts[1]);
+			packageManager.getOrLoadPackage(subPath, packageFile, false, scmPath);
+		}
+	}
+}
+
+private alias lines = text => text.split("\n").map!strip.filter!(a => !a.empty);
```
wilzbach: Can we at least split on https://dlang.org/phobos/std_ascii.html#newline?

source/dub/git.d
```diff
@@ -0,0 +1,40 @@
...
+import dub.internal.vibecompat.core.file;
+import dub.package_;
+import dub.packagemanager;
+
+import std.algorithm;
+import std.exception : enforce;
+import std.range;
+import std.string;
+
+/// adds the git submodules checked out in the root path as direct packages
```
wilzbach: Full Ddoc + sentences start with a capital letter.

source/dub/git.d
```diff
@@ -0,0 +1,40 @@
...
+		"-C", rootPath.toNativeString,
+		"--git-dir=" ~ (rootScmPath.relativeTo(rootPath)).toNativeString,
+		"submodule", "status"]);
+
+	enforce(submoduleInfo.status == 0,
+		format("git submodule status exited with error code %s: %s", submoduleInfo.status, submoduleInfo.output));
+
+	foreach (line; submoduleInfo.output.lines) {
+		const parts = line.split(" ").map!strip.filter!(a => !a.empty).array;
+		const subPath = rootPath ~ parts[1];
```
wilzbach: Can't we do sth. less fragile?

source/dub/git.d
```diff
@@ -0,0 +1,40 @@
...
+		const packageFile = Package.findPackageFile(subPath);
+
+		if (!packageFile.empty) {
```
wilzbach:
```suggestion
		if (packageFile != NativePath.init)
```

wilzbach:
> Since @wilzbach doesn't seem to want anything to do with this PR...

Sorry, I have other priorities in life at the moment. I only wanted to avoid that sth. without tests get merged. 

> Right, but as I said, I don't understand why this switch. If dub learns to recognize git submodules' version, it should be enough for the use case you describe, without needing to add them as packages (instead using path dependency for this).

Yeah, `dub` DOES NOT download submodules. This is due to the way dependencies are currently checked out as zip files from GitHub and we should definitely fix this, but it also implies that this "submodules" can't be used for anything published on the dub registry. Hence, this feature can only be used when you do a `git clone` of your package yourself. However, in this case path-based dependencies work as well.

Additionally, path-based dependencies have a few more advantages:
- no overhead for detection required
- selective control (e.g. for the dlang-tour we use quite a lot of submodules, but we don't want to build any of them even though they are buildable)

> That aside, adding version requirements for path dependencies seems like it would do the same thing. I'm not sure if it'd even be any different implementation-wise. Especially since to recognize the path of a git dependency, we'd also need to read the list of git submodules. 

Why doesn't `version="*"` work for you? If not, have you considered allowing path-based dependencies without `version`?

FeepingCreature:
> Why doesn't version="*" work for you?

Because we want the version checking logic to apply when we are using the repository for a direct build as well as when we're using it as a dependency in another repository, which happens very often.

source/dub/git.d
FeepingCreature: I'm also not enthused about it, but I'm not sure how to improve on it.

source/dub/git.d
FeepingCreature: done

source/dub/git.d
FeepingCreature: done

source/dub/git.d
FeepingCreature: done

source/dub/package_.d
FeepingCreature: done

FeepingCreature:
@wilzbach I know you have other priorities, but then could you please resolve your change request for tests now that there's a test?

FeepingCreature:
Ping.

FeepingCreature:
~ p i n g ~

Can somebody please give me a pass/fail/change request on this? This PR is holding up work.

FeepingCreature:
@thewilsonator Thank you!

wilzbach:
@thewilsonator: there was clearly NO consensus on merging this flag with multiple contributors being unsure about the usefulness of this flag. What was your motivation for merging this?

FeepingCreature:
There was clearly no functioning mechanism for arriving at a consensus either.

wilzbach:
> To be quite fair, there was clearly no functioning mechanism for arriving at a consensus either.

Yes, but that doesn't NOT mean that Nicholas can just arbitrarily merge things. Especially without providing any motivation or reasoning!

> This PR was open for two months. Over this time, the following pattern emerged: somebody comes by and leaves unsure feedback, I answer the feedback, and the person never shows up again.

Yes, but it doesn't mean that this PR can be merged without motivation / reasoning.

> What exactly was I supposed to be doing here?

You should have pinged someone with actual authority and time to approve things -> @atilaneves. 

> Can somebody please give me a pass/fail/change request on this? 

My [review](https://github.com/dlang/dub/pull/1780#issuecomment-544825217) was a FAIL (!)

1) This feature is specially tailored for your internal use-case. It can't be used by any dub package on the dub registry (as dub packages can't have submodules)  and thus adds more baggage to dub that needs to be supported.
2) It should be possible with `path`-based dependencies.

> If Dep was added as a path-based dependency in Foo, we would have trouble getting it to use the Dep in Bar.

How so?
What you're trying to achieve can be handled with the following:

```
dependency "dep1" version="0.10.0" path="./dep1"
dependency "dep2" version=">=0.5.0" path="./dep2"
```

Which uses an existing feature and is more general-purpose as it doesn't solely apply to git submodules.

FeepingCreature:
> My review was a FAIL (!)

No it was not! Your review was a request for a specific enhancement, ie. a test. The test was added.

> It should be possible with path-based dependencies.

I think that mixes two things: *accepted* versions and *provided* versions. Path-based dependencies will break once the package is used as a submodule in another package.

wilzbach:
> Path-based dependencies will break once the package is used as a submodule in another package.

Why?

FeepingCreature:
> Why?

I mean, I guess you could ignore the path? Not sure if dub actually does this right now? The point is that if you have two submodules that both use the same transitive dependency, you need some way to advertise this dependency in the root package that uses both of them.

wilzbach:
> To be quite fair, there was clearly no functioning mechanism for arriving at a consensus either.

BTW I do understand the problem, but there's nothing I can do about (as you have noticed my time is very limited). Maybe a NG discussion is a good start?

> The point is that if you have two submodules that both use the same transitive dependency, you need some way to advertise this dependency in the root package that uses both of them.

Are we talking about this?

```
- main package
+ submodules
  - dep1 (depends on dep3)
  - dep2 (depends on dep3)
  - dep3
```

1) Vibe.d solves this problem neatly with subpackages which can depend on each other

https://github.com/vibe-d/vibe.d/blob/master/web/dub.sdl

2) Path-based dependencies declared by the main package could overwrite the paths for the same respective package dependent upon in e.g. dep{1,2}. I understand that dub doesn't do this correctly at the moment, but it's what I understood this PR was trying to achieve.

I do think that the mindset of "this feature could work, but it's broken for this specific use-case. Hence, let's add a new feature" doesn't not scale long-term

FeepingCreature:
> Are we talking about this?

Yes, exactly.

> Vibe.d solves this problem neatly with subpackages which can depend on each other

I don't believe that subpackages can have their own versions.

> Path-based dependencies declared by the main package could overwrite the paths for the same respective package dependent upon in e.g. dep{1,2}. I understand that dub doesn't do this correctly at the moment, but it's what I understood this PR was trying to achieve.

I mean, I could implement that, but at this point I'm expecting that some people give feedback, the PR will lie dead for months of insistent pings, then somebody click merge and you'll revert it. My only hope is that you'll be more reluctant to revert a PR that you like.

FeepingCreature:
Update: I remember what the problem with `path=` was! Submodules don't work like that; the version cannot be determined from the path of the submodule alone, you'd need to look back and find the SCM root path which is in `.git/modules`. So it'd be "this PR but worse"; you'd need to detect that the path points *into* a submodule then walk parent folders until you can find the main `.git` dir.

FeepingCreature:
## Action plan for submodule support

Going by feedback, the path most likely to have success is augmenting the `path=` override on dependencies to support determining submodule versions to use them for the dependency, and handle cases where `path=` itself appears in dependencies correctly if those paths are not available, ie. because they're not checked out. As such, there's several steps that will be separate PRs:

1. Augment `determineVersionFromSCM` to support determining versions of `path=` dependencies that are submodules, by reading the `.git` file that git creates as a pointer at the actual git scm folder.
2. Override path-based dependencies if the root package defines path-based dependencies for the same path. This should allow dub to tolerate cases where a transitive dependency defines a `path=` override that points at a folder that doesn't exist, as is the case in nontransitive submodule checkouts.

FeepingCreature:
Hang on, `path` doesn't work like that.

FeepingCreature:
## Action plan for submodule support

Going by feedback, the path most likely to have success is augmenting the `path=` override on dependencies to support determining submodule versions to use them for the dependency, and handle cases where `path=` itself appears in dependencies correctly if those paths are not available, ie. because they're not checked out. As such, there's several steps that will be separate PRs:

1. Change `path` behavior to not ignore `version` but determine an actual version from inspecting the specified path, or else defaulting to `master`.
2. Augment `determineVersionFromSCM` to support determining versions of `path=` dependencies that are submodules, by reading the `.git` file that git creates as a pointer at the actual git scm folder.
3. Override path-based dependencies if the root package defines path-based dependencies for the same path. This should allow dub to tolerate cases where a transitive dependency defines a `path=` override that points at a folder that doesn't exist, as is the case in nontransitive submodule checkouts.

@wilzbach Does this seem likely to have success?

atilaneves:
What's deficient with adding the submodule as a path package in the package description?

FeepingCreature:
`path=` is not compatible with `version=` _or_ submodule versions.

source/dub/git.d
```diff
@@ -0,0 +1,47 @@
...
+		rootPath = the root path of the git repository to check for submodules.
+ */
+public void addGitSubmodules(PackageManager packageManager, NativePath rootPath) {
+	import std.process : execute;
+
+	auto rootScmPath = rootPath ~ ".git";
+	const submoduleInfo = execute([
+		"git",
+		"-C", rootPath.toNativeString,
+		"--git-dir=" ~ (rootScmPath.relativeTo(rootPath)).toNativeString,
```
CyberShadow: Why is `--git-dir` necessary here? It was necessary in conjunction with `--work-tree` before `-C` was introduced.

source/dub/git.d
CyberShadow: Bash code to enumerate submodules:

```bash
	git config -lz | \
		while IFS=$'\n' read -r -d '' name value
		do
			if [[ "$name" =~ ^submodule\.(.*)\.url$ || ( "$name" =~ ^submodule\.(.*)\.active$ && "$value" == true ) ]]
			then
				printf '%s\0' "${BASH_REMATCH[1]}"
			fi
		done | \
			sort -uz | \
			mapfile -d '' -t list
```

See https://git-scm.com/docs/gitsubmodules#_active_submodules for reference.

To get the submodule git dir:

```bash
git rev-parse --git-path modules/"$m"
```

where `$m` is the submodule name as per above list.

To get the submodule work dir:

```bash
git -C "$md" rev-parse --show-toplevel
```

where `md` is the git dir above.

source/dub/package_.d
```diff
@@ -75,20 +75,24 @@ class Package {
 			root = The directory in which the package resides (if any).
 			parent = Reference to the parent package, if the new package is a
 				sub package.
+			scm_path = The directory in which the VCS (Git) stores its state.
+				Different than root/.git for submodules.
```
CyberShadow: This seems unnecessary

source/dub/package_.d
```diff
@@ -75,20 +75,24 @@ class Package {
...
 
 		PackageRecipe recipe;
 		parseJson(recipe, json_recipe, parent ? parent.name : null);
-		this(recipe, root, parent, version_override);
+		this(recipe, root, parent, version_override, scm_path);
 	}
 	/// ditto
-	this(PackageRecipe recipe, NativePath root = NativePath(), Package parent = null, string version_override = "")
+	this(PackageRecipe recipe, NativePath root = NativePath(), Package parent = null,
+		 string version_override = "", NativePath scm_path = NativePath())
```
CyberShadow: Ditto

source/dub/package_.d
```diff
@@ -98,15 +102,15 @@ class Package {
 
 		// try to run git to determine the version of the package if no explicit version was given
 		if (recipe.version_.length == 0 && !parent) {
-			try recipe.version_ = determineVersionFromSCM(root);
+			try recipe.version_ = determineVersionFromSCM(root, scm_path);
```
CyberShadow: Ditto

source/dub/package_.d
```diff
@@ -146,11 +150,15 @@ class Package {
 				empty, the `root` directory will be searched for a recipe file.
 			parent = Reference to the parent package, if the new package is a
 				sub package.
+			scm_path = The directory in which the VCS (Git) stores its state.
+				Different than root/.git for submodules!
```
CyberShadow: Ditto

source/dub/package_.d
```diff
@@ -146,11 +150,15 @@ class Package {
...
 			version_override = Optional version to associate to the package
 				instead of the one declared in the package recipe, or the one
-				determined by invoking the VCS (GIT currently).
+				determined by invoking the VCS (Git currently).
 	*/
-	static Package load(NativePath root, NativePath recipe_file = NativePath.init, Package parent = null, string version_override = "")
+	static Package load(NativePath root,
+		NativePath recipe_file = NativePath.init, Package parent = null,
+		string version_override = "", NativePath scm_path = NativePath.init)
```
CyberShadow: Ditto

source/dub/package_.d
```diff
@@ -163,7 +171,7 @@ class Package {
 
 		auto recipe = readPackageRecipe(recipe_file, parent ? parent.name : null);
 
-		auto ret = new Package(recipe, root, parent, version_override);
+		auto ret = new Package(recipe, root, parent, version_override, scm_path);
```
CyberShadow: Ditto

source/dub/package_.d
```diff
@@ -742,21 +750,24 @@ class Package {
 	}
 }
 
-private string determineVersionFromSCM(NativePath path)
+private string determineVersionFromSCM(NativePath path, NativePath scm_path)
 {
+	if (scm_path.empty) {
+		scm_path = path ~ ".git";
+	}
```
CyberShadow: Don't do this

source/dub/package_.d
```diff
@@ -742,21 +750,24 @@ class Package {
...
-	// GIT to speed up the initialization phase.
+	// Git to speed up the initialization phase.
 	version (Windows) {
 		import std.file : exists, readText;
 
-		// quickly determine head commit without invoking GIT
+		// quickly determine head commit without invoking Git
 		string head_commit;
-		auto hpath = (path ~ ".git/HEAD").toNativeString();
+		auto hpath = (scm_path ~ "HEAD").toNativeString();
```
CyberShadow: This was bad code and this is not an improvement :)

source/dub/git.d
CyberShadow: For this purpose probably we want to look at unregistered submodules too (i.e. submodule entries in the tree).

In which case, the equivalent of:

```bash
git ls-tree -r --full-tree HEAD -z | grep -z '^160000 ' | cut -z -d $'\t' -f 2 | xargs -0 printf '%s\n'
```
would be more appropriate.

Then, just check for a `dub.sdl` at each entry.

PR #1783, opened 2019-10-10
[refactor] Add struct representing semantic versions, replacing string.

FeepingCreature:
Regex is taken from https://semver.org/ .

edit: Note that this is prep work for another PR that will introduce a new version operator, `~1.2.3`, to represent "semantic version compatible" (ie. identical version if `0.x.y`, up to next major if `x.y.z`), and can I just say how weird it is that Dub supports every conceivable version operation *except* "compatible semver"?

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -668,7 +668,16 @@ struct Version {
 		static immutable UNKNOWN_VERS = "unknown";
 		static immutable masterString = "~master";
 		enum branchPrefix = '~';
-		string m_version;
+		enum Type {
+			Branch,
+			SemVer,
+			Unknown
+		}
+		Type type;
+		union {
+			SemVer semver_;
+			string branch_;
+		}
 	}
 
 	static immutable Version minRelease = Version("0.0.0");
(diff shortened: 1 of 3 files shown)


wilzbach:
> I think dub has bigger things slowing it down, like the DirEntry implementation.

Sure, but this isn't an argument to push a performance regression.

FeepingCreature:
I'll go see if I can get it faster.

FeepingCreature:
It did not help. Looks like `std.conv.parse` is a nonstarter. Closing.

PetarKirov:
@FeepingCreature Modulo the potential problems with std.regex, I liked very much the direction you were going with this pull request.

FeepingCreature:
Yeah I would have preferred it for the clarity of code alone, but it isn't worth it if it leads to a slowdown, since dub can't afford the speed loss.

PetarKirov:
@FeepingCreature what did you measure? Was it a micro benchmark (measuring only the parsing and comparison of versions time) or running the whole dub application? Did you try using LDC with optimizations?
If you got 1% perf degradation but that was running the whole app, I don't think we should worry about that, as dub in mainly IO/waiting for other processes bound.

rikkimax:
Agreed, those performance numbers are massive.

I just ran a quick benchmark via run.dlang.io, even with writeln'ing I had to run the 5 tests 100 times to get in the ms range.

I say reopen and use it, its just noise in running an executable.

```d
import std.regex;
import std.stdio;
import std.conv;
import std.datetime.stopwatch;

enum semVerRegex = `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)` ~
		`(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))` ~
		`?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`;

void main() {
    string[] versions = ["0.0.0", "1.2.3", "4.5.6-data1+data2", "7.8.9+data2", "9.10.11-data1"];
    
    StopWatch sw;
    sw.start;
    
    foreach(_; 0 .. 100) {
        foreach(version_; versions) {
            auto match = matchFirst(version_, ctRegex!semVerRegex);

            writeln(match["major"].to!int,
                    match["minor"].to!int,
                    match["patch"].to!int,
                    match["prerelease"],
                    match["buildmetadata"]);
        }
    }
    
    sw.stop;
    writeln(sw.peek, " \t <=> \t", sw.peek / versions.length);
}
```

PR #1784, opened 2019-10-10
Add semver compatibility operator "^x.y.z".

FeepingCreature:
Matches all versions that are compatible with the public API x.y.z.

This is basically equivalent to `>=x.y.z <(x+1).y.z`, but since it's arguably the defining operator of semantic versioning it deserves its own operator.

Unittests are added.

@wilzbach I threw in a 10ms speed improvement for you too. :)

### DIFF
--- a/changelog/semver-compatibility-operator.dd
+++ b/changelog/semver-compatibility-operator.dd
@@ -1,0 +1,8 @@
+Added SemVer compatibility operator "^"
+
+Dub now supports version specifications of the form `^x.y.z`.
+This corresponds to a "semver compatible version", ie. any version up from
+`x.y.z` with the same major number. If the major number is 0, only the same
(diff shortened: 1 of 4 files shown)


s-ludwig:
It's definitely good to have this operator, but the ambiguity with branch specifications is a bit worrying. Especially a branch named "2.0" is supposedly not uncommon.

Also, is there a precedence for the proposed behavior? NPM apparently defines `~1.2.3` as `>=1.2.3 <1.3.0-0` instead of `>=1.2.3 <2.0.0-0`, it has `^1.2.3` for that. Following that route, we'd just have to add the `^` syntax with a full version following it.

FeepingCreature:
Thanks! I didn't know about the `^` syntax. I'll switch to that.

FeepingCreature:
Split them out.

PR #1789, opened 2019-10-25
Fix incomplete string import overrides

### LINKED ISSUE #1788: String import paths not properly pushed to dependencies
This is a regression introduced in 1.7.0 (#1263) and affects [vibed.org](https://github.com/vibe-d/vibed.org/) in particular.

### System information
- dub 1.17.0

### Bug Description
Not all string import paths that might be required by a string import file that overrides a corresponding file in a dependency are added to the dependency's build settings.

### How to reproduce?

Given three packages, **a**, **b**, **c**, where **a** depends on **b** and **c**, the following string import files exist:
- a: layout.diet (references fancylayout.diet)
- b: layout.diet
- c: fancylayout.diet

**a**'s layout.diet correctly overrides the one from **b**. However, if **a**'s layout.diet contains a reference to fancylayout.diet, the build fails, because during the build of **b** (which uses **a**'s layout.dt), the string import paths of **c** are not added to the compiler command line.

### Expected Behavior

Package **b** should have package **c**'s string import paths added on its compiler command line.

s-ludwig:
Fixes #1788.


PetarKirov:
Azure fails with:

> Error: out of memory dmd failed with exit code 1.

Most likely switching from a Win32 built dmd to one built in `-m64` mode will fix this.

source/dub/generators/generator.d
```diff
@@ -383,36 +383,30 @@ class ProjectGenerator
...
 			// special support for overriding string imports in parent packages
 			// this is a candidate for deprecation, once an alternative approach
 			// has been found
-			if (ti.buildSettings.stringImportPaths.length) {
+			if (pti && ti.buildSettings.stringImportPaths.length) {
 				// override string import files (used for up to date checking)
-				foreach (ref f; ti.buildSettings.stringImportFiles)
+				if (ti.buildSettings.stringImportFiles
+					.map!(f => NativePath(f).head)
+					.any!(f => overrides.any!(of => NativePath(of).head == f)))
```
PetarKirov: I would split this change, as it looks like pure refactoring, in a separate commit.

source/dub/generators/generator.d
```diff
@@ -383,36 +383,30 @@ class ProjectGenerator
 		}
 
 		// 5. override string import files in dependencies
-		static void overrideStringImports(ref TargetInfo ti, TargetInfo[string] targets, string[] overrides)
+		static void overrideStringImports(ref TargetInfo ti, TargetInfo* pti, TargetInfo[string] targets, string[] overrides)
```
PetarKirov: I know it would be a bit verbose, but I'd rename `pti` to `parentTi` (or `parent_ti`, I'm not well-versed with vibe.d's style guide), so that it's clear further down that we're (lazily) pretending the dependent package's import files to those each dependency if we detect that we need to perform an override.

On second thought, if what I wrote above is correct, a comment along those lines would be quite helpful for other people looking at the code at a later time.

source/dub/generators/generator.d
```diff
@@ -383,36 +383,30 @@ class ProjectGenerator
...
 			// add to overrides for recursion
 			overrides ~= ti.buildSettings.stringImportFiles;
 			// override dependencies
 			foreach (depname; ti.dependencies)
-				overrideStringImports(targets[depname], targets, overrides);
+				overrideStringImports(targets[depname], &ti, targets, overrides);
 		}
 
-		overrideStringImports(targets[rootPackage.name], targets, null);
+		overrideStringImports(targets[rootPackage.name], null, targets, null);
```
PetarKirov:
```suggestion
		/*
		 * Pass `null` as parent target info, because the root package
		 * has no parent.
		 */
		overrideStringImports(targets[rootPackage.name], null, targets, null);
```

source/dub/generators/generator.d
```diff
@@ -383,36 +383,30 @@ class ProjectGenerator
...
-							f = o;
-							ti.buildSettings.prependStringImportPaths(op.parentPath.toNativeString);
-						}
-					}
+					ti.buildSettings.prependStringImportPaths(pti.buildSettings.stringImportPaths);
 				}
 			}
 			// add to overrides for recursion
 			overrides ~= ti.buildSettings.stringImportFiles;
 			// override dependencies
```
PetarKirov:
```suggestion
			// override dependencies by inheriting string imports from this `ti` (target info)
```

s-ludwig:
I redid the change now, keeping the previous nested `for` loop, because the previous rewrite actually accidentally skipped the `stringImportFiles` overriding performed within that loop.

s-ludwig:
> Azure fails with:
>>    Error: out of memory dmd failed with exit code 1.
> Most likely switching from a Win32 built dmd to one built in -m64 mode will fix this.

BTW, is there any valid reason why the default distribution package still contains a 32-bit executable these days? I can't imagine that any D developer is still working with a 32-bit Windows.

wilzbach:
> BTW, is there any valid reason why the default distribution package still contains a 32-bit executable these days? I can't imagine that any D developer is still working with a 32-bit Windows.

Yes, DMD does _NOT_ ship an official 64-bit executable for Windows.

PR #1791, opened 2019-10-31
Fix compilation with DIP1000 and compatibility with future `in=scope const` dmd change

PetarKirov:


source/dub/compilers/dmd.d
```diff
@@ -217,7 +217,7 @@ config    /etc/dmd.conf
 			case TargetType.executable:
 				if (platform.platform.canFind("windows"))
 					return settings.targetName ~ ".exe";
-				else return settings.targetName;
+				else return settings.targetName.idup;
```
MoonlightSentinel: Couldn't this copy be avoided by annotating `settings` with `return`?

PetarKirov: Yes, that should work and I thought about that, but I decided not to change the interface. I think it's more of compiler deficiency that it doesn't recognize that `settings.targetName` in GC allocated and so it has infinite lifetime, and so it's safe to escape it.

PetarKirov:
One of the AppVeyor builds is failing with out of memory since several other PRs, so it's unrelated.

wilzbach:
Thanks!

source/dub/compilers/dmd.d
PetarKirov: Actually I now remember why I didn't add the `return` attribute. The reason is that in all, but this single case, this function doesn't escape any part of it's parameter. Although this may sound too theoretical, in principle adding `return` attribute essentially limits callers of the function, meaning that just because of this single case they will need ensure that lifetime of `settings` is as long as the one of the result, which is mildly annoying.
A better solution than the `idup` is a function (with potentially the same name) that returns the argument as it is (without copying) if it is immutable with infinite lifetime.

PR #1794, opened 2019-11-03
Parse and store platform attribute of dependencies

GoaLitiuM:
This is to just avoid loss of information when converting between recipe types with `dub convert`. These changes do not affect build process (yet?), but can be used as a base for future support of platform specific dependencies.


andre2007:
As side information, there is a faq explaining why dependencies are not platform specific https://github.com/dlang/dub/wiki/FAQ

source/dub/recipe/json.d
```diff
@@ -251,10 +252,14 @@ private Json toJson(in ref BuildSettingsTemplate bs)
...
 	auto ret = Json.emptyObject;
 	if( bs.dependencies !is null ){
-		auto deps = Json.emptyObject;
-		foreach( pack, d; bs.dependencies )
-			deps[pack] = serializeToJson(d);
-		ret["dependencies"] = deps;
+		Json[string] deps;
+		foreach( pack, d; bs.dependencies ){
+			string key = d.platform ? "-"~d.platform : "";
+			deps.require(key, Json.emptyObject);
```
wilzbach:
```
source/dub/recipe/json.d(258,8): Error: no property require for type Json[string]
```

tl;dr: you can't use `require` here as it was only recently added and dub is supposed to support compilation for many D compilers.

thewilsonator:
Are travis and appveyor expected to fail here?

PR #1796, opened 2019-11-10
Print the platform probe atomically

SSoulaimane:
DMD may print some stuff in between calls of `pragma(msg)` and corrupt the json data.

This is blocking DMD PR https://github.com/dlang/dmd/pull/10539, all buildkite tests which depend on dub are failing.

After debugging, I found out It's printing a probe like this:
```json
__dub_probe_begin__
{
  "compiler": "dmd",
  "frontendVersion": 2088,
  "compilerVendor": "Digital Mars D",
  "platform": [
import    core.internal.array.utils	(/home/ss/Downloads/dmd/generated/linux/debug/64/../../../../../druntime/import/core/internal/array/utils.d)
    "linux", "posix"
  ],
  "architecture": [
    "x86_64"
   ],
}
__dub_probe_end__
```
It fails at parsing the probe as json and aborts everything.

### DIFF
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -298,6 +298,7 @@ NativePath generatePlatformProbeFile()
 	import dub.internal.vibecompat.core.file;
 	import dub.internal.vibecompat.data.json;
 	import dub.internal.utils;
+	import std.string : format;
 
 	// try to not use phobos in the probe to avoid long import times
 	enum probe = q{
@@ -314,7 +315,7 @@ NativePath generatePlatformProbeFile()
 			return res;
 		}
 
-		pragma(msg, } ~ `"` ~ probeBeginMark ~ `"` ~q{
+		pragma(msg, `%1$s`
 			~ '\n' ~ `{`
 			~ '\n' ~ `  "compiler": "`~ determineCompiler() ~ `",`
 			~ '\n' ~ `  "frontendVersion": ` ~ toString!__VERSION__ ~ `,`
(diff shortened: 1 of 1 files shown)


WebFreak001:
this code was extremely difficult to read before and after. Now with even more differently formatted string literals inside that string literal it's confusing.

Can we maybe change parts of the generated code to ``` `` ``` strings or `q"EOS <multiple lines> EOS"` for this big block where it's not obvious at a first glance which parts are "strings outside the strings", "strings inside the strings", "code inside the strings" or which parts are "code outside the strings"

wilzbach:
> this code was extremely difficult to read before and after. Now with even more differently formatted string literals inside that string literal it's confusing.

The proper solution is to use the new `-Xi=compilerInfo` anyhow, so I don't see any problems with this as the only maintainance step should be to remove this.

SSoulaimane:
@wilzbach dub has to support older compiler versions, so I don't think a this method will be replaced by anything recent.

wilzbach:
> dub has to support older compiler versions, so I don't think a this method will be replaced by anything recent.

Well, the last ten releases have this new interface which is incidentally also what we officially support as the lowest bootstrap compiler, so that's not really a big concern.

Anyhow, my point was that there's no point in investing more time into making this old interface look nicer as it's maintainance only.

SSoulaimane:
@WebFreak001 it looks fine to me, the only confusing bit is the to figure out the `q{ }` where it starts and ends.

PR #1798, opened 2019-11-20
Support dependencies as git url with exact commit

belka-ew:
https://github.com/dlang/dub/pull/1765

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -882,7 +882,7 @@ struct Version {
 		if (isSCM || other.isSCM) {
 			if (!isSCM) return -1;
 			if (!other.isSCM) return 1;
-			return (m_version == m_version) ? 0 : 1;
+			return m_version == other.m_version ? 0 : 1;
 		}
 
 		if (isBranch || other.isBranch) {
@@ -973,14 +973,15 @@ unittest {
 /// Determines whether the given string is a Git hash.
 bool isGitHash(string hash) @nogc nothrow pure @safe
 {
-	import std.ascii : isAlphaNum;
+	import std.ascii : isHexDigit;
 	import std.utf : byCodeUnit;
 
-	return hash.length >= 7 && hash.length <= 40 && hash.byCodeUnit.all!isAlphaNum;
+	return hash.length >= 7 && hash.length <= 40 && hash.byCodeUnit.all!isHexDigit;
 }
 
 @nogc nothrow pure @safe unittest {
 	assert(isGitHash("73535568b79a0b124bc1653002637a830ce0fcb8"));
 	assert(!isGitHash("735"));
 	assert(!isGitHash("73535568b79a0b124bc1-53002637a830ce0fcb8"));
+	assert(!isGitHash("73535568b79a0b124bg1"));
 }
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -34,7 +34,6 @@ struct PackageDependency {
 	Dependency spec;
 }
 
-
 /**
 	Represents a dependency specification.
 
@@ -56,6 +55,7 @@ struct Dependency {
 		NativePath m_path;
 		bool m_optional = false;
 		bool m_default = false;
+		Repository m_repository;
 	}
 
 	/// A Dependency, which matches every valid version.
@@ -93,10 +93,30 @@ struct Dependency {
 		m_path = path;
 	}
 
+	/** Constructs a new dependency specification that matches a specific
+		Git reference.
+	*/
+	this(Repository repository, string spec) {
+		this.versionSpec = spec;
+		this.repository = repository;
+	}
+
 	/// If set, overrides any version based dependency selection.
 	@property void path(NativePath value) { m_path = value; }
 	/// ditto
 	@property NativePath path() const { return m_path; }
+
+	/// If set, overrides any version based dependency selection.
+	@property void repository(Repository value)
+	{
+		m_repository = value;
+	}
+
+	/// ditto
+	@property Repository repository() const
+	{
+		return m_repository;
+	}
 
 	/// Determines if the dependency is required or optional.
 	@property bool optional() const { return m_optional; }
@@ -110,6 +130,9 @@ struct Dependency {
 
 	/// Returns true $(I iff) the version range only matches a specific version.
 	@property bool isExactVersion() const { return m_versA == m_versB; }
+
+	/// Determines whether it is a Git dependency.
+	@property bool isSCM() const { return !repository.empty; }
 
 	/// Returns the exact version matched by the version range.
 	@property Version version_() const {
@@ -167,7 +190,7 @@ struct Dependency {
 			ves = ves[1..$].expandVersion;
 			m_versA = Version(ves);
 			m_versB = Version(bumpIncompatibleVersion(ves) ~ "-0");
-		} else if (ves[0] == Version.branchPrefix) {
+		} else if (ves[0] == Version.branchPrefix || ves.isGitHash) {
 			m_inclusiveA = true;
 			m_inclusiveB = true;
 			m_versA = m_versB = Version(ves);
@@ -218,7 +241,6 @@ struct Dependency {
 		string r;
 
 		if (this == invalid) return "invalid";
-
 		if (m_versA == m_versB && m_inclusiveA && m_inclusiveB) {
 			// Special "==" case
 			if (m_versA == Version.masterBranch) return "~master";
@@ -272,7 +294,12 @@ struct Dependency {
 	*/
 	string toString()()
 	const {
-		auto ret = versionSpec;
+		string ret;
+
+		if (!repository.empty) {
+			ret ~= repository.toString~"#";
+		}
+		ret ~= versionSpec;
 		if (optional) {
 			if (default_) ret ~= " (optional, default)";
 			else ret ~= " (optional)";
@@ -296,12 +323,13 @@ struct Dependency {
 	Json toJson()
 	const @trusted { // NOTE Path and Json is @system in vibe.d 0.7.x and in the compatibility layer
 		Json json;
-		if( path.empty && !optional ){
+		if( path.empty && repository.empty && !optional ){
 			json = Json(this.versionSpec);
 		} else {
 			json = Json.emptyObject;
 			json["version"] = this.versionSpec;
 			if (!path.empty) json["path"] = path.toString();
+			if (!repository.empty) json["repository"] = repository.toString;
 			if (optional) json["optional"] = true;
 			if (default_) json["default"] = true;
 		}
@@ -314,6 +342,15 @@ struct Dependency {
 		d = fromJson((fromJson(d.toJson())).toJson());
 		assert(d == Dependency("1.0.0"));
 		assert(d.toJson() == Json("1.0.0"), "Failed: " ~ d.toJson().toPrettyString());
+	}
+
+	@trusted unittest {
+		Dependency dependency = Dependency(Repository("git+http://localhost"), "1.0.0");
+		Json expected = Json([
+			"repository": Json("git+http://localhost"),
+			"version": Json("1.0.0")
+		]);
+		assert(dependency.toJson() == expected, "Failed: " ~ dependency.toJson().toPrettyString());
 	}
 
 	/** Constructs a new `Dependency` from its JSON representation.
@@ -330,6 +367,12 @@ struct Dependency {
 
 				dep = Dependency.any;
 				dep.path = NativePath(verspec["path"].get!string);
+			} else if (auto repository = "repository" in verspec) {
+				enforce("version" in verspec, "No version field specified!");
+				enforce(repository.length > 0, "No repository field specified!");
+
+				dep = Dependency(Repository(repository.get!string),
+						verspec["version"].get!string);
 			} else {
 				enforce("version" in verspec, "No version field specified!");
 				auto ver = verspec["version"].get!string;
@@ -411,6 +454,7 @@ struct Dependency {
 		A specification is valid if it can match at least one version.
 	*/
 	bool valid() const {
+		if (this.isSCM) return true;
 		return m_versA <= m_versB && doCmp(m_inclusiveA && m_inclusiveB, m_versA, m_versB);
 	}
 
@@ -440,6 +484,7 @@ struct Dependency {
 	/// ditto
 	bool matches(ref const(Version) v) const {
 		if (this.matchesAny) return true;
+		if (this.isSCM) return true;
 		//logDebug(" try match: %s with: %s", v, this);
 		// Master only matches master
 		if(m_versA.isBranch) {
@@ -463,6 +508,13 @@ struct Dependency {
 	*/
 	Dependency merge(ref const(Dependency) o)
 	const {
+		if (this.isSCM) {
+			if (!o.isSCM) return this;
+			if (this.m_versA == o.m_versA) return this;
+			return invalid;
+		}
+		if (o.isSCM) return o;
+
 		if (this.matchesAny) return o;
 		if (o.matchesAny) return this;
 		if (m_versA.isBranch != o.m_versA.isBranch) return invalid;
@@ -677,6 +729,80 @@ unittest {
 	assert(Dependency("^1.2").versionSpec == "~>1.2"); // equivalent; prefer ~>
 }
 
+/**
+	Represents an SCM repository.
+*/
+struct Repository
+{
+	private string m_remote;
+
+	private Kind m_kind;
+
+	enum Kind
+	{
+		git,
+	}
+
+	/**
+		Params:
+			remote = Repository remote.
+	 */
+	this(string remote)
+	{
+		if (remote.startsWith("git+"))
+		{
+			m_remote = remote["git+".length .. $];
+			m_kind = Kind.git;
+		}
+		else
+		{
+			throw new Exception("Unsupported repository type");
+		}
+	}
+
+	string toString() nothrow pure @safe
+	{
+		if (empty) return null;
+		string kindRepresentation;
+
+		final switch (kind)
+		{
+			case Kind.git:
+				kindRepresentation = "git";
+		}
+		return kindRepresentation~"+"~remote;
+	}
+
+	/**
+		Returns:
+			Repository URL or path.
+	*/
+	@property string remote() @nogc nothrow pure @safe
+	in { assert(m_remote !is null); }
+	body
+	{
+		return m_remote;
+	}
+
+	/**
+		Returns:
+			Repository type.
+	*/
+	@property Kind kind() @nogc nothrow pure @safe
+	{
+		return m_kind;
+	}
+
+	/**
+		Returns:
+			Whether the repository was initialized with an URL or path.
+	*/
+	@property bool empty() const @nogc nothrow pure @safe
+	{
+		return m_remote.empty;
+	}
+}
+
 
 /**
 	Represents a version in semantic version format, or a branch identifier.
@@ -705,7 +831,7 @@ struct Version {
 	this(string vers)
 	{
 		enforce(vers.length > 1, "Version strings must not be empty.");
-		if (vers[0] != branchPrefix && vers.ptr !is UNKNOWN_VERS.ptr)
+		if (vers[0] != branchPrefix && !vers.isGitHash && vers.ptr !is UNKNOWN_VERS.ptr)
 			enforce(vers.isValidVersion(), "Invalid SemVer format: " ~ vers);
 		m_version = vers;
 	}
@@ -719,6 +845,9 @@ struct Version {
 
 	bool opEquals(const Version oth) const { return opCmp(oth) == 0; }
 
+	/// Tests if this represents a hash instead of a version.
+	@property bool isSCM() const { return m_version.isGitHash; }
+
 	/// Tests if this represents a branch instead of a version.
 	@property bool isBranch() const { return m_version.length > 0 && m_version[0] == branchPrefix; }
 
@@ -730,7 +859,7 @@ struct Version {
 		Note that branches are always considered pre-release versions.
 	*/
 	@property bool isPreRelease() const {
-		if (isBranch) return true;
+		if (isBranch || isSCM) return true;
 		return isPreReleaseVersion(m_version);
 	}
 
@@ -749,6 +878,13 @@ struct Version {
 		if (isUnknown || other.isUnknown) {
 			throw new Exception("Can't compare unknown versions! (this: %s, other: %s)".format(this, other));
 		}
+
+		if (isSCM || other.isSCM) {
+			if (!isSCM) return -1;
+			if (!other.isSCM) return 1;
+			return (m_version == m_version) ? 0 : 1;
+		}
+
 		if (isBranch || other.isBranch) {
 			if(m_version == other.m_version) return 0;
 			if (!isBranch) return 1;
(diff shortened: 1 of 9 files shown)


source/dub/dependency.d
```diff
@@ -273,7 +295,12 @@ struct Dependency {
 	*/
 	string toString()()
 	const {
-		auto ret = versionSpec;
+		string ret;
+
+		if (!repository.empty) {
+			ret ~= "#";
+		}
+		ret ~= versionSpec;
```
WebFreak001: depending on where this toString value is presented to the user, it might be worth implementing parsing such a path pattern so for example `"foo": "https://github.com/bla.git#abc"` would work which would be epic.

However I think this might also fit in another PR.

source/dub/dependency.d
```diff
@@ -303,6 +330,7 @@ struct Dependency {
 			json = Json.emptyObject;
```
WebFreak001: I don't know where toJson is called right now, however this else block with your new repository value will not be used because repository is used and thus `path` is empty and most likely non-optional, which makes it go into the other part of this if.

(I think this might be a SDL <-> JSON conversion thing which would not work here?)

source/dub/dependency.d
```diff
@@ -678,6 +719,34 @@ unittest {
...
+		return m_remote;
+	}
+
+	/**
+		Returns:
+			Whether the repository was initialized with an URL or path.
+	*/
+	@property bool empty() const @nogc nothrow pure @safe
+	{
+		return m_remote is null;
```
WebFreak001: `return !m_remote.length;` or `return m_remote.empty;` might be better because empty will currently be false if Repository was initialized with `""`

It will slightly change behavior as currently an `""` repository will actually serialize and deserialize properly as empty string instead of not being there however I think such a case should rather be handled as edge case in the parser and not be allowed in the first place.

source/dub/dependency.d
```diff
@@ -331,6 +359,10 @@ struct Dependency {
 
 				dep = Dependency.any;
 				dep.path = NativePath(verspec["path"].get!string);
+			} else if (auto repository = "repository" in verspec) {
+				enforce("version" in verspec, "No version field specified!");
+				dep = Dependency(Repository(repository.get!string),
```
WebFreak001: this should probably error out if repository is empty (see below)

source/dub/dependency.d
belka-ew: Added `enforce`.

source/dub/dependency.d
belka-ew: Added `enforce`.

source/dub/dependency.d
belka-ew: Thanks. Fixed and added a test.

source/dub/dependency.d
belka-ew: I personally don't like it that much. It is something npm.js does and I find separate fields more readable and more informative.

But if it has to be implemented, I'd prefer a separate PR. I'd like to make as few changes as possible to see if this feature has any future at all.

belka-ew:
> not sure how the others feel about this but I think having the json name be called "commit" would make more sense and also look better in SDL. After all "repository" is also a new field.

IMHO "commit" isn't a good name either. "commit" suggests a commit hash. But it also can be a `~branch` or a version (translated to a tag name). Supporting the last one wasn't probably a great idea, now things like `~>0.16.1` work and they shouldn't... A better name would be probably "ref", or "reference", or "refname".

It was already discussed before and a few requested to allow an arbitrary string as "version" (or "commit", or "refname"), so SCM can decide what is valid and what not.

So should I change the field name and/or remove the constraints (validation before passing it to a git call)?

source/dub/dub.d
```diff
@@ -582,7 +587,9 @@ class Dub {
 			fetchOpts |= (options & UpgradeOptions.preRelease) != 0 ? FetchOptions.usePrerelease : FetchOptions.none;
 			if (!pack) fetch(p, ver, defaultPlacementLocation, fetchOpts, "getting selected version");
 			if ((options & UpgradeOptions.select) && p != m_project.rootPackage.name) {
-				if (ver.path.empty) m_project.selections.selectVersion(p, ver.version_);
+				if (!ver.repository.empty) {
+					m_project.selections.selectVersion(p, ver.repository, ver.versionSpec);
+				} else if (ver.path.empty) m_project.selections.selectVersion(p, ver.version_);
```
WebFreak001: please use consistent brace styles

source/dub/dub.d
```diff
@@ -1580,6 +1587,8 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 	protected override Dependency[] getSpecificConfigs(string pack, TreeNodes nodes)
 	{
 		if (!nodes.configs.path.empty && getPackage(pack, nodes.configs)) return [nodes.configs];
+		else if (!nodes.configs.repository.empty && getPackage(pack, nodes.configs))
+			return [nodes.configs];
```
WebFreak001: I think this is becoming bad code style here with another if. How about:
```d
if (!nodes.configs.path.empty || !nodes.config.repository.empty) {
	if (getPackage(pack, nodes.configs)) return [nodes.configs];
	else return null;
}
else return null;
```

source/dub/project.d
```diff
@@ -1542,6 +1550,18 @@ final class SelectedVersions {
 		m_dirty = true;
 	}
 
+	/// Selects a certain Git reference for a specific package.
+	void selectVersion(string package_id, Repository repository, string spec)
```
WebFreak001: Please pick a more descriptive method name such as `selectVersionWithRepository` or `selectGitVersion`. dub.d:591 is extremely difficult to read because it is an overload of selectVersion using repository + versionSpec instead of version_ which is already confusing enough.

Stating it another time here in the method name that this is for git would make sense imo to make the overload clearer.

source/dub/dependency.d
```diff
@@ -412,6 +455,7 @@ struct Dependency {
 		A specification is valid if it can match at least one version.
 	*/
 	bool valid() const {
+		if (this.isGit) return true;
```
WebFreak001: do these `isGit` calls make sense for only git? I think having a generic `isSCM` method (which for now would do the same thing as isGit) would make this code clearer

source/dub/dependency.d
```diff
@@ -441,6 +485,7 @@ struct Dependency {
 	/// ditto
 	bool matches(ref const(Version) v) const {
 		if (this.matchesAny) return true;
+		if (this.isGit) return true;
```
WebFreak001: ditto

source/dub/dependency.d
```diff
@@ -464,6 +509,13 @@ struct Dependency {
 	*/
 	Dependency merge(ref const(Dependency) o)
 	const {
+		if (this.isGit) {
+			if (!o.isGit) return this;
+			if (this.m_versA == o.m_versB) return this;
+			return invalid;
+		}
+		if (o.isGit) return o;
```
WebFreak001: ditto

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+
+		Params:
+			name = Package name
+			dependency = Dependency that contains the repository URL and a specific commit
+
+		Returns:
+			The package loaded from the given Git repository or null if the
+			package couldn't be loaded.
+	*/
+	Package clonePackage(string name, Dependency dependency)
```
WebFreak001: all the `git` references in the documentation should probably be replaced with just generic description that this method will download a repository if `repository` is set

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+	in { assert(!dependency.repository.empty); }
+	body {
+		import std.process : escapeShellCommand, executeShell;
+
+		string gitReference = dependency.versionSpec;
```
WebFreak001: this should first check if repository isGit (even if right now the `in` block does check if the repository is empty, in case of other SCM all callers of this method assume this works with everything)

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+		if (gitReference.startsWith("~")) gitReference.popFront;
+
+		const destination = m_repositories[LocalPackageType.user].packagePath ~
+			NativePath(name~"-"~gitReference) ~ name;
+		const nativeDestination = destination.toNativeString();
+
+		if (!exists(nativeDestination)) {
+			auto command = escapeShellCommand("git", "clone",
```
WebFreak001: can we use spawnProcess or just execute instead of executeShell in this whole method? It makes the escaping much easier and less likely for errors.

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+			}
+
+			command = escapeShellCommand("git",
+					"-C", nativeDestination,
+					 "checkout", gitReference);
+			if (executeShell(command).status != 0) {
+				rmdirRecurse(nativeDestination);
+				return null;
+			}
+		}
```
WebFreak001: the whole git specific block here should be moved into dub/internal/git.d for readability and portability

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+		auto pack = Package.load(destination);
+		addPackages(m_temporaryPackages, pack);
```
WebFreak001: can you comment and document why you are calling addPackages inside the clone function?

source/dub/packagemanager.d
```diff
@@ -203,6 +203,55 @@ class PackageManager {
...
+			The package loaded from the given Git repository or null if the
+			package couldn't be loaded.
+	*/
+	Package clonePackage(string name, Dependency dependency)
+	in { assert(!dependency.repository.empty); }
+	body {
+		import std.process : escapeShellCommand, executeShell;
+
+		string gitReference = dependency.versionSpec;
+		if (gitReference.startsWith("~")) gitReference.popFront;
```
WebFreak001:
```suggestion
		gitReference = gitReference.chompPrefix("~");
```

source/dub/project.d
```diff
@@ -256,7 +256,7 @@ class Project {
 		enforce(!m_rootPackage.name.canFind(' '), "Aborting due to the package name containing spaces.");
 
 		foreach (d; m_rootPackage.getAllDependencies())
-			if (d.spec.isExactVersion && d.spec.version_.isBranch) {
+			if (d.spec.isExactVersion && d.spec.version_.isBranch && !d.spec.isGit) {
```
WebFreak001:
```suggestion
			if (d.spec.isExactVersion && d.spec.version_.isBranch && d.spec.repository.empty) {
```

not inverting things is much better readable imo and this is not git specific.

source/dub/project.d
```diff
@@ -354,7 +354,10 @@ class Project {
 					}
 				} else if (m_selections.hasSelectedVersion(basename)) {
 					vspec = m_selections.getSelectedVersion(basename);
-					if (vspec.path.empty) p = m_packageManager.getBestPackage(dep.name, vspec);
+					if (!vspec.repository.empty) {
+						p = m_packageManager.clonePackage(basename, vspec);
+						if (subname.length) p = m_packageManager.getSubPackage(p, subname, true);
```
WebFreak001: this getSubPackage line is now copy-pasted 5 times into this function, let's add a function inside this function which does the if check + assignment for us, such as `resolveSubPackage(p, true);`

source/dub/project.d
```diff
@@ -354,7 +354,10 @@ class Project {
...
+					} else if (vspec.path.empty) p = m_packageManager.getBestPackage(dep.name, vspec);
```
WebFreak001: I would add braces here and move this into a separate line for readability.

source/dub/project.d
```diff
@@ -354,7 +354,10 @@ class Project {
 					}
 				} else if (m_selections.hasSelectedVersion(basename)) {
 					vspec = m_selections.getSelectedVersion(basename);
-					if (vspec.path.empty) p = m_packageManager.getBestPackage(dep.name, vspec);
+					if (!vspec.repository.empty) {
```
WebFreak001: the precedence is probably wrong here. Everywhere else before you first checked path, then repository. Here you first check repository.

source/dub/project.d
```diff
@@ -1604,16 +1624,23 @@ final class SelectedVersions {
...
 	static Dependency dependencyFromJson(Json j)
 	{
 		if (j.type == Json.Type.string)
 			return Dependency(Version(j.get!string));
-		else if (j.type == Json.Type.object)
+		else if (j.type == Json.Type.object && "path" in j)
 			return Dependency(NativePath(j["path"].get!string));
+		else if (j.type == Json.Type.object && "repository" in j)
+			return Dependency(Repository(j["repository"].get!string), j["version"].get!string);
 		else throw new Exception(format("Unexpected type for dependency: %s", j.type));
```
WebFreak001: The error message can now print "object" as unexpected type if neither path or repository are specified. It should probably just print the json at this point to avoid confusion.

source/dub/project.d
```diff
@@ -1604,16 +1624,23 @@ final class SelectedVersions {
...
 
 	static Dependency dependencyFromJson(Json j)
 	{
 		if (j.type == Json.Type.string)
 			return Dependency(Version(j.get!string));
-		else if (j.type == Json.Type.object)
+		else if (j.type == Json.Type.object && "path" in j)
 			return Dependency(NativePath(j["path"].get!string));
+		else if (j.type == Json.Type.object && "repository" in j)
+			return Dependency(Repository(j["repository"].get!string), j["version"].get!string);
```
WebFreak001:
```suggestion
			return Dependency(Repository(j["repository"].get!string), enforce("version" in j, "Expected \"version\" field in repository version object").get!string);
```

rikkimax:
> You are assuming everywhere that everything is git if `repository` is set. I think for forwards-compatibility the Repository struct should contain an `isGit` method and imo also contain the version spec under a (internal) name such as `gitRef`.

Use this pattern in the future. Doing it that way implies that there will only ever be two types of repositories.

```d

struct Struct {
  Type type;
  
  enum Type {
    Error, // optional in this case probably isn't needed
    Git,
    Mercurial,
  }
}
```

source/dub/packagemanager.d
belka-ew: I renamed the function to `loadSCMPackage`. Is it clearer?
It is actually analogous to `getOrLoadPackage`, but `getOrLoadPackage` loads from the filesystem and SCM packages should be cloned first if it isn't already done.

belka-ew:
Rebased on the latest master, fixed the related tests, made requested changes.

belka-ew:
This feature is industry-related; it was assigned to me by Funkwerk.

Geod24:
What happens when one app depends on two libraries which depends on a different version of the same library ?

WebFreak001:
there is precedence with SCM versions over normal versions right now, otherwise the versions must match or it will be invalid when trying to merge the dependencies: https://github.com/dlang/dub/pull/1798/files#diff-d5a3ded5a606c330dbd2ae1ba36a3044R514

WebFreak001:
actually looking at it now, something doesn't seem right in that merge function:
it checks this.m_versA and o.m_versB but not this.m_versB or o.m_versA

maybe this struct needs to be looked at more precisely again

Geod24:
For a new functionality the release notes and tests are quite underwhelming.
My personal preference would be to rely on `git submodules` for this (https://github.com/dlang/dub/issues/1840) but if current dub users implemented this solution and need it, so be it, however it should be a bit more tested and documented.

And TBH I'm worried that this becomes the path of least resistance, and we end up with a bunch of incompatible libraries. Slippery slope argument I know, but this is essentially completely bypassing the version resolution algorithm.

belka-ew:
> actually looking at it now, something doesn't seem right in that merge function:
it checks this.m_versA and o.m_versB but not this.m_versB or o.m_versA

I overlooked it and changed it in the last commit to `this.m_versA == o.versA`. But is should practically be the same, since `versA == versB` in the case of exact versions and branches.

belka-ew:
> My personal preference would be to rely on git submodules for this (#1840)

I'm not sure, anyone will implement it.  The first attempt to implement the git-feature from this PR (much simpler feature) was done in 2018 (https://github.com/dlang/dub/pull/1403) and nobody had interest so far. Leaving aside my private opinions DLF expressed an intention to drop dub in favour of something like dud. With a second package manager dub development would become even less active.

> And TBH I'm worried that this becomes the path of least resistance, and we end up with a bunch of incompatible libraries. Slippery slope argument I know, but this is essentially completely bypassing the version resolution algorithm.

I'd like to make one thing clear: It is not intended to be used in public packages. Such SCM repositories are supported everywhere. Ruby Bundler supports them, PHP Composer, NodeJS. And none of them has the problem that repositories become the path of least resistance.

So I have a few ruby projects which share some (private) code. Since I'm a lonely developer I don't have the capacity to maintain my private gem registry, so I put the shared code into a git repository and here is the code from one of my Gemfiles:

```ruby
gem 'infra', '1.0.1', git: 'git@here.is.my.ip:/and/path/to/the.git'
```

This way I can continue the development without the whole ceremony with setting up a private registry.

Funkwerk has a private dub registry. The use case there was dependency testing. I have a repository "MySharedD" - I make some changes there, but want to test that these changes work in real end-products "productA" and "productB". So I include "MySharedD" in "productA" as git reference, so that it overrides everything and can check that "productA" works with the modified "MySharedD". I don't want to release "productA" with the git reference; instead I merge my changes into MySharedD, make a new release and update the dependency in "productA". Currently I have to go to `~/.dub` and make my changes there and everything collapses around me.

Geod24:
> I'm not sure, anyone will implement it.

Yes, I understand the frustration.
I will set some time aside this week to experiment with this (and the submodules idea), and give it another review by next Monday. Hope that works for you.

andre2007:
I would like to enhance Dub to also specify a single maven artefact as dependency. I see 1 issue with your pull request. A new syntax is introduced which can only be used for protocol git:

```
"dependencies": {
		"gitcompatibledubpackage": {
			"repository": "https://github.com/dlang-community/gitcompatibledubpackage.git",
			"version": "ccb31bf6a655437176ec02e04c2305a8c7c90d67"
		}
	}
```

If I now want to provide s.th. similiar for protocol Maven, I have to reinvent an additional syntax. I would really like to see here a protocol neutral syntax, therefore it should be open for additional protocols. As an idea:

```
"dependencies": {
    "gitpackage": {
        "repository": "git+https://github.com/dlang-community/gitpackage.git",
        "version": "ccb31bf6a655437176ec02e04c2305a8c7c90d67"
    },
    "mavenpackage": {
        "repository": "mvn+http://localhost:8040/maven/libs-release/dubpackages/mavenpackage",
        "version": "1.0.3"
    },
    "dubpackage": {
        "repository": "dub+http://dub-server/dubpackage",
        "version": "1.0.3"
    },
    "httppackage": {
        "repository": "http://localhost:8888/httppackage-5.2.4.zip"
    }
}
```

WebFreak001:
I agree, having a `git+` prefix would be a good addition to this for future compatibility

belka-ew:
I added `git+`.

source/dub/internal/git.d
```diff
@@ -101,3 +101,35 @@ private string determineVersionWithGitTool(NativePath path)
...
+		if (wait(command) != 0) {
+			return false;
+		}
+	}
+
+	command = spawnProcess(["git", "-C", destination, "checkout", reference]);
+	if (wait(command) != 0) {
+		rmdirRecurse(destination);
+		return false;
+	}
```
WebFreak001, line 132: Do you think we should clone recursively here? Currently dub downloaded packages don't support submodules but I think this would be a great place to support them.

NilsLankila: yes because the dep specified by the git+... url may contain a a sub dependency that specifies a "path" property that points to where git submodule would be cloned.

belka-ew: I wouldn't make it mandatory. Saying carefully, there are some people who use submodules that can form a cycle :)

So `repository a` has `repository b` as submodule and vice versa. Such submodules can still be used If I add both, `repository a` and `repository b` as dependencies and don't clone resursively.

Though this can be easely extended with an additional option, like `recursively: true` in `dub.json`.

NilsLankila: i don't know, I understood the question of webfreak as a suggestion to execute the command to pull submodule inconditionally (as this command would not fail), i.e without changing anything more.

belka-ew: `git submodule update --init`? This would be fine; `git submodule update --init --recursive` not.

source/dub/internal/git.d
```diff
@@ -101,3 +101,35 @@ private string determineVersionWithGitTool(NativePath path)
...
+		Whether the cloning succeeded.
+*/
+bool cloneRepository(string remote, string reference, string destination)
+{
+	import std.process : Pid, spawnProcess, wait;
+
+	Pid command;
+
+	if (!exists(destination)) {
+		command = spawnProcess(["git", "clone", remote, destination]);
```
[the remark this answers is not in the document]
belka-ew: added `--no-checkout` unconditionally, `-q` if not `--verbose`.

source/dub/internal/git.d
```diff
@@ -101,3 +101,35 @@ private string determineVersionWithGitTool(NativePath path)
...
+		if (wait(command) != 0) {
+			return false;
+		}
+	}
+
+	command = spawnProcess(["git", "-C", destination, "checkout", reference]);
```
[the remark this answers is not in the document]
belka-ew: added `--detach` unconditionally, since the output isn't useful even for debugging; `-q` if not `--verbose`.

belka-ew:
SDL should be fixed now.

belka-ew:
?

belka-ew:
ping @WebFreak001

belka-ew:
:'-(

NilsLankila:
Yeah what's goin on ? I'd like to site this : "every great journey starts with the first step", i.e aiming at perfection will not help.

WebFreak001:
oh sorry, I kept putting it off, losing it in the eternal void that is my open firefox tabs. Gonna check and review now

source/dub/dependency.d
```diff
@@ -829,4 +966,21 @@ unittest {
...
+	assert(Version("73535568b79a0b124bc1653002637a830ce0fcb8").isSCM);
+}
+
+/// Determines whether the given string is a Git hash.
+bool isGitHash(string hash) @nogc nothrow pure @safe
+{
+	import std.ascii : isAlphaNum;
+	import std.utf : byCodeUnit;
+
+	return hash.length >= 7 && hash.length <= 40 && hash.byCodeUnit.all!isAlphaNum;
```
WebFreak001, line 979: this should be `isHexDigit` (as commit hashes are hex encoded)

If we want to support tag names I would instead suggest adding some kind of additional prefixed syntax in a later PR

belka-ew: True that.

belka-ew:
@WebFreak001 Your assumption is correct, version comparison was wrong.

belka-ew:
Is something still missing?

andre2007:
@belka-ew thanks a lot!

NilsLankila:
woohoo finally.

andre2007:
While trying this functionality today I noticed a small difference to dub packages downloaded from a dub repository. The package folder on local file system is missing a lock file (for dub packages downloaded with the new functionality). Is this by purpose? I am not sure for what the file is needed (concurrent dub package downloads?)

nordlow:
Would it be possible to add support for specifying git-url+hash in dub.selections. @Geod24?

WebFreak001:
I think the dub selections already contain this. Add a git repo dependency and run `dub upgrade`, it will put it in there.

PR #1801, opened 2019-11-22
Revert "Add flag --submodules to add git submodules in the project folder as available packages."

wilzbach:
Reverts dlang/dub#1780


FeepingCreature:
The history of this feature is clearly dissatisfactory, in many ways. (Two-month PRs?)

Look, if you think this doesn't fit in dub, then say it doesn't fit in dub. If you don't want this to go in, then say you don't want it to go in. Don't sit around noncommittally for two months, then immediately revert it when it gets merged.

Either outcome would be acceptable to us. Feature is upstreamed, great, we can rely on dub supporting it. (Can we?) Feature is rejected, great, we can (in)officially fork dub and ship our own binary internally. But please, commit to one or the other!

wilzbach:
> The history of this feature is clearly dissatisfactory, in many ways. (Two-month PRs?)

I fully understand, but I am not a dub maintainer.

> Look, if you think this doesn't fit in dub, then say it doesn't fit in dub. If you don't want this to go in, then say you don't want it to go in. 

Well, I do think that this feature is mostly useless for the dub community, but it's my opinion and I was hoping other people might share there opinion.

> Don't sit around noncommittally for two months, then immediately revert it when it gets merged.

1) I gave feedback a few times
2) I have other priorities
3) It's not my responsibility. There are other people too!
4) I reverted because it was an unfinished discussion and there was _no_ reason given why Nicholas merged the PR (and he has a pretty track record on merging things without reviewing them)

FeepingCreature:
In summary, what I'm getting from this is that dub is dead?

You don't hold it as your responsibility to review pull requests, but you hold it as your responsibility to ensure that pull requests don't get merged without maintainer agreement, and you don't accept nick as a maintainer, and afaics there isn't anybody else.

The effect of this is, as far as I can see, that there _isn't anybody_ who could make this kind of non-obvious decision, and as a result dub is stuck in place.

wilzbach:
> In summary, what I'm getting from this is that dub is dead?

Well, there's definitely no active development.

> You don't hold it as your responsibility to review pull requests

Yes, I try to when I get a chance, but I don't think I have any responsibility here. Remember that unlike Nicholas we aren't paid for this.

> you don't accept nick as a maintainer

Yes, he has shown a very bad track record on merging things without reviewing them (over all dlang repositories). As he hasn't provided a reasoning for his decision I think he just clicked on merge for your PR.

>  but you hold it as your responsibility to ensure that pull requests don't get merged without maintainer agreement

Yes, when I accidentally manage to catch a PR that shouldn't have been merged, I revert. 

> The effect of this is, as far as I can see, that there doesn't exist a person who could make this kind of non-obvious decision, and as a result dub is stuck, ironically, in "maintenance mode". Does that seem correct?

More or less, though there's @atilaneves @s-ludwig @PetarKirov @Geod24 @MartinNowak @CyberShadow

FeepingCreature:
> Well, there's definitely no active development.

Can't imagine why. /snark

> More or less, though there's @atilaneves @s-ludwig @PetarKirov @Geod24 @MartinNowak @CyberShadow

I really do think this is an ownership problem. Isn't this something the D foundation should take a look at? Critical infrastructure is left unmaintained; this seems like the sort of thing that should be in the first row of problems to solve rather than yet another half-baked feature.

atilaneves:
As I just mentioned in the PR that got merged in the first place, why does dub need this?

FeepingCreature:
(See the other PR for response.)

CyberShadow:
I am inclined to agree with reverting only because (as per my review there) the implementation seems vastly suboptimal to me and mostly consists of unnecessary complexity (new configuration options and function parameters added throughout call stacks), but I might be wrong.

FeepingCreature:
I have a follow up that uses path selection via `dub.selection.json` instead. https://github.com/dlang/dub/pull/1802 It seems the better approach, but I'm not sure if there's problems with requiring hardcoding `dub.selections.json` to the extent that the project won't build if you delete it.

PR #1804, opened 2019-11-25
Replace m_inclusiveA, m_versA, m_inclusiveB, m_versB with easier-to-understand LowerBound/UpperBound types

FeepingCreature:
The aim here is to push the comparison logic into a single type. As a bonus, a bug in version comparison is also fixed.


Geod24:
Somewhat done in #2271

PR #1807, opened 2019-11-29
Recognize git semantic version tags without leading "v".

FeepingCreature:
Note that `code.dlang.org` still requires `v`, so this only affects version detection on internal repositories.

https://forum.dlang.org/thread/cdtzzufsejkmdwmnsoul@forum.dlang.org seems broadly in favor, though I forgot to ask about the specific case of repository-local version detection.

This is part of my effort to get git submodules supported in dub for internal use, since we have a lot of internal projects that don't use dub-type 'v' tags - simply because there has been zero reason to use 'v' tags, because we weren't using dub anyways.

### DIFF
--- a/source/dub/internal/git.d
+++ b/source/dub/internal/git.d
@@ -87,10 +87,11 @@ private string determineVersionWithGitTool(NativePath path)
 		auto commit = parts[$-1];
 		auto num = parts[$-2].to!int;
 		tag = parts[0 .. $-2].join("-");
-		if (tag.startsWith("v") && isValidVersion(tag[1 .. $])) {
-			if (num == 0) return tag[1 .. $];
-			else if (tag.canFind("+")) return format("%s.commit.%s.%s", tag[1 .. $], num, commit);
-			else return format("%s+commit.%s.%s", tag[1 .. $], num, commit);
+		if (tag.startsWith("v")) tag = tag[1 .. $];
+		if (isValidVersion(tag)) {
+			if (num == 0) return tag;
+			else if (tag.canFind("+")) return format("%s.commit.%s.%s", tag, num, commit);
+			else return format("%s+commit.%s.%s", tag, num, commit);
 		}
 	}
 


FeepingCreature:
preemptively ping @wilzbach to save you the bother of having to revert this in 73h...

FeepingCreature:
Oh yeah it won't automerge cause of the AppVeyor thing... `Error: out of memory`. Not sure how to fix that in this PR... :-)

s-ludwig:
Sorry, but why is this even going to be merged? The forum thread was all but clear about this being a good idea. I think the rationale for the change is pretty weak (the original one even being wrong) and having a clear guideline/rule for naming things is generally a valuable thing that should require a good reason for it to be weakened. I also don't think the consequences of the change have been analyzed sufficiently.

Anyway, since I'm not able to devote enough time or energy to this project at this point, this is just an opinion.

FeepingCreature:
> Sorry, but why is this even going to be merged?

There's been a pattern here where @thewilsonator merges without sufficient review and @wilzbach reverts again. As I said in the PR, I'm okay with either an acceptance or a clear rejection; unfortunately, it seems I'm going to get the "merge immediately-revert immediately" pattern again...

> The forum thread was all but clear about this being a good idea. 

The forum thread at least showed a broad agreement, if not consensus, that this was a good thing. There was some disagreement, and I see the reason for it, but frankly the forum often takes stupid and self-contradictory positions, so I don't think it's realistic to convince everybody on it. (Look what they did to named arguments!) I think aside really obvious cases, "many people agree" is the best you *can* plausibly get. Of course, this is made more difficult by the fact that there's no final authority on dub right now, there's only the two-faced deity of @thewilsonator and @wilzbach (one merges, the other reverts). It's positively mythic.

wilzbach:
>  As I said in the PR, I'm okay with either an acceptance or a clear rejection; unfortunately, it seems I'm going to get the "merge immediately-revert immediately" pattern again...

I don't have an opinion about this PR.

> The forum thread at least showed a broad agreement, if not consensus, that this was a good thing.

At the end of the day at the moment the decision is up to @atilaneves or @s-ludwig.

atilaneves:
I don't have a strong opinion on this.

FeepingCreature:
I don't think anybody has a strong opinion on anything with dub. The only one on this repo who has a strong opinion is wilzbach that thewilsonator should not merge things without adequate review.

Geod24:
I started looking at this PR by thinking we should make it optional. But reading the forum post actually made me change my mind. I simply don't see a strong argument in favor of it.

If we were to start all over, sure, we should make it optional, but currently I don't think this change would carry its own weight (because to be useful it has to come with changes to other parts of the ecosystem, as you mentioned). So I would say leave it as is.

FeepingCreature:
I already know opinions are mixed. Adding more opinions on top will not fix that problem.

I need a dedicated maintainer slash product owner to tell me "yes" or "no". Unfortunately, this fails at "I need a dedicated maintainer slash product owner."

atilaneves:
> Thanks, but I already know opinions are mixed. Adding more opinions on top will not fix that problem.
> 
> I need a dedicated maintainer slash product owner to tell me "yes" or "no". Unfortunately, this fails at "I need a dedicated maintainer slash product owner."

Let's go with "no" then. Thanks for the effort!

WebFreak001:
imo I have seen people mistag multiple times in the past, I would be for a yes because this shouldn't break most users tagging system.

The only case where I can imagine this would break stuff is when people tag releases like `2019.10.15` which would be picked up. But I think in 99.98% of cases this should be no issue.

However for this to really be effective, it would also need to be allowed on the dub registry. Then there is the issue with how dub will handle multiple tags with the same name which would need to be tested.

I have cloned all dub repositories and listed all tags for all repositories here: [https://wfr.moe/f6LWVZ/tags.txt](https://wfr.moe/f6LWVZ/tags.txt)

And here is a list of dub packages which have previously tagged without the v but with valid version strings with their current versions: [https://wfr.moe/f6LWvQ/fails.txt](https://wfr.moe/f6LWvQ/fails.txt)

FeepingCreature:
I think some of those are projects that switched to dub 'v' style when they gained a dub file, not necessarily mistags.

Geod24:
> Let's go with "no" then. Thanks for the effort!

I think that settles it.

PR #1808, opened 2019-11-29
Mildly refactor dependency resolver

FeepingCreature:
DependencyResolver is a template for no good reason. Replace its parameters with concrete types.
Change the unittest to use the actual Dependency type instead of a weird, int-based imitation.
Rename TreeNodes to PackageDependency, because we're no longer pretending to be a generic pseudo-array type.

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -27,14 +27,25 @@ import std.string;
 */
 class DependencyResolver {
 	/** Encapsulates a list of outgoing edges in the dependency graph.
-
 		A value of this type represents a single dependency with multiple
 		possible configurations for the target package.
 	*/
-	static struct TreeNodes {
+	static struct PackageDependency {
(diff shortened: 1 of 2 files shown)


s-ludwig:
The plan is to potentially use `DependencyResolver` also for configuration resolution. I am with you with regards to naming, where the current names can be confusing when thining about versions instead of configurations. However, instead of dropping the whole generality of the algorithm, it should be enough to change the naming convention here.

I don't care much about the unit tests using `Dependency` instead of a smaller test type, although it does introduce an unnecessary dependency and makes the tests more verbose.

Calling this a "mild" refactor stretches the meaning of the word quit a bit though ;-)

FeepingCreature:
I called it a mild refactor because it (arguably) doesn't change the structure of the code, only the syntax.

Oh, so the template logic is intended to see more use later on? Fair enough.

PR #1810, opened 2019-12-05
Meson build

DanySpin97:


DanySpin97:
> though I believe that if this isn't added to a CI it will get outdated rather quickly

Is it okay if I use `meson build` in place of ./build.sh? Or should I just add it along side?

wilzbach:
> Is it okay if I use meson build in place of ./build.sh? Or should I just add it along side?

I don't think that we can expect all packagers to have meson installed, so it would need to be alongside. However, just doing as one additional job on e.g. Travis should be more than enough.

DanySpin97:
> I don't think that we can expect all packagers to have meson installed

It fairly reasonable to depend on meson. According to ArchLinux packages database, meson is required by ~400 other packages. Among them, there are some core packages like glib2, gtk3, mesa and xorg.

wilzbach:
> It fairly reasonable to depend on meson. [..]

Well, Dub is not only shipped to ArchLinux, but a lot of other distributions. Additionally, it is part of LDC and GDC. I'm not certain whether they use the `build.sh` at the moment.

Anyhow, even if there's a consensus to go fully with meson, we still couldn't immediately replace the `build.sh`, but we would need to give all packagers a reasonable transition time for a few releases.

Vild:
> I'm not certain whether they use the build.sh at the moment.

Arch uses a custom ldmd command right now, and it gets the build files from `build-files.txt`: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/dub#n26
As long as it is CI tested to make sure it is never breaks, I would use it for Arch.

One thing I would really like to be added to this meson file is the ability to run the unittests.

I'm not meson expert, but I think it would be a good idea to try and make it look more similar to the other meson.build files that exist for other D project. Like https://github.com/dlang-community/containers/blob/master/meson.build
But this suggestion might not be applicable as those projects are often shared libraries.

DanySpin97:
> One thing I would really like to be added to this meson file is the ability to run the unittests.

I wanted to have this MR merged and then create a new one to run unittest over meson. If you prefer I can add unittest support directly in this MR.

> I'm not meson expert, but I think it would be a good idea to try and make it look more similar to the other meson.build files that exist for other D project.

Which things is dub meson.build missing? Is some flag missing? Some files not installed? Can you please list the issues so I can address them?

If you are referring to the single file over multiple files, it is just personal preference, as I found clearer to have multilpe files instead of a big one; it does not make a difference on what it does.

> I'm not certain whether they use the build.sh at the moment.

This exactly why meson should be preferred imho, because it is way more standard than a custom build script and everyone would use the same method.

> Anyhow, even if there's a consensus to go fully with meson, we still couldn't immediately replace the build.sh, but we would need to give all packagers a reasonable transition time for a few releases.

I think packagers would be happy to switch to meson already. The reasonable transition would be justified in breaking changes to the API, imho.

PR #1811, opened 2019-12-05
Fix unnecessary rebuild of single file package

andre2007:
In case a single file package has dependencies, isUpToDate method checks whether `dub.selections.json` exists. As single file package has the UpgradeOptions.noSaveSelections `dub.selections.json` won't be created and therefore isUpToDate always returns false. 
This pull request passes the UpgradeOptions.noSaveSelections to the build generator and avoid the check of `dub.selections.json`

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -814,6 +814,7 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.rdmd = m_rdmd;
 		gensettings.tempBuild = m_tempBuild;
 		gensettings.parallelBuild = m_parallel;
+		gensettings.single = m_single;
 
 		logDiagnostic("Generating using %s", m_generator);
 		dub.generateProject(m_generator, gensettings);
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -591,6 +591,10 @@ struct GeneratorSettings {
 
 	// only used for generator "build"
 	bool run, force, direct, rdmd, tempBuild, parallelBuild;
+
+	/// single file dub package
+	bool single;
+
 	string[] runArgs;
 	void delegate(int status, string output) compileCallback;
 	void delegate(int status, string output) linkCallback;
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -396,7 +396,8 @@ class BuildGenerator : ProjectGenerator {
 		foreach (p; packages)
 			allfiles ~= (p.recipePath != NativePath.init ? p : p.basePackage).recipePath.toNativeString();
 		foreach (f; additional_dep_files) allfiles ~= f.toNativeString();
-		if (main_pack is m_project.rootPackage && m_project.rootPackage.getAllDependencies().length > 0)
+		bool checkSelectedVersions = !settings.single;
+		if (checkSelectedVersions && main_pack is m_project.rootPackage && m_project.rootPackage.getAllDependencies().length > 0)
 			allfiles ~= (main_pack.path ~ SelectedVersions.defaultFile).toNativeString();
 
 		foreach (file; allfiles.data) {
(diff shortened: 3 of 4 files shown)


andre2007:
version.d was added by mistake, i will remove it from pr

source/dub/commandline.d
```diff
@@ -599,6 +599,7 @@ abstract class PackageBuildCommand : Command {
 		bool m_forceRemove = false;
 		bool m_single;
 		bool m_filterVersions = false;
+		bool m_noSaveSelections;
```
aliak00: nit, but imo, using positives is easier to parse (gets rid of having to deal with double negatives)

aliak00: (i.e. making the var `saveSelections` as opposed to `noSaveSelections`)

andre2007: done:)

andre2007:
Ping:)

semaphoreci & appveyor issues are unrelated.

andre2007:
Reopening pr to trigger voters again

source/dub/generators/generator.d
```diff
@@ -590,7 +590,7 @@ struct GeneratorSettings {
 	bool filterVersions;
 
 	// only used for generator "build"
-	bool run, force, direct, rdmd, tempBuild, parallelBuild;
```
WebFreak001: can you add a line of documentation to your new field? I think it's the least obvious out of all fields here.

andre2007: Fixed :)

source/dub/generators/generator.d
```diff
@@ -591,6 +591,10 @@ struct GeneratorSettings {
 
 	// only used for generator "build"
 	bool run, force, direct, rdmd, tempBuild, parallelBuild;
+
+	/// Selected dub packages are stored in dub.selections.json
+	saveSelections;
```
WebFreak001: isn't this missing `bool`?

also for backwards compatibility and API use this should probably be ` = true`

WebFreak001: and actually how this is used, the comment should probably say something like:

```
/// checks if dub.selections.json is updated for rebuild check if true
```

if I read the code correctly.

andre2007: Fixed

WebFreak001:
The only thing I think should be improved now is that the variable is still called "saveSelections". I think some name that would better represent the documentation and the case what it actually does (rebuild if dub.selections.json is modified) would be more appropriate. Makes it easier to construct and read `GeneratorSettings` variables in new code (which is done both by dub and by third party libraries)

andre2007:
@WebFreak001 rebuilding is a decision at some specific point in the program flow based on the information that the selections are saved. Therefore I am not sure about changing the variable name, because.. But if you like I change the variable name, do you have a proposal?

WebFreak001:
hm you are right but the variable isn't used for anything else and I don't really see any other uses in it in the future.

Maybe this shouldn't be stored in GeneratorSettings at all like this, but rather only store "single" in generator settings (because generators might be interested in single more) and then determine `bool checkSelectedVersions = !settings.single;` in the `isUpToDate` function.

`single` would also fit the other existing build variables in there more imo

PetarKirov:
@andre2007 The lastest revision of your PR looks like a pretty clean change, but unfortunately right now dub-registry decided to take a break and the CIs haven't been able to test it, which I would prefer before merging. Please ping me (here, or better or on Slack) whenever at least some of the CIs are green again.

P.S. Thanks for following through on the problem and taking it to the finish line. I meant to work on this myself, but I haven't had enough free time lately.

andre2007:
reopen to force a restart of pr voters

andre2007:
@WebFreak001 @PetarKirov 
Thanks for you input.

Unfortunately the pr voters are all broken. See https://forum.dlang.org/post/iqbsiwwzcqylnzznbemx@forum.dlang.org

andre2007:
B

Issue #1775, opened 2019-09-24
`preBuildCommand` does not take `--root` into account


Geod24:
### System information

- **dub version**: DUB version 1.17.0, built on Sep  7 2019
- **OS Platform and distribution**: Mac OSX 10.14
- **compiler version**: DMD64 D Compiler v2.087.1

### Bug Description

When using `dub --root foo/bar`, the cwd is wrong.

### How to reproduce?

Add this to your `dub.json`:
```json
"preBuildCommands": [ "echo `pwd`" ],
```
And run it from another directory.

### Expected Behavior

The `cwd` is the value provided to `--root`

Geod24:
A fix for this would probably need to take https://github.com/dlang/dub/issues/658 into account.

Geod24:
Duplicate of #384

PR #1820, opened 2019-12-18
Fix issue #1775: preBuildCommand does not take root into account

Geod24:
```
This fixes the case reported, by passing the 'Package.path()'
(which is forwarded by 'Dub.rootPath') to runCommand.
```

Since the tests are not so simple to run locally, I set it to originally fail, to be sure it's run by Travis.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -813,7 +813,7 @@ void runBuildCommands(in string[] commands, in Package pack, in Project proj,
 
 	auto depNames = proj.dependencies.map!((a) => a.name).array();
 	storeRecursiveInvokations(env, proj.rootPackage.name ~ depNames);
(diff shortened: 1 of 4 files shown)


Geod24:
So it took me way longer than it should have to realize I forgot to `git add` the "marker" file.
Fixed now, but since `code.dlang.org` is down the build is still going to be red, heh.

Geod24:
The Travis failure seems relevant, so do not merge yet, but testing locally is a constant struggle.

Geod24:
Scratch that, https://github.com/dlang/dub/pull/1821 also fails and it doesn't touch anything that could cause this failure. Feel free to merge.

Geod24:
This one too @thewilsonator ? :D

PR #1821, opened 2019-12-18
Replace `build.sh` and `build.cmd` with `build.d`

Geod24:
```
Since dub needs a D compiler anyway, we might follow in DMD's step and use D for building,
as it is more reliable and expressive than a shell script as features pile up.

This unifies the two scripts to expose the same set of functionalities on all platforms,
with a behavior that is hopefully more sensible
(tradeoff and reasoning are explained through the code / documentation).
It is also completely independent from the current working directory.

Some discrepancies included:
- Name of the variable to override to select the compiler (DC vs DMD);
- Version file: Windows was always regenerating it;
- Version file: Windows would display "unknown" if `dub describe` failed,
  POSIX would use an empty string;
- Compiler detection: POSIX would detect ldmd2 / gdmd if dmd is not present,
  Windows would just fail;

All those functionalities are now unified in a single script.
The old script are kept for compatibility, but forward to the new script,
and can be removed in a few releases.
```

This does not need to be urgently merged, as I don't know *who* uses those scripts. One possibility would be to leave them in place so nothing is broken. My only experience with them is limited to [packaging DUB for Alpine](https://github.com/alpinelinux/aports/blob/27b8dc5bd034f91683012dcb5ad680e64c72c712/testing/dub/APKBUILD#L15) and some early local testing, but nowadays it's much saner to compile DUB with DUB.

Note: If this does not encounter strong opposition, I also plan to go ahead and modernize the testsuite.

### DIFF
--- a/build.d
+++ b/build.d
@@ -1,0 +1,196 @@
+#!/usr/bin/env rdmd
+/*******************************************************************************
+
+    Standalone build script for DUB
+
+    This script can be called from anywhere, as it deduces absolute paths
+    based on the script's placement in the repository.
+
+    Invoking it while making use of all the options would like like this:
+    DMD=ldmd2 DFLAGS="-O -inline" ./build.d my-dub-version
+    Using an environment variable for the version is also supported:
+    DMD=dmd DFLAGS="-w -g" GITVER="1.2.3" ./build.d
+
+    Copyright: D Language Foundation
+    Authors: Mathias 'Geod24' Lang
+    License: MIT
+
+*******************************************************************************/
+module build;
+
+private:
+
+import std.algorithm;
+static import std.file;
+import std.format;
+import std.path;
+import std.process;
+import std.stdio;
+import std.string;
+
+/// Root of the `git` repository
+immutable RootPath = __FILE_FULL_PATH__.dirName;
+/// Path to the version file
+immutable VersionFilePath = RootPath.buildPath("source", "dub", "version_.d");
+/// Path to the file containing the files to be built
+immutable SourceListPath = RootPath.buildPath("build-files.txt");
+/// Path at which the newly built `dub` binary will be
+immutable DubBinPath = RootPath.buildPath("bin", "dub");
+
+// Flags for DMD
+immutable OutputFlag = "-of" ~ DubBinPath;
+immutable IncludeFlag = "-I" ~ RootPath.buildPath("source");
+
+
+/// Entry point
+int main(string[] args)
+{
+    // This does not have a 'proper' CLI interface, as it's only used in
+    // special cases (e.g. package maintainers can use it for bootstrapping),
+    // not for general / everyday usage by newcomer.
+    // So the following is just an heuristic / best effort approach.
+    if (args.length > 2 ||
+        (args.length == 2 && (args[1].canFind("help", "?") || args[1] == "-h")))
+    {
+        writeln("USAGE: ", args[0], " [version]");
+        writeln("  In order to build DUB, a version module must first be generated.");
+        writeln("  If one is already existing, it won't be overriden. " ~
+                "Otherwise this script will use the first argument, if any, " ~
+                "or the GITVER environment variable.");
+        writeln("  If both are empty, `git describe` will be called");
+        writeln("  Build flags can be provided via the `DFLAGS` environment variable.");
+        writeln("  LDC or GDC can be used by setting the `DMD` value to " ~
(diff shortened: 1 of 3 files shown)


s-ludwig:
Unless I'm mistaken, there are now no CI tests on Windows anymore? Seems like someone was a bit overzealous there to get the build green again.

Geod24:
Indeed. The last build was 3 days ago: https://ci.appveyor.com/project/s-ludwig/dub
However I do not see something in the commit history that could have led to this.
Can you check on the AppVeyor side ? Looks like it's registered under your account.

PetarKirov:
AppVeyor has been failing with out of memory for more than 1 month I think.

PetarKirov:
@Geod24 another candidate script to replace is release.sh

wilzbach:
> AppVeyor has been failing with out of memory for more than 1 month I think.

Yep, that's why I disabled it recently. We might want to switch to Azure pipelines like DMD did, but for now having every PR fail because of an unrelated OOM was too much :/

s-ludwig:
Wasn't it just one or two of the pipelines that failed that way? Going without any Windows testing is quite likely to build up more hidden dept. Also, isn't the OOM DMD-32 related and not caused by Travis-CI?

wilzbach:
> Going without any Windows testing is quite likely to build up more hidden dept

Fully agreed, but a Windows CI that always fails while building doesn't help either.

> Wasn't it just one or two of the pipelines that failed that way?

IIRC the first job failed and then Appveyor would cancel all following configurations.

> isn't the OOM DMD-32 related and not caused by Travis-CI?

Yes, but I'm not sure I follow. Travis was/is used to be OSX + Linux only, so no changes there. We haven't enabled their new Windows support yet, but we probably should.

s-ludwig:
Sorry, I meant AppVeyor!

>>  Wasn't it just one or two of the pipelines that failed that way?
>
> IIRC the first job failed and then Appveyor would cancel all following configurations.

Hm, that would be quite different from the behavior that I know from AppVeyor. I would probably have just added the failing configuration(s) as an `allow_failures:` entry to at least keep the issue visible.

Of course it's a rather sad state that DMD is not able to build its own package manager on Windows anymore, especially since it doesn't use any considerable amount of CTFE/template magic (unless someone started to use `CtRegex` or something similar again).

Geod24:
So where are we on this ?

build.d
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/**
+ * Generate the version file describing DUB's version / commit
+ *
+ * Params:
+ *   dubVersion = User provided version file. Can be `null` / empty,
+ *                in which case the existing file (if any) takes precedence,
+ *                or the version is infered with `git describe`.
+ *                A non-empty parameter will always override the existing file.
```
wilzbach: Nit: official D style is without leading stars and it's also about consistency within the file (see too).

Geod24: Right, I'll fix.

On this point: `dub` doesn't follow the official D style *at all*. Be it in the placement of brackets, the usage of tabs, etc... 
What would be the acceptable way to deal with this:
1) Keep it as is (provided the style is not consistent, and under-documented);
2) Gradually improve it when a piece of code is worked on;
3) Improve it all at once;

I'm not fan of 1) because it confuses contributors, and 3 makes `git blame` a bit harder (but not impossible, especially since Github support the "blame before this revision" feature).

Geod24:
> This does not need to be urgently merged

Actually, if we can get this merged before 1.19, that'd be great, as I will be able to rebuild the Alpine Linux package using GDC (once https://gitlab.alpinelinux.org/alpine/aports/merge_requests/1139 is merged).

build.d
Geod24: Fixed style

Geod24:
So I didn't even realize there was a [build-gdc.sh](https://github.com/dlang/dub/blob/800dd8031ef1a640ca95ef874a5cb3b653f4c1fd/build-gdc.sh) script.
Of course it has a different set of issues. Namely, it calls `pkg-config`, uses a different set of flags, and depends on `git` (no way to specify `GITVER`).
In my [Alpine package](https://gitlab.alpinelinux.org/alpine/aports/merge_requests/2484), I ended up using 2 `sed` over `build.sh` over using it, as it's the simpler approach.

Geod24:
Since there seem to be a clear agreement (and no dissent) in favor of going with a D build script, can we move forward ? @thewilsonator ?

PR #1827, opened 2019-12-20
Remove dub.packagesupplier

Geod24:
It has been deprecated since v1.10.0,
more than 18 months ago.


s-ludwig:
Wait, this will require a version bump to 2.0.0 due to SemVer! We should probably have some more meat for such a version jump.

wilzbach:
Argh. I forgot that dub does actual SemVer releases and doesn't just deprecate symbols for a few versions like Phobos. Sorry about that. I guess we will have to revert that one then.

Geod24:
Arg, that is very good to know.
So what is considered public interface ? Because there are a LOT of public symbols in `dub`, and I'm fairly certain many PR just change interface without concern for dub as a library, which is why I went with the remove.

I'm all for going SemVer, but to be worth something, it'll require much more than just "not removing this symbol". Is there any known user of dub as a library that I can inspect ?

s-ludwig:
Whenever I see a pull request I actually take care that no public symbols are changed in a backwards-incompatible manner, but things may of course have slipped through. So generally, indeed everything apart from the `internal` package is part of the public interface, as well as the CLI.

These are some projects that depend on the `dub` package: https://libraries.io/dub/dub/dependents

We also have some internal projects that for example use `DependencyResolver`.

We should probably add a note about this topic to CONTIRBUTING.md to create more awareness of the issue.

Issue #1795, opened 2019-11-04
Wrong Windows binaries in GitHub releases


GoaLitiuM:
All the x86 and x86_64 releases for Windows dating all the way back to 1.14.0 do not run on Windows machines because they are actually Linux binaries.

Geod24:
Confirmed with v1.18.0:
```console
file ~/Downloads/dub.exe
/Users/geod24/Downloads/dub.exe: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.24, BuildID[sha1]=c08f0a5898d0c533ce27c85d7fefafbded071b6c, with debug_info, not stripped
```

saaadel:
same for newer versions:
dub-v1.18.0-windows-x86 
dub-v1.18.0-windows-x86_64 
dub-v1.19.0-beta.1-windows-x86 
dub-v1.19.0-beta.1-windows-x86_64

ZHANITEST:
Can't use `dub`-#1795 and can't build `DUB`-#1657 on Windows now.
Only `v1.15` work now what come with on dmd installer for Windows.

I hope it will be resolved soon 😢

wilzbach:
@ZHANITEST you should be able to use the dub binary bundled/installed with either your DMD or LDC release. This issue only affects the manual installation of dub for Windows.

Geod24:
https://github.com/dlang/dub/releases/tag/v1.20.0

Problem fixed.

PR #1833, opened 2019-12-23
Improve travis-ci.sh

Geod24:
This will probably break horribly, but there's only one way to find out...

### DIFF
--- a/travis-ci.sh
+++ b/travis-ci.sh
@@ -24,6 +24,17 @@ if [ "$COVERAGE" = true ]; then
     dub test --compiler=${DC} -b unittest-cov
     ./build.sh -cov
 
+    wget https://codecov.io/bash -O codecov.sh
+    bash codecov.sh
+else
+    ./build.sh
+    DUB=`pwd`/bin/dub DC=${DC} test/run-unittest.sh
+fi
+
+## Checks that only need to be done once per CI run
+## Here the `COVERAGE` variable is abused for this purpose,
+## as it's only defined once in the whole Travis matrix
+if [ "$COVERAGE" = true ]; then
     # run tests with different compilers
     DUB=`pwd`/bin/dub DC=${DC} test/run-unittest.sh
     deactivate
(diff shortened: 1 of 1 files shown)


Geod24:
Ugh this stage setup is also terrible. If something fails in the first stage and is retriggered (and pass), the second stage never starts so the whole build is marked as 'cancelled' and one has to restart *everything*.

Also the reason why #1795 happens is pretty clear, looking at `.travis-ci.yaml`...
@wilzbach : Since you are our resident Azure pipeline expert, do you think you'll have some time to do your magic before the end of the month ? Otherwise I might take a look at it.

wilzbach:
> @wilzbach : Since you are our resident Azure pipeline expert, do you think you'll have some time to do your magic before the end of the month ? Otherwise I might take a look at it.

Yeah it was on my list already. I will be travelling a lot for the next few days, but I think I can get this before New Year's Eve.

wilzbach:
> Ugh this stage setup is also terrible. If something fails in the first stage and is retriggered (and pass), the second stage never starts so the whole build is marked as 'cancelled' and one has to restart everything

BTW the reason for running the second stage for every PR was that once we enabled the Travis builds which upload to GitHub we weren't sure how reliable they are and it was also the only way to test the PRs. However, a solution would be to enable them only for a PR which is from the dub repository and not with branches from forks.

Geod24:
> Travis builds which upload to GitHub

I don't quite get why we do that ?
Those 6 builds are pretty much useless  right now. The OSX one is useful, but it should be part of test. They also use a different set of flags (release flags) but that could be just 1 entry in the test matrix.

IIUC, the original intent was for this stage to trigger only on tags: https://github.com/dlang/dub/blob/2e9a26acf388dcd54cdc3f4332ee4cfd9b28047a/.travis.yml#L105-L107

Which makes much more sense to me. I couldn't find much in the [original PR](https://github.com/dlang/dub/pull/1433#discussion_r181036384). I guess whatever bug you were working around at the time might have been worked out, since LTO has seen a lot of love over the last 2 years ?

wilzbach:
> I don't quite get why we do that ?

So that the people can download dub binaries (from GitHub releases). This is interesting for example if you want to have a newer version of dub with an older compiler.

> IIUC, the original intent was for this stage to trigger only on tags:

Yes, though as I mentioned this wasn't done as otherwise the scripts couldn't be tested. I fully agree that there should be no need to run them for a normal PR build. I was just explaining the historical reason ... or at least trying to.


> I guess whatever bug you were working around at the time might have been worked out, since LTO has seen a lot of love over the last 2 years ?

Yes.

Geod24:
Okay, I think revisiting the travis config can be done later though. This is good to go as is and already gave me useful insight.

PR #1841, opened 2019-12-27
add flush before reading input for Windows msys2/mingw

ShigekiKarita:
When I use dub with msys2 and mingw, `dub init` shows nothing. This is because the message prompt (e.g., `Package recipe format (sdl/json) [json]:`) does not flush stdout before reading input in that environment (I found powershell and cmd are OK). Therefore, I propose to add a manual `stdout.flush()` before the `readln()`.


Geod24:
I think you could extend it to all platforms.

ShigekiKarita:
I agree. I removed `version (Windows)`.

PR #1842, opened 2019-12-29
Remove leftover and outdated CHANGELOG.md file

Geod24:
```
DUB switched to using fragments from the 'changelog' directory
in PR1455 (v1.10, 2018-07-04), but the CHANGELOG.md was left
in place with the assumption a script to maintain it would
be written.
However, no such script has been written, and there doesn't
need to be a need for one either.
```


wilzbach:
Might be nice to keep for historical purposes and just reference that changes since 1.10 are part of the official dlang changelog?

Geod24:
> Might be nice to keep for historical purposes and just reference that changes since 1.10 are part of the official dlang changelog?

You could say the same for any historical artifact. It will still be available for historical reference - just `git checkout v1.10.0` and you'll get the changelog, like you do for other dlang project.

PR #1845, opened 2020-01-02
Fix issue 1277: Missing dub.json causes infinite loop

Geod24:
For some reason, having a 'source/app.d' was recognized as being in a package.
This would load an empty package with no manifest and send the dependency resolver
in an infinite loop.
Remove the special detection of 'source/app.d' and rely only on manifest.
Another option would have been to detect 'source/app.d' as a single-file package,
but that is a bit too much magic (the user can just 'dub run --single source/app.d').


Geod24:
Self merging since this has been approved (and is quite trivial)

PR #1847, opened 2020-01-02
Add zsh completion script

Geod24:
Currently supports the main dub commands: build, init, run, test, clean, dustmite...
Reads the local packages from dub whenever possible.
A few improvements to dub will be necessary to make it completely useful (e.g. display configurations).

### DIFF
--- a/scripts/zsh-completion/_dub
+++ b/scripts/zsh-completion/_dub
@@ -1,0 +1,265 @@
+#compdef dub
+
+# Useful help:
+# https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org
+# http://zsh.sourceforge.net/Doc/Release/Completion-System.html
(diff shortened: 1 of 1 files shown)


Geod24:
Guess we don't have that many people that can review a zsh completion script. Self merging because there's no way this breaks anything.

PetarKirov:
@Geod24 Can we generate shell completion scripts automatically? AFAIR we already generate CLI help, DDoc for dlang.org and man pages out of dmd's CLI.

Geod24:
For posterity (discussed on Slack): I would love to, but it seemed like a faster way to just fo with a manual script at first. Also the real value will be in completing things based on the user / project settings (e.g. completing `-c` / `--config`), for which we first need to add support to dub.

TL;DR: Ideally yes, but not soon (from me).

wilzbach:
We already generate the man pages for dub automatically - https://github.com/dlang/dub/blob/master/scripts/man/gen_man.d.

This might help.

Issue #1822, opened 2019-12-19
Released DUB binaries should be built with the latest compiler


Geod24:
I was playing with `install.sh` and realized that the binaries we distribute are different from the one packaged with DMD, which is obviously problematic for reporting bug.

Additionally, since the compiler used is < 2.087.1, the binaries are simply broken on OSX:

```console
$ ~/dlang/dub-1.18.0/dub --version
dyld: lazy symbol binding failed: Symbol not found: _dyld_enumerate_tlv_storage
  Referenced from: /Users/geod24/dlang/dub-1.18.0/dub
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: _dyld_enumerate_tlv_storage
  Referenced from: /Users/geod24/dlang/dub-1.18.0/dub
  Expected in: /usr/lib/libSystem.B.dylib

[3]    44459 abort      ~/dlang/dub-1.18.0/dub --version
$ ~/dlang/dmd-2.089.1/osx/bin/dub --version
DUB version 1.18.0, built on Dec 14 2019
```

Panke:
What's the state of this?

Geod24:
Fixed. We'll need to do things like https://github.com/dlang/dub/pull/1879 from time to time but at least it works on OSX now...

PR #1849, opened 2020-01-03
Use build.d everywhere, fix windows build

Geod24:
It seems easier to forward arguments from the build script this way,
and it's much more common to change the build arguments than the version.


Geod24:
Should fix https://github.com/dlang/dub/issues/1795

Geod24:
And https://github.com/dlang/dub/issues/1822

PR #1865, opened 2020-01-21
VisualD project generation now account for the debugVersions key (deb…

p0nce:
…ug identifiers)


Geod24:
Is there any way to test this ?  And I assume you are using it ?

p0nce:
Hello, yes I've tested it locally before submitting :

image

p0nce:
Seems the cmake generator emit those, it's just the VisualD generator that doesn't.

Geod24:
> yes I've tested it locally before submitting

I mean is there a way to have integration tests ?

Geod24:
@p0nce : If the answer is no, I'm still happy to go ahead with this change :)
Just wanted to avoid any future regression.

PR #1868, opened 2020-01-26
ldc: support --arch=aarch64 and powerpc64

myfreeweb:
Because Meson calls e.g. `dub describe something --arch aarch64`.

I gueeeeessss we could teach Meson to use the full triple (#1755), but this way works with `x86_64`, why shouldn't it work with other architectures.


Geod24:
Can you amend / push ?

myfreeweb:
@Geod24 amend with what? you mean rebase?

Geod24:
Yes, in order to retrigger the CI.

wilzbach:
The failure just happens because the registry is down. It can safely be ignored.

Geod24:
Rebased for ya.

PR #1869, opened 2020-01-29
Emit the Posix signal name on app run failure

AndrejMitrovic:
I haven't yet figured out how to add this to the test-suite.

So here's a screenshot of it in action instead:

Capture


source/dub/generators/build.d
```diff
@@ -546,7 +546,9 @@ class BuildGenerator : ProjectGenerator {
 				auto result = prg_pid.wait();
 				settings.targetExitStatus = result;
 				runPostRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
-				enforce(result == 0, "Program exited with code "~to!string(result));
+				string error = "Program exited with code "~to!string(result);
+				version (Posix) error ~= " (" ~ signalToName(-result) ~ ")";
+				enforce(result == 0, error);
```
AndrejMitrovic: :facepalm: I just realized I am too eagerly allocating the message. I'll fix it. ._.

AndrejMitrovic:
Update: fixed the brace styling and the over-eager string allocation.

source/dub/generators/build.d
```diff
@@ -29,7 +29,7 @@ import std.encoding : sanitize;
 
 string getObjSuffix(in ref BuildPlatform platform)
 {
-    return platform.platform.canFind("windows") ? ".obj" : ".o";
+	return platform.platform.canFind("windows") ? ".obj" : ".o";
```
Geod24: ?

source/dub/generators/build.d
```diff
@@ -645,3 +650,40 @@ unittest { // issue #1235 - pass no library files to compiler command line when
...
+		"SIGUSR2",
+		"SIGCHLD",
+		"SIGCONT",
+		"SIGSTOP",
+		"SIGTSTP",
+		"SIGTTIN",
+		"SIGTTOU",
+	];
+
+	final switch (signal)
```
AndrejMitrovic: Hmm.. maybe this should be a switch with a default, I might have missed some signals here. Unfortunately I couldn't find a simple array/tuple of all signals anywhere in Druntime, they are all scattered around.

source/dub/generators/build.d
```diff
@@ -645,3 +650,40 @@ unittest { // issue #1235 - pass no library files to compiler command line when
 	auto gen = new BuildGenerator(prj);
 	gen.generate(settings);
 }
+
+version (Posix)
+private string signalToName (int signal) {
+	import core.stdc.signal;
+	import core.sys.posix.signal;
+
+	enum sig_types = [
```
Vild: This function could be improved by look into what `core.stdc.signal` and `core.sys.posix.signal` exports instead of just hard-coding the list.

For example something this seems to work. But the codes layout / readability could be improved.
```d
string translate(int signal) {
    import core.sys.posix.signal;
    import std.algorithm : filter, sort;
    import std.array : array;
    import std.string : startsWith, indexOf;

    enum allSignals = [__traits(allMembers, core.sys.posix.signal), __traits(allMembers, core.stdc.signal)]
        .filter!(x => x.startsWith("SIG") && x.indexOf("_") < 0 )
        .array;

    switch (signal) {
        static foreach (sig; allSignals)
            static if (__traits(isIntegral, mixin(sig))) // SIGRTMIN / SIGRTMAX are functions on some systems, so ignore those.
                case mixin(sig):
                    return sig;
        default:
        	return "<Unknown signal>";
    }
}
```

source/dub/generators/build.d
```diff
@@ -546,7 +546,12 @@ class BuildGenerator : ProjectGenerator {
 				auto result = prg_pid.wait();
 				settings.targetExitStatus = result;
 				runPostRunCommands(m_project.rootPackage, m_project, settings, buildsettings);
-				enforce(result == 0, "Program exited with code "~to!string(result));
+				string error() {
+					string error = "Program exited with code "~to!string(result);
+					version (Posix) error ~= " (" ~ signalToName(-result) ~ ")";
+					return error;
```
Vild: `core.sys.posix.string.strsignal` should probably also be added to the output, as that gives a text explanation of the signal.
https://linux.die.net/man/3/strsignal

source/dub/generators/build.d
AndrejMitrovic: It seems the codebase uses tabs. So I made it consistent.

AndrejMitrovic:
Right, `strsignal` is somewhat useful, but it really isn't much:

```
SIGINT desc: Interrupt
SIGQUIT desc: Quit
SIGILL desc: Illegal instruction
SIGABRT desc: Aborted
SIGFPE desc: Floating point exception
SIGKILL desc: Killed
SIGSEGV desc: Segmentation fault
SIGPIPE desc: Broken pipe
SIGALRM desc: Alarm clock
SIGTERM desc: Terminated
SIGUSR1 desc: User defined signal 1
SIGUSR2 desc: User defined signal 2
SIGCHLD desc: Child exited
SIGCONT desc: Continued
SIGSTOP desc: Stopped (signal)
SIGTSTP desc: Stopped
SIGTTIN desc: Stopped (tty input)
SIGTTOU desc: Stopped (tty output)
```

I could still add it I suppose.

source/dub/generators/build.d
AndrejMitrovic: Hehe, a bit extreme. We can do something like this for now, but I hope we'll refactor those bindings in Druntime at some point. :)

PR #1870, opened 2020-01-30
Add dub.bytecraft.nl as mirror and remove herokuapp

wilzbach:
CC @skoppe - would you be okay with your mirror being added as a backup to the dub client?


skoppe:
That would be great. But be aware that usage has been very low, there are no automated tests and no one can modify it besides me (bus factor == 1). On the other hand, it has good uptime because it is running on cloudflare's global edge network and no one can change it besides me :)

But only using it when code.dlang.org is down means it can't fill its cache. It would thus make more sense to put it before code.dlang.org (but I wouldn't do that at this point).

At the end of the day, more people need to use it directly, since that will keep its cache up-to-date.

So yeah, all considered, I think having it as an backup option would be good.

Long term, I really think we should fix/redesign dub's infra and run it on the cloud/edge.

But why remove the heroku one? Doesn't it work?

Geod24:
@wilzbach : How does it work when it comes to package management ? Mirrors are read only ?
Wouldn't it be better to use DNS round robin for mirrors (when the request comes from dub), and fallback URLs when the DNS is down ?

wilzbach:
> But be aware that usage has been very low, there are no automated tests and no one can modify it besides me (bus factor == 1). On the other hand, it has good uptime because it is running on cloudflare's global edge network and no one can change it besides me :)

That's fine. There are also no strings attached. The other two existing mirrors aren't doing well either.

> But only using it when code.dlang.org is down means it can't fill its cache. It would thus make more sense to put it before code.dlang.org (but I wouldn't do that at this point).

Oh I see. Then it doesn't help much.
How hard would it be to periodically fill the cache with the full package dump?

```
curl https://code.dlang.org/api/packages/dump | gunzip > mirror.json 
```

> But why remove the heroku one? Doesn't it work?

Yes.

> How does it work when it comes to package management ? Mirrors are read only ?

Yes.

> Wouldn't it be better to use DNS round robin for mirrors (when the request comes from dub), and fallback URLs when the DNS is down ?

Yes, in theory, but the mirrors have a worse track record than the main code.dlang.org instance. 

> Long term, I really think we should fix/redesign dub's infra and run it on the cloud/edge.

Fully agreed.
Another option that could help a lot would be to send the full package index to the client, s.t. they basically only need the registry to update their package cache every now and then.

skoppe:
> > But only using it when code.dlang.org is down means it can't fill its cache. It would thus make more sense to put it before code.dlang.org (but I wouldn't do that at this point).
> 
> Oh I see. Then it doesn't help much.
> How hard would it be to periodically fill the cache with the full package dump?

There is more logic in the registry than just piping the db to the client. For instance the getPackageInfosRecursive function that the dub client depends on for its package dependency resolution (so it knows what transitive dependencies it needs to fetch).

My script doesn't do any of that. Just caching responses + redirecting to gitlab/github.

It is on purpose, to keep stuff simple.

> > Long term, I really think we should fix/redesign dub's infra and run it on the cloud/edge.
> 
> Fully agreed.
> Another option that could help a lot would be to send the full package index to the client, s.t. they basically only need the registry to update their package cache every now and then.

Yes, and the logical conclusion of that would be to host the full package index on git, and only keep the web interface around for registering packages, requesting permissions to create webhooks to gitlab/github change-feed and hosting said webhook.

skoppe:
@wilzbach I think you can go ahead with this one.

I have updated my cloudflare dub-registry to be standalone.

I made [skoppe/dub-packages-index](https://github.com/skoppe/dub-packages-index) that contains the latest dub package metadata. It is updated by a 15min cronjob running on github actions (at [skoppe/dub-packages-indexer](https://github.com/skoppe/dub-packages-indexer)) that also updates the cloudflare dub-registry KV-store with latest packages.

Geod24:
@skoppe : Can you rebase ? There's a conflict and one of the CI is stuck.

skoppe:
> @skoppe : Can you rebase ? There's a conflict and one of the CI is stuck.

It is not my PR, and since I am not part of the owners I don't think I can.

ping @wilzbach

thewilsonator:
Done.

s-ludwig:
It should probably be added after codemirror.dlang.org.

PR #1873, opened 2020-02-07
Add: recursive option on test command

Sobaya007:


thewilsonator:
Thanks! could you please add a changelog entry for this.

wilzbach:
... and test ;-)

andre2007:
Yes, the test is important. I am not sure wheter it works as expected because the dependencies are compiled not in unittest mode when running dub test on a dub package.

Also what is the expected behavior if the test of a dependency fails, will the unittests for the other dependencies still be executed?

Sobaya007:
How can I add test?
Write some unittest block or putting any shell script in test directory?

wilzbach:
> some unittest block or putting any shell script in test directory?

Yes in `test` most of the existing tests can be found. They are mostly shell scripts. You could check for the tests for `dub add` if you need help on how to do it exactly.

andre2007:
If I am not wrong,  there is a *.sh file missing which call dub test -r for the test package you added.

Sobaya007:
According to [run-unittest.sh](https://github.com/dlang/dub/blob/45b00b4d14600974f2bc81551001dfb30bf79251/test/run-unittest.sh#L50), just adding package directory into test will be OK.
Actually I ran [travis.sh](https://github.com/dlang/dub/blob/master/scripts/ci/travis.sh) on my PC and confirmed the test is executed.

andre2007:
I can't see how your tests triggers your new recursive functionality. Where do you pass the --recursive argument?

Sobaya007:
Oh, sorry. It's my mistake.

source/dub/commandline.d
```diff
@@ -1009,6 +1010,9 @@ class TestCommand : PackageBuildCommand {
 		args.getopt("f|force", &m_force, [
 			"Forces a recompilation even if the target is up to date"
 		]);
+		args.getopt("r|recursive", &m_recursive, [
+			"Runs test on all packages including subpackages."
```
shoo: Unfortunately, this command-line help seems different from fact.
This option recursive tests for dependencies, not subpackages.
(I'd rather test the subpackages recursive.)

andre2007:
A proposal: the test libraries could output in unittest mode a text to stdout. In your test bash script you can grep for this text. Or even better call one time dub test with -r and one time without -r and compare the output.

A sample how to to grep and compare dub output in bash you can see here https://github.com/dlang/dub/pull/1811/files

andre2007:
Thanks. I thought of s.th. like this in the test script:
Pseudo code
```
dub test root_package -r | output contains root_pack and liba and libb and libc
dub test root_package | output contains only root_pack
```

Sobaya007:
Anything else?

andre2007:
Only a nitpick, I am not sure whether 4 additional git ignore files are needed. Also please merge the 11 commits into 1 commit.

andre2007:
Otherwise looks good for me. Thanks a lot for your effort.

Sobaya007:
OK?

PR #1876, opened 2020-02-12
Add --recurse-subpackages option on test command

shoo:
Related: #1873 
I'd rather test subpackages recursively.

This will conflict with #1873. Therefore, further editing is required after merging #1873.
Maybe it needs to have a consistent command name as well.


andre2007:
I am not sure whether I understand the new functionality. The term `recurse` confuses me because I thought a dub root package can only have a flat list of sub packages, no hierarchy.

Also if I am not completely wrong,  by executing `dub test` on a root package, the unittests of sub package will already run today.

shoo:
>I am not sure whether I understand the new functionality. The term recurse confuses me because I thought a dub root package can only have a flat list of sub packages, no hierarchy.

I propose `--all-subpackages` as the next idea.

>Also if I am not completely wrong, by executing dub test on a root package, the unittests of sub packages will already run today.

Unfortunately, it doesn't seem to be the case. The test I wrote shows that.
https://github.com/shoo/dub/blob/7891ab47fd98df0b385ef7800069248dec4df770/test/test-recurse-subpackages.sh#L28

If the root package depends on a subpackage, and build with `--build-mode=allAtOnce`, the subpackage will be tested as well. And also, in that case, it can test with #1873.
But, the root package does not always depend on subpackages.

---
PS. It doesn't matter much about this pull request, but please let me know for later study. Why are subpackages flat? Is there a design reason for not having a nested structure like sub-sub-package?

PR #1887, opened 2020-03-01
Add codemirror.dlang.org as the primary fallback registry.

s-ludwig:
I've migrated the registry to a dedicated server with a decent amount of RAM (64G), so that the high memory usage of the registry is not a problem anymore. The old server is now running the registry in mirror mode (which does not exhibit the memory problematic) and is available at https://codemirror.dlang.org/ as the official fallback.

This should finally provide virtually 100% uptime, especially with #1870 also present.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -61,6 +61,7 @@ deprecated("use defaultRegistryURLs") enum defaultRegistryURL = defaultRegistryU
 /// The URL to the official package registry and it's default fallback registries.
 static immutable string[] defaultRegistryURLs = [
 	"https://code.dlang.org/",
+	"https://codemirror.dlang.org/",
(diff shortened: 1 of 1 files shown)


s-ludwig:
Thinking about it, let me re-target this at stable, so that it gets into 1.20/2.090.0.

wilzbach:
> Thinking about it, let me re-target this at stable, so that it gets into 1.20/2.090.0.

Thought the same. I just pushed this.

s-ludwig:
Or rather redo it for stable then ;)

s-ludwig:
Okay, great, thanks!

wilzbach:
> Or rather redo it for stable then ;)

It's easier to merge stable back into master ;-)

https://github.com/dlang/dub/pull/1888

PR #1890, opened 2020-03-07
Default registry: add Skoppe's, remove Herokuapp

Geod24:
Replacement for https://github.com/dlang/dub/pull/1870 because the CI seems to be AWOL.
@s-ludwig comment addressed.


Geod24:
Ah, AppVeyor has been disabled on master but not on stable. See https://github.com/dlang/dub/pull/1821#issuecomment-567873383 . So we just need @wilzbach to remove it from the required list, I guess.

s-ludwig:
I've changed the stable CI settings to match master now. Not really a good state to be in, in general, but I guess for now it will do.

PR #1894, opened 2020-03-11
fix #1556: dub --cache=local fetch + build

### LINKED ISSUE #1556: build command ignores '--cache'
### System information

- **dub version**: 1.11.0+13-gb837cee
- **OS Platform and distribution**: GNU/Linux 4.18, Ubuntu 18.04
- **compiler version**: doesn't matter

### Bug Description
The build command (`dub build`) ignores the package cache option specified by `--cache`.

### How to reproduce?
Double-check that the used package does **not exist in the user cache**, otherwise you'll get fooled by DUB (for simplicity, this examples uses the package `_` that acutally exists):
```sh
dub --cache=local fetch _
dub --cache=local build _
```
```
Failed to find a package named '_'.
```

### Expected Behavior
Build that package THAT IS ACTUALLY THERE from the SPECIFIED local cache.

Panke:
Building a package that was locally cached failed, because
the dependencies whould be fetched at the wrong destination.

For example if one tries to build the package 'optional':

$ dub --cache=local fetch optional
-> optional is now at ./.dub/package/optional
$ dub --cache=local build optional
-> fails because the dependency bolts was fetched to
./.dub/package/optional/.dub/package/bolts.

This is fixed by not overwriting the root path if dub builds
a package from the cache.


Geod24:
Seems to do what it is intended to, but fails because `dub` tests with a wide range of compilers, and `bolts` does not support 2.077.
Look at other tests to see how they're doing with this. Ideally we should avoid fetching from `code.dlang.org` for our tests, because it usually makes things very brittle.

Panke:
I'll take a look.

Panke:
I changed the test to use local packages similar to other tests.

Issue #1889, opened 2020-03-04
dub generate bazel


Panke:
Bazel [1] already has rudimentary D support, but an easy way to use third party packages is missing. 

I want to improve the situation in a way that allows easy integration of dub packages into a bazel workspace. I currently have two ideas. First is something along the lines of  `dub describe | dub2bazel` to automatically create the project files needed by bazel and the second is `dub generate bazel`. 

My hope would be that one could just `dub generate bazel` and check the resulting files in (preferably upstream) to build and use the package with bazel. 

Anyhow I am neither proficient in dub nor bazel and wanted to ask what you think is the best way to integrate the two?

s-ludwig:
Unfortunately I'm not familiar with Bazel, but judging by the first impression, the `dub generate` route sounds like a good fit. Writing a generator is pretty straight forward, the one in dub/generators/cmake.d could be used as a starting point, although it doesn't handle all possible build settings right now.

PR #1898, opened 2020-03-14
Fetch with dependencies

Panke:


Panke:
I am needing this or something similar for bazel support (#1889)

Geod24:
Needs a rebase. And for the name, `--recursive` is more common (e.g. it's what `git` and many CLI tools use).

Panke:
> Needs a rebase. And for the name, `--recursive` is more common (e.g. it's what `git` and many CLI tools use).

--recursive was my first thought, but we got some use of recursive that is quite dissimilar to this one, like https://github.com/dlang/dub/pull/1873 which means sub packages and not dependencies. 

I couldn't care less for the name though. If no one else chimes in, it will be recursive.

Geod24:
I think for `dub fetch`, `--recursive` is pretty unambiguous. You cannot fetch sub-packages independently.

Geod24:
Also need a release notes entry, this is a change that might be interesting to a few people (e.g. in build scripts).

WebFreak001:
should we make this a switch? imo this should be the default

workflow for someone with data caps:

connect to internet
fetch all required dependencies
disconnect from internet
build stuff locally -- oops, dependencies of fetched dependencies aren't there by default

Geod24:
We can't have self-reference / circular reference, can we ? That would be one argument against it.
Having it the default for `dub fetch` makes sense. IDK about `--cache=local` though.

WebFreak001:
definitely needs a test for recursive dependencies here then

Panke:
> We can't have self-reference / circular reference, can we ? That would be one argument against it.
> Having it the default for `dub fetch` makes sense. IDK about `--cache=local` though.

I don't think --cache=local should be a default.

Geod24:
> I don't think --cache=local should be a default.

Oh I phrased my message very poorly. I meant "IDK if having it the default for cache local makes as much sense". But after reviewing again, I think it does too.

> what do we call the switch?

I'd say, in doubt, do what `git` does. So `--shallow`.

Panke:
What we could do is: download everything that is not in the --cache=<thiscache>

Panke:
This turns out to be more annoying than I thought. This works only for dependencies that are dependencies of the default configuration, but won't fetch anything for e.g. vibe-d.

I could use some pointers on how to get a list of transitive dependencies.

WebFreak001:
try checking the selections and/or generating them for the fetched package, they are recursive

andre2007:
When I execute following commands:
```
dub fetch vibe-d
dub describe vibe-d
```

vibe-d and all dependencies are downloaded. Therefore `dub describe vibe-d` does the fetch with dependencies. Maybe you could check how it works here and reuse the already available functionality?

andre2007:
Related https://github.com/dlang/dub/issues/1930

Geod24:
As far as I can tell this was done by #2786

PR #1901, opened 2020-03-18
Style fixes and refactorings

WebFreak001:
I recommend looking at the commits individually


test/issue564-invalid-upgrade-dependency/a-1.0.0/source/a.d
```diff
@@ -1,3 +1,3 @@
-void test()
-{
+void test()
+{
 }
```
Geod24, line 3: Want to address those missing newlines while at it ?

Geod24:
Could you merge the last commit into 1a8ad57 ?

source/dub/packagesuppliers/packagesupplierlist.d
```diff
@@ -0,0 +1,131 @@
+module dub.packagesuppliers.packagesupplierlist;
```
Geod24, line 1: Module DDOC please

Geod24: Also it's already in the `packagesuppliers` package, which is a mouthful, so maybe just `list` ?
I'm not so strongly opinionated on this, just find the repetition pointless.

source/dub/packagesuppliers/packagesupplierlist.d
```diff
@@ -0,0 +1,131 @@
...
+	on multiple package suppliers safely.
+
+	Provides utility iterator to wrap common try-catch loops with providers.
+
+	Contains common patterns used with multiple package suppliers.
+*/
+struct PackageSupplierList {
+	private {
+		PackageSupplier[] m_suppliers;
+	}
```
Geod24, line 19: Could we perhaps use the D style for new code ?

source/dub/packagesuppliers/packagesupplierlist.d
```diff
@@ -0,0 +1,131 @@
...
+
+	this(PackageSupplier[] suppliers)
```
Geod24, line 21: I guess there is no hope for `@safe` or any of those attributes in `dub` yet ?

source/dub/packagesuppliers/packagesupplierlist.d
```diff
@@ -0,0 +1,131 @@
...
+	auto tryAllOrWarnRaw(lazy string doingWhat)
+	{
+		return tryAllOrCall(delegate(ref ps, e) {
+			logWarn("%s using %s: %s", doingWhat, ps.description, e.msg);
+			logDebug("Full error: %s", e.toString().sanitize);
+		});
+	}
+
+	/// Same as $(LREF tryAllOrWarn) but calls an error action instead of logging.
+	auto tryAllOrCall(void delegate(ref PackageSupplier, Exception) errorCallback)
```
Geod24, line 73:
```suggestion
	auto tryAllOrCall(scope void delegate(ref PackageSupplier, Exception) errorCallback)
```

Geod24: Also might want to make that `Exception` `scope`.

source/dub/packagesuppliers/packagesupplierlist.d
```diff
@@ -0,0 +1,131 @@
...
+	{
+		foreach (ps; m_suppliers) {
+			try {
+				auto pinfo = ps.fetchPackageRecipe(package_id, dep, pre_release);
+				if (pinfo.type == Json.Type.null_)
+					continue;
+				supplier = ps;
+				return pinfo;
+			} catch(Exception e) {
+				logWarn("Package metadata for %s %s could not be downloaded from %s: %s", package_id, dep, ps.description, e.msg);
```
Geod24, line 123: It's not necessarily downloaded though, is it ?

WebFreak001: hm true, looking at the source code now it is already mentioned too.

I think I had some edge case where it didn't throw an intentional exception but rather previously something like
```
Package foo not found for registry at https://code.dlang.org: Operation timeout on handle 0x1234
```
or something like that.

I think I would keep the change to %s %s (package_id, dep) because also showing the dependency version in all error messages, even though it might already be in the error message, is a little bit more friendly.

Or maybe just throw different types of exceptions?

WebFreak001:
overscoped PR, base branch too old now - best to redo this PR in smaller steps.

PR #1905, opened 2020-03-22
fix #1867 - Add lowmem for settings

### LINKED ISSUE #1867: Add commandline option -lowmem
The `-lowmem` option is effective when building in an environment with low memory capacity.
When specifying this using dub, it is necessary to describe it in `dflags` of dub.sdl/.json which is a configuration file now.
However, since this option is not dependent on the project itself, but an option that needs to be switched depending on the building environment, it should not be described in a configuration file that applies regardless of the environment.
To solve this, I propose to allow the `-lowmem` option to be specified in the command line argument of dub itself.

Thanks!

shoo:
I first suggested an addition to the command-line options at #1867. But, I finally agreed to add to the [settings file](https://dub.pm/settings). So, this PR does not add any command-line options tentatively.
I think it is possible to add it if needed, but what do you think about adding command-line options?

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -998,6 +998,7 @@ abstract class PackageBuildCommand : Command {
 	{
 		if (!m_compilerName.length) m_compilerName = dub.defaultCompiler;
 		if (!m_arch.length) m_arch = dub.defaultArchitecture;
+		if (dub.defaultLowMemory) m_buildSettings.options |= BuildOption.lowmem;
 		m_compiler = getCompiler(m_compilerName);
 		m_buildPlatform = m_compiler.determinePlatform(m_buildSettings, m_compilerName, m_arch);
 		m_buildSettings.addDebugVersions(m_debugVersions);
@@ -1592,6 +1593,7 @@ class DescribeCommand : PackageBuildCommand {
 		settings.buildType = m_buildType;
 		settings.compiler = m_compiler;
 		settings.filterVersions = m_filterVersions;
+		settings.buildSettings.options |= m_buildSettings.options & BuildOption.lowmem;
 
 		if (m_importPaths) { m_data = ["import-paths"]; m_dataList = true; }
 		else if (m_stringImportPaths) { m_data = ["string-import-paths"]; m_dataList = true; }
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -132,6 +132,7 @@ class Dub {
 		NativePath m_overrideSearchPath;
 		string m_defaultCompiler;
 		string m_defaultArchitecture;
+		bool m_defaultLowMemory;
 	}
 
 	/** The default placement location of fetched packages.
@@ -288,6 +289,7 @@ class Dub {
 		determineDefaultCompiler();
 
 		m_defaultArchitecture = m_config.defaultArchitecture;
+		m_defaultLowMemory = m_config.defaultLowMemory;
 	}
 
 	@property void dryRun(bool v) { m_dryRun = v; }
@@ -331,6 +333,13 @@ class Dub {
 		configuration file will be used. Otherwise null will be returned.
 	*/
 	@property string defaultArchitecture() const { return m_defaultArchitecture; }
+
+	/** Returns the default low memory option to use for building D code.
+
+		If set, the "defaultLowMemory" field of the DUB user or system
+		configuration file will be used. Otherwise false will be returned.
+	*/
+	@property bool defaultLowMemory() const { return m_defaultLowMemory; }
 
 	/** Loads the package that resides within the configured `rootPath`.
 	*/
@@ -765,6 +774,7 @@ class Dub {
 		settings.compiler = getCompiler(compiler_binary);
 		settings.platform = settings.compiler.determinePlatform(settings.buildSettings, compiler_binary, m_defaultArchitecture);
 		settings.buildType = "debug";
+		if (m_defaultLowMemory) settings.buildSettings.options |= BuildOption.lowmem;
 		settings.run = true;
 
 		foreach (dependencyPackage; m_project.dependencies)
@@ -1261,6 +1271,7 @@ class Dub {
 		settings.buildType = "debug";
 		settings.run = true;
 		settings.runArgs = runArgs;
+		if (m_defaultLowMemory) settings.buildSettings.options |= BuildOption.lowmem;
 		initSubPackage.recipe.buildSettings.workingDirectory = path.toNativeString();
 		template_dub.generateProject("build", settings);
 	}
@@ -1331,6 +1342,7 @@ class Dub {
 		settings.compiler = getCompiler(compiler_binary); // TODO: not using --compiler ???
 		settings.platform = settings.compiler.determinePlatform(settings.buildSettings, compiler_binary, m_defaultArchitecture);
 		settings.buildType = "debug";
+		if (m_defaultLowMemory) settings.buildSettings.options |= BuildOption.lowmem;
 		settings.run = true;
 
 		auto filterargs = m_project.rootPackage.recipe.ddoxFilterArgs.dup;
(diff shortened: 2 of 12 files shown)


source/dub/generators/targetdescription.d
```diff
@@ -47,6 +47,7 @@ class TargetDescriptionGenerator : ProjectGenerator {
 			d.buildSettings = ti.buildSettings.dup;
 			d.dependencies = ti.dependencies.dup;
 			d.linkDependencies = ti.linkDependencies.dup;
+			if (settings.lowmem) d.buildSettings.options |= BuildOption.lowmem;
```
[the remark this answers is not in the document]
shoo: It looks like it will, but it actually isn't.
I looked into the reasons why it isn't.

In L47, the `target` is used.
https://github.com/shoo/dub/blob/a0090af56f062d67fc949aa6beee01198317d4b0/source/dub/generators/targetdescription.d#L47

The settings of `target` look as if they were passed from `GeneratorSettings.buildSettings` on this line:
https://github.com/shoo/dub/blob/a0090af56f062d67fc949aa6beee01198317d4b0/source/dub/generators/generator.d#L107

If I look inside the function, the following line shows the addition of settings:
https://github.com/shoo/dub/blob/a0090af56f062d67fc949aa6beee01198317d4b0/source/dub/generators/generator.d#L560

However, there is no code in this function to add `options` (and `requirements`).
https://github.com/shoo/dub/blob/a0090af56f062d67fc949aa6beee01198317d4b0/source/dub/compilers/buildsettings.d#L67-L90

If I add the code to add the options here, I get the expected behavior.

```d
		...
		addPreRunCommands(bs.preRunCommands);
		addPostRunCommands(bs.postRunCommands);
		addRequirements(bs.requirements);
		addOptions(bs.options);
```

However, I can't speculate on the extent of the impact of making this change.
Is this change reasonable?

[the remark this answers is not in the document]
WebFreak001: hm no I don't think it's a good place in the add function. I think it should work if it's added into https://github.com/dlang/dub/blob/800dd8031ef1a640ca95ef874a5cb3b653f4c1fd/source/dub/project.d#L1148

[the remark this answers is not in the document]
shoo: The role of `processVar` seems to be to expand the variables contained in dub.sdl/dub.json, which exist in each project by using GeneratorSettings, etc.
The purpose of this PR, to transfer the lowmem option included in GeneratorSettings to BuildSettings, seems to be contrary to the purpose of this function.

Another idea is calling `BuildSettings.addOptions` immediately after the `processVar` execution. What do you think of this idea?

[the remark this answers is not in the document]
WebFreak001: hm that's true but the processVars function already adds options from other, which doesn't have anything to do with that, so I think it would be fine.

cc @s-ludwig

apz28:
Why not creating a pass-through (as-is) DMD command flag? it will have advantage of DUB does not need to maintain when DMD add/remove/change the syntax.  Let user setup whatever the flag they want and DMD will complain if it not follow its' rule

WebFreak001:
@apz28 this is an addition to the dub settings file, which is for example in ~/.dub/settings.json and is generally not tied to any project. Having a lowmem setting makes a lot of sense there, for example on a raspberry pi or other low spec device you always want to compile with lowmem then

test/issue1867-lowmem.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+DIR=$(dirname "${BASH_SOURCE[0]}")
+
+if ! { ${DUB} build --root ${DIR}/issue1867-lowmem -v -f 2>&1 || true; } | grep -cF " -lowmem " > /dev/null; then
+    die $LINENO 'DUB build with lowmem did not find -lowmem option.'
```
kinke: This is not sufficient because of https://issues.dlang.org/show_bug.cgi?id=20699.

shoo: Thank you for the information !
This PR of passing the `-lowmem` from dub to dmd and the Issue20699 that the `-lowmem` in the passed response file is not actually used seems to be two different things.
So, I choose not to do anything to this point.
And also, it seems to me that Issue20699 is a dmd issue, but if it has to be fixed in the dub, a separate Issue should be created. Well, I'm hoping that the dmd improvements will solve this problem.

kinke: Sure, I just meant to let you know that this won't work with DMD.

shoo:
I've made changes in response to review points from @WebFreak001.
However, I was hesitant to change the option of BuildSettings in proccecssVars, so I was careful to minimize the impact.
Also, please note that in the case of the describe command, the `buildSettings` are not currently taken over from the member variables. Only in regards to lowmem, I have changed it to take over.
These should be refactored in the future to ensure that it is handled in more appropriate functions.

source/dub/commandline.d
```diff
@@ -1232,6 +1233,7 @@ class DescribeCommand : PackageBuildCommand {
 		settings.buildType = m_buildType;
 		settings.compiler = m_compiler;
 		settings.filterVersions = m_filterVersions;
+		settings.buildSettings.options |= m_buildSettings.options & BuildOption.lowmem;
```
WebFreak001: What exactly does this change in the describe command? I would imagine because m_buildSettings isn't used at all, that it would miss other stuff too.

source/dub/dub.d
```diff
@@ -1319,6 +1330,7 @@ class Dub {
 		settings.compiler = getCompiler(compiler_binary); // TODO: not using --compiler ???
 		settings.platform = settings.compiler.determinePlatform(settings.buildSettings, compiler_binary, m_defaultArchitecture);
 		settings.buildType = "debug";
+		if (m_defaultLowMemory) settings.buildSettings.options |= BuildOption.lowmem;
 		settings.run = true;
```
WebFreak001: the "run tool" functions should really be refactored, but that won't be part of this PR, so I'm fine with this change.

source/dub/generators/generator.d
```diff
@@ -97,6 +97,7 @@ class ProjectGenerator
 			BuildSettings buildSettings;
 			auto config = configs[pack.name];
 			buildSettings.processVars(m_project, pack, pack.getBuildSettings(settings.platform, config), settings, true);
+			if (settings.buildSettings.options & BuildOption.lowmem) buildSettings.options |= BuildOption.lowmem;
```
WebFreak001: adding on what we previously discussed: the lowmem addition could be copied into processVars inside the `if (include_target_settings)` while still keeping the functionality of the function sane.

Below where it calls finalizeGeneration right now it doesn't include the lowmem flag for example. Even though it isn't used right now, it might be subject to change in the future (because it processes the environment variables in there)

What would you think about that approach?

shoo: I rather put it in its current location, being careful that the lowmem only applies to the build and not to finalizeGeneration/performPostGenerateActions.
The reason for doing so is that I didn't see the vision of lowmem being used in finalizeGeneration/performPostGenerateActions.
I made it a priority to minimize the impact.

Also, there still doesn't seem to be a sensible reason to put this process in a function for variable expansion.
From this perspective, we thought it was reasonable to handle it outside of the function.

source/dub/commandline.d
shoo: You can get lowmem by running `dub describe --data=options --data-list`.

>I would imagine because m_buildSettings isn't used at all, that it would miss other stuff too.

I suspect that the fact that m_buildSettings is not being used at all is a bug...
For example, the following does not show the expected debug version(TEST_VER) because of this:
```
dub describe --data=debug-versions --data-list -d=TEST_VER
```

WebFreak001: made an issue #1911

shoo:
Is there anything I can still do about this pull request?

thewilsonator (approved):

PR #1906, opened 2020-03-25
Refactor and test the `runDubCommandLine` function

gedaiu:
I needed to update `trial` to the latest dub version, but i got stuck in reusing the `runDubCommandLine`. I moved to some other structs to be easier to integrate in other libs.

I hope this works for you too.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -174,8 +174,13 @@ struct CommandLineHandler
 
 	/** Get an instance of the requested command.
 
-	If there is no command requested the `run` command is returned.
-	If the `--help` argument was set, the help command will be returned as default.
+	If there is no command in the argument list, the `run` command is returned
+	by default.
+
+	If the `--help` argument previously handled by `prepareOptions`,
+	`this.options.help` is already `true`, with this returning the requested
+	command. If no command was requested (just dub --help) this returns the
+	help command.
 
 	Params:
 		name = the command name
@@ -184,11 +189,13 @@ struct CommandLineHandler
 		Returns the command instance if it exists, null otherwise
 	*/
 	Command getCommand(string name) {
-		if(name == "help" || (name == "" && options.help)) {
+		if (name == "help" || (name == "" && options.help))
+		{
 			return new HelpCommand();
 		}
 
-		if(name == "") {
+		if (name == "")
+		{
 			name = "run";
 		}
 
@@ -215,7 +222,8 @@ struct CommandLineHandler
 	Command prepareCommand(string name, CommandArgs args) {
 		auto cmd = getCommand(name);
 
-		if(cmd !is null || cmd.name != "help") {
+		if (cmd !is null || cmd.name != "help")
+		{
 			// process command line options for the selected command
 			cmd.prepare(args);
 			enforceUsage(cmd.acceptsAppArgs || !args.hasAppArgs, name ~ " doesn't accept application arguments.");
@@ -587,10 +595,11 @@ class CommandArgs {
 	@property bool hasAppArgs() { return m_appArgs.length > 0; }
 
 
-	/** Checks if the `--version` argument is present.
+	/** Checks if the `--version` argument is present on the first position in
+	the list.
 
 	Returns:
-		true if the application version argument is present
+		true if the application version argument was found on the first position
 	*/
 	@property bool hasAppVersion() { return m_args.length > 1 && m_args[1] == "--version"; }
 
@@ -632,7 +641,7 @@ class CommandArgs {
 	}
 
 	/** Returns the list of unprocessed arguments, ignoring the app arguments,
-	and calls `dropAllArgs`.
+	and resets the list of available source arguments.
 	*/
 	string[] extractRemainingArgs()
 	{
@@ -643,15 +652,13 @@ class CommandArgs {
 
 	/** Returns the list of unprocessed arguments, including the app arguments
 		and calls `dropAllArgs`.
-
-	Returns:
-		Returns the list of unprocessed arguments with the app arguments
 	*/
 	string[] extractAllRemainingArgs()
 	{
 		auto ret = extractRemainingArgs();
 
-		if(this.hasAppArgs) {
+		if (this.hasAppArgs)
+		{
 			ret ~= "--" ~ m_appArgs;
 		}
 
@@ -807,15 +814,17 @@ class HelpCommand : Command {
 		];
 	}
 
-	/// There is nothing to prepare
+	/// HelpCommand.prepare is not supposed to be called, use
+	/// cast(HelpCommand)this to check if help was requested before execution.
 	override void prepare(scope CommandArgs args)
 	{
-		assert(false, "There is nothing to prepare");
+		assert(false, "HelpCommand.prepare is not supposed to be called, use cast(HelpCommand)this to check if help was requested before execution.");
 	}
 
-	/// There is nothing to execute
+	/// HelpCommand.execute is not supposed to be called, use
+	/// cast(HelpCommand)this to check if help was requested before execution.
 	override int execute(Dub dub, string[] free_args, string[] app_args) {
-		assert(false, "There is nothing to execute");
+		assert(false, "HelpCommand.execute is not supposed to be called, use cast(HelpCommand)this to check if help was requested before execution.");
 	}
 }
 
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -86,6 +86,9 @@ struct CommandNameArgument
 	private string[] m_args;
 
 	/** Separates the command name from the arguments.
+
+	Params:
+		args = a list of string arguments that will be processed
 	*/
 	this(string[] args)
 	{
@@ -97,14 +100,20 @@ struct CommandNameArgument
 		}
 	}
 
-	/** Returns the command name.
+	/** Retreive the command name
+
+	Returns:
+		Returns the command name if it is set, otherwise an empty string
 	*/
 	string value()
 	{
 		return m_value;
 	}
 
-	/** Returns the list of unprocessed arguments and calls `dropAllArgs`.
+	/** Returns the list of unprocessed arguments.
+
+	Returns:
+		Returns the list of aguments, without the command name
 	*/
 	string[] extractAllRemainingArgs()
 	{
@@ -149,8 +158,8 @@ struct CommandLineHandler
 
 	/** Create the list of all supported commands
 
-		Returns:
-			Returns the list of the supported command names
+	Returns:
+		Returns the list of the supported command names
 	*/
 	string[] commandNames()
 	{
@@ -186,6 +195,12 @@ struct CommandLineHandler
 
 	If there is no command requested the `run` command is returned.
 	If the `--help` argument was set, the help command will be returned as default.
+
+	Params:
+		name = the command name
+
+	Returns:
+		Returns the command instance if it exists, null otherwise
 	*/
 	Command getCommand(string name) {
 		if(name == "help" || (name == "" && options.help)) {
@@ -207,8 +222,15 @@ struct CommandLineHandler
 
 	/** Get an instance of the requested command after the args are sent.
 
-		It uses getCommand to get the command instance and then calls prepare.
-		*/
+	It uses getCommand to get the command instance and then calls prepare.
+
+	Params:
+		name = the command name
+		args = the command arguments
+
+	Returns:
+		Returns the command instance if it exists, null otherwise
+	*/
 	Command prepareCommand(string name, CommandArgs args) {
 		auto cmd = getCommand(name);
 
@@ -222,6 +244,9 @@ struct CommandLineHandler
 	}
 
 	/** Get a configured dub instance.
+
+	Returns:
+		A dub instance
 	*/
 	Dub prepareDub() {
 		Dub dub;
@@ -572,7 +597,10 @@ class CommandArgs {
 		m_args = "dummy" ~ (app_args_idx >= 0 ? args[0..app_args_idx] : args);
 	}
 
-	/** Checks if the `--version` argument is present.
+	/** Checks if the app arguments are present.
+
+	Returns:
+		A list of arguments that are present after `--` argument
 	*/
 	@property bool hasAppArgs() { return m_appArgs.length > 0; }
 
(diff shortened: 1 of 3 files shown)


source/dub/commandline.d
```diff
@@ -77,6 +78,300 @@ CommandGroup[] getCommands()
...
+	/** Parses the general options and sets up the log level
+		and the root_path
+	*/
+	void prepareOptions(CommandArgs args) {
+		LogLevel loglevel = LogLevel.info;
+
+		options.prepare(args);
+
+		if (options.vverbose) loglevel = LogLevel.debug_;
+		else if (options.verbose) loglevel = LogLevel.diagnostic;
```
thewilsonator: dupilcate argument to `if` an `else if`

gedaiu: for which values? here is `vverbose` and `verbose`

thewilsonator: Ah.

source/dub/commandline.d
```diff
@@ -383,16 +650,90 @@ class CommandArgs {
 		m_args = null;
 	}
 
-	/** Returns the list of unprocessed arguments and calls `dropAllArgs`.
+	/** Returns the list of unprocessed arguments, ignoring the app arguments,
+	and calls `dropAllArgs`.
```
WebFreak001: but does it call dropAllArgs? While not really likely, it's possible someone inherits from CommandArgs and expects dropAllArgs to be called (for whatever reason) because of this documentation.

source/dub/commandline.d
```diff
@@ -383,16 +650,90 @@ class CommandArgs {
...
 		auto ret = m_args[1 .. $];
 		m_args = null;
 		return ret;
 	}
+
+	/** Returns the list of unprocessed arguments, including the app arguments
+		and calls `dropAllArgs`.
+
+	Returns:
+		Returns the list of unprocessed arguments with the app arguments
```
WebFreak001: this "Returns" section is a bit useless with the documentation before it saying the same already.

source/dub/commandline.d
```diff
@@ -383,16 +650,90 @@ class CommandArgs {
...
+	*/
+	string[] extractAllRemainingArgs()
+	{
+		auto ret = extractRemainingArgs();
+
+		if(this.hasAppArgs) {
```
WebFreak001: inconsistent style (no space after if)

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+	If the `--help` argument was set, the help command will be returned as default.
+
+	Params:
+		name = the command name
+
+	Returns:
+		Returns the command instance if it exists, null otherwise
+	*/
+	Command getCommand(string name) {
+		if(name == "help" || (name == "" && options.help)) {
```
WebFreak001: inconsistent style (no space after if)

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+			return new HelpCommand();
+		}
+
+		if(name == "") {
```
WebFreak001: inconsistent style (no space after if)

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+		name = the command name
+		args = the command arguments
+
+	Returns:
+		Returns the command instance if it exists, null otherwise
+	*/
+	Command prepareCommand(string name, CommandArgs args) {
+		auto cmd = getCommand(name);
+
+		if(cmd !is null || cmd.name != "help") {
```
WebFreak001: inconsistent style (no space after if)

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+	CommonOptions options;
+
+	/** Create the list of all supported commands
+
+	Returns:
+		Returns the list of the supported command names
+	*/
+	string[] commandNames()
+	{
+		return commands.map!(g => g.commands.map!(c => c.name).array).join.array;
```
WebFreak001: .join already returns an array

source/dub/commandline.d
```diff
@@ -383,16 +650,90 @@ class CommandArgs {
...
+
+	/** Returns the list of unprocessed arguments, including the app arguments
+		and calls `dropAllArgs`.
+
+	Returns:
+		Returns the list of unprocessed arguments with the app arguments
+	*/
+	string[] extractAllRemainingArgs()
+	{
+		auto ret = extractRemainingArgs();
```
WebFreak001: The "calling dropAllArgs" part here starts not making any sense anymore. Just looks like a trap someone could fall into because this function will return something different the second time it's called. In the current usage in this file this becomes extremely error prone because there are 2 different `extractAllRemainingArgs` functions (in here and in CommandNameArgument) which do completely different things.

source/dub/commandline.d
```diff
@@ -470,6 +811,32 @@ struct CommandGroup {
...
+		this.description = "Shows the help message";
+		this.helpText = [
+			"Shows the help message and the supported command options."
+		];
+	}
+
+	/// There is nothing to prepare
+	override void prepare(scope CommandArgs args)
+	{
+		assert(false, "There is nothing to prepare");
```
WebFreak001: this will result in something like
```
core.exception.AssertError@commandline.d(832): There is nothing to prepare
----------------
??:? _d_assert_msg [0x559093d2d9ce]
commandline.d:832 void dub.commandline.HelpCommand.prepare() [0x559093d2cb18]
```

To someone not having seen this code this isn't very helpful. A more helpful message would be something like:

HelpCommand.prepare is not supposed to be called, use `cast(HelpCommand)this` to check if help was requested before execution.

source/dub/commandline.d
```diff
@@ -470,6 +811,32 @@ struct CommandGroup {
...
+	}
+
+	/// There is nothing to execute
+	override int execute(Dub dub, string[] free_args, string[] app_args) {
+		assert(false, "There is nothing to execute");
```
WebFreak001: ditto

source/dub/commandline.d
```diff
@@ -126,155 +445,76 @@ int runDubCommandLine(string[] args)
...
 
-	if (options.help) {
+	if (cmd is null) {
+		logError("Unknown command: %s", command_name_argument.value);
+		writeln();
+		showHelp(handler.commands, common_args);
+		return 1;
+	}
+
+	if (cmd.name == "help") {
```
WebFreak001: `cast(HelpCommand)cmd` better describes the relation between this check and HelpCommand

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+}
+
+/// It returns the `help` command when there is none set and the --help arg
+/// was set
+unittest {
+	CommandLineHandler handler;
+	auto args = new CommandArgs(["--help"]);
+	handler.prepareOptions(args);
+	handler.commands = getCommands();
+	assert(handler.getCommand("").name == "help");
```
WebFreak001: ditto to line 488

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+}
+
+/// It returns the `help` command when the `help` command is sent
+unittest {
+	CommandLineHandler handler;
+	handler.commands = getCommands();
+	assert(handler.getCommand("help").name == "help");
```
WebFreak001: ditto to line 488

source/dub/commandline.d
```diff
@@ -77,6 +78,325 @@ CommandGroup[] getCommands()
...
+		LogLevel loglevel = LogLevel.info;
+
+		options.prepare(args);
+
+		if (options.vverbose) loglevel = LogLevel.debug_;
+		else if (options.verbose) loglevel = LogLevel.diagnostic;
+		else if (options.vquiet) loglevel = LogLevel.none;
+		else if (options.quiet) loglevel = LogLevel.warn;
+		else if (options.verror) loglevel = LogLevel.error;
+		setLogLevel(loglevel);
```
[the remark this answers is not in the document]
gedaiu: I'm not sure why a library would use this structure. My goal, here was to have a structure that you can reuse in other apps, and have the same behavior as dub.

source/dub/commandline.d
gedaiu: There is no `CommandNameArgument` struct anymore. Is this still a problem?

gedaiu:
@WebFreak001 ready for another view

source/dub/commandline.d
```diff
@@ -77,6 +78,314 @@ CommandGroup[] getCommands()
...
+		name = the command name
+		args = the command arguments
+
+	Returns:
+		Returns the command instance if it exists, null otherwise
+	*/
+	Command prepareCommand(string name, CommandArgs args) {
+		auto cmd = getCommand(name);
+
+		if (cmd !is null || cast(HelpCommand)cmd !is null)
```
WebFreak001: wrong code? if cmd is null the second part will always be `null !is null` and otherwise the second part won't be run.

You probably wanted to use
```suggestion
		if (cmd !is null && !(cast(HelpCommand)cmd))
```

source/dub/commandline.d
WebFreak001: hm I think it's ok now. However now because potentially you could more easily misuse the API (calling both extractRemainingArgs and extractAllRemainingArgs), can you add an assert into extractRemainingArgs to check if m_args has been extracted yet and otherwise error that you should only call one of the functions once?

source/dub/commandline.d
```diff
@@ -126,155 +434,76 @@ int runDubCommandLine(string[] args)
...
 
-	if (options.help) {
+	if (cmd is null) {
+		logError("Unknown command: %s", command_name_argument.value);
+		writeln();
+		showHelp(handler.commandGroups, common_args);
+		return 1;
+	}
+
+	if (cast(HelpCommand)cmd !is null) {
```
WebFreak001: while an overload of opCast!bool could break it, I think just `if (cast(HelpCommand)cmd)` is more readable than adding `!is null`

gedaiu: I think if opCast brakes this comparison, it's a bug in the ompiler.... anyway.. I updated the code.

source/dub/commandline.d
gedaiu: done

gedaiu:
@WebFreak001 ready for another look

gedaiu:
@WebFreak001 is the `help.sh` good enough?

test/help.sh
```diff
@@ -0,0 +1,27 @@
...
+if ! { ${DUB} --help | grep "Manages the DUB project in the current directory."; } then
+    die 'DUB did not print the default help message, with the `--help` argument.'
+fi
+
+### It shows the build command help
+if ! { ${DUB} build -h | grep "Builds a package"; } then
+    die 'DUB did not print the build help message, with the `-h` argument.'
+fi
+
+if ! { ${DUB} build -h | grep "Builds a package"; } then
```
WebFreak001, line 25:
```suggestion
if ! { ${DUB} build --help | grep "Builds a package"; } then
```

both this one and the one above are `-h` right now

source/dub/commandline.d
```diff
@@ -77,6 +78,314 @@ CommandGroup[] getCommands()
...
+}
+
+/// It returns an empty string on when there are no args
+unittest {
+	assert(extractCommandNameArgument([]).value == "");
+	assert(extractCommandNameArgument([]).remaining == []);
+}
+
+/// It returns the first argument when it does not start with `-`
+unittest {
```
MoonlightSentinel, line 111: These unittests could be merged because they test the same function.

source/dub/commandline.d
```diff
@@ -383,16 +640,90 @@ class CommandArgs {
...
+	}
+}
+
+/// It should not find the app version for an empty arg list
+unittest {
+	assert(new CommandArgs([]).hasAppVersion == false);
+}
+
+/// It should find the app version when `--version` is the first arg
+unittest {
```
MoonlightSentinel, line 677: Ditto

source/dub/commandline.d
```diff
@@ -126,155 +434,76 @@ int runDubCommandLine(string[] args)
...
+		cmd = handler.prepareCommand(command_name_argument.value, command_args);
 	} catch (Throwable e) {
 		logError("Error processing arguments: %s", e.msg);
 		logDiagnostic("Full exception: %s", e.toString().sanitize);
 		logInfo("Run 'dub help' for usage information.");
 		return 1;
 	}
 
-	if (options.help) {
+	if (cmd is null) {
```
MoonlightSentinel, line 470: Could cmd ever be an empty string (and thus != null)?

gedaiu: cmd is a `Command` object. How can be an empty string? I am not sure if I understand your question.

source/dub/commandline.d
gedaiu: Why do you think that would be better? I see an advantage of having separate  unit tests, because in case of a change, you can easily see which assets failed. When you have everything in one unit test block, you see only the first failing assert.

WebFreak001: I agree that more separate unittests are in general better, but here you are extracting the same `["test"]` input parameter so what MoonlightSentinel is trying to say is that the .value and .remaining test for the same input should be in the same unittest block I think.

source/dub/commandline.d
gedaiu: done :)

source/dub/commandline.d
gedaiu: ok. done.

gedaiu:
ping @MoonlightSentinel 

@WebFreak001 I don't understand why the `issue1416-maven-repo-pkg-supplier.sh` fails. I don't see why my change broken this test. Is this happening on master too?

source/dub/commandline.d
MoonlightSentinel: Sorry, mixed something up there.

WebFreak001:
test fail seems unrelated, can someone restart buildkite test if possible?

gedaiu:
@WebFreak001 I did a rebase with master. Do I need do to anything else to get this pr merged?

gedaiu:
Is there anything that blocks this pr to be merged?

WebFreak001:
@s-ludwig want to take a look at this? Otherwise would merge soon

PR #1916, opened 2020-04-05
Fix issue 1372 -- hidden directories should be ignored unless explicitly included

schveiguy:
This is built off of #1700.

I have modified a bit so if a path is given from the package file, it should be included by removing the base part of the path from the part that's searched for hidden directories.

I also finished out the test so it should work.

### DIFF
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -251,8 +251,11 @@ struct BuildSettingsTemplate {
 					}
 
 					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
-						import std.path : baseName;
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						if (baseName(d.name)[0] == '.' || d.isDir ||
+							pathSplitter(d).canFind!(name => name[0] == '.'))
+							continue;
 						auto src = NativePath(d.name).relativeTo(base_path);
 						files ~= src.toNativeString();
 					}
--- a/source/dub/recipe/packagerecipe.d
+++ b/source/dub/recipe/packagerecipe.d
@@ -250,11 +250,15 @@ struct BuildSettingsTemplate {
 						continue;
 					}
 
-					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
+					auto pstr = path.toNativeString();
+					foreach (d; dirEntries(pstr, pattern, SpanMode.depth)) {
 						import std.path : baseName, pathSplitter;
-						import std.algorithm.searching: canFind;
-						if (baseName(d.name)[0] == '.' || d.isDir ||
-							pathSplitter(d).canFind!(name => name[0] == '.'))
+						import std.algorithm.searching : canFind;
+						// eliminate any hidden files, or files in hidden directories. But always include
+						// files that are listed inside hidden directories that are specifically added to
+						// the project.
+						if (d.isDir || pathSplitter(d.name[pstr.length .. $])
+								   .canFind!(name => name.length && name[0] == '.'))
 							continue;
 						auto src = NativePath(d.name).relativeTo(base_path);
 						files ~= src.toNativeString();
--- a/test/issue1372-ignore-files-in-hidden-dirs.sh
+++ b/test/issue1372-ignore-files-in-hidden-dirs.sh
@@ -2,4 +2,15 @@
 
 cd ${CURR_DIR}/issue1372-ignore-files-in-hidden-dirs/
 
+echo "Compile and ignore hidden directories"
+rm dub.json
+ln -s dub_json_no_hidden.json dub.json
 ${DUB} build --force
+
+rm dub.json
+
+echo "Compile and explcitly include file in hidden directories"
+ln -s dub_json_hidden.json dub.json
+${DUB} build --force
+
+rm dub.json
--- a/test/issue1372-ignore-files-in-hidden-dirs.sh
+++ b/test/issue1372-ignore-files-in-hidden-dirs.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd ${CURR_DIR}/issue1372-ignore-files-in-hidden-dirs/
+
+${DUB} build --force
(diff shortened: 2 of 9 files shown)


source/dub/recipe/packagerecipe.d
```diff
@@ -250,9 +250,17 @@ struct BuildSettingsTemplate {
...
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+					auto pstr = path.toNativeString();
+					foreach (d; dirEntries(pstr, pattern, SpanMode.depth)) {
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						import std.string: startsWith;
+						// eliminate any hidden files, or files in hidden directories. But always include
+						// files that are listed inside hidden directories that are specifically added to
+						// the project.
+						if (baseName(d.name)[0] == '.' || d.isDir ||
```
schveiguy, line 261: I have a feeling I can just remove the first part of this expression. But let's see if it passes first.

PetarKirov:
An alternative way to address this issue is with a `.vibeignore` file, a la `.gitignore`. If no such file is provided, vibe can have a fallback one internally, that ignores Apple files like this.

The reason I'm not 100% with ignoring all `.*` files is that there various legitimate uses, like serving a `.git` repo.

Though, I agree with not serving hidden files in the Win32 sense.

PetarKirov:
An alternative way to address this issue is with a `.dubignore` file, a la `.gitignore`. If no such file is provided, vibe can have a fallback one internally, that ignores Apple files like this.


~The reason I'm not 100% with ignoring all `.*` files is that there various legitimate uses, like serving a `.git` repo.
Though, I agree with not serving hidden files in the Win32 sense.~

Edit: nevermind, I somehow thought that this issue was about vibe.d's serve static files function.

schveiguy:
@PetarKirov Yeah, I thought of something like that at first, but as it turns out, dub ALREADY ignores hidden files (.file) but not hidden directories. So this really is an incremental change that makes sense in that context.

schveiguy:
I'm not sure what the travis failure comes from, is that expected?

schveiguy:
Anyone?

Geod24:
Can you rebase on master ? Petar should have solved it with https://github.com/dlang/dub/pull/1920

schveiguy:
@Geod24 I read your message about 10 times and was very confused until I realized you were answering my question above 😆 

Will rebase.

schveiguy:
Yay, it's green! Anything else that anyone wants for this?

source/dub/recipe/packagerecipe.d
WebFreak001: sounds like a good idea to do now

source/dub/recipe/packagerecipe.d
```diff
@@ -250,9 +250,17 @@ struct BuildSettingsTemplate {
...
 					}
 
-					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
-						import std.path : baseName;
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+					auto pstr = path.toNativeString();
+					foreach (d; dirEntries(pstr, pattern, SpanMode.depth)) {
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						import std.string: startsWith;
```
WebFreak001, line 257: unused import

schveiguy: Yep. I think I intended to use it and then just used `name[0] == '.'`

source/dub/recipe/packagerecipe.d
schveiguy: OK, I reworked it. I also realized that checking some weird directory like "a/b//c" could cause an array bounds error, so I added a check for that.

Geod24:
> @Geod24 I read your message about 10 times and was very confused until I realized you were answering my question above 😆

That Phobos review really _did_ numb you ;)

source/dub/recipe/packagerecipe.d
```diff
@@ -250,9 +250,16 @@ struct BuildSettingsTemplate {
...
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+					auto pstr = path.toNativeString();
+					foreach (d; dirEntries(pstr, pattern, SpanMode.depth)) {
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
+						// eliminate any hidden files, or files in hidden directories. But always include
+						// files that are listed inside hidden directories that are specifically added to
+						// the project.
+						if (d.isDir || pathSplitter(d.name[pstr.length .. $])
+                                                                    .canFind!(name => name.length && name[0] == '.'))
```
WebFreak001, line 261: using spaces here but all previous lines have tabs

maybe you would like this to be like `\t\t\t\t\t\t__________.canFind` where _ is space or only tabs.

or instead of having such a long line in the first place, make a helper function to check if a given directory name should be ignored. Maybe we want to add `node_modules` to this in the future or stuff like that, having one function for it would make it easier to modify.

source/dub/recipe/packagerecipe.d
```diff
@@ -250,9 +250,16 @@ struct BuildSettingsTemplate {
 						continue;
 					}
 
-					foreach (d; dirEntries(path.toNativeString(), pattern, SpanMode.depth)) {
-						import std.path : baseName;
-						if (baseName(d.name)[0] == '.' || d.isDir) continue;
+					auto pstr = path.toNativeString();
+					foreach (d; dirEntries(pstr, pattern, SpanMode.depth)) {
+						import std.path : baseName, pathSplitter;
+						import std.algorithm.searching: canFind;
```
WebFreak001, line 256: while you are formatting: there is a space missing before the colon `:`

source/dub/recipe/packagerecipe.d
schveiguy: I hate tabs. So my editor defaults to expanded tabs (all spaces). Fixed.

I could potentially make a separate function, or I could have used `startsWith`, which would probably be shorter. If you want me to do that, I will.

source/dub/recipe/packagerecipe.d
schveiguy: Fixed.

WebFreak001:
@schveiguy oh I forgot, can you make a PR with a changelog entry for this? Might be interesting for users to see that they can now specify hidden paths manually

schveiguy:
Yes, I can do that.

PR #1920, opened 2020-04-12
Update test/*.min_frontend for several vibe.d tests to 2.077

PetarKirov:
Vibe.d officially stopped supporting DMDFE 2.076.0 as of: vibe-d/vibe.d@4755eb4

While vibe.d didn't break compatibility immediately, the latest release fails to build with DMD 2.076, which causes pull requests to fail. This commit bumps the min_frontend in several tests that use vibe.d to reflect this.


PetarKirov:
Thanks!

PR #1924, opened 2020-04-16
run nonet build with travis test

### LINKED ISSUE #1923: Support nonet application
Currently there might be issues in the library-nonet implementation like in #1910 which are not caught by the simple unittests. Running the full tests directory on a dub executable using library-nonet (vibe.d instead of curl) would prevent such issues from happening much better.

gedaiu:
fix #1923

@WebFreak001 would this change be enough?

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -14,6 +14,12 @@ configuration "application" {
 	versions "DubUseCurl" "DubApplication"
 }
 
+configuration "application-nonet" {
+	dependency "vibe-d:http" version=">=0.7.30 <=0.9.0"
+	targetType "executable"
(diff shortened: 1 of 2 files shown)


WebFreak001:
> would this change be enough?

```
./scripts/ci/travis.sh: line 33: --config=application-nonet: command not found
```
no

WebFreak001:
looks like this isn't enough, I think you will need to create more travis test runs which have some environment variable set to test using the nonet app instead of the normal app. You can just keep the build code for the nonet library like it is right now, I think that part is good, and then just conditionally call run-unittests.sh

PR #1935, opened 2020-05-01
Up the dub optional dependency to workaround code.dlang.org limitations.

schveiguy:
buildkite is failing everywhere because of issue dlang/dub-registry#458 and dub. The altered test here just increases the tag version to a later tag that will be included in the first 100 tags from github.

Obviously dub-registry fix is preferred (and without one, this will keep happening), but this could at least get buildkite working again.

### DIFF
--- a/test/issue672-upgrade-optional.sh
+++ b/test/issue672-upgrade-optional.sh
@@ -3,9 +3,9 @@
 . $(dirname "${BASH_SOURCE[0]}")/common.sh
 cd ${CURR_DIR}/issue672-upgrade-optional
 rm -rf b/.dub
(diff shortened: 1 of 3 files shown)


schveiguy:
Crap, I just realized, we should have made this into a 1.20.2 release, since there's already a 1.21.0 in beta. Is there a way to get a release out now so we can fix the failures? I think a cherry pick can work, right?

schveiguy:
Or is the release impending soon enough?

Geod24:
Can you make a PR to `stable` ?

schveiguy:
@Geod24 #1936

PR #1938, opened 2020-05-02
fix #1040 - dub run <package>@x.y.z ignores the specified version

### LINKED ISSUE #1040: dub run <package> blindy runs first package it can find in `.dub/packages` folder without regards to version
In my case I have multiple versions of the unit-threaded package in the package folder.

What I need is to be able to specify the package version dub should run. E.g. like so:

`dub run unit-threaded@0.7.2`

shoo:


source/dub/commandline.d
```diff
@@ -645,7 +645,7 @@ abstract class PackageBuildCommand : Command {
 		]);
 	}
 
-	protected void setupPackage(Dub dub, string package_name, string default_build_type = "debug")
+	protected void setupPackage(Dub dub, string package_name, Version ver = Version.unknown, string default_build_type = "debug")
```
WebFreak001, line 648: to avoid API breakage for people who extend the PackageBuildCommand class, can you move this new argument after default_build_type?

source/dub/packagemanager.d
```diff
@@ -188,9 +188,11 @@ class PackageManager {
 	/// ditto
 	Package getPackage(string name, Version ver, NativePath path)
 	{
-		auto ret = getPackage(name, path);
-		if (!ret || ret.version_ != ver) return null;
-		return ret;
+		foreach( p; getPackageIterator(name) )
```
WebFreak001, line 191: mixing code styles here (space after `foreach`)

source/dub/packagemanager.d
```diff
@@ -188,9 +188,11 @@ class PackageManager {
...
+			if (p.version_ == ver)
+				if (p.path.startsWith(path))
```
WebFreak001, line 193: a little bit picking here, but I think writing this with `&&` instead of a nested `if` would be better style

shoo:
> Thanks for this change! Just one minor compatibility thing and 2 code style things

Thanks for the code review!
I have responded to the points made, so please check back.

WebFreak001:
very good change, thanks for the work on this feature! Will definitely make dub more intuitive to use. (I had this issue multiple times too with fetching new tools and I think new users couldn't fix this themselves easily)

PR #1941, opened 2020-05-03
gdc: Build dub_platform_probe with -fsyntax-only

ibuclaw:
Not quite related to #576, but the use of -S can leave tons of dub_platform_probe.S files in the project directory if left unchecked.


Geod24:
I don't think we have a test with GDC in the CI though :(

ibuclaw:
@Geod24 sadly there's a patch needed to be added to std.net.curl in order to get it working, see bugzilla 19367.

Geod24:
Well you could try using the nonet library, which uses Vibe.d instead of Curl.

Issue #1829, opened 2019-12-21
Make the test-suite cross platform


Geod24:
Currently our test-suite is very tied to bash / POSIX.
The Windows CI [did not test it](https://github.com/dlang/dub/blob/7336602d36fdf711499a88f1ea40e52c2acbe452/appveyor.yml#L117-L123), nor is OSX:  https://github.com/dlang/dub/pull/1055

Moreover, it is quite hard for a newcomer to use it locally without diving into the travis configuration.
One of the possible way to approach this is to rewrite the runner in D.
I started doing this with `travis-ci.sh` to see how it would pan out. Any other suggestions ?

PR #1946, opened 2020-05-18
Make the testsuite re-runnable on OSX

Geod24:
This does two things:
- It makes the test-suite pass on OSX (or at least, on my machine);
- It makes some tests re-runnable so they don't fail miserably after the first run;

There's still a lot to do, like cleaning up the mess the testsuite create, or not messing up with the users packages, but one step at a time.

Part of https://github.com/dlang/dub/issues/1829


Geod24:
Fixed the micro failure (missing `|| true` after the `remove`).

PR #1948, opened 2020-05-19
Remove unused imports [on hold].

veelo:


thewilsonator:
```
source/dub/compilers/compiler.d(131,77): Error: no property toNativeString for type GenericPath!(PosixPathFormat)
source/dub/generators/sublimetext.d(56,39): Error: no property toNativeString for type GenericPath!(PosixPathFormat)
```

source/dub/compilers/compiler.d
```diff
@@ -8,21 +8,14 @@
 module dub.compilers.compiler;
 
 public import dub.compilers.buildsettings;
-public import dub.dependency : Dependency;
```
veelo, line 11: This is technically a breaking change I guess, which according to CONTRIBUTING.md cannot go into master. However, this line could be removed without adding any other line into the dub source itself, as no files that import `dub.compilers.compiler` use `Dependency`. Can a public import be deprecated?

Let me know if this line should be left as is.

WebFreak001: removing public imports should be considered a no-go. There are applications (most often IDEs/plugins) which depend on dub as a library and might have imported dub.compilers.compiler and used Dependency from it.

Deprecating public imports seems to work exactly the way you expect it to and should be done here instead (with a message telling you to import dub.dependency)

WebFreak001:
@thewilsonator this kind of change needs a lot more checking, the CI output will probably not be enough here.

When removing imports the templates which aren't covered might break because they could have used symbols which are now removed. So for these changes you need to go through every template definition and look at all the used symbols for every file where an import is removed. If mixins are involved this gets a lot more work to check correctly too. Dub is used as a library too so something like this can easily cause issues.

(I'm just assuming here that you didn't do this thorough review because your approval is 2 minutes after the PR opened)

veelo:
@WebFreak001 Thanks for pointing this out. I have grepped for "template" and "static if", which doesn't give hits in the files from this changeset. I just scanned the first 1k lines of `source.dub.commandline.d` without spotting an eponymous template, then I decided that I wouldn't have time to do the same for the rest...

WebFreak001:
you can fairly reliably find function templates by searching for `)(`, but it won't be all templates and there will also be a few false positives.

veelo:
> Actually checking if the removed symbols are then really gone through that import is another challenge though. I looked at dependency.d where the `dub.internal.vibecompat.inet.url` import is possibly needed in line 279 but when looking at if toString is called it looks like that shouldn't actually be possible if the tests pass.

I'm not sure I understand. You mean you would have expected compilation to fail because the template that uses the import is instantiated in the unittest? Adding to that, `NativePath` is used on line 52 of `dub.dependency.d`. I suspect `dub.internal.vibecompat.inet.path` is imported through some other public import, but I don't see it. I have tried running [https://code.dlang.org/packages/depend](https://code.dlang.org/packages/depend) on it, but couldn't get it to work.

I'm getting less enthusiastic about all this, maybe we should just close this PR... I don't understand the failing check either.

veelo:
I have split this up into smaller PRs.

Geod24:
Thanks!

PR #1950, opened 2020-05-19
Resolve deprecation warnings.

veelo:


Geod24:
We really should get rid of that duplication.

PR #1951, opened 2020-05-19
Fix failing test on native Windows.

veelo:


Geod24:
Great! Isn't there another failing unittest for the same reason though ?

PR #1960, opened 2020-06-06
Dub root package´target env vars added

andre2007:
A postGenerateCommands script of a dependency can access these new environment variables to calculate the path of the built binary.


changelog/dub_root_package_target_env_vars_added.dd
```diff
@@ -0,0 +1,4 @@
+Dub root package target environment variables added
+
+Environment variables `DUB_ROOT_PACKAGE_TARGET_TYPE`, `DUB_ROOT_PACKAGE_TARGET_PATH` 
+and `DUB_ROOT_PACKAGE_TARGET_NAME` added.
```
wilzbach, line 4: Maybe sth. like?

```suggestion
and `DUB_ROOT_PACKAGE_TARGET_NAME` have been added and exposed.
```

andre2007: done

PR #1964, opened 2020-06-11
fix trailing white character which has broken  CyberShadow's DAutoTest

NilsLankila:


NilsLankila:
I think you'll have to merge by hand, just like #1960, given that DautoTest is a required check.

NilsLankila:
well #1960 hadn't a SemaphoreCI check but it has one that didn't start even 30 hours after the labeling.

Geod24:
Right, thanks for the ping. Damn CIs are pesky, and I think dlang-bot is AWOL too.

Issue #1832, opened 2019-12-23
Changing the format of `settings.json`


Geod24:
I believe `settings.json` was a great addition to dub, something I've personally been wanting for a while. Allowing users to e.g. choose their default compiler without having to compile `dub` itself.

I'd like to extend it, for example by giving it the ability to use a specific Copyright / authors string by default, and by providing a default value for `--cache`. There are also a few bugs that need fixing (#1472 is trivial, for example).

However, currently, it suffers from one great downside which should make anyone wary of extending it: it is written in JSON.

JSON has proven to be a bad format for human-written configuration. Be it the lack of comments, noisy syntax, lack of typing... So much that a push was done to move to SDLang a few years ago.
However JSON enjoys wide support among editors and tools, and is still familiar to people than SDLang, the later being yet another skill a newcomer to D has to pick up.

Hence, I would like to bring forth the idea of changing `settings.json` to use a different format.
Out of all the markup languages out there, I my vote would go to YAML:
- Almost as widely supported as JSON;
- Easy enough to read;
- [We have a great library to read/write it already](https://github.com/dlang-community/D-YAML).

[I think the comparison is fairly simple](https://www.json2yaml.com/).

Would love some opinion from DUB / `settings.json` users.

jacob-carlborg:
I would like to add. That most importantly, YAML is a superset of JSON. All existing JSON files would still be readable by Dub if the format is changed to YAML. No need for multiple parser inside Dub. Perhaps only the file extension (`.json`) need to be deprecated.

wilzbach:
My vote goes for YAML as well :+1:

andre2007:
Also my vote goes for YAML. It is well known in the industry.

PetarKirov:
I agree with @Geod24 high-level goals of extending `setting.json`'s uses. I'm ok with using YAML (in addition to SDLang) as it's also growing in popularity among various other projects.
However, unless we target this feature for 2.0, we need to keep backwards compatibility and continue supporting JSON.

Geod24:
After https://github.com/dlang/dub/pull/2310 we'll be able to do this fairly easily and in a backward-compatible manner.

PR #1972, opened 2020-06-23
[RFC] Make dub self hosted

Geod24:
There are a bunch of places I need to go through and polish, mostly CI scripts, to make sure the `build.d` script is correctly called, but I wanted to get an early-ish feedback on what people think about this before I put the 80% of work required to get those last 20% of feature.

This PR will make `dub` depend on `dub` for building. This mean that we will need to have a bootstrap step in the CI to ensure we do not depend on new versions "too soon". As a concession to user-friendliness, the `build.d` script was left in place as a shell to call `dub`. It also allows backward compatibility if one needs that.

The main motivation to make `dub` self-hosted is to open it up to the usage of libraries. Currently `dub` has a few modules that are copy-pasted from Vibe.d because no dependency management way could be decided on for the dependency-management tool. I personally wanted to experiment with a few things (including #1832) but was blocked by the addition of a dependency. I considered using `git submodules` but the usage felt like reinventing the wheel because of the logic needed in `build.d`. @PetarKirov suggested self-hosting, and while originally hostile to the idea, I had grown convinced that it might be the time to do so.

I hope the diff is convincing enough: Literally no addition to dub, only a sed on import, and -5k LoCs.

CC @s-ludwig @WebFreak001 @MartinNowak @wilzbach @andre2007 : Destroy (or 👍 )!


WebFreak001:
cc @Vild, I think this would break the ArchLinux package

Geod24:
It would likely break every packages out there which is why I'm in no hurry to pull this.
I maintain the packages on Alpine, and so I think I will just download `v1.21.0`, build it, then build the latest (a two-stage process). For packages that use a `git` repository, that might be even simpler. But if there's any problem for packagers, I'm not against adding more logic to `build.d` (it already tries *very hard* to be user friendly, e.g. by making sure the binaries given are executable).

andre2007:
Thanks a lot for working on this. I actually needs this to due to the undead movement of std.xml. Also it would ease the `dub watch` feature Martin Nowak has created pr some years ago. Here also another dub dependency is needed.

Geod24:
> Can you describe the bootstrapping process?

Checkout dub v1.21.0, build it with `build.d`, copy the binary somewhere safe, checkout `master` or the tag you want to build, build using the previously-build dub.

> The latter, which is the only available option for me, will likely break in the future, leaving us with no way to bootstrap dub. Am I missing something?

I will add a bootstrapping check in the CI, ensuring that master can be built with v1.21.0 for a relatively long time. How do you guys handle DMD at the moment ?

Alternatively, I could add an option to `build.d` to do the bootstrap by itself. Something like `DC=compiler ./build.d bootstrap`.

FFY00:
Having a build system would also allow you to dynamically link to vibe.d and all other dependencies. This is *very* desirable for distributions.

The current approach goes against arch guidelines, but I guess there's no option so it will have to do.

FFY00:
> ensuring that master can be built with v1.21.0 for a relatively long time

This does not give me enough confidence. "relatively long time" is not enough for me. At some point things will change enough that this will be very difficult to maintain. Might be in 1 year, might be in 5 years, but it will likely happen at some point.

> Alternatively, I could add an option to `build.d` to do the bootstrap by itself. Something like `DC=compiler ./build.d bootstrap`.

I want to manage sources myself, so there is no need. I fetch the sources, validate the checksums, and proceed with the build.

WebFreak001:
I have a fear of dependencies down the dependency chain introducing breaking issues like using toolchainRequirements or new potential future dub features (like git submodules, new repository types or preBuildCommands with new variables) which wouldn't necessarily break the build, causing headaches.

I think personally I would prefer the most if dependencies were normally specified and built using dub, however for the bootstrapping process you are expected to install all the packages (like `dub upgrade --cache=local`) and the build tool would find them in the .dub folder or similar. This would be pretty similar to what we currently have, while allowing us to use dub dependencies and optionally include them for building when bootstrapping.

Geod24:
> Having a build system would also allow you to dynamically link to vibe.d and all other dependencies. This is very desirable for distributions.
>
> The current approach goes against arch guidelines, but I guess there's no option so it will have to do.

I agree and I'm well aware of how at odds the D ecosystem is with traditional Linux packaging practices. However, this change does not make the situation worse, does it ?

> This does not give me enough confidence. "relatively long time" is not enough for me. At some point things will change enough that this will be very difficult to maintain. Might be in 1 year, might be in 5 years, but it will likely happen at some point.

The wording was intentionally vague, as I'm not the only stakeholder, and I can't tell you how long it will be without gathering opinion. This is also why I asked how you are handling DMD, as it requires a D compiler to build.

FFY00:
> I think personally I would prefer the most if dependencies were normally specified and built using dub, however for the bootstrapping process you are expected to install all the packages (like `dub upgrade --cache=local`) and the build tool would find them in the .dub folder or similar. This would be pretty similar to what we currently have, while allowing us to use dub dependencies and optionally include them for building when bootstrapping on another system.

Copying or symlinking from the system to `.dub` is a bad hack IMO, but it would work if you really want to do it.

> I agree and I'm well aware of how at odds the D ecosystem is with traditional Linux packaging practices. However, this change does not make the situation worse, does it ?

It wouldn't. But you have a situation here that would be much better resolved by adhering the standard practices, and I am still getting pushback. What are the cons of having a proper build system?

> The wording was intentionally vague, as I'm not the only stakeholder, and I can't tell you how long it will be without gathering opinion.

This is exactly the problem. And opinions change with time. I have had to deal with with similar issues in the past, and I am advising that from my experience going down this route is not a great option.

> This is also why I asked how you are handling DMD, as it requires a D compiler to build.

We use LDC to build DMD. LDC can be bootstrapped without a D compiler.

https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/dmd#n51

Anyway, it's not my call. I am just giving my opinion as it was asked from me.

Geod24:
> We use LDC to build DMD.

Then the bootstrap issue is with `ldc`. And from what I can see, it simply depends on itself ?

> What are the cons of having a proper build system?

You mean switching to `pkg-config` ? We use it inside `dub` for C libraries, for for D libraries, no one uses it. `dub` is supposed to be the standard D build system, so I'm not sure what you are suggesting.

> This is exactly the problem. And opinions change with time. I have had to deal with with similar issues in the past, and I am advising that from my experience, I think going down this route is not a great option.

I understand the frustration (and experienced it myself). I'll make sure we reach an agreement before this gets merged (if it does). But I'd like an answer on the `ldc` bootstrapping issue.

> Anyway, it's not my call. I am just giving my opinion as it was asked from me.

And it is very welcome!

FFY00:
> Then the bootstrap issue is with `ldc`. And from what I can see, it simply depends on itself ?

Yes, they went for the same approach as you propose. I do not think it is very maintainable, but they will have to deal with it.

While I think the approach the LDC developers choose isn't great, it is 

> You mean switching to `pkg-config` ? We use it inside `dub` for C libraries, for for D libraries, no one uses it. `dub` is supposed to be the standard D build system, so I'm not sure what you are suggesting.

Dub is a decent package manager but an awful build system. I could have a lengthy discussion about this, but when it comes to our issue I think it all boils down to dub not being able to resolve dependencies on the system.

Nobody uses pkgconfig for D dependencies because dub, the de-facto "build system", does not support them. Dub also doesn't support installing to the system which does not help in any way. There is a real need for this, `vibe.d` for example is forced to also maintain meson definitions because of this.

The main motivation for this change is to be able to have external dependencies, instead of using dub and making everything more complicated just provide a simple Makefile or meson definition that finds the dependencies via pkgconfig and builds dub. If you also want to maintain a dub file for dub to be able to build itself, go ahead, but do not make it so that dub is un-bootstrapable by itself, especially since the Makefile/meson definition would be super simple.

veelo:
I wonder whether this alternative idea for bootstrapping would work:

1. On a fully functional system do `dub generate cmake`.
2. Copy the CMake scripts to the system to be bootstrapped.
3. Use CMake to build dub.

If CMake is not available on the system to be bootstrapped, Makefiles can be used that are generated by CMake on the fully functional system.

Does this solve any bootstrapping concerns?

FFY00:
It would be much better if the cmake files were committed, but overall, it would be a reasonable approach.

What I am afraid of is in the future the cmake generation be changed in a way that it relies on dub, like calling dub for something. This is something I feel would be reasonable to do otherwise, but would absolutely break the bootstrapping.

PR #1974, opened 2020-06-30
Fix bogus error message for the path used to reference sub packages.

s-ludwig:
Sub packages should ("must") always be referenced through the path of their base package instead of their own path (in case it differs). However, the dependency resolution code assume the opposite. Both way are now accepted, but referencing the sub package path directly results in a warning.


s-ludwig:
@wilzbach: Do you have an idea why the SemaphoreCI status is not being reported anymore? I'd probably switch back to making TravisCI the required test, since I couldn't see what's wrong.

wilzbach:
> @wilzbach: Do you have an idea why the SemaphoreCI status is not being reported anymore? I'd probably switch back to making TravisCI the required test, since I couldn't see what's wrong.

I don't know what went wrong. Maybe they retired their 1.0 offering (and switched fully to 2.0).
Anyhow, I removed it as required check, s.t. this can be merged.

PR #1977, opened 2020-07-20
Remove usage of deprecated std.xml

andre2007:
In this case, parsing the xml with std.regex seems to make most sense.
Example XML
``` xml
<metadata modelVersion="1.1.0">
  <groupId>dubpackages</groupId>
  <artifactId>maven-dubpackage</artifactId>
  <versioning>
    <latest>1.0.6</latest>
    <release>1.0.6</release>
    <versions>
      <version>1.0.5</version>
      <version>1.0.6</version>
    </versions>
    <lastUpdated>20180317184845</lastUpdated>
  </versioning>
</metadata>
```

### DIFF
--- a/source/dub/packagesuppliers/maven.d
+++ b/source/dub/packagesuppliers/maven.d
@@ -79,7 +79,7 @@ class MavenRegistryPackageSupplier : PackageSupplier {
 
 	private Json getMetadata(string packageId)
 	{
-		import std.xml;
+		import dub.internal.undead.xml;
 
 		auto now = Clock.currTime(UTC());
 		if (auto pentry = packageId in m_metadataCache) {
@@ -113,6 +113,7 @@ class MavenRegistryPackageSupplier : PackageSupplier {
 			 xml.parse();
 		};
 		xml.parse();
+
 		m_metadataCache[packageId] = CacheEntry(json, now);
 		return json;
 	}
--- a/build-files.txt
+++ b/build-files.txt
@@ -39,6 +39,7 @@ source/dub/internal/sdlang/parser.d
 source/dub/internal/sdlang/symbol.d
 source/dub/internal/sdlang/token.d
 source/dub/internal/sdlang/util.d
+source/dub/internal/undead/xml.d
 source/dub/internal/utils.d
 source/dub/internal/vibecompat/core/file.d
 source/dub/internal/vibecompat/core/log.d
--- a/source/dub/internal/undead/xml.d
+++ b/source/dub/internal/undead/xml.d
@@ -1,0 +1,3109 @@
+// Written in the D programming language.
+
+/**
+$(RED Warning: This module is considered out-dated and not up to Phobos'
+      current standards. It will remain until we have a suitable replacement,
+      but be aware that it will not remain long term.)
+
+Classes and functions for creating and parsing XML
+
+The basic architecture of this module is that there are standalone functions,
+classes for constructing an XML document from scratch (Tag, Element and
+Document), and also classes for parsing a pre-existing XML file (ElementParser
+and DocumentParser). The parsing classes <i>may</i> be used to build a
+Document, but that is not their primary purpose. The handling capabilities of
+DocumentParser and ElementParser are sufficiently customizable that you can
+make them do pretty much whatever you want.
+
+Example: This example creates a DOM (Document Object Model) tree
+    from an XML file.
+------------------------------------------------------------------------------
+import dub.internal.undead.xml;
+import std.stdio;
+import std.string;
+import std.file;
+
+// books.xml is used in various samples throughout the Microsoft XML Core
+// Services (MSXML) SDK.
+//
+// See http://msdn2.microsoft.com/en-us/library/ms762271(VS.85).aspx
+
+void main()
+{
+    string s = cast(string) std.file.read("books.xml");
+
+    // Check for well-formedness
+    check(s);
+
+    // Make a DOM tree
+    auto doc = new Document(s);
+
+    // Plain-print it
+    writeln(doc);
+}
+------------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


source/dub/packagesuppliers/maven.d
```diff
@@ -104,15 +104,12 @@ class MavenRegistryPackageSupplier : PackageSupplier {
...
-		auto xml = new DocumentParser(xmlData);
-
-		xml.onStartTag["versions"] = (ElementParser xml) {
-			 xml.onEndTag["version"] = (in Element e) {
-				json["versions"] ~= serializeToJson(["name": packageId, "version": e.text]);
-			 };
-			 xml.parse();
-		};
-		xml.parse();
+		enum versionRegEx = `<version>(.*?)<\/version>`;
```
WebFreak001, line 107: this doesn't parse e.g. `<version>1&#46;0&#46;6</version>` properly anymore.

I would imagine this isn't a huge deal as the SemVer version strings can all be represented easily without escaping, but this does make it no longer strictly conform to XML and servers could do anything.

I would have also expected this to change comment and whitespace behavior, however it seems the previous xml parser solution didn't care about that either, it just unescaped.

andre2007: You are right. A simple copy of the `decode` functionality from undead is about 200 lines. I wonder wheter to copy&paste this, to be on the save side?

On the other side I am not really sure decode is necessary at all here.

wilzbach: > copy&paste this, to be on the save side?

Yeah, that's fine.

andre2007: Module `dub.internal.undead.xml` added. It contains the decode function and all code dependencies in the order and the format as defined in the original file.

WebFreak001:
I just noticed one other thing: before it would only search for `<version>...</version>` inside a `<versions>...</versions>` block. Can it happen that in maven there is anywhere else a `<version>` than in a `<versions>` block? Like some metadata version or something.

andre2007:
You are right, I have to extend the regex to find `version` tags only within a `versions` parent tag. (Just saw an example where `version` tag is after `artifactId`).

WebFreak001:
I'm really not quite sold on the conversion to regex and manual string operations. XML could contain a lot more weird stuff like whitespace in the tags, random attributes, the tags being in a completely different nesting level (like in some sub-tags instead of the expected root tag) than expected, etc.

How big would putting the entire undeaD xml package in be?

andre2007:
3100 lines or 90kb. Should I add it?

WebFreak001:
hm I don't think I should decide on that, what does @wilzbach think?

andre2007:
@wilzbach should I add the whole undead.xml file, what do you think?

rikkimax:
Alternatively do what DCD and friends do. Git submodule for script (i.e. makefile) and dub dependency for dub.

wilzbach:
> @wilzbach should I add the whole undead.xml file, what do you think?

Hmm, I unfortunately don't see a better alternative as well. This really looks like the easiest solution to this problem as long as we don't trust dub enough to bootstrap itself or want to maintain easy bootstrapping of dub on new platforms.

Though we should add a big comment on the header stating that it's dead code, link to the undead package and maybe even put it in under an undead folder as well?

Re git submodules: I don't think it matters much whether it's a git submodule or just a copy/paste as I don't expect this file to change. The git submodule will require small adaptions to the CIs, build script and dub distro packages, but that's marginal as long as it's mentioned in a changelog entry.

If no one else has a strong opinion, I would leave the final choice on copy/paste or git submodule to you. I'm fine either way.

andre2007:
As the other dependencies are also code copies and in the long run we will have a better solution, when dub bootstrapping is available, I favour copying the file into it.
I will change the pr.

andre2007:
macOS-10.15 LDC master fails to an Vibe-d SSL linker issue:

d_tls.a(vibe.stream.openssl.o)
  "_get_rfc3526_prime_2048", referenced from:
      __D4vibe6stream7openssl14OpenSSLContext11setDHParamsMFNeAyaZv in libvibe-d_tls.a(vibe.stream.openssl.o)
  "_sk_num", referenced from:
      __D6deimos7openssl9safestack__T10SKM_sk_numTSQBqQBm6x509v315GENERAL_NAME_stZ__TQBwZQCaFNbPSQDkQDgQDb__T8STACK_OFTQCrZQoZi in libvibe-d_tls.a(vibe.stream.openssl.o)
     (maybe you meant: __D6deimos7openssl9safestack__T10SKM_sk_numTSQBqQBm6x509v315GENERAL_NAME_stZ__TQBwZQCaFNbPSQDkQDgQDb__T8STACK_OFTQCrZQoZi)
  "_sk_value", referenced from:
      __D6deimos7openssl9safestack__T12SKM_sk_valueTSQBsQBo6x509v315GENERAL_NAME_stZ__TQByZQCcFNbPSQDmQDiQDd__T8STACK_OFTQCrZQoiZPQDa in libvibe-d_tls.a(vibe.stream.openssl.o)
     (maybe you meant: __D6deimos7openssl9safestack__T12SKM_sk_valueTSQBsQBo6x509v315GENERAL_NAME_stZ__TQByZQCcFNbPSQDmQDiQDd__T8STACK_OFTQCrZQoiZPQDa)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error: /usr/bin/cc failed with status: 1
ldc2 failed with exit code 1.
[ERROR] /Users/runner/work/dub/dub/test/ddox.sh:11 command failed
[ERROR] Script failure.

andre2007:
Ping :)

Geod24 (approved):

Issue #1535, opened 2018-08-21
x86_mscoff matches x86 platform specs


John-Colvin:
Which means that you can't have do e.g. link to one file for optlink and one for ms linker

s-ludwig:
Although x86_mscoff is really just a hack and we still need proper selection mechanics based on the object format (#228), the semantics are consistent with the DMD pre-defined versions. We could add another hack on top and introduce x86_omf as an intermediate solution, but it's also currently possible to use configurations to work around the issue:

```
configuration "coff" {
  platforms "windows-x86_mscoff"
  libs "mycofflib"
}
configuration "default" {
  libs "myomflib" platform="windows-x86"
}
```

This can still cause a link failure when building with "dub -a x86_mscoff -c omf" manually, but by default it will do the right thing.

shoo:
I've seen a lot of solutions to this problem with the addition of configuration.
However, this has made configuration extremely complex and created a lot of confusion. And the confusion is compounded by multiplication if the project has libraries that it depends on.
The addition of x86_omf will alleviate some of the confusion.

Geod24:
Upstream has removed OMF support so this is now working as expected.

PR #1983, opened 2020-07-27
Add architecture x86_omf for windows-dmd

shoo:
This pull request is in support of #1535.
#1535 is meant to point out x86_mscoff will also match x86, so it cannot specialize in omf.

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -156,12 +156,6 @@ interface Compiler {
 		// cmdline option does not lead to the same string being found among
 		// `build_platform.architecture`, as it's brittle and doesn't work with triples.
 		if (build_platform.compiler != "ldc") {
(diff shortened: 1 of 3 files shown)


wilzbach:
No changelog?

shoo:
I forgot to update the changelog. Should I add the changelog from now?
And, need to change the documentation too? (From what I've seen, x86_mscoff isn't mentioned either, so the documentation may not be necessary.)

wilzbach:
> Should I add the changelog from now?

That would be great, s.t. other users are aware of this change. A simple PR to the changelog folder should be all that's needed.

> And, need to change the documentation too? (From what I've seen, x86_mscoff isn't mentioned either, so the documentation may not be necessary.)

Yes, the documentation sources can be found here: https://github.com/dlang/dub-docs

PR #1988, opened 2020-08-03
Update dependencies to the latest release

Geod24:
Needed for https://github.com/dlang/phobos/pull/7545

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
       if: runner.os == 'macOS'
       run: |
         brew install pkg-config coreutils
+        echo ::set-env name=PKG_CONFIG_PATH::/usr/local/opt/openssl@1.1/lib/pkgconfig/
 
     - name: '[Linux] Install dependencies'
       if: runner.os == 'Linux'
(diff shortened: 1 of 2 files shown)


wilzbach:
The newest eventcore seems to be incompatible with 2.076 - https://travis-ci.com/github/dlang/dub/jobs/367979885

IIRC we need to maintain building from scratch with 2.076 for GDC, but that's only for dub from source, but not for dub built with dub.

Geod24:
That's odd, because the diff between v0.9.6 and v0.9.7 is [a single commit](https://github.com/vibe-d/eventcore/commit/862b5d470c7b19019811963596f9da624913147d) which cannot possibly be the source of this...

wilzbach:
At the moment eventcore 0.8.48 is used here.

Geod24:
Note to self: Don't comment on Github issues before taking your morning coffee.
Alright, I'll try to look into it a bit later this week.

wilzbach:
As I said, I think it's fine to simply ignore the dub build of dub for older versions. AFAIK GDC only builds dub from source.

Geod24:
Looking at the CI, this is actually currently red on master. But this PR fixes a failure in the Github Actions (because it fails to link without the fix in this PR, so force merging.

PR #1992, opened 2020-08-09
Update supported compiler versions

Geod24:
Update LDC used for release from v1.20 to v1.22,
move coverage target to be the latest compiler
(so hopefully the less buggy too),
and make sure we only support the last 10 releases.


Geod24:
Okay, rebased on master, this should fix it. Master is currently broken on one Travis-ci dimension (the v2.076).

PR #1994, opened 2020-08-13
Update Vibe.d to v0.9.0 in test_registry & `dub init`

Geod24:
Needed for https://github.com/dlang/phobos/pull/7545

### DIFF
--- a/source/dub/init.d
+++ b/source/dub/init.d
@@ -113,7 +113,7 @@ void main()
 private void initVibeDPackage(NativePath root_path, ref PackageRecipe p, scope void delegate() pre_write_callback)
 {
 	if ("vibe-d" !in p.buildSettings.dependencies)
-		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.8.2");
+		p.buildSettings.dependencies["vibe-d"] = Dependency("~>0.9");
 	p.description = "A simple vibe.d server application.";
 	pre_write_callback();
 
--- a/test/test_registry.d
+++ b/test/test_registry.d
@@ -1,6 +1,6 @@
 #!/usr/bin/env dub
 /+dub.sdl:
-dependency "vibe-d" version="~>0.8.5"
+dependency "vibe-d" version="~>0.9"
 versions "VibeNoSSL"
 +/
 


jblachly:
@Geod24 do you have commit access to dub as well? `dub init -t vibe.d` still emits version 0.8.6 as well.

Geod24:
Thanks for the info, added an extra commit. Yes I have push access but I'm sure someone will be able to review this today.

jblachly:
Didn't realize we were already in dub repo *heh* -- thanks for adding the extra commit to the PR

Geod24:
But it's `dub init`. You won't break anything by creating a new repo.

WebFreak001:
even if it's a new repo, new vibe.d updates will be released with the code that was once written based on this generated `dub init` then. Furthermore seeing the `~>0.x` syntax will promote its usage possibly also by people without knowledge that it means `>=0.x.0 <1.0.0` and how minor versions in SemVer may break code with a major v0

The `~>0.x.y` syntax is more stable for that as it means `>=0.x.y <0.z.0` where z is x+1 (`~>0.9.0` => `>=0.9.0 <0.10.0`)

Geod24:
> The `~>0.x.y` syntax is more stable for that as it means `>=0.x.y <0.z.0` where z is x+1 (`~>0.9.0 => >=0.9.0 <0.10.0`)

That's not quite true:

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

The argument for not having `~>0.x` is that  according to SemVer, major == 0 means "anything goes". So that same reasoning applies to patch releases.

andre2007:
Regarding the ctrl+c issue of vibe-d (still reproducible with vibe-d 0.9.0 according to a user in the dlang forum), what do you think about defaulting `VibeHighEventPriority` in dub init?

wilzbach:
> So that same reasoning applies to patch releases.

Yes, but in practice patch releases are more stable.

Anyhow, I don't really think it matters and 0.9 means that we don't have to update dub constantly (or won't break when it can't be updated).

wilzbach:
> what do you think about defaulting VibeHighEventPriority in dub init?

It would take months for most users to update to the newest dub release (the next release this could be in is 1.11). And no one immediately updates their compiler, so I would say at least half a year until this has an impact on beginners. Furthermore, I think they would be confused because tutorials don't mention the version flag.
So I think the issue should just be fixed or in the worst case vibed's default configuration could default to adding the version flag.

Geod24:
So what do I do with  this PR ?

PR #1997, opened 2020-08-24
Updates to Github CI setup

Geod24:
I think this is why master is currently broken (first commit). Second is just bonus.


Geod24:
Yeah that was it, will merge when green

PR #1999, opened 2020-08-26
Make 'dub remove' remove all packages by default in non-interactive mode 

Geod24:
This is a new feature that simplify writing scripts a tiny bit.
Also included one fix from my Alpine PR since otherwise the lack of `--version` could error out.

PR #1987, opened 2020-07-28
Added version spec ability for dub commands

deviator:
fix #1979
ping @andre2007

PR #1998, opened 2020-08-25
CI: Add test for building on Alpine Linux / Musl

Geod24:
Tried to upgrade the Alpine Musl package, realized it was broken. So might as well do this.
It also found some issues in the testsuite, so it's an instant win.
This will fail the CI, just putting it up there in case someone is interested / has suggestions.

PR #2000, opened 2020-08-26
Add a changelog entry and cleanup test suite for --version deprecation

Geod24:
PR 2000 🎉 
Based on #1999 (first two commits) for simplicity, but can easily remove it if people reject it.

The changelog / cleanup was missing from #1987 (my bad for merging it).
CC @deviator

### DIFF
--- a/changelog/version.dd
+++ b/changelog/version.dd
@@ -1,0 +1,8 @@
+All commands now support `package[@<version-spec>]`, `--version` is deprecated
+
+`dub` commands had some disparity in their version handling.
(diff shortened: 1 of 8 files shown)


Geod24:
Fixed the failures. That should also help with #1998 .

PetarKirov:
Nicely done! One question: in some places you removed `--non-interactive`, was that intentional?

Geod24:
Yes: When doing #1999 I realized that `dub remove PKG --version=X` was never interactive anyway (check the code).

PR #2001, opened 2020-08-26
Remove redundant COND || false statement in tests

Geod24:
I'm seeing some weird failures in my other PRs so splitting this to make sure `master` still works...


thewilsonator (approved):

Geod24:
Looking at https://github.com/dlang/dub/commit/33d7047f4378b8ad14f776206d582c244c064482 I got a bad feeling...

wilzbach (approved):
Yeah this works because of `set -e` in `comon.sh`

Geod24:
We should really fix the bot to take actions into account...

PR #2002, opened 2020-08-26
version-spec: Use non-obsolete syntax for head/tail

Geod24:
```
This syntax is hardly documented, with the Linux manpage saying:
> For backwards compatibility, `-number` is equivalent to `-n number`.
http://schillix.sourceforge.net/man/man1/head.1.html
While some other man pages either don't mention it or call it obsolete.
For maximum compatibility, just use `-n 1`.
```

Doubt that's the cause of the failure, but had to go to stackoverflow to learn about this syntax: https://unix.stackexchange.com/questions/467157/why-isnt-head-1-equivalent-with-head-n-1-but-instead-its-the-same-as-h

### DIFF
--- a/test/version-spec.sh
+++ b/test/version-spec.sh
@@ -5,29 +5,29 @@
 $DUB add-local "$CURR_DIR/version-spec/newfoo"
 $DUB add-local "$CURR_DIR/version-spec/oldfoo"
 
-[[ $($DUB describe foo | grep path | head -1) == *"/newfoo/"* ]]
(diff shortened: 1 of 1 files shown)


wilzbach (approved):
I don't think that's the cause as it only fails with the dmd-master build.

Geod24:
Yeah me neither, just wanted to weed out the oddity. Also, sometimes it's LDC master...

Geod24:
Going ahead with this since you approved.

Issue #1991, opened 2020-08-07
Add dub where command


schveiguy:
This is an enhancement.

A really simple, yet useful feature, would be to provide a path to where a dub-built executable exists.

As an example of what go does, see this forum discussion: https://forum.dlang.org/umgzqzybzpenguobnoog@forum.dlang.org

I propose a `dub where` command which will return the project bin path.

This way someone can do:

```sh
> dub build foo
> export PATH=$PATH:`dub where foo`
> foo --help
```

wilzbach:
I'm not arguing against this feature (it has good use cases), but I'm not sure you would want to use it for exposing dub program executables:
- as upgrading or downgrading versions will be hard (in fact it even applies to keeping the current version as dub by default takes the newest version so you would need to manually add `dub where foo@1.9.2` to prevent accidental breakage when installing other packages)
- you will need to manually run add executable paths to your `PATH` for every new program

So I think  what the original poster probably wants is sth. like `dub install foo` which by default copies (or symlinks) the executable(s) in `~/.local/bin` (like e.g. Python) or in `~/.dub/bin`. It's also the most-wished-for feature-request at dub
-> https://github.com/dlang/dub/issues/839. It gets a lot more complicated with libraries, but an initial implementation which focuses only on binaries should be a lot less controversial and might get the ball running.

schveiguy:
dub install is doable with this command as well, as you can copy the file using the path:

```sh
> cp `dub where foo`/* ~/bin
```

But my thought is that this is a simple thing that doesn't need to deal with how the user's environment is set up -- dub knows where it puts files, it's just telling you that, so you can do whatever you want with them (install, run, etc). It should be a very small change to dub, that is not prone to breakage.

Definitely, there are some subtleties. The versioning thing is something that might be difficult to deal with.

andre2007:
I just wonder whether we need an additional command or `dub describe` can be enhanced to return this information.
Actually I am not sure wheter is isn't already implemented as the documentation lists `target-path` as value for `--data-list` argument.
Dub describe has also version support.

(Still we need a install command)

PR #2004, opened 2020-08-29
added run/test '--gdb' command line argument

foerdi:
If `--gdb` is used, then the executable will be started with gdb attached.
The gdb executable can be specified with the DEBUGGER environment variable.

### DIFF
--- a/changelog/gdb.dd
+++ b/changelog/gdb.dd
@@ -1,0 +1,4 @@
+added run/test `--gdb` command line argument
+
+If `--gdb` is used, then the executable will be started with gdb attached.
+The gdb executable can be specified with the DEBUGGER environment variable.
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1080,6 +1080,7 @@ class GenerateCommand : PackageBuildCommand {
 		bool m_rdmd = false;
 		bool m_tempBuild = false;
 		bool m_run = false;
+		bool m_gdb = false;
 		bool m_force = false;
 		bool m_combined = false;
 		bool m_parallel = false;
@@ -1164,6 +1165,7 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.combined = m_combined;
 		gensettings.filterVersions = m_filterVersions;
 		gensettings.run = m_run;
+		gensettings.gdb = m_gdb;
 		gensettings.runArgs = app_args;
 		gensettings.force = m_force;
 		gensettings.rdmd = m_rdmd;
@@ -1303,6 +1305,10 @@ class RunCommand : BuildCommand {
 			"Builds the project in the temp folder if possible."
 		]);
 
+		args.getopt("gdb", &m_gdb, [
+			"Attach gdb. The gdb executable can be specified with the DEBUGGER environment variable (default 'gdb' is used)"
+		]);
+
 		super.prepare(args);
 		m_run = true;
 	}
@@ -1319,6 +1325,7 @@ class TestCommand : PackageBuildCommand {
 		bool m_combined = false;
 		bool m_parallel = false;
 		bool m_force = false;
+		bool m_gdb = false;
 	}
 
 	this() @safe pure nothrow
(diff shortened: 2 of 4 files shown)


wilzbach:
Hi. Thanks a lot for your PR.
Two quick things:
1) it would be great to have a simple test to avoid regressions
2) I often use gdb with custom arguments, so this should be supported

More importantly I think this approach is just working around the actual problem: not being able to access the generated executable easily. What about a user who wants to use valgrind? Or perf?

Geod24 (changes requested):
I think this is too specific. GDB only works on Linux (not on OSX / Windows) and those platforms would need their own flag. Plus, as mentioned by @wilzbach , you can just `dub build` and access the executable.

foerdi:
I was afraid that this was too specific. Here are two alternative:

1. `dub run --show-executable` then everyone can create their own tools or commands. like: `gdb --args $(dub run --show-executable)`
2. Something like: `dub run --wrapper|-w`, with `--wapper`  multiple arguments can be given, first one is the wrapper executable and the others are are arguments for the wrapper. like: `dub run -w gdb,--args`

Would that be a better approach?

@Geod24: I know I can just `dub build` and then access the binary. This PR aims to achieve some user improvements. In my case: I have a larger multi package project with a common target destination. My current workflow is:
1. `dub test`
2. `cgdb ../../foo/bar/exec`
3. code fixes and go to 1

With this PR my workflow is reduced to:
1. `dub test --gdb` (or friends)
2. code fixes and go to 1

wilzbach:
> dub run --show-executable then everyone can create their own tools or commands. like: gdb --args $(dub run --show-executable)

1) Shouldn't it be `dub build`? (as dub is no longer running the executable)
2) I'm not sure this is a good idea as this would easily break when there's a compile error. 

We have something similar for the install script (the `-a` flag), and it's not a good UX as currently you don't see what's happening, so I'm a bit reluctant to add something which is entirely silent. Dub would need to ensure all build output is redirected to stderr when such a flag is added.

A simpler solution would to implement the already requested `where` command: https://github.com/dlang/dub/issues/1991
It would then be: `dub build && gdb --args $(dub where)`.

> Something like: dub run --wrapper|-w, with --wapper multiple arguments can be given

This looks rather weird. Is there any other package manager that supports something like this?
I think a good approach would be to look at a few common package managers and see how they deal with this problem?

> My current workflow is:

Are you aware of `dub build -c unittest`?

```
dub build -c unittest && cgdb ./<project>-test-<config>
```

> cgdb ../../foo/bar/exec

BTW are you aware of `-checkaction=context` and `-verrors=context` (this is for build errors)? At least the first one might reduce the need to use a debugger?

Geod24:
> Shouldn't it be dub build? (as dub is no longer running the executable)

Exactly my thought. Docker does something similar when you use `-q` IIRC, it only gives you the hash of the executable, which you can use to run as part of a command (using `whatever $(docker build -q .) args`);

foerdi:
My initial inspiration was `coredumpctl debug` from `systemd-coredump`, this start a debugger session form a specified coredump. For example. :`coredumpctl debug -1 --debugger cgdb` start cgdb with the last coredump.

> 1. Shouldn't it be dub build? (as dub is no longer running the executable)
> 2. I'm not sure this is a good idea as this would easily break when there's a compile error.

Indeed.

> BTW are you aware of -checkaction=context and -verrors=context (this is for build errors)? At least the first one might reduce the need to use a debugger?

Yes. I know these options, but last time I tried that, it was not working in my `@nogc` environment, Maybe I should try it again.
Nonetheless I think a option to start a debugger from dub would be a nice usability feature.

> This looks rather weird.

Indeed, this was a fast Idea. First I was thinking about `--wrapper gdb --wrapper-args -d,src,--args` this is more verbose and looks also weird (the part of `--wrapper-args`). The `--wrapper gdb,-d,src,--args` was a simplified version of this.

A different option would be like `find -exec`: `--wrapper gdb -d "src" --args \;` but this needs bigger changes to the argument parser.

At last I was thinking about a environment variable solution, like `DUB_WRAPPER_ARGS='-d "src" --args' dub test -w gdb`, the difficulty here is the correct parsing and splitting of `DUB_WRAPPER_ARGS` variable.
A statically solution could be to store the extra arguments in to the dub config file, everyone could setup their own dub wrappers and execute them with `dub test -w gdb` or `dub run -w valgrind`.

This are some Ideas, I'm curious what you think about it.
If this is to specific in general or not suitable for dub, than I can understand this and I will think more about `dub where` and friends.

> Are you aware of dub build -c unittest?

Did you mean `dub build -b unittest`? This is not working for a library. And I need to type the whole path again ;).

foerdi:
Maybe I will work on a suitable solution in the future, but for now I will close this PR.

PR #2013, opened 2020-09-17
Improve output of toolchain requirements test, workaround heisenbug

Geod24:
```
There is currently an Heisenbug in Github CI where dub fails to run the program.
E.g. we see the following in the logs:
---
Performing "debug" build using ldc2 for x86_64.
dubtest1531 ~master: building configuration "application"...
Linking...
Running ../../../../../tmp/dubtest1531_2enU78/dubtest1531
Failed to execute program (No such file or directory)
[ERROR] -e:79 Did not pass with ldc="<=1.23.0"
[ERROR] Script failure.
---
This works around said bug by only building the package,
but not running it.
```

Also using `-q` to reduce noise.

### DIFF
--- a/test/issue1531-toolchain-requirements.sh
+++ b/test/issue1531-toolchain-requirements.sh
@@ -75,12 +75,14 @@ EOF
 
 # pass test compiler requirement given as $1
 function test_cl_req_pass {
+    echo "Expecting success on '$DC $1'" 2>&1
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

Geod24:
Ergh we should set those status as required...

PR #2016, opened 2020-09-28
[stable] Deduplicate dflags and lflags

Geod24:
Before:
Screen Shot 2020-09-28 at 12 03 40

After:
Screen Shot 2020-09-28 at 12 04 49

There are still some improvements needed (notice the duplicated `-L-framework`), but the diff / benefit ratio is already great.
Packed in another tiny cleanup to avoid endless CI retriggers.
Targeting stable because I'm working on propagating dflags downwards and obviously this will come in handy.

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -89,10 +89,10 @@ struct BuildSettings {
 		addPostRunCommands(bs.postRunCommands);
 	}
 
-	void addDFlags(in string[] value...) { dflags ~= value; }
+	void addDFlags(in string[] value...) { add(dflags, value); }
 	void prependDFlags(in string[] value...) { prepend(dflags, value); }
 	void removeDFlags(in string[] value...) { remove(dflags, value); }
-	void addLFlags(in string[] value...) { lflags ~= value; }
+	void addLFlags(in string[] value...) { add(lflags, value); }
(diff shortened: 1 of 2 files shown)


Geod24:
I believe that was also a pet peeves of yours @UplinkCoder

thewilsonator (approved):

s-ludwig:
But `-L-framework` *needs* to be duplicated for every framework - won't this result in a wrong outcome?

`lflags "-L-framework" "-LCoreServices" "-L-framework" "-LCoreFoundation"`

should get `-L-framework -LCoreServices -LCoreFoundation` with these changes, meaning that "CoreFoundation" is passed to the linker as a normal file argument instead of as a framework name.

As far as I can see, this operates on the wrong level and needs to be reverted.

Geod24:
Look at the second screenshot: It is still duplicated.
However, I didn't knew this bit  about `-L-framework`, and I have no idea why this is still duplicated. I'll take a look.

s-ludwig:
Yes, I've seen the screenshot, but that does not look like the outcome of the algorithm, but of some coincident external effect.

s-ludwig:
By the way, I know how annoying it is to miss a release, but in this case it was quite lucky that it came too late for the latest release. It really shouldn't have gone to the stable branch ;-)

PR #2020, opened 2020-10-01
Revert "Deduplicate dflags and lflags" (#2016)

s-ludwig:
PR #2016 introducted a de-duplication strategy that is too eager and corrupts argument chains, such as "-framework CoreServices -framework CoreFoundation" where the "-framework" argument needs to be duplicated for each framework.

This reverts commit 10bea6f07fa460c806cbec56323d600af00d9f09.


Geod24 (approved):
Right. Sorry I didn't have time to look at it. Let's revert and I'll try to come up with a better solution.

PR #2030, opened 2020-10-22
ci: build release .tar.gz without travis owner and group id

ljmf00:

---

The current release `.tar.gz` have `dub` binary under `travis:travis` owner. We should set it as `root:root` which is described as anonymous on GNU tar documentation . https://www.gnu.org/software/tar/manual/html_section/tar_33.html#SEC69

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,11 @@ jobs:
     - stage: deploy
       d: ldc-1.22.0,dub
       os: osx
+      addons:
+        homebrew:
+          packages:
+            - gnu-tar
+          update: true
       script: echo "Deploying to GitHub releases ..." && ./scripts/ci/release.sh
       deploy:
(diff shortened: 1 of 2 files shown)


ljmf00:
CI is falling because seems that macOS doesn't use GNU tar and use BSD tar instead. There's a simple way to use GNU tar on macos in travis-ci ? I'm not familiar with macOS.

Geod24:
Try to `brew install gnu-tar` in the `.travis.yml` and use `gtar` in the script.

ljmf00:
> Try to `brew install gnu-tar` in the `.travis.yml` and use `gtar` in the script.

@Geod24 Done. Can you review again?

wilzbach (approved):
Looks good and makes sense. Thanks a lot!

PR #2033, opened 2020-11-11
dub.compilers.ldc: Fix --rdmd with --compiler=ldc2

CyberShadow:
Use two dashes to make rdmd ignore the option. ldc2 doesn't care.

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -112,7 +112,7 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 		}
 
 		// since LDC always outputs multiple object files, avoid conflicts by default
-		settings.addDFlags("-oq", "-od=.dub/obj");
+		settings.addDFlags("--oq", "-od=.dub/obj");
 
 		if (!(fields & BuildSetting.versions)) {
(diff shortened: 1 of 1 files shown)


Geod24:
Test ?

CyberShadow:
How to test this?

Do you know an existing test that I can start from?

Geod24:
Just drop a shell script in `test`. Look at any shell-script there to see what they do.

CyberShadow:
OK, and the test should assume that ldc and rdmd are installed? Or should it mock them somehow?

I'm looking at `.travis.yml` and it seems the D compiler is part of the matrix? So the test should check if ldc is installed and vacuously succeed otherwise? Or is that just for building Dub?

Geod24:
> OK, and the test should assume that ldc and rdmd are installed? Or should it mock them somehow?

You can check they are installed via an environment variable.

> So the test should check if ldc is installed and vacuously succeed otherwise?

Indeed.

But taking a second look at it, it's quite painful to set up (there's also, like, tons of issues with it).
I'll just merge this, and if you feel like adding tests in the future, you're very welcome.

Geod24 (approved):

PR #2035, opened 2020-11-17
Fix Issue #1474 - Regenerate build settings after preGenerateCommands

omerfirmak:
Regenerate buildSettings after preGenerateCommands.
Handle cases where `filterDuplicates` input 'vals` has duplicates.

### DIFF
--- a/examples/generated-sources/dub.json
+++ b/examples/generated-sources/dub.json
@@ -1,11 +1,7 @@
 {
 	"name": "generated-sources",
 	"description": "Example of using pre generate commands to generate source code.",
-	"sourceFiles": [ "source/test.d" ],
 	"preGenerateCommands": [
-		"echo 'Currently, a workaround is necessary to support source-code generate through pre*Commands'",
-		"echo 'Generated source files need to be added to sourceFiles, otherwise they wont be picked up on first run'",
-		"echo 'See dlang/dub issue 1474 for more details / discussions.'",
 		"echo 'int fun() { return 42; }' > source/test.d"
 	]
 }
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -133,7 +133,9 @@ private:
 	// Append vals to arr without adding duplicates.
 	static void add(ref string[] arr, in string[] vals, bool noDuplicates = true)
 	{
-		arr ~= filterDuplicates(arr, vals, noDuplicates);
+		// vals might contain duplicates, add each val individually
+		foreach (val; vals)
+			arr ~= filterDuplicates(arr, [val], noDuplicates);
 	}
 
 	unittest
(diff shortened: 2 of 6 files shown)


Geod24:
Previous attempt: https://github.com/dlang/dub/pull/1708

omerfirmak:
> Previous attempt: #1708

Yep, forgot to mention that. #1708 fails to populate `targets` with the updated `buildSettings`. It just updates the local copy.

Geod24:
Can you rebase on master please ?

source/dub/compilers/buildsettings.d
```diff
@@ -133,7 +133,9 @@ private:
 	// Append vals to arr without adding duplicates.
 	static void add(ref string[] arr, in string[] vals, bool noDuplicates = true)
 	{
-		arr ~= filterDuplicates(arr, vals, noDuplicates);
+		// vals might contain duplicates, add each val individually
+		foreach (val; vals)
+			arr ~= filterDuplicates(arr, [val], noDuplicates);
```
PetarKirov, line 138: Please add a unittest covering this case.

source/dub/compilers/buildsettings.d
```diff
@@ -148,7 +150,10 @@ private:
 	// Prepend arr by vals without adding duplicates.
 	static void prepend(ref string[] arr, in string[] vals, bool noDuplicates = true)
 	{
-		arr = filterDuplicates(arr, vals, noDuplicates) ~ arr;
+		import std.algorithm.mutation : reverse;
+		// vals might contain duplicates, add each val individually
+		foreach (val; vals.dup.reverse)
```
PetarKirov, line 155: Would the following work:

```suggestion
		import std.range : retro;
		// vals might contain duplicates, add each val individually
		foreach (val; vals.retro)
```

If it it does, it should avoid the memory allocation caused by `.dup`.

PetarKirov (commented):
Looks good to me, modulo these comments:

source/dub/compilers/buildsettings.d
omerfirmak: Done

source/dub/compilers/buildsettings.d
omerfirmak: Yep, it works. Thanks!

omerfirmak:
> Can you rebase on master please ?

Done

examples/generated-sources/dub.json
```diff
@@ -1,11 +1,7 @@
 {
 	"name": "generated-sources",
 	"description": "Example of using pre generate commands to generate source code.",
-	"sourceFiles": [ "source/test.d" ],
```
Geod24, line 4: Could you add a test with globbing as well ?
e.g.:
```json
"sourceFiles": [ "something/*.d" ]
```

And have 2 modules there, one that is already present, one that is generated afterwards.
Those two modules should have an `extern(C) void someModulePrefix_func() {}` function, that is forward declared in the main, to avoid the need for an `import` (which would then require you to use another `dub` directive).

Geod24:
Perfect, thanks! Just extend the test, and we're good to go.

omerfirmak:
> Perfect, thanks! Just extend the test, and we're good to go.

Done! :+1:

Geod24 (approved):

veelo:
This is great. Thank you!

veelo:
Are dub fixes part of the dmd release changelog? This is definitely worth a mention.

Geod24:
Only if they come with a changelog entry at the moment, e.g. https://dlang.org/changelog/2.094.0.html

Issue #279, opened 2014-03-12
[enh] Generic way to hijack dependencies' dflags


Geod24:
There's currently no way to change a dependency dflags, other than changing the said dependency dub.json.
That is problematic for specific use case:
- If one wants to use shared libraries, the code has to be PIC. I'm currently writing a PAM module which use vibe.d, and I had to fork it, branch it, and add a libevent-pic configuration. So that currently forces library writer to provide at least 3 versions of said library: static, static PIC, dynamic.
- Aside from PIC, one may want to disable specific features (like  "-mno-sse" on gdc).

It could be useful if we had a way to edit [d|l]flags  (add / remove / do it from scratch) in the "subConfiguration".

Geod24:
Solved by #2040

PR #2040, opened 2020-11-25
Allow custom dflags to be defined for dependencies

omerfirmak:
For example;
```
{
	"authors": [
		"omer"
	],
	"copyright": "Copyright © 2020, omer",
	"description": "A simple vibe.d server application.",
	"license": "proprietary",
	"name": "dubtest",

	"dependencies": {
		"vibe-d": { "version" : "~>0.9.2", "dflags" : ["-preview=in"] }
	}
}
```

`-preview=in` will be applied to `vibe-d` and all of its dependencies. Any `BuildSettings` field can be used, parsing support is present. But, for now only, `dflags` is taken into account when compiling.

### DIFF
--- a/changelog/dependency-build-settings.dd
+++ b/changelog/dependency-build-settings.dd
@@ -1,0 +1,15 @@
+Allow custom build settings to be defined for dependencies
+
+For example:
+---
+{
+    "name": "example",
+    "dependencies": {
+        "vibe-d": { "version" : "~>0.9.2", "dflags" : ["-preview=in"] }
+    }
+}
+---
+
+In this example, `-preview=in` will be applied to `vibe-d` and all of its dependencies.
+Any $(LINK2 build settings, https://dub.pm/package-format-json.html#build-settings) field will be parsed,
+however only `dflags` is taken into account when compiling for now.
--- a/test/depen-build-settings/dub.json
+++ b/test/depen-build-settings/dub.json
@@ -1,0 +1,8 @@
+{
+	"description": "A minimal D application.",
+	"name": "depen-build-settings",
+
+	"dependencies": {
+		"depend1": { "version" : "*", "dflags" : ["-X"] }
+	}
+}
--- a/test/depen-build-settings/source/app.d
+++ b/test/depen-build-settings/source/app.d
@@ -1,0 +1,12 @@
+import std.stdio;
+import std.file;
+
+extern (C) void depend1_func();
+
+void main()
+{
+	writeln("Edit source/app.d to start your project.");
+	depend1_func();
+	assert(exists("depend2.json"));
+	assert(exists("depend.json"));
+}
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -314,20 +314,45 @@ class ProjectGenerator
 		visited.clear();
 
 		// 1. downwards inherits versions, debugVersions, and inheritable build settings
-		static void configureDependencies(const scope ref TargetInfo ti, TargetInfo[string] targets, size_t level = 0)
+		static void configureDependencies(const scope ref TargetInfo ti, TargetInfo[string] targets,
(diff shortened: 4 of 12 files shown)


omerfirmak:
Would this need a changelog entry?

Geod24:
Definitely. CC @s-ludwig , how do you feel about this ? The issue is that currently, we cannot propagate flags from the parent packages to the child packages. Things like `-fPIC` or `-preview` switches.

Geod24:
And a test :)

test/depen-build-settings/dub.json
```diff
@@ -0,0 +1,8 @@
+{
+	"description": "A minimal D application.",
+	"name": "depen-build-settings",
+
+	"dependencies": {
+		"depend1": { "version" : "*", "dflags" : ["-version=depen_build_settings"] }
```
Geod24, line 6: Using `-version` is a pretty bad example, because DUB handles it, and will give you a warning on this.
Use something like `-fPIC` instead, something that needs to be propagated and won't go away.

Geod24:
Also the test needs to make sure that the `dflags` is transitive otherwise it's not of much use to the user.

Geod24:
Would solve https://github.com/dlang/dub/issues/279

test/depen-build-settings/dub.json
omerfirmak: I dont see any warning while building, at which should I receive it? Both `depen1` and `depend2` need `-version=depen_build_settings` to successfully compile. If I remove or change it to something else, compilation fails as expected.

`depen1` depends on `depen2` so, test also checks that the flags are applied transitively to all of the dependency chain.

Geod24:
```json
{
    "name": "mytest",
    "dflags": [ "-version=Hello" ]
}
```

Leads to:
```console
$ dub build                                                                                                                                                          :(

## Warning for package mytest ##

The following compiler flags have been specified in the package description
file. They are handled by DUB and direct use in packages is discouraged.
Alternatively, you can set the DFLAGS environment variable to pass custom flags
to the compiler, or use one of the suggestions below:

-version=Hello: Use "versions" to specify version constants in a compiler independent way

mytest ~master: building configuration "application"...
Linking...
```

andre2007:
Changing the dflags of a dependency should also trigger a new compilation of the dependency. Maybe this a check for this could also be added to the test.

test/depen-build-settings/source/app.d
```diff
@@ -0,0 +1,12 @@
...
+import std.file;
+
+extern (C) void depend1_func();
+
+void main()
+{
+	writeln("Edit source/app.d to start your project.");
+	depend1_func();
+	assert(exists("depend2.json"));
+	assert(exists("depend.json"));
```
Geod24, line 11: Gitignore them ?

Geod24 (approved):
LGTM, any opinion @PetarKirov @s-ludwig @wilzbach ?

changelog/dependency-build-settings.dd
```diff
@@ -0,0 +1,18 @@
...
+{
+	"description": "A simple vibe.d server application.",
+	"license": "proprietary",
+	"name": "dubtest",
+
+	"dependencies": {
+		"vibe-d": { "version" : "~>0.9.2", "dflags" : ["-preview=in"] }
+	}
+}
+`
```
Geod24, line 14:
```suggestion
---
{
	"description": "A simple vibe.d server application.",
	"license": "proprietary",
	"name": "dubtest",

	"dependencies": {
		"vibe-d": { "version" : "~>0.9.2", "dflags" : ["-preview=in"] }
	}
}
---
```

And  there seems to be some tabs in there

Geod24:
Fixed the changelog (and slightly improved it), merging since no opposition was voiced.

Geod24:
@omerfirmak : This is missing the implementation for SDL.

PR #2041, opened 2020-11-26
Make `sourcePaths` work as documented

### LINKED ISSUE #1913: `sourcePaths` does not work as documented
### System information
- **dub version**: DUB version 1.20.0, built on Mar  9 2020
- **OS Platform and distribution**: Mac OSX 10.15.4
- **compiler version**: DMD 2.091.0

```console
% dub init dub-bug -n
Successfully created an empty project in '/Users/geod24/projects/dub-bug'.
Package successfully created in dub-bug
% cd dub-bug
% mkdir source/app1/ source/app2
% echo "void main () {}" > source/app1/main.d
% echo "void main () {}" > source/app2/main.d
% rm source/app.d
```

dub.json edited to:
```JSON
{
    "name": "dub-bug",
    "configurations": [
        {
            "name": "app1",
            "targetName": "bug-app1",
            "sourcePaths": [ "source/app1" ]
        },
        {
            "name": "app2",
            "targetName": "bug-app2",
            "sourcePaths": [ "source/app2" ]
        }
    ]
}
```

Result:
```console
% dub build -c app1
dub-bug ~master: building configuration "app1"...
Error: module main from file source/app2/main.d conflicts with another module main from file source/app1/main.d
dmd failed with exit code 1.
% dub build -c app2
dub-bug ~master: building configuration "app2"...
Error: module main from file source/app2/main.d conflicts with another module main from file source/app1/main.d
dmd failed with exit code 1.
```

However, according to [the documentation](https://dub.pm/package-format-json.html#build-settings):
> Allows to customize the path where to look for source files (any folder "source" or "src" is automatically used as a source path **if no `sourcePaths` setting is specified**)

Emphasis mine. This only works if `sourcePaths` is at the top level:
```
{
    "name": "dub-bug",
    "targetName": "bug-app1",
    "sourcePaths": [ "source/app1" ]
}
```

Compiles successfully.

omerfirmak:
`sourcePaths` is only default initialized when both
top level and configuration level `sourcePaths` is empty.

Fixes #1913

### DIFF
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -671,55 +671,62 @@ class Package {
 			}
 		}
 
-		// check for default source folders
-		immutable hasSP = ("" in bs.sourcePaths) !is null;
-		immutable hasIP = ("" in bs.importPaths) !is null;
-		if (!hasSP || !hasIP) {
-			foreach (defsf; ["source/", "src/"]) {
-				if (existsFile(m_path ~ defsf)) {
-					if (!hasSP) bs.sourcePaths[""] ~= defsf;
-					if (!hasIP) bs.importPaths[""] ~= defsf;
-				}
-			}
-		}
-
-		// check for default app_main
-		string app_main_file;
-		auto pkg_name = m_info.name.length ? m_info.name : "unknown";
-		foreach(sf; bs.sourcePaths.get("", null)){
-			auto p = m_path ~ sf;
-			if( !existsFile(p) ) continue;
-			foreach(fil; ["app.d", "main.d", pkg_name ~ "/main.d", pkg_name ~ "/" ~ "app.d"]){
-				if( existsFile(p ~ fil) ) {
-					app_main_file = (NativePath(sf) ~ fil).toNativeString();
-					break;
-				}
-			}
-		}
-
 		// generate default configurations if none are defined
 		if (m_info.configurations.length == 0) {
 			if (bs.targetType == TargetType.executable) {
 				BuildSettingsTemplate app_settings;
 				app_settings.targetType = TargetType.executable;
-				if (bs.mainSourceFile.empty) app_settings.mainSourceFile = app_main_file;
 				m_info.configurations ~= ConfigurationInfo("application", app_settings);
 			} else if (bs.targetType != TargetType.none) {
 				BuildSettingsTemplate lib_settings;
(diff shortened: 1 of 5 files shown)


omerfirmak:
Apparently, some projects in buildkite relied on the old behavior.

Geod24:
> Apparently, some projects in buildkite relied on the old behavior.

Probably most people that use this will implicitly depend on this behavior.
@atilaneves what do you make of this change / issue ?

atilaneves:
I think that, unfortunately, breaking projects that relied on the old behaviour is a no-no. As I mentioned in [the issue itself](https://github.com/dlang/dub/issues/1913), putting both apps in, say, `foo/app1` and `foo/app2` instead of in a top-level `source` directory seems to sidestep this issue.

Geod24:
> I think that, unfortunately, breaking projects that relied on the old behaviour is a no-no.

Yeah, that's a given.

> As I mentioned in the issue itself, putting both apps in, say, foo/app1 and foo/app2 instead of in a top-level source directory seems to sidestep this issue.

The question was more about whether you considered it a valid issue as well,  being a user of the current behavior.
Just so there isn't one single person (me) providing feedback. If that's something you agree on, then we can talk about deprecating this behavior, by asking people to add `source` in their configurations as well.

atilaneves:
If it didn't break any projects I'd say ok, but it does, so...

rikkimax:
There are enough issues with dub doing things that people don't expect that this should be fixed.

But @atilaneves is right, it could break projects.

So how about a big honking warning (for projects it will break) for a couple of years and not make it go live until then?

PetarKirov:
I guess the simplest way to resolve matters like this is by adding a settings.json option to revert to the old behavior that would be available for a deprecation period of 2 years and could be enabled by default for next 3 releases or something like that.

s-ludwig:
> I guess the simplest way to resolve matters like this is by adding a settings.json option to revert to the old behavior that would be available for a deprecation period of 2 years and could be enabled by default for next 3 releases or something like that.

A global setting won't cut it, as it may need to be set in some packages, but must not be set in others. But adding a deprecation warning (along with mitigation instructions) would be a good idea in any case. We could also add that setting to the package recipe to allow silencing the warning by forcing the new behavior.

PR #2042, opened 2020-11-26
[CI]: Remove old AppVeyor, TravisCI, SemaphoreCI and build-gdc.sh files

PetarKirov:
As they're not used anymore.


PetarKirov:
@Geod24 Hmm, I still see Travis here:
image

Perhaps it has to be disabled from one more place?

PetarKirov:
As far as GDC support goes, we shouldn't keep one CI (semaphore) alive just for that. I suggest we include it in the GitHub Actions matrix and handle is specially (it's not supported by the [setup-dlang](https://github.com/dlang-community/setup-dlang) action) by installing it manually on some linux distro.

Geod24 (approved):

PR #2043, opened 2020-11-26
More precise platform variables

foerdi:
The environment variables `$PLATFORM` and `$PLATFORM_POSIX` contains now the most specific platform identifier.
As an example, `$PLATFORM` is now "android" instead of "linux" on a android target.

Before these changes, I couldn't separate my targetPath with `"../build-$PLATFORM-$ARCH/lib"` for different Posix platforms.
I am aware that these changes could affect other projects. Maybe this point needs a discussion?

Alternatively I could use `targetPath "../build-android-$ARCH/lib" platform="android"`, but I like the `$PLATFORM` variant more.


foerdi:
now I added some test cases

source/dub/project.d
```diff
@@ -1378,18 +1378,20 @@ private string getVariable(Project, Package)(string name, in Project project, in
 
 	if (name == "PLATFORM") {
 		import std.algorithm : filter;
-		foreach (p; gsettings.platform.platform.filter!(p => p != "posix"))
+		import std.range : retro;
+		foreach (p; gsettings.platform.platform[].retro.filter!(p => p != "posix"))
```
maxhaton, line 1382: Could you explain these changes to me?

PetarKirov:
I'm in favor of having the `$PLATFORM` variable be more specific, though this may break some projects.

An alternative solution would be to introduce a new variable that follows the new behavior in order to leave the old variable as it is.

@s-ludwig @Geod24 @schveiguy what so you think?

Geod24:
Yeah, unfortunately any change could break someone's workflow. Perhaps now would be a good time to use the `DUB_` prefix, e.g. `DUB_PLATFORM` ?

source/dub/project.d
foerdi: sure, in the `gsettings.platform.platform` Array is stored all platform specifier from the current target, in the least specified order.For example, if you are compiling a project on an Android target, the platform array will contain something like: `[ "posix", "linux", "android" ]`. By reversing this array we get the most specific platform first. We could build the array reversed but I don't know the source code well enough for side effects.

foerdi:
> Yeah, unfortunately any change could break someone's workflow. Perhaps now would be a good time to use the `DUB_` prefix, e.g. `DUB_PLATFORM` ?

`DUB_PLATFORM` could be confused with` PLATFORM`, maybe `DUB_PLATFORM_OS` is clearer?

source/dub/project.d
foerdi: oh my mistake, the filter is necessary because in the platform array is `"linux"` before `"posix"`, like:  [ "linux", "posix", "android" ]

maxhaton: If I understand what you're saying, and ignoring the [out of the scope of this PR] issue that neither Linux nor Android are technically POSIX compatible nor compliant, this is a bit of a kludge. If possible, assigning a priority (number) to each platform name would be a better (and still simple) solution, then simply take the highest priority (either in place or from some other table).

At very least it needs a comment explaining this code because the use of retro seems like it has the potential to be extremely fragile (Is this ordering actually enforced in dub, I couldn't find anywhere)? If you want to be fancy a DAG could be constructed at compile time specifying the relations between platforms but I wouldn't bother.

s-ludwig:
+1 for the `DUB_` prefix. Also I think that the contents should probably be a comma/(semi-)colon separated list of identifiers, rather than an arbitrary one that got picked out. While that makes the resulting scripts a little bit more complex, it is a general solution that follows what the compiler provides, instead of a very specific functionality that may be too specific or too broad for a particular application.

s-ludwig:
@maxhaton 
> ignoring the [out of the scope of this PR] issue that neither Linux nor Android are technically POSIX compatible nor compliant

Just note that this simply follows the compiler, which means that the POSIX assumption is deeply rooted in the runtime and many external packages. We really shouldn't change it here.

maxhaton:
> @maxhaton
> 
> > ignoring the [out of the scope of this PR] issue that neither Linux nor Android are technically POSIX compatible nor compliant
> 
> Just note that this simply follows the compiler, which means that the POSIX assumption is also deeply rooted in the runtime and many external packages. We really shouldn't change it here.

I wasn't pointing fingers!

foerdi:
> +1 for the `DUB_` prefix. Also I think that the contents should probably be a comma/(semi-)colon separated list of identifiers, rather than an arbitrary one that got picked out. While that makes the resulting scripts a little bit more complex, it is a general solution that follows what the compiler provides, instead of a very specific functionality that may be too specific or too broad for a particular application.

I understand your point and normally I would argue this myself, but my intention behind this MR is: to have a simple and nice way to describe an output directory for a multi-target project.

Would be an additional variable a solution? Maybe `DUB_PLATFORM_OS` for all platform Identifiers `-`-separated and `DUB_PLATFORM_OS_SHORT` for the most significant platform identifier. 
What do you think?

> At very least it needs a comment explaining this code because the use of retro seems like it has the potential to be extremely fragile.

That is absolutely correct, I'll add at least one comment on the place where the platform order is defined.

PR #2045, opened 2020-11-27
[stable] Use compiler version, not frontend version for build IDs and paths

kinke:
This prevents artifacts from being reused after upgrading the compiler from a beta to a final or to a newer patch release (well, I hope).


Geod24:
That will also prevent stale artifact when testing with a locally built compiler, right ?

kinke:
Hmm I don't think so - that might only have some `~dirty` suffix or so in its version string. The timestamp of the resolved compiler executable would make it truly unique I guess.

Geod24 (approved):

Issue #2012, opened 2020-09-17
Nested dub invocations behaves randomly


Geod24:
I have the following in my `dub.json` file:
```json5
{
    "preGenerateCommands": [ "$DUB source/scpp/build.d" ],
}
```
Permalink: https://github.com/bpfkorea/agora/blob/952cd01b14778d5ca5e34c049c8217c59037614f/dub.json#L14

This calls a build file that builds a C++ library. The build file itself doesn't use any of `dub`'s features, it's only here to be cross platform (because `rdmd` wasn't working well for a reason I forgot).
This is literally the package definition of the `build.d` file:
```D
#!/usr/bin/env dub
/+
 dub.json:
 {
     "name": "cpp_build"
 }
 +/
```
Permalink: https://github.com/bpfkorea/agora/blob/952cd01b14778d5ca5e34c049c8217c59037614f/source/scpp/build.d#L1-L7

I put a `-v` in my `$DUB` invocation just to see what happens, and lo and behold found a few issues.
This is the flags I see when the compiler is invoked:
```console
$ ldc2 -extern-std=c++14 -preview=in -extern-std=c++14 -c -ofsource/scpp/.dub/build/application-$DFLAGS-posix.osx.darwin-x86_64-ldc_2093-CEAA0795D0F4AF6821A572E766E19528/cpp_build.o -w -oq -od=.dub/obj -d-version=Have_cpp_build source/scpp/build.d -vcolumns
Linking...
ldc2 -ofsource/scpp/.dub/build/application-$DFLAGS-posix.osx.darwin-x86_64-ldc_2093-CEAA0795D0F4AF6821A572E766E19528/cpp_build source/scpp/.dub/build/application-$DFLAGS-posix.osx.darwin-x86_64-ldc_2093-CEAA0795D0F4AF6821A572E766E19528/cpp_build.o
```
My original call, in the parent package, was `dub test --compiler=dmd`. `dub` points to the one distributed with `dmd-2.094.0-beta.1`, through the install script. So by default this bug should select `dmd`, right ?

Issue 1: `ldc2` is used. Because my `dub.settings.json` uses `ldc2` as default compiler. Obviously I override this with `--compiler`, but the nested one doesn't pick it up. That might be a sensible behavior, but it's a very surprising one. Conversely, if I had no `dub.settings.json`, `dub` would always use `dmd` regardless of what `--compiler` I provide.
[...]

PR #2047, opened 2020-11-30
Use DC env variable as the default compiler

omerfirmak:
When DUB recursively invokes itself, it sets up the DC
env var with the compiler it is using. The `callee` dub should
use that compiler if it is not overriden with "--compiler". 

`DC=ldc2 dub build` is equivalent to `dub build --compiler=ldc2`
`DC=ldc2 dub build --compiler=dmd` would use `dmd`

This addresses the issue 1 mentioned in #2012

### DIFF
--- a/changelog/env-d-compiler.dd
+++ b/changelog/env-d-compiler.dd
@@ -1,0 +1,6 @@
+Use DC environment variable as default D compiler
+
+dub now respects the `DC` environment variable, meaning that `DC=ldc2 dub build` will behave as `dub build --compiler=ldc2`.
+In case both are supplied, the `--compiler` switch still has priority.
+Note that when DUB recursively invokes itself, for example in `preGenerateCommands`,
+it sets the `DC` variable to the compiler it is using, meaning that nested dub invocation will now use the same compiler.
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1380,7 +1380,11 @@ class Dub {
 		import std.process : environment;
 		import std.range : front;
 
-		m_defaultCompiler = m_config.defaultCompiler.expandTilde;
+		// Env takes precedence
+		if (auto envCompiler = environment.get("DC"))
+			m_defaultCompiler = envCompiler;
+		else
+			m_defaultCompiler = m_config.defaultCompiler.expandTilde;
 		if (m_defaultCompiler.length && m_defaultCompiler.isAbsolute)
 			return;
 
--- a/test/dc-env.sh
+++ b/test/dc-env.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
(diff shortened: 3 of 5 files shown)


omerfirmak:
Any ideas on how to unittest this?

PetarKirov:
Probably there are cleaner approaches, but the first thing that comes to my mind is this:

```D
void main()
{
    version (LDC)
        return 1;
    else version (DigitalMars)
        return 2;
    // and so on.
}

File: source/dub/dub.d [hunk 1]
```diff
@@ -1380,7 +1380,11 @@ class Dub {
 		import std.process : environment;
 		import std.range : front;
 
-		m_defaultCompiler = m_config.defaultCompiler.expandTilde;
+		// Env takes precedence
+		if (auto envCompiler = environment.get("DC"))
+			m_defaultCompiler = envCompiler;
+		else
+			m_defaultCompiler = m_config.defaultCompiler.expandTilde;
```
Geod24, line 1387: Does `env` take precedence, or is it just looked up first ? IIUC `--compiler` takes precedence over everything else ?

changelog/env-d-compiler.dd
```diff
@@ -0,0 +1,5 @@
+Use DC environment variable as default D compiler
+
+When DUB recursively invokes itself, it sets up the DC
+env var with the compiler it is using. The `callee` dub should
+use that compiler if it is not overriden with "--compiler".
```
Geod24, line 5:
```suggestion
dub now respects the `DC` environment variable, meaning that `DC=ldc2 dub build` will behave as `dub build --compiler=ldc2`. 
In case both are supplied, the `--compiler` switch still has priority.
Note that when DUB recursively invokes itself, for example in `preGenerateCommands`,
it sets the `DC` variable to the compiler it is using, meaning that nested dub invocation will now use the same compiler.
```

I think we should focus more on the general case than to what brought this issue.
It's really a bug that `DC` is not respected in the general case.

Geod24 (approved):
Another way to test it is:
```D
void main (string[] args)
{
    version (LDC)
        immutable expected = "ldc2";
    // ...
    assert(expected == args[1], format!"Expected '%s' but got '%s'"(expected, args[1]); 
}
```

source/dub/dub.d
omerfirmak: it takes precedence over the m_config. But --compiler overrides everything, yes.

omerfirmak:
Added the test.

changelog/env-d-compiler.dd
```diff
@@ -0,0 +1,6 @@
+Use DC environment variable as default D compiler
+
+dub now respects the `DC` environment variable, meaning that `DC=ldc2 dub build` will behave as `dub build --compiler=ldc2`.
```
Geod24, line 3: Trailing whitespace

omerfirmak: Fixed

omerfirmak:
Travis seems to be stuck.

PR #2048, opened 2020-11-30
Dont export DFLAGS & LFLAGS by default

omerfirmak:
DFLAGS and LFLAGS will not be exported as environment variables by default.
They can be accessed with $DFLAGS and $LFLAGS variables in dub.json

Eg;

`preGenerateCommands : ["DFLAGS=$DFLAGS env | grep DFLAGS"]`

will output DFLAGS environment variable with all the dflags used.

### DIFF
--- a/changelog/dont-propagate-flags.dd
+++ b/changelog/dont-propagate-flags.dd
@@ -1,0 +1,10 @@
+`DFLAGS` and `LFLAGS` no longer propagate to nested `dub` invocations
+
+`DFLAGS` and `LFLAGS` will no longer be exported as environment variables by default
+when invoking pre-generate, pre-build, pre-run, post-generate, post-build, or post-run commands.
+
+If the previous behavior is still desired, they can be accessed using `$DFLAGS` and `$LFLAGS` in dub.json
+E.g.:
+`preGenerateCommands : ["DFLAGS=$DFLAGS env | grep DFLAGS"]`
+
+will output DFLAGS environment variable with all the dflags used.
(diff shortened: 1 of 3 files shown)


changelog/dont-propagate-flags.dd
```diff
@@ -0,0 +1,10 @@
+Dont propagate DFLAGS and LFLAGS to sub commands by default
```
Geod24, line 1: The wording seems a bit off to me. "Don't ..." feels imperative and a little ambiguous. Something like "`DFLAGS` and `LFLAGS` are no longer propagated to [...]" sounds way more readable from a changelog PoV. Will edit myself.

source/dub/project.d
```diff
@@ -1396,6 +1396,15 @@ private string getVariable(Project, Package)(string name, in Project project, in
 
 	if (name == "BUILD_TYPE") return gsettings.buildType;
 
+	if (name == "DFLAGS" || name == "LFLAGS")
+	{
+		auto buildSettings = pack.getBuildSettings(gsettings.platform, gsettings.config);
+		if (name == "DFLAGS")
+			return join(cast(string[])buildSettings.dflags," ");
+		else if (name == "LFLAGS")
+			return join(cast(string[])buildSettings.lflags," ");
```
Geod24, line 1405: Those casts weren't needed. I don't know why they were here in the first place, maybe because before another type was used.Anyways, I removed them.

Geod24 (approved):

PR #2050, opened 2020-12-01
In single mode the current working directory is added to import paths by default

drug007:
Fix #2051 

This PR adds the current working dir to the list of import paths if user run `test` command in single file mode like:
```
dub test --single path/to/file.d
```
to let dub to find the file.

Important:
the file should not be named `main.d` because the file is imported as module `main` and there will be conflict between `static import main` and `function D main`:
```
/tmp/dub_test_root_test.d(9,12): Error: function D main conflicts with static import dub_test_root.main at /tmp/dub_test_root_test.d(3,15)
```

PR #2052, opened 2020-12-02
Fix #2051: Running unit tests from DUB single file packages fails

### LINKED ISSUE #2051: Running unit tests from DUB single file packages fails
### System information

- **dub version**: dub 1.17.0
- **OS Platform and distribution**: Debian 9
- **compiler version** dmd-2.088.1

### Bug Description

Running unit tests from DUB single file packages fails.
See [forum post](https://forum.dlang.org/post/rq5a3n$gcs$1@digitalmars.com) for additional information

### How to reproduce?
Running this:
```D
#!/usr/bin/env dub
/+ dub.sdl:
    name: "main"
+/

version(unittest) {}
else void main()
{
}

unittest
{
	auto input = [1721];
	assert(input[0] == 1721);
}
```
gives:
```
user@user:~/dub$ dub test --single ./test.d
Package main (configuration "application") defines no import paths, use {"importPaths": [...]} or the default package directory structure to fix this.
Generating test runner configuration 'test-test-application' for 'application' (executable).
Source file '/home/user/dub/test/test.d' not found in any import path.
```

### Expected Behavior
the result should be:
```
user@user:~/dub$dub test --single ./test.d
Package main (configuration "application") defines no import paths, use {"importPaths": [...]} or the default package directory structure to fix this.
Generating test runner configuration 'test-test-application' for 'application' (executable).
Performing "unittest" build using dmd for x86_64.
main ~master: building configuration "main-test-application" (default build policy: time)...
Linking...
Running ../../../../tmp/.dub/build/test-~master/test-test-application-unittest-linux.posix-x86_64-dmd_2088-9FDC9F7EC06E24A4F67D7D1E5B88EBD1/test-test-application 
All unit tests have been run successfully.
```

The reason of the bug is that in single mode dub, running test command, try to import the main source file as module and failed because can not find it in any import paths. The main source file should be excluded from this importing but current check for this fails because it can't compare absolute and relative paths correctly (it compares them by char).

drug007:
Performing test command dub imports every source file as module to build the project excluding the main source file. But this excluding may fail and in this case dub try to import the main source file as module. Single file package building fails because the main source file does not belong to any import path. Now both the current file and the main source file paths are compared in relative form to prevent importing the main source file as module.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,6 +63,8 @@ jobs:
         # The value doesn't matter as long as it's > 2.087
         FRONTEND: 2.093.0
       run: |
+        dub build --compiler=${{ env.DC }}
+        dub run   --compiler=${{ env.DC }} --single test/issue_2051_running_unittests_from_dub_single_file_packages_fails.d
         ./scripts/ci/travis.sh
 
     - name: '[Windows] Test'
@@ -72,3 +74,4 @@ jobs:
       run: |
         dub build --compiler=${{ env.DC }}
         dub test  --compiler=${{ env.DC }}
+        dub run   --compiler=${{ env.DC }} --single test\issue_2051_running_unittests_from_dub_single_file_packages_fails.d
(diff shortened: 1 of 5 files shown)


drug007:
@thewilsonator I opened this PR instead of #2050 and it's ready to review. How to remove `WIP` label?

thewilsonator (approved):

PR #2068, opened 2020-12-25
Remove buildRequirements: relaxProperties

Geod24:
Two small commits that should be trivial to review.


Geod24:
Hum, actually this breaks Semver...

PR #2077, opened 2021-01-12
[WIP] Build cache policy

drug007:
### Preface

This is revamped #2044 
The PR is the proof of concept, do not review the implementation. Only API, usage etc is subject to review.

#### Dub build modes

Dub uses three way to build the project - rdmd, direct and cached. Consider the last one, that is enabled by default. In cached build case when you build your project the first time the build cache is created. Next time you try to rebuild the project dub compares timestamp of every source file to the timestamp of the target file of the project (in general your executable or library binary).  If any of the source files has more recent timestamp than the target file dub rightly considered that the target file is old and rebuild the project to create new target file and caches all build artifacts. Let's call this mode _time dependent build_. This PR adds an other mode, so called _hash dependent build_ that uses the hash sum of a file instead of its timestamp, i.e. in this mode dub checks if the hash sum of the file has been changed comparing to the previous value on the last successful build, if so then rebuild is triggered. If no previous value exists rebuild is triggered too, of course. 

#### Difference between time dependent and hash dependent build mode

Drawback of time dependent build is that file timestamp and file content are loosely coupled. You can easily change file timestamp  and keep file content the same. Moreover you can change file content and keep its timestamp. So using file timestamp to detect changes is easy to implement and in trivial case is usable in non trivial cases it is pretty annoying.

Hash dependent build avoids this drawback because the hash sum of the file tightly coupled and it is impossible to change file content or its hash sum independently. Moreover it allows to have many build caches simultaneously because there is strong correspondence between the hash sum of the build cache and the hash sum of the source files. You can build your project several times and then redo all your changes and dub won't rebuild the project but will use corresponding build caches.

Of course hash dependent build has its own drawback and first of all it is storage space used by build caches, it is much more than in case of time dependent build.

### TL;DR
Build cache policy added. Now you can use hash dependent build to detect changes in the source files, including string imports (there is a nuacense), by means of their hash sum. This allows to have a build cache per every source file change and revert to the previous state without rebuilding at the cost of storage space occupied by build caches.

### Feature usage

Build cache policy can be set using command line option, environment var or settings file. Option has four values:

    - default means there is no explicit value set so default one will be used
    - time means time dependent build
    - sha1 mean hash dependent build using sha1 algorithm
    - sha256 mean hash dependent build, using sha256 algorithm
Default value lets you explicitly reset the option to undefined value and can be useful in automated workflow. Currently time dependent build is default one.

To set the option using command line interface use `build-cache-policy` argument:

dub build --build-cache-policy=sha1

To set the option using the environment variable use `DUB_BUILD_CACHE_POLICY` var:

$ DUB_BUILD_CACHE_POLICY=sha1 dub build # linux

To set the option using settings file add:

{
    "buildCachePolicy : "sha1"
}

to your system wide, user wide or project specific setting file, for example `dub.settings.json` file in the project root. Now the project will be built using sha1 algorithm every time. But you can override it using command line or env var:

```
# hash dependent build using sha256, not sha1 due to options priority
DUB_BUILD_CACHE_POLICY=sha1 dub build --build-cache-policy=sha256
```

##### Note
This feature is intended to be set once in a setting file or a script and is not intended to be use frequently in command line interface so long but easy names for options were used.

### Option priority

 Options set by different ways have different priority (strong to weak):

* command line option
* environment variable
* settings file
* default value (currently it is time dependent build but can be changed later)

Both sha1 and sha256 are implemented because sha1 is less secure than sha256 but faster so user can select what (s)he wants, some users discussed that in the D forum. Also other algorithms can be added too, especially those implemented in Phobos can be added easily. By experience sha1 much faster than sha256 and I would recommend use it if you have no strong reason to use sha256.


drug007:
@CyberShadow @John-Colvin

rikkimax:
dub.selections.json should also be supported (and overrides global settings file).
If you care about reproducibility of builds, this file will be committed.

drug007:
I'm not sure about dub.selection.json
Build cache policy defines how to build but not what to build. Build artifacts should not depend on how they were built, i.e. no matter if you use time dependent build or hash dependent one you should get the same artifacts.

rikkimax:
Yes, but times are not generally preserved when things get copied around between machines.
So it could be useful for workers who may be passing around build artifacts.

PR #2081, opened 2021-01-20
Allow adding dflags on dependencies when using SDL format

### LINKED ISSUE #2079: Cannot add dflags on dependencies when using SDL format
Dub v1.24.0 introduced [dflags on dependencies](https://dlang.org/changelog/2.095.0.html#dependency-build-settings), but this is only supported for the JSON format, not the SDL format.

CC @omerfirmak @mdparker

omerfirmak:
Fixes #2079

### DIFF
--- a/source/dub/recipe/sdl.d
+++ b/source/dub/recipe/sdl.d
@@ -204,6 +204,10 @@ private void parseDependency(Tag t, ref BuildSettingsTemplate bs, string package
 		dep.default_ = attrs["default"][0].value.get!bool;
 
 	bs.dependencies[pkg] = dep;
+
+	BuildSettingsTemplate dbs;
+	parseBuildSettings(t, dbs, package_name);
+	bs.dependencyBuildSettings[pkg] = dbs;
 }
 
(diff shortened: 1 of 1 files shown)


Geod24:
Can you rebase to target stable ? `git rebase --onto upstream/stable upstream/master`.
Also can you add a test that only `dflags` is accepted ?

omerfirmak:
Rebased.

>  Also can you add a test that only dflags is accepted ?

What do you mean exactly? Currently, other build settings are accepted but will have no affect.

Geod24:
> What do you mean exactly? Currently, other build settings are accepted but will have no affect.

Hum, I thought that they would print an error message. Well, if it's the same for JSON, and there's no easy way to warn about it, let's keep it as is. This is for power users (still) anyways.

omerfirmak:
> > What do you mean exactly? Currently, other build settings are accepted but will have no affect.
> 
> Hum, I thought that they would print an error message. Well, if it's the same for JSON, and there's no easy way to warn about it, let's keep it as is. This is for power users (still) anyways.

Yes, it's the same for JSON as well.

Geod24 (approved):

PR #2082, opened 2021-01-20
Improve 'linker-D-flag' detection for ldmd2

kinke:
`ldmd2` supports all `ldc2` ones (and e.g. `-link-defaultlib-shared` is rather important).

Making these helper functions public will come in handy for reggae as well.

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -352,7 +352,7 @@ config    /etc/dmd.conf
 		return args.map!(s => s.canFind(' ') ? "\""~s~"\"" : s);
 	}
 
-	private static bool isLinkerDFlag(string arg)
+	static bool isLinkerDFlag(string arg)
 	{
 		switch (arg) {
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):
LGTM. I suppose this is not changelog-worthy as it's only beneficial for using dub as a library, rather than for end-users.

PetarKirov:
@kinke I suppose you guys can use the [git-paths](https://github.com/dlang/dub/blob/v1.23.0/changelog/git-paths.dd) feature to depend on dub @ https://github.com/dlang/dub/commit/24ca22d6e9f627883f0833b4914b8a885119c187 in reggae.

kinke:
> I suppose this is not changelog-worthy as it's only beneficial for using dub as a library, rather than for end-users.

The first commit, yes, the 2nd benefits normal usage with `ldmd2` as well. I consider that more as a bug fix though, hence no changelog entry.

PR #2089, opened 2021-01-24
Selectively import symbols from std.array.

veelo:
Merged two imports from `std.array` into one. I manually verified that no other symbols from `std.array` are used anywhere in this file, so it should be safe to merge. Ref https://github.com/dlang/dub/pull/1948#issuecomment-630777513.


PetarKirov (approved):

veelo:
Thanks.

PR #2090, opened 2021-01-24
Remove unused imports from compiler.d.

veelo:
No symbols from `dub.internal.vibecompat.core.file` are used in this file, nor from `dub.internal.vibecompat.inet.url` which is publicly imported. `dub.internal.vibecompat.inet.url` publicly imports `dub.internal.vibecompat.inet.path`, but that is also imported privately here.

No symbols from `dub.internal.vibecompat.data.json` are used in this file, nor from `dub.internal.vibecompat.data.serialization` which it publicly imports.

No symbols from `std.conv` are used in this file.


PetarKirov (approved):
LGTM, thanks for analyzing the imports and cleaning then up!

PR #2095, opened 2021-01-24
Deprecate unused but public import.

veelo:
`Dependency` is publicly imported, but not used in this file. Mark for future removal by deprecating it.

Ref comment by @WebFreak001 in https://github.com/dlang/dub/pull/1948/files#r427248952

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -8,7 +8,7 @@
 module dub.compilers.compiler;
 
 public import dub.compilers.buildsettings;
-public import dub.dependency : Dependency;
+deprecated(`Please "import dub.dependency : Dependency" instead.`) public import dub.dependency : Dependency;
 public import dub.platform : BuildPlatform, matchesSpecification;
 
 import dub.internal.vibecompat.core.file;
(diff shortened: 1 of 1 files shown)


source/dub/compilers/compiler.d
```diff
@@ -8,7 +8,7 @@
 module dub.compilers.compiler;
 
 public import dub.compilers.buildsettings;
-public import dub.dependency : Dependency;
+deprecated(`Please "import dub.dependency : Dependency" instead.`) public import dub.dependency : Dependency;
```
Geod24, line 11:
```suggestion
deprecated("Please `import dub.dependency : Dependency` instead") public import dub.dependency : Dependency;
```

This gives you syntax highlight, as DMD does for error messages.

veelo: Interesting. Where is that documented?

Geod24 (approved):

PetarKirov (approved):

source/dub/compilers/compiler.d
PetarKirov: I don't think it's documented explicitly, but it follows from the fact that the error/diagnostic message printing logic in the frontend is shared. The same trick also applies to `static assert`.

PR #2098, opened 2021-01-25
Fix #2085: [visuald] spurious library project

### LINKED ISSUE #2085: [visuald] Root package (targetType: none) produces spurious library project in the VS solution
### System information
- **dub version**: dub 1.23.0
- **OS Platform and distribution**: Windows 10
- **compiler version** dmd-v2.094.1

### Bug Description

If a root package's purpose is solely as a container for sub packages, `dub generate visuald` creates projects not only for the sub packages but one for the root package as well. This root project is configured to produce a library, but without any source files and missing target name this will fail to build. Therefore building the solution fails. The desired behaviour is for the root project to not produce any output.

### How to reproduce?

`root\dub.json`:
```json
{
	"name": "root",
	"targetType": "none",
	"dependencies": {
		"root:sub": "*"
	},
	"subPackages": [
		"sub"
	]
}
```

`root\sub\dub.json`:
```json
{
	"description": "A minimal D application.",
	"name": "sub",
	"targetType": "executable"
}
```

`root\sub\source\app.d`:
```d
import std.stdio;

void main()
{
	writeln("Edit source/app.d to start your project.");
}
```

1.  `cd root`
2.  `mkdir .dub` (Without this the next step fails, which is another issue.)
3.  `dub generate visuald`
4.  `start root.sln`
5.  Ctrl+Shift+B

### Expected Behavior

The generated `root\.dub\root.visualdproj` contains a `<Config>` section. Removing that section entirely seems to give the desired behaviour: building the solution results in all sub packages to be built, without errors. The resulting file then becomes:
```xml
<DProject>
  <ProjectGuid>{5D0E23E5-2FA1-4829-8841-1289CE482F2D}</ProjectGuid>
  <Folder name="root">
    <Folder name="root">
      <File tool="None" path="..\dub.json" />
    </Folder>
  </Folder>
</DProject>
```
Note: it looks like one `Folder` nest level can be removed as well.

### Discussion

An alternative is to not generate a project for the root package at all, but then the root `dub.json` should be put in a folder inside the solution, which is a different mechanism (different file format, different icon). For consistency I think it is fine to have a project for the root package, even though it does not resemble the directory structure.

veelo:
If no output is generated (TargetType.none) then don't generate a project configuration.

This prevents VS from attempting to build a library with no name from no sources.

### DIFF
--- a/source/dub/generators/visuald.d
+++ b/source/dub/generators/visuald.d
@@ -256,6 +256,9 @@ class VisualDGenerator : ProjectGenerator {
 				return ret;
 			}
 
+			if (buildsettings.targetType == TargetType.none)
+				return;
+
 			foreach(architecture; settings.platform.architecture) {
 				auto arch = architecture.vsArchitecture;
 				ret.formattedWrite("  <Config name=\"%s\" platform=\"%s\">\n", to!string(type), arch);
--- a/test/issue2085-target-none-visuald.sh
+++ b/test/issue2085-target-none-visuald.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd "${CURR_DIR}/issue2085-target-none-visuald" || die "Could not cd."
+
+"$DUB" generate visuald
+
+if grep -c -e \"</Config>\" .dub/root.visualdproj; then
+	die $LINENO 'Regression of issue #2085.'
+fi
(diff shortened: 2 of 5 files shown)


PetarKirov (commented):
LGTM, but could you add a test for this change?

PetarKirov (approved):
Thanks!

veelo:
Thanks for all your reviews!

PetarKirov:
It's easy to review small, but well-thought out changes ;)

PR #2099, opened 2021-01-25
Include copyFiles and extraDependencyFiles in visuald projects.

### LINKED ISSUE #1053: Can't add generic files to ide solutions
Hi,

It seems that it is not possible to add some files to ide solutions that will not be compiled.
I want to add my shader files,... to a visuald solution.

I took a look and it seems that it is partially possible, the dub.sdl file is added with a special flag to avoid his compilation.

I would create a pull request with "genericFiles" and "genericPaths" sdl commands.
I think that I will able to do it for sdl and json parser and visuald generator + the documentation update.

Will it be ok?

veelo:
Resolves #1053.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -669,6 +669,62 @@ ProjectGenerator createProjectGenerator(string generator_type, Project project)
 
 
 /**
+	Calls delegates on files and directories in the given path that match any globs.
+*/
+void findFilesMatchingGlobs(in NativePath path, in string[] globList, void delegate(string file) addFile, void delegate(string dir) addDir)
+{
+	import std.path : globMatch;
+
+	string[] globs;
+	foreach (f; globList)
+	{
+		if (f.canFind("*", "?") ||
+			(f.canFind("{") && f.balancedParens('{', '}')) ||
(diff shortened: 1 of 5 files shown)


source/dub/generators/generator.d
```diff
@@ -668,6 +668,62 @@ ProjectGenerator createProjectGenerator(string generator_type, Project project)
...
+					break;
+				}
+}
+
+
+/**
+	Calls delegates on files in the given path that match any globs.
+
+	If a directory matches a glob, the delegate is called on all existing files inside it recursively
+	in depth-first pre-order.
```
maxhaton, line 713: https://run.dlang.io/is/cvvhB9 Comment says depth-first, all span modes are by breadth?

veelo: The comment follows the documentation of [breadth](https://dlang.org/phobos/std_file.html#breadth.) Seems correct to me looking at the output of your test run?

maxhaton: OK - I wasn't familiar with the SpanMode semantics

maxhaton (approved):

veelo:
Thanks!

PR #2101, opened 2021-01-27
generators: add ninja generator

ljmf00:

---
A revive of PR #1625 .


redthing1:
Looks very interestng!

source/dub/generators/ninja.d
```diff
@@ -0,0 +1,597 @@
...
+
+		// to enable colors on ninja
+		string[] dflags = [
+			settings.platform.compiler == "gdc"
+				? "-fdiagnostics-color=always"
+				: "-color=on"
+		];
+		string[] lflags;
+
+		switch(settings.buildType)
```
WebFreak001, line 362: this switch should probably use the Compiler class instead of reinventing these values

ljmf00: Thanks. There is a lot of stuff I'm not aware of about dub internals and this is one of the things. I'm going to fix this when I have some time to touch this.

Geod24:
Would you like to pursue this further ? I think it would be of great value to many users.

PR #2104, opened 2021-02-05
Bump FRONTEND, move dmd to latest by default.

maxhaton:


.github/workflows/main.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       env:
         COVERAGE: false
         # The value doesn't matter as long as it's > 2.087
-        FRONTEND: 2.093.0
+        FRONTEND: 2.095.0
```
Geod24, line 64: Those might get out of sync though

maxhaton: They *did* get out of sync - it seems to be an environment variable though, so I didn't see a simple way of doing it automatically at 5AM

.github/workflows/main.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       matrix:
         # Latest stable version, update at will
         os: [ macOS-10.15, ubuntu-18.04, windows-2019 ]
-        dc: [ dmd-2.093.1, ldc-1.23.0, dmd-master, ldc-master ]
+        dc: [ dmd-latest, ldc-latest, dmd-master, ldc-master ]
```
veelo, line 32: Isn't it better to use a fixed release? #2105 is failing and I don't know if the problem is with my PR or because the test is now using a different compiler.

maxhaton: I suppose, but it fell several releases behind when it wasn't pegged to the latest release. Also dmd is tested upstream so *hopefully* the it should be on the PR not the compiler (in theory...)

PR #2105, opened 2021-02-09
Actually copy copyFiles when building in Visual Studio.

veelo:
From the changelog:

copyFiles can now be used in VisualD projects.
---------------------------------------------------

Files that are needed for the application to run can be specified in the `copyFiles` build option, which causes these files to be copied to the target path. The dub generator for VisualD now replicates this behaviour, so that this kind of applications can be successfully debugged in Visual Studio.

### DIFF
--- a/test/issue1053-extra-files-visuald.sh
+++ b/test/issue1053-extra-files-visuald.sh
@@ -21,6 +21,6 @@ if [ `grep -c -e "README.txt" .dub/extra_files.visualdproj` -ne 1 ]; then
 	die $LINENO 'Regression of issue #1053.'
 fi
 
-if [ `grep -e "README.txt" .dub/extra_files.visualdproj | grep -c -e "copy /Y $(InputPath) $(TargetDir)"` -ne 1 ]; then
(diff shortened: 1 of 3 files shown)


test/issue1053-extra-files-visuald.sh
```diff
@@ -20,3 +20,7 @@ fi
 if [ `grep -c -e "README.txt" .dub/extra_files.visualdproj` -ne 1 ]; then
 	die $LINENO 'Regression of issue #1053.'
 fi
+
+if [ `grep -e "README.txt" .dub/extra_files.visualdproj | grep -c -e "copy /Y $(InputPath) $(TargetDir)"` -ne 1 ]; then
```
Geod24, line 24:
```suggestion
if [ `grep -e "README.txt" .dub/extra_files.visualdproj | grep -c -e 'copy /Y $(InputPath) $(TargetDir)'` -ne 1 ]; then
```

maxhaton:
@veelo ready for merge?

veelo:
Yes this is ready, thank you.

maxhaton (approved):

maxhaton:
> Yes this is ready, thank you.

Good to go

PR #2109, opened 2021-02-23
Replacing rununittest.sh

drug007:
### Purpose

Now on all platforms the test suite will run d files in the `test` subdirectory that have the suffix ".script.d" after running shell scripts. Replacing ".sh" files by "script.d" files lets us to make the test suite crossplatform one and run the test suite on both posix systems (macos and linux) and windows.

Note: packages are still tested only on posix systems

#### Details
1. The suffix "script.d" is used to distinguish existed d files from new d scripts.
2. ~~Like rununittest.sh imports `common.sh` using `source common.sh`, rununittest.d does `mixin(import("common.d"));`. I do not like it but it just mimics the current bash scripts design.~~ common.d is a package so regular import is used
3. I'm not sure these two strings are equivalent:

bash:
```shell
echo "$supplemental" | >&2 sed 's|^|        |g'
```
dlang:
```d
stderr.writeln("        ", supplemental);
```
d version of `die` is just a `writeln` + `throw exception`


test/run-unittest.d
```diff
@@ -2,44 +2,11 @@
...
-	import std.stdio : File, stderr;
-
-	const str = text("[ERROR] ", args);
-	version(Windows) stderr.writeln(str);
-	else stderr.writeln("\033[0;31m", str, "\033[0m");
-	stderr.flush;
-	File(logFile, "a").writeln(str);
-	any_errors = true;
-}
+mixin(import("common.d"));
```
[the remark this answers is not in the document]
drug007: Of course it does, but for that a package should be created to import common from it and I did what is simpler (no additional package, just a single file). But I'll create the package if it is preferable way.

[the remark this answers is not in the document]
Geod24: Why would a package be needed ?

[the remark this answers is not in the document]
Geod24: I also think a regular import is preferable.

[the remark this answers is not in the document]
drug007: I didn't like that mixin too. Done.

.github/workflows/main.yml
```diff
@@ -75,3 +77,6 @@ jobs:
         dub build --compiler=${{ env.DC }}
         dub test  --compiler=${{ env.DC }}
         dub run   --compiler=${{ env.DC }} --single test\issue2051_running_unittests_from_dub_single_file_packages_fails.d
+        $env:DUB="$pwd\bin\dub.exe"
+        $env:DC="dmd"
```
Geod24, line 81: `DC` is already exported by the action

test/0-init-fail-json.script.d
```diff
@@ -0,0 +1,37 @@
...
+int main()
+{
+	enum packname = "0-init-fail-pack";
+	enum deps = "logger PACKAGE_DONT_EXIST"; // would be very unlucky if it does exist...
+
+	auto dub = environment.get("DUB");
+	if (dub == "")
+	{
+		die(`Environment variable "DUB" must be defined to run the tests.`);
+	}
```
Geod24, line 23:
```suggestion
	if (!dub.length)
		die(`Environment variable "DUB" must be defined to run the tests.`);
```

test/common/source/common.d
```diff
@@ -0,0 +1,57 @@
...
+	import std.file : getcwd, thisExePath;
+	import std.path : baseName, buildPath;
+	import std.stdio : File, stderr;
+	import core.stdc.stdlib : exit;
+
+	auto str = format!"[ERROR] %s:%s %s"(thisExePath.baseName, line, msg);
+	stderr.writeln(str);
+	auto f = File(buildPath(getcwd, logFile), "w");
+	f.writeln(str);
+	if (supplemental != "")
```
Geod24, line 51:
```suggestion
	if (!supplemental.length)
```

test/common/source/common.d
```diff
@@ -0,0 +1,57 @@
...
+
+	const str = text("[ERROR] ", args);
+	version(Windows) stderr.writeln(str);
+	else stderr.writeln("\033[0;31m", str, "\033[0m");
+	stderr.flush;
+	File(logFile, "a").writeln(str);
+	any_errors = true;
+}
+
+auto die(string msg = "command failed", string supplemental = "", size_t line = __LINE__)
```
Geod24, line 39:
```suggestion
void die(string msg = "command failed", string supplemental = null, size_t line = __LINE__)
```

Although I'm not sure why you don't pass `file` as well to get rid of `thisExePath` ?

test/run-unittest.d
```diff
@@ -81,13 +47,13 @@ int main(string[] args)
...
 	const dc_bin = baseName(dc);
-	const curr_dir = args[0].absolutePath.dirName.buildNormalizedPath;
+	const curr_dir = getcwd;
 	const frontend = environment.get("FRONTEND", "");
 
 	//** if [ "$#" -gt 0 ]; then FILTER=$1; else FILTER=".*"; fi
 	auto filter = (args.length > 1) ? args[1] : "*";
 
-	version(linux)
+	version(posix)
```
Geod24, line 56:
```suggestion
	version (Posix)
```

test/common/source/common.d
Geod24: Well, actually, the `common.sh` design was because of bash's limitations. Nowadays `die` could just be:
```D
void die(Args...)(Args args)
{
    stderr.writeln(args);
    throw new Exception("Test failed");
}
```

test/run-unittest.d
```diff
@@ -81,13 +47,13 @@ int main(string[] args)
 	//** CURR_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 	//** FRONTEND="${FRONTEND:-}"
 	const dc_bin = baseName(dc);
-	const curr_dir = args[0].absolutePath.dirName.buildNormalizedPath;
+	const curr_dir = getcwd;
```
Geod24, line 50:
```suggestion
	const curr_dir = __FILE_FULL_PATH__.dirName();
```

Geod24 (commented):
I think the code should be much more D-like, and much less shell-like.

test/common/source/common.d
drug007: I would prefer to just port all bash scripts to D initially and then enhance them or redesign all test suit as needed. I have no time to make it at once, so this slow process step-by-step.

test/run-unittest.d
```diff
@@ -107,5 +73,16 @@ int main(string[] args)
 		}
 	}
 
+	foreach(DirEntry script; dirEntries(curr_dir, (args.length > 1) ? args[1] : "*.script.d", SpanMode.shallow))
```
Geod24, line 76:
```suggestion
	foreach (DirEntry script; dirEntries(curr_dir, (args.length > 1) ? args[1] : "*.script.d", SpanMode.shallow))
```

test/run-unittest.d
```diff
@@ -107,5 +73,16 @@ int main(string[] args)
...
+	{
+		const min_frontend = script.name ~ ".min_frontend";
+		if (exists(min_frontend) && frontend.length && frontend < min_frontend.readText) continue;
```
Geod24, line 79:
```suggestion
		if (frontend.length && exists(min_frontend) && frontend < min_frontend.readText) continue;
```

Since `frontend.length` is constant through the program, and does not require file IO, put this condition first.

test/run-unittest.d
```diff
@@ -107,5 +73,16 @@ int main(string[] args)
...
+		log("Running " ~ script ~ "...");
+		if (spawnProcess([dub, script.name], ["DUB":dub, "DC":dc, "CURR_DIR":curr_dir]).wait)
```
Geod24, line 81: Usage of `DC` was only fixed in the latest version, meaning that anyone testing with an older compiler might get a different result. It won't show up in result, but it will for developers.

test/common/source/common.d
Geod24: Well bash script and D are quite different, so one can't just copy-paste the code and make it work with minor syntax changes like you would with C / C++. So it's not so much redesign here, it's just recognizing that the `die` function is meant to print an error message to `stderr` and exit. And doing this with varargs and Exception is much more natural to do in D than shoehorning the signature given to this function in bash.

But if you don't want to change it with this PR, I don't mind as long as the other comments are addressed.

test/common/dub.sdl
```diff
@@ -0,0 +1,5 @@
+name "common"
+description "utility package for test suit"
```
Geod24, line 2:
```suggestion
description "Utility package for test suite"
```

test/common/dub.sdl
```diff
@@ -0,0 +1,5 @@
...
+authors "drug007"
+copyright "The D foundation"
```
Geod24, line 4:
```suggestion
copyright "The D Language Foundation"
```

test/common/dub.sdl
```diff
@@ -0,0 +1,5 @@
...
+license "BSL"
```
Geod24, line 5:
```suggestion
license "BSL-1.0"
```

For software licenses, you usually want to use the SPDX identifier: https://spdx.org/licenses/BSL-1.0.html

test/common/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+common.so
+common.dylib
+common.dll
+common.a
+common.lib
+common-test-*
+*.exe
+*.o
+*.obj
+*.lst
```
Geod24, line 15: Make the package a `sourceLibrary` and we can remove this.

Geod24:
Thanks for doing this BTW! It's going to be very helpful to have a non-flakey CI :)

test/common/source/common.d
drug007: It is too defensive may be, but I try to make less intrusive changes as much as possible to avoid breaking something because I don't know code very well.  But if the reviewer says how  something can be improved I don't see the reason to resist it. That's what code review intended for.

drug007:
@Geod24 thank you for your review! BTW, could you take a look at #2111 ?

maxhaton (approved):

.github/workflows/main.yml
```diff
@@ -75,3 +75,5 @@ jobs:
         dub build --compiler=${{ env.DC }}
         dub test  --compiler=${{ env.DC }}
         dub run   --compiler=${{ env.DC }} --single test\issue2051_running_unittests_from_dub_single_file_packages_fails.d
+        $env:DUB="$pwd\bin\dub.exe"
+        cd test; dub --single run-unittest.d; $env:DUB=''; $env:DC=''
```
Geod24, line 79: I don't quite follow what those `env` steps are doing ?

test/common/source/common.d
```diff
@@ -0,0 +1,45 @@
...
+	File(logFile, "a").writeln(str);
+}
+
+/// prints error message to standard error stream and log file
+/// and set any_errors var to true value to indicate that some
+/// test fails
+void logError(Args...)(Args args)
+{
+	import std.conv : text;
+	import std.stdio : File, stderr;
```
Geod24, line 29: Move those to module level, no point in using function-local imports when all of them import the same thing.

.github/workflows/main.yml
Geod24: My understanding is that we're setting `DUB` before, then resetting it after.
I don't see the point of resetting it after since the test ends and it wouldn't affect a new step if we were to add one.
Regarding setting it before, I'd prefer if we use the Github syntax, that is:
```yaml
env:
  DUB: ${{ github.workspace }}\bin\dub.exe
run: [...]
```

Geod24 (approved):
Nice, just one last queston

.github/workflows/main.yml
drug007: Frankly speaking when I reviewed my code I didn't see any reasons for that except it was the artifact of the past. I left it just because it didn't hurt in general. But now I'm going to fix that.

test/common/source/common.d
drug007: I write PoCs very frequently. It's much easier to refactor the code when imports are local and that's my habit now. But I agree, here it is useless.

Geod24 (approved):
Amazing, thanks!

PR #2112, opened 2021-02-24
Fix single file package dynamic library

andre2007:
This PR fixes the issue that main sourcefile was not added to sourceFiles while building dynamic link library from dub single file package.

### DIFF
--- a/test/issue1505-single-file-package-dynamic-library.sh
+++ b/test/issue1505-single-file-package-dynamic-library.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd ${CURR_DIR}
+rm -f libsingle-file-test-dynamic-library.{so,dylib}
+rm -f single-file-test-dynamic-library.dll
+
+${DUB} build --single issue1505-single-file-package-dynamic-library.d
+if [ ! -f libsingle-file-test-dynamic-library.{so,dylib} ] && [ ! -f single-file-test-dynamic-library.dll ]; then
+	die $LINENO 'Normal invocation did not produce a dynamic library in the current directory'
+fi
+rm -f libsingle-file-test-dynamic-library.{so,dylib}
+rm -f single-file-test-dynamic-library.dll
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -245,7 +245,7 @@ class ProjectGenerator
 		// add main source files to root executable
 		{
 			auto bs = &roottarget.buildSettings;
-			if (bs.targetType == TargetType.executable) bs.addSourceFiles(mainSourceFiles);
+			if (bs.targetType == TargetType.executable || genSettings.single) bs.addSourceFiles(mainSourceFiles);
 		}
 
 		if (genSettings.filterVersions)
--- a/test/issue1505-single-file-package-dynamic-library.d
+++ b/test/issue1505-single-file-package-dynamic-library.d
@@ -1,0 +1,13 @@
+/+ dub.sdl:
+   name "single-file-test-dynamic-library"
+   targetType "dynamicLibrary"
++/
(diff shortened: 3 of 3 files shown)


test/issue1505-single-file-package-dynamic-library.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd ${CURR_DIR}
+rm -f libsingle-file-test-dynamic-library.so
+rm -f single-file-test-dynamic-library.dll
+
+${DUB} build --single issue1505-single-file-package-dynamic-library.d
+if [ ! -f libsingle-file-test-dynamic-library.so ] && [ ! -f single-file-test-dynamic-library.dll ]; then
```
PetarKirov, line 8:
```suggestion
if [ ! -f libsingle-file-test-dynamic-library.{dll,so,dylib} ]; then
```

test/issue1505-single-file-package-dynamic-library.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd ${CURR_DIR}
+rm -f libsingle-file-test-dynamic-library.so
+rm -f single-file-test-dynamic-library.dll
```
PetarKirov, line 5:
```suggestion
rm -f single-file-test-dynamic-library.{dll,so,dylib}
```

test/issue1505-single-file-package-dynamic-library.sh
```diff
@@ -0,0 +1,12 @@
...
+
+${DUB} build --single issue1505-single-file-package-dynamic-library.d
+if [ ! -f libsingle-file-test-dynamic-library.so ] && [ ! -f single-file-test-dynamic-library.dll ]; then
+	die $LINENO 'Normal invocation did not produce a dynamic library in the current directory'
+fi
```
PetarKirov, line 10: The test would be more robust if it checks whether an exported function defined by the single-file shared library yields the expected result when called from another single file executable.

PetarKirov (commented):
A few comments, otherwise the change looks good:

test/issue1505-single-file-package-dynamic-library.sh
andre2007: The windows `dll` and the linux `so` have different names. The linux so starts with `lib`.

Yes, I was also wondering why the MacOS test pipeline were succeeding although the `dylib` was missing.

test/issue1505-single-file-package-dynamic-library.sh
andre2007: @PetarKirov 
My gut feeling is, this rather complex logic is not necessary in this case. When dub and dmd returns with return code 0 and dmd produced a dynamic library, it is safe to assume, the file is valid. A test whether dmd is producing valid dynamic libraries I would see in the test suite of dmd.

andre2007:
@PetarKirov I added the `dylib` extension. As already written the filename on windows and posix are different (`lib` prefix on posix) therefore I cannot combine them into single statements.

test/issue1505-single-file-package-dynamic-library.sh
PetarKirov: > The linux so starts with lib

Ah yes, then this should do it:

```
{lib,}single-file-test-dynamic-library.{dll,so,dylib}
```

test/issue1505-single-file-package-dynamic-library.sh
PetarKirov: If we had a test helper script (also used for the rest of the tests) that did such checks my case would be easier to make. In the absence of such, I'm fine with leaving it like this.

PetarKirov (approved):

PR #2124, opened 2021-04-03
Colored output

Zardoz89:
I retake the code from the old PR  #1490 and updated, so all merge conflicts are fixed and it's working against the actual ~master branch of DUB.

Colored output vs ~master output : 
imagen

Another example : 
imagen

Showing an error at compile time : 
imagen


Zardoz89:
Could someone with OSX check why the OSX tests fails ?

I think that it's here, but I don't have idea why this happens and why only on OSX : 
```
 [INFO]     Testing too slow registry
     Warning Package dub not found for registry at http://localhost:3744/: Failed to download http://localhost:3744/api/packages/infos?packages=%5B%22dub%22%5D&include_dependencies=true&minimize=true
             Getting a release version failed: No package dub was found matching the dependency >=0.0.0
             Retry with ~master...
     Warning Package dub not found for registry at http://localhost:3744/: Failed to download http://localhost:3744/api/packages/infos?packages=%5B%22dub%22%5D&include_dependencies=true&minimize=true
       Error No package dub was found matching the dependency ~master
        8.21 real         0.05 user         0.03 sys
[INFO] Running /Users/runner/work/dub/dub/test/issue1556-fetch-and-build.sh...
Trying to fetch fs-sdl-dubpackage
/Users/runner/work/dub/dub/test/timeout.sh: line 34: echo: write error: Broken pipe
Error:  pipefail:33 command failed
    Fetching main-package 1.0.0
             Storing package 'main-package', version requirements: >=0.0.0
    Finished main-package fetched
             Please note that you need to use `dub run <pkgname>` or add it to dependencies of your package to actually use/run it. dub does not do actual installation of packages outside of its own ecosystem.
```

s-ludwig:
Not sure about macOS, I can try to test later locally when I have more time.

I haven't looked at the details, but one thing should be removed from the PR, moving the logging functionality into a public module. Adding a public API should be thought out well and it's hard to argue why DUB would provide its own logging API (plus it breaks vibe.d compatibility).

Zardoz89:
> Not sure about macOS, I can try to test later locally when I have more time.
> 
> I haven't looked at the details, but one thing should be removed from the PR, moving the logging functionality into a public module. Adding a public API should be thought out well and it's hard to argue why DUB would provide its own logging API (plus it breaks vibe.d compatibility).

The code isn't from me. I can't say anything about why the author did it public. I simply rebased his code and fixed any merge errors that I find. I can try to changes that you see necessary, like making the logger private module.

veelo:
@Zardoz89 Do you have any plans for picking this up again?

Zardoz89:
Nope

WebFreak001:
moved to #2268

PR #2134, opened 2021-05-21
Revert "Mutually exclude mainSourceFile of different configurations"

schveiguy:
Reverts dlang/dub#2039

This change needs to go. Changing the existing semantics of directives is not reasonable in a stable ecosystem.

Instead, implement via a new directive.


thewilsonator (approved):

Imperatorn:
> Reverts #2039
> 
> This change needs to go. Changing the existing semantics of directives is not reasonable in a stable ecosystem.
> 
> Instead, implement via a new directive.

For us not 100% updated, what would the consequences of reverting this be? 🤔

schveiguy:
The PR being reverted changed the meaning of the `mainSourceFile` directive to exclude the file in OTHER configurations for building. Essentially, it tied the file to the configuration it was declared in. Reverting it shouldn't affect the vast majority of projects, because this directive is rarely used. For projects where it was used, it should restore the original usage, and in some cases (i.e. mysql-native) fix the project, or it might break projects that are using the new feature, but those should be fairly new.

More importantly it should make dub perform consistently across compiler versions.

Geod24:
> Reverting it shouldn't affect the vast majority of projects, because this directive is rarely used.

The same reasoning apply to the original change.

> For projects where it was used, it should restore the original usage, and in some cases (i.e. mysql-native) fix the project, or it might break projects that are using the new feature, but those should be fairly new.

It will break our project. And in the case of `mysql-native`, the fix was trivial: https://github.com/mysql-d/mysql-native/pull/229 . I don't understand why you didn't do it ?

If it was reverted in the same release or in a point release, I'd understand. But it's been 3 releases.

schveiguy:
> The same reasoning apply to the original change.

My statement was not a justification, just an answer to the question "how does this affect me?". The revert should be done even if it breaks current projects, because the change was unsound in the first place.

What may need to happen before this is a way to do what the original change does, just using a different directive.

wilzbach:
> The revert should be done even if it breaks current projects,

I'm afraid we can't do this anymore (it's too late).

schveiguy:
So what is the solution? Leave the current hack in place because it sneaked into a release?

schveiguy:
I'll note that buildkite passes, so none of the projects that we are looking at actually use this feature. I feel the code breakage would be extremely minimal (vastly outnumbered by fixing projects, including historical ones, that were broken in the first place), and those who depend on it can still use the hacky version of dub from a recent compiler until the correct fix (new directive) is released.

schveiguy:
> I don't understand why you didn't do it ?

Turns out that "trivial" fix doesn't work. https://travis-ci.com/github/mysql-d/mysql-native/jobs/507100666

Geod24:
> Turns out that "trivial" fix doesn't work.

Yep, my analysis of the file was slightly off, and I should also add it to `excludedSourceFiles` because `mysql-native` usage was just odd.

Note that in the process of doing that, I noticed that `mysql-native` actually has `excludedSourceFiles "source/app.d"` in all configurations but one:
https://github.com/mysql-d/mysql-native/blob/e52eacfdd41a1972e033d2e56367658a49e0a268/dub.sdl#L12-L15

This is yet another proof that people need a way to have configuration-specific files because that's how they use it. With the improved definition of `mainSourceFile`,  this kind of use case actually scales.

schveiguy:
> This is yet another proof that people need a way to have configuration-specific files because that's how they use it.

I have no problem with the idea, just the fact that it used a directive that already meant something completely different (and in the process broke existing code).

Geod24:
Closing because this behavior have been there for over 18 months and a revert no longer makes sense.

PR #2139, opened 2021-06-01
Fix dependency build settings flooding dflags

omerfirmak:
from `depen-build-settings` test

```
/usr/bin/dmd -X -X -X -X -c -of.dub/build/application-debug-linux.posix-x86_64-dmd_v2.096.1-25506C1B97B12F5C72F39C8661659746/depen-build-settings.o -debug -g -w -version=Have_depen_build_settings -version=Have_depend1 -version=Have_depend2 -Isource/ -Idepend/source/ -Idepend/depend2/source/ source/app.d -vcolumns
```
this fixes the barrage of -X flags you can see here

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -13,7 +13,8 @@ import std.array : array;
 import std.algorithm : filter, any;
 import std.path : globMatch;
 import std.typecons : BitFlags;
-
+import std.algorithm.iteration : uniq;
+import std.range : chain;
 
 /// BuildPlatform specific settings, like needed libraries or additional
 /// include paths.
@@ -105,7 +106,7 @@ struct BuildSettings {
 		addPostRunCommands(bs.postRunCommands);
 	}
 
-	void addDFlags(in string[] value...) { dflags ~= value; }
+	void addDFlags(in string[] value...) { dflags = chain(dflags, value.dup).uniq.array; }
 	void prependDFlags(in string[] value...) { prepend(dflags, value); }
 	void removeDFlags(in string[] value...) { remove(dflags, value); }
 	void addLFlags(in string[] value...) { lflags ~= value; }


s-ludwig:
This is unfortunately not a generally valid fix, as there are arguments that may actually have to appear multiple times, because they are paired with their successor arguments (some arguments can be written as `-arg=value` or `-arg value` and there are more complex ones, such as `-L-framework -LFoundation`.

I haven't seen the dependency/dflags addition so far, so I can't really suggest an alternative fix right now.

Geod24:
@omerfirmak : Tried this too: https://github.com/dlang/dub/pull/2016

omerfirmak:
> This is unfortunately not a generally valid fix, as there are arguments that may actually have to appear multiple times, because they are paired with their successor arguments (some arguments can be written as `-arg=value` or `-arg value` and there are more complex ones, such as `-L-framework -LFoundation`.
> 
> I haven't seen the dependency/dflags addition so far, so I can't really suggest an alternative fix right now.

This only clears consecutive duplicate flags, so your examples should still work. See https://dlang.org/library/std/algorithm/iteration/uniq.html

s-ludwig:
> This only clears consecutive duplicate flags, so your examples should still work. See https://dlang.org/library/std/algorithm/iteration/uniq.html

Indeed, I had overlooked that. Off the top of my head, I can't think of an example that would break with this then.

source/dub/compilers/buildsettings.d
```diff
@@ -105,7 +106,7 @@ struct BuildSettings {
 		addPostRunCommands(bs.postRunCommands);
 	}
 
-	void addDFlags(in string[] value...) { dflags ~= value; }
+	void addDFlags(in string[] value...) { dflags = (dflags ~ value).uniq.array; }
```
s-ludwig, line 109:
```suggestion
	void addDFlags(in string[] value...) { dflags = chain(dflags, value).uniq.array; }
```

Geod24 (approved):

PR #2144, opened 2021-06-09
[stable] Don't reduce the max path length on Windows even further

### LINKED ISSUE #2143: Long paths on Windows: Cannot even use 260 chars
I'm seeing issues with long paths on a recent Windows 10, with LDC and Visual Studio 2019 in this case.

### Errors with the MS linker

E.g.:
```
LINK : fatal error LNK1181: cannot open input file '..\..\..\..\AppData\Local\dub\packages\unit-threaded-1.0.15\unit-threaded\subpackages\integration\.dub\build\library-debug-windows-x86_64-ldc_v1.27.0-beta1-sym1-55CBD351797D62157A88F9C8E869018F\unit-threaded_integration.lib'
```

Interesting because the relative path length is only 223, and the absolute one (`C:\Users\packer\AppData\…`) is 227 chars, nowhere near the 260 limit. **Edit**: Ah, `C:\Users\packer\dev\SIL\lang\test_plugin1\..\..\..\..\AppData\…` exceeds 260.

One workaround is using `-link-internally` with LDC, which makes it use the LLD linker integrated in LDC.

~~Another workaround is converting the relative path to `\\?\C:\Users\packer\AppData\…` (in the LDC cmdline issued by dub; these are simply forwarded to the linker). This could be done by LDC as well, but see below.~~ **Edit**: Nope, `link.exe` doesn't support the `\\?\` prefix for absolute paths > 260.

### Superfluous rebuilds

I was about to tackle this in LDC, but then came across another very annoying bit: such dependencies are rebuilt again and again, even if nothing has changed. Running dub with `-v` shows:
```
File ..\..\..\..\AppData\Local\dub\packages\unit-threaded-1.0.15\unit-threaded\subpackages\integration\.dub\build\library-debug-windows-x86_64-ldc_v1.27.0-beta1-sym1-4B9CC688E7DF7F085E7CCAC75893DE7E\unit-threaded_integration.lib doesn't exist, triggering rebuild.
```

(The file obviously exists.) So handling this in the compiler alone isn't sufficient.

kinke:
Fixes #2143.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -600,13 +600,21 @@ private string shrinkPath(NativePath path, NativePath base)
 {
 	auto orig = path.toNativeString();
 	if (!path.absolute) return orig;
-	auto ret = path.relativeTo(base).toNativeString();
-	return ret.length < orig.length ? ret : orig;
+	version (Windows)
+	{
+		// avoid relative paths starting with `..\`: https://github.com/dlang/dub/issues/2143
+		if (!path.startsWith(base)) return orig;
+	}
+	auto rel = path.relativeTo(base).toNativeString();
+	return rel.length < orig.length ? rel : orig;
 }
 
 unittest {
 	assert(shrinkPath(NativePath("/foo/bar/baz"), NativePath("/foo")) == NativePath("bar/baz").toNativeString());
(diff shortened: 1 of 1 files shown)


Geod24 (approved):
Looks slightly hackish but if it works 🤷

PR #2146, opened 2021-06-12
Create folders before copying file during generator finalization.

JudgeFish:
Should be a simple start for my first contribution on github :)

Only tested on https://github.com/dayllenger/beamui.git so far.
   dub generate visuald :opengl
now works and project files are generated.

Other older issues with relative paths in visuald generated subprojects are still hindering it to build though.
Might have a solution for this later...

regards
/Thomas

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -865,6 +865,7 @@ private void finalizeGeneration(in Package pack, in Project proj, in GeneratorSe
 				}
 				logDiagnostic("  %s to %s", src.toNativeString(), dst.toNativeString());
 				try {
+					mkdirRecurse(dst.parentPath().toNativeString());
 					hardLinkFile(src, dst, true);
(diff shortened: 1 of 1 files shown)


Geod24:
Wouldn't it make more sense to have it be part of `hardLinkFile` ?

JudgeFish:
Looking at hardLinkFile(), it looks rather low-level so I would propose keeping it clean from extra behavior.

source/dub/generators/generator.d
```diff
@@ -865,6 +865,7 @@ private void finalizeGeneration(in Package pack, in Project proj, in GeneratorSe
 				}
 				logDiagnostic("  %s to %s", src.toNativeString(), dst.toNativeString());
 				try {
+					mkdirRecurse(dst.parentPath().toNativeString());
```
Geod24, line 868: Can you change this to `ensureDirectory(dst.parentPath())` ?

Geod24 (approved):
Sorry, this fell off the radar. If you want to fix it, I've left a comment, otherwise I'll just submit this fix in a week or so. Having a test to ensure we don't regress would be good, too.

PR #2150, opened 2021-06-22
Generate paths relative to visuald project files

### LINKED ISSUE #1477: Generating a VisualD solution for as subPackage uses incorrect relative path
### System information

- **dub version**: 1.9.0-beta.1
- **OS Platform and distribution**: Windows 8
- **compiler version** v2.080.0-beta.1

### Bug Description

When a subpackage is generated, the project files utilize the relative path assuming they were placed within the subpackage directory, but they get placed in the root .dub folder. The Subpackage also specifies a targetPath "../.."

### How to reproduce?

dub generate visuald :mysubpackage

### Expected Behavior

Opening the solution in D will result in the compiler being able to find all files.

JudgeFish:
Fixes #1477
Visuald internally generated paths should be relative to project file path. Actually relative to "working dir"/.dub
Current implementation is relative to the dub-file but this does not work correctly for sub projects when generating from base project.

Tested on project https://github.com/dayllenger/beamui.git
1. From root dir in beamui :
  dub generate visuald :opengl
generates "beamui_opengl.sln" as before but now with correct paths.

2. From subproject beamui/examples/opengl path:
  dub generate visuald
generates "opengl.sln" as before, also with correct paths.

Note that beamui dub project is incomplete regarding windows paths for pixman lib.

An alternative solution would be to change where visuald files are generated for sub projects but that would be the subject for another issue.


thewilsonator (approved):
looks fine, can you please add a test case for this

JudgeFish:
Looking into tests how to add tests.

JudgeFish:
Added a bash script in "test" folder to build and check output files.
Unsure when and how this script is run though... (no scripts are not run in "run-unittest.sh" ).

thewilsonator:
Hmm, could you please force push to try restarting the CI

thewilsonator:
The test seems to have permission problems executing. Did you set executable permissions on that shell script? If not please do.

thewilsonator:
Thanks!

PR #2151, opened 2021-07-06
add username and password to URL.toString()

Gagege:


WebFreak001:
At least in dub registry the equivalent class had them omitted by design, where it rather added a constant `***@***` string if credentials are present, to not expose it in error messages.

However in dub (running on the users PC with the users settings) I think it's not harmful to include credentials in any case. I think from the start it would maybe have been better to have an explicit build method that includes all the data, but I think adding it now would be unnecessary.

It's worthy thinking of the security considerations of this in any case - if any malicious party for example can look over the output of the users run (log of a CI run with secret dub registry credentials) and the connection fails and causes the credentials to be printed - could this be considered a security risk?

You mixed tabs and spaces btw.

Gagege:
I see what you're saying @WebFreak001. What if I concatenated the username and password where it's needed in the download function instead? I'll update my PR anyway.

Gagege:
Made an update. This should be safer, and backward compatible.

source/dub/internal/vibecompat/inet/url.d
```diff
@@ -217,6 +217,11 @@ struct URL {
 		{
 			dst.put("//");
 		}
+		if( !m_username.empty ) {
+            dst.put(m_username);
```
s-ludwig: This needs to use `std.uri.encodeComponent` to properly work for user names/passwords with special characters (interestingly, this is also wrong in the vibe.d implementation).

Gagege: Got it. Made the change.

Gagege:
@WebFreak001 @s-ludwig Do either of you know what happened with that failed build, DAutoTest? I can't find the actual error in it.

source/dub/internal/utils.d
```diff
@@ -236,19 +236,22 @@ version (Have_vibe_d_http)
 
 	Note: Timeouts are only implemented when curl is used (DubUseCurl).
```
WebFreak001, line 237: this note should now also say that authentication only works with curl.

See https://github.com/vibe-d/vibe.d/blob/94c19a4e7839eb30a613424bf13606a623079bd4/inet/vibe/inet/urltransfer.d#L34

source/dub/internal/utils.d
```diff
@@ -260,6 +260,9 @@ void download(string url, string filename, uint timeout = 8, string safeUrl = nu
 /// ditto
 void download(URL url, NativePath filename, uint timeout = 8)
 {
+	version (Have_vibe_d_http) {
+		assert(url.username.length == 0 && url.password.length == 0, "Auth not supported yet.");
+	}
```
WebFreak001, line 265: I have been thinking about if this assert is a good idea - I think it would actually help us keep the docs up-to-date and also allow to write a changelog, once the vibe.d urltransfer module supports HTTP auth. However I think there should be a note here, that this restriction comes from vibe.d not supporting it initially and to remove it (and the docs) once it does

Gagege: Is the reason the same for timeout? Vibe-d doesn't support it?

WebFreak001: I think that's gonna be the case here, but didn't check source

WebFreak001 (approved):

PR #2155, opened 2021-07-21
PackageManager must know placement dir

denizzzka:
upd: For now if you try to replace PackageManager arguments (representing repositories dirs) you will catch inconsistency between provided dirs where dub will search packages and dirs where dub stores fetched packages.

This PR:
* Removes duplicate definition of paths to local repositories
* Adds possibility for user defined PackageManager (PR what opens this ability is coming soon)

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -236,7 +236,7 @@ struct CommandLineHandler
 		if (options.bare) {
 			dub = new Dub(NativePath(getcwd()));
 			dub.rootPath = NativePath(options.root_path);
-			dub.defaultPlacementLocation = options.placementLocation;
+			dub.placementLocation = options.placementLocation;
 
 			return dub;
 		}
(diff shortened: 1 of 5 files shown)


rikkimax:
It might be a good idea to add a second package manager instance to verify that everything has been caught.

e.g. regex match packages names into a custom directory.

denizzzka:
@rikkimax can you explain more detail?

rikkimax:
> @rikkimax can you explain more detail your idea, what problem can be solved by second PackageManager?

To verify that this change works correctly both today and will continue to work in the future.

There could be some hidden assumptions in some place that can only be found by doing this.

denizzzka:
> There could be some hidden assumptions in some place that can only be found by doing this.

For now in ~master if you try to replace PackageManager arguments (representing repositories dirs) you will catch inconsistency between provided dirs where dub will search packages and dirs where dub stores fetched packages.

Now the very possibility of this is excluded - this PR totally removes this paths from Dub

So, I think, it isn't necessary.

Geod24:
So, after countless refactorings of the `PackageManager`, I don't think it should know the default `PlacementLocation`.
Instead, the API offers 2 possibilities:
- Act on all possible `PlacementLocation` (usually reads);
- Act on a specific `PlacementLocation` (usually write);

The `Dub` class is responsible for choosing the `PlacementLocation` and storing the default. Feel free to re-open if you disagree and have a use case I am not seeing.

PR #2158, opened 2021-07-31
Fix #2141 - When used as a library, the default compiler is different from the expected.

### LINKED ISSUE #2141: When used as a library, the default compiler is different from the expected.
The current dub may not detect the default compiler as intended due to the mechanism of detecting the preferred compiler by thisExePath when used in libraries.
When used as a library, the following code, which assumes that the executable file path and the compiler are located in the same place, will not work:
https://github.com/dlang/dub/blob/4c1ce741d69316093d69721cdf73fc13666d6f83/source/dub/dub.d#L1495

The current determineDefaultCompiler function ignores the environment variable DC. The environment variable DC suggests which compiler should be preferred, as specified in dlang-setup in GitHub Actions or in install.sh.

It also doesn't seem to be a good idea for the compiler type to take precedence over the path priority.
If `PATH=C:\ldc\bin;C:\dmd\bin;%PATH%`, then ldc2 should have priority over dmd.

https://github.com/dlang/dub/blob/4c1ce741d69316093d69721cdf73fc13666d6f83/source/dub/dub.d#L1508

shoo:
If a default compiler is given, eg. DC, the compiler is searched from the paths, and if not, the compiler is searched according to the priority of the paths.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1502,11 +1502,67 @@ class Dub {
 
 		// If nothing found next to dub, search the user's PATH, starting
 		// with the compiler name from their DUB config file, if specified.
-		if (m_defaultCompiler.length)
-			compilers = m_defaultCompiler ~ compilers;
 		auto paths = environment.get("PATH", "").splitter(sep).map!NativePath;
-		auto res = compilers.find!(bin => paths.canFind!(p => existsFile(p ~ (bin~exe))));
-		m_defaultCompiler = res.empty ? compilers[0] : res.front;
+		if (m_defaultCompiler.length && paths.canFind!(p => existsFile(p ~ (m_defaultCompiler~exe))))
+			return;
+		foreach (p; paths) {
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

source/dub/dub.d
```diff
@@ -1502,11 +1502,67 @@ class Dub {
...
+		version (Posix) enum sep = ":", exe = "";
+		
+		immutable dmdpath = testdir.buildPath("dmd", "bin");
+		immutable ldcpath = testdir.buildPath("ldc", "bin");
+		mkdirRecurse(dmdpath);
+		mkdirRecurse(ldcpath);
+		immutable dmdbin = dmdpath.buildPath("dmd"~exe);
+		immutable ldcbin = ldcpath.buildPath("ldc2"~exe);
+		std.file.write(dmdbin, null);
+		std.file.write(ldcbin, null);
```
Geod24, line 1546: Unittest should not do IO!

PR #2160, opened 2021-08-15
dub: json: improve messages on type missmatch when reading values

ljmf00:

---

Running a malformed JSON config gives `Expected JSON <expected-type>, got <atual-type>`. This is very annoying and hard to debug which value is malformed.

This PR changes that message to be: `'<actual-key-name>': Expected JSON <expected-type>, got <atual-type>`.

### DIFF
--- a/source/dub/internal/vibecompat/data/json.d
+++ b/source/dub/internal/vibecompat/data/json.d
@@ -1484,7 +1484,7 @@ struct JsonSerializer {
 	//
 	void readDictionary(T)(scope void delegate(string) field_handler)
 	{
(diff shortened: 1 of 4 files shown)


WebFreak001 (changes requested):
I think you should also add the line number which is included with `.line`.

Does this include member fields of things printing `.configurations.foo.dependencies` instead of just `.dependencies` too? If yes: add it to the test case

Geod24:
@ljmf00 : Closing as we're not using the JSON parser anymore, and the message has been improved.
Feel free to re-open if the quality of the current message is still lacking.

PR #2164, opened 2021-08-24
Replace deprecated keyword `body` with `do`

nordlow:


nordlow:
Getting scope errors in CI jobs. What to do?

thewilsonator (approved):

thewilsonator:
https://github.com/dlang/undeaD/pull/47

Geod24:
> source/dub/internal/undead/xml.d

The problematic code is in dub, so you need to apply the diff here as well.

RazvanN7 (approved):

RazvanN7:
> The problematic code is in dub, so you need to apply the diff here as well.

@nordlow This.

nordlow:
I don't follow. I can't find any more references to the body keyword on this branch.

RazvanN7:
The patch that @thewilsonator applied to undead needs to be applied to the local copy of undead in dub. See `source/dub/internal/undead/xml.d` : it's the same code as in dlang/undeaD#47

nordlow:
> The patch that @thewilsonator applied to undead needs to be applied to the local copy of undead in dub. See `source/dub/internal/undead/xml.d` : it's the same code as in [dlang/undeaD#47](https://github.com/dlang/undeaD/pull/47)

Ahh, sorry. I was searching for the body keyword. I'll fix.

nordlow:
Changes applied.

WebFreak001:
it says the toType function in xml.d is missing `scope` on its argument, not sure how it works upstream? (should probably even be `scope return`)

nordlow:
> it says the toType function in xml.d is missing `scope` on its argument, not sure how it works upstream? (should probably even be `scope return`)

Done.

thewilsonator (approved):

WebFreak001 (approved):

nordlow:
Read now.

RazvanN7 (approved):

Issue #2118, opened 2021-03-22
Package namespacing for private dub registry


skoppe:
At symmetry we use https://github.com/skoppe/private-dub to host a private dub registry based off of our private gitlab.

We intentionally prefix all packages with `symmetry.`. This is to ensure no public package can shadow a private one or vice versa.

This works because code.dlang.org doesn't accept packages with a `.` in their name.

The only thing left to do would be to 'teach' dub about private package namespacing, so it avoids printing a warning, and to ensure the `.` stays off-limits on code.dlang.org in the future.

What is the stance on this? I will likely open an MR in a week or 2.

rikkimax:
Something like this has been wanted for a while now. So I'm all for a solution to be found.

One thing I think that should be considered is if the "account name" on the repository side, must match the namespace name. Let the registry whitelist certain namespaces when hosted privately by configuration to ignore certain namespaces requirements.

PR #2167, opened 2021-09-17
Allow '.' as separator in package names

nordlow:
FYI, @John-Colvin @skoppe

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -249,9 +249,9 @@ class Project {
 		}
 		if (m_rootPackage.name != m_rootPackage.name.toLower()) {
 			logWarn(`WARNING: DUB package names should always be lower case. %s`, nameSuggestion());
-		} else if (!m_rootPackage.recipe.name.all!(ch => ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' || ch == '-' || ch == '_')) {
+		} else if (!m_rootPackage.recipe.name.all!(ch => ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' || ch == '-' || ch == '_' || ch == '.')) {
 			logWarn(`WARNING: DUB package names may only contain alphanumeric characters, `
-				~ `as well as '-' and '_'. %s`, nameSuggestion());
+				~ `as well as '-', '_' and '.'. %s`, nameSuggestion());
 		}
 		enforce(!m_rootPackage.name.canFind(' '), "Aborting due to the package name containing spaces.");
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Please no! This

nordlow:
We are using this behaviour company internally.

skoppe:
Yes see, https://github.com/dlang/dub/issues/2118

I think what is missing is to set a policy that it is allowed as a package name, but cannot be uploaded to code.dlang.org

nordlow:
> Yes see, #2118
> 
> I think what is missing is to set a policy that it is allowed as a package name, but cannot be uploaded to code.dlang.org

Feel free to take over the PR and adjust the code as needed, @skoppe.

s-ludwig:
Before the package name rules are relaxed, the possible conflicts should be evaluated really carefully. Other symbols, such as `:`, `/` and `@`, are already used to separate package names and we should be very certain that `.` can be sacrificed without limiting other functionality in the future.

s-ludwig:
One example that would break due to this today is the dub.pm redirection feature (e.g. http://dubtestproject.dub.pm).

skoppe:
> Before the package name rules are relaxed, the possible conflicts should be evaluated really carefully. Other symbols, such as `:`, `/` and `@`, are already used to separate package names and we should be very certain that `.` can be sacrificed without limiting other functionality in the future.

Do you know where `:`, `/` and `@` are already used? I remember trying the `/` but failed to get it working (IIRC dub failed to download it).

> One example that would break due to this today is the dub.pm redirection feature (e.g. http://dubtestproject.dub.pm).

We should disallow any uploading of package to `code.dlang.org` that contain the separate. This is to prevent packages shadowing each other; we don't want to run into a situation where a private package is registered by an attacker on `code.dlang.org` This has happened to other package managers and can be very damaging.

s-ludwig:
> Do you know where :, / and @ are already used? I remember trying the / but failed to get it working (IIRC dub failed to download it).

I mean it is being used outside of the package name as a separator:
- `dub --override-config eventcore/select`
- `dub fetch eventcore@0.9.18`
- `dub build vibe-d:web`

> We should disallow any uploading of package to code.dlang.org that contain the separator (any really).

This is done [here](https://github.com/dlang/dub-registry/blob/master/source/dubregistry/registry.d#L487).

skoppe:
> > Do you know where :, / and @ are already used? I remember trying the / but failed to get it working (IIRC dub failed to download it).
> 
> I mean it is being used outside of the package name as a separator:
> 
>     * `dub --override-config eventcore/select`
> 
>     * `dub fetch eventcore@0.9.18`
> 
>     * `dub build vibe-d:web`

Ah yes, didn't know about the `/` one.

> > We should disallow any uploading of package to code.dlang.org that contain the separator (any really).
> 
> This is done [here](https://github.com/dlang/dub-registry/blob/master/source/dubregistry/registry.d#L487).

Perfect. Perhaps it would be wise to add an explicit case [in the switch](https://github.com/dlang/dub-registry/blob/master/source/dubregistry/registry.d#L723) statement to make it clear to future contributors the `.` character is reserved for private package.

That brings up the question whether the dub registry ever wants to allow namespaces and what character to use for that, since all the goods ones are taken.

Issue #2171, opened 2021-09-27
Dustmite Run Program Execution Timeout


nordlow:
I'm missing a way to specify a timeout for the program when doing `dub dustmite`.

Preferrably using --program-time=TIMEOUT.

I'd like to add this.

Alternative we could add a way to specify a prefix wrapper command that as a special case can run command using a timeout as 

```sh
timeout N DUB_RUN_BINARY
```

.

PR #2172, opened 2021-09-27
Draft: Add dustmite timeout support

nordlow:
I want to be sure that I'm not waiting in vain typically for dub run calls during dustmite. This adds support for this. I'm thinking of 
defaulting the timeout to the initial executable run time times some factor >=1.

Closes https://github.com/dlang/dub/issues/2171.

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -18,7 +18,7 @@ import std.typecons : BitFlags;
 import std.algorithm.iteration : uniq;
 import std.range : chain;
 
-enum timeoutExitStatus = 11111; // exit status for process when timed out
+enum timeoutExitStatus = 124; // exit status for process when timed out. `timeout` also uses 124.
 
 /// BuildPlatform specific settings, like needed libraries or additional
 /// include paths.


source/dub/compilers/buildsettings.d
```diff
@@ -16,9 +16,12 @@ import std.typecons : BitFlags;
 import std.algorithm.iteration : uniq;
 import std.range : chain;
 
+enum timeoutExitStatus = 11111; // exit status for process when timed out
```
nordlow, line 19: This should be some carefully chosen value. Do you have any suggestions, @thewilsonator and @CyberShadow?

nordlow:
Conflicts resoved.

source/dub/compilers/compiler.d
```diff
@@ -124,18 +126,63 @@ interface Compiler {
...
+				import std.conv : to;
+				return ["timeout", timeout.total!"seconds".to!string] ~ args;
+			}
+			return args;
+		}
+
 		int status;
 		if (output_callback) {
-			auto result = executeShell(escapeShellCommand(args), env);
+			auto result = executeShell(escapeShellCommand(timeouted(args)), env); // TODO: avoid using timeoutArgs
```
WebFreak001, line 145: I don't think we should rely on timeout being installed, use D native timeout code like below here too, instead of relying on `timeout` as system command

WebFreak001 (changes requested):

CyberShadow:
This implementation of process execution timeout is going to not be useful in many circumstances, and can easily crash a user's PC. I don't think I would recommend incorporating this implementation and making it available for general use.

The problem is that it only kills the top-level process. What you want to do instead is kill the entire process tree, suspending them all first. If you don't do that, you're likely going to get an ever-growing pile of processes, which are likely to continue remaining stuck and eating resources, running in the background, creating something not too unlike a fork bomb.

Implementing a correct process killing mechanism for a timeout feature is quite platform specific, and is not implemented in DustMite itself for this reason. I recommend using a proper port of the `timeout` utility as part of the oracle command instead. (Dub's integration could conceivably do that as well.)

source/dub/compilers/compiler.d
CyberShadow, line 145: Oh, I see that's what this does already. What's with the other branch, though?

source/dub/compilers/buildsettings.d
CyberShadow: The `timeout` tool uses 124.

source/dub/compilers/compiler.d
nordlow: What branch?

source/dub/compilers/buildsettings.d
nordlow: Adjusted at 28c516458484c5520d2014bd4921e5bcd1f40696.

source/dub/compilers/compiler.d
CyberShadow: The `else` part of `if (output_callback) {`.

CyberShadow: I explained above why that is not a good idea. The two are not synonymous. We should move in the other direction instead.

PR #2174, opened 2021-09-28
Forward dub dustmite parameter -j

nordlow:
This does not mimic DustMites behaviour of `-j` to mean `-jN` where `N` is the number of cores online. As that syntax is currently not supported by `std.getopt` which dub uses opposite to DustMite which has its own parser.

Ready for merge, @thewilsonator.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -2281,6 +2281,7 @@ class DustmiteCommand : PackageBuildCommand {
 		bool m_combined;
 		bool m_noRedirect;
 		string m_strategy;
+		uint m_jobCount;		// zero means not specified
 	}
 
 	this() @safe pure nothrow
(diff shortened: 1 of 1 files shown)


source/dub/commandline.d
```diff
@@ -2281,6 +2281,7 @@ class DustmiteCommand : PackageBuildCommand {
 		bool m_combined;
 		bool m_noRedirect;
 		string m_strategy;
+		uint m_jobCount = uint.min; // zero means not specified
```
thewilsonator, line 2284: why is this set to `uint.min`?

nordlow: Ahh, it's really not needed. I'll remove.

nordlow: Done.

nordlow:
auto-merge @thewilsonator?

source/dub/commandline.d
thewilsonator: You dan't get notifications for edit, but as long as you are either subscribed to notifications to the repository, participate in, or are mentioned either directly or requested for review you get an email.

Take a look at the footer of email threads from GitHub and you'll see things like `You are receiving this because you were mentioned.` or `You are receiving this because you are subscribed to this thread` or some such.

As for the delay in reply. Timezones are a thing, and people need to sleep ;)

thewilsonator (approved):

source/dub/commandline.d
nordlow: I see. Thanks! :)

PR #2181, opened 2021-10-14
Expand tilde in --root path

nordlow:
Bugfix.

Makes dub able to support tilde character in argument passed too `--root` CLI flag.


WebFreak001:
no test case needed imo, because this is just stdlib function

PR #2188, opened 2021-10-27
Draft: Reduce memory usage of Json

nordlow:
This reduces `Json.sizeof` from 64 to 40 bytes.

The memory usage could be reduced further by making use of an `Algebraic` data type or another more efficient json library.

Optionally, a `Json` node could be represented using a dual-pointer; one pointing to the tree and one to the current node. That would enable efficient storage of the filename inside the tree part.

For even denser tree representations, mir-ion provides a solution.

The `m_fileName` is only used in three diagnostics calls to `enforceJson`. Need a decision about whether we should drop that field or not. If it should be kept we need to find a way to propagate it without storing it in the node. That's trivial to solve except for `Node.opIndex` and `Node.opCast`.

### DIFF
--- a/source/dub/internal/vibecompat/data/json.d
+++ b/source/dub/internal/vibecompat/data/json.d
@@ -67,7 +67,6 @@ struct Json {
 			int line;
 		version (VibeJsonFieldNames)
 			string m_name;
-		pragma(msg, __FILE__, "(", __LINE__, ",1): Debug: ", Json.sizeof);
 	}
 
 	/** Represents the run time type of a JSON value.
--- a/source/dub/internal/vibecompat/data/json.d
+++ b/source/dub/internal/vibecompat/data/json.d
@@ -63,12 +63,11 @@ struct Json {
 		@property ref inout(Json[]) m_array() inout { return getDataAs!(Json[])(); }
 
 		Type m_type = Type.undefined;
-
-		version (VibeJsonFieldNames) {
-			uint m_magic = 0x1337f00d; // works around Appender bug (DMD BUG 10690/10859/11357)
+		version(JsonLineNumbers)
+			int line;
+		version (VibeJsonFieldNames)
 			string m_name;
-			string m_fileName;
-		}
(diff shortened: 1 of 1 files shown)


source/dub/internal/vibecompat/data/json.d
```diff
@@ -398,15 +398,15 @@ struct Json {
 	*/
 	inout(T) opCast(T)() inout { return get!T; }
 	/// ditto
-	@property inout(T) get(T)()
+	@property inout(T) get(T)(in string fileName = "DUMB_FILE_NAME")
```
WebFreak001, line 401: I'm not sure if `DUMB_FILE_NAME` is such a good name because:
- it's a valid filename, although a dumb one
- it does not give much information when it would suddenly appear in an error message

Additionally this should be a `static immutable string defaultJsonFileName` at the start of the file, to avoid writing this magic value out everywhere.

I think something like `(.json)` would be better.

source/dub/internal/vibecompat/data/json.d
```diff
@@ -948,7 +948,7 @@ Json parseJson(R)(ref R range, int* line = null, string filename = null)
 
 	assert(ret.type != Json.Type.undefined);
 	version(JsonLineNumbers) ret.line = curline;
-	ret.m_fileName = filename;
+	// ret.m_fileName = filename;
```
WebFreak001, line 951: better to delete than to comment out such a trivial line I think

source/dub/internal/vibecompat/data/json.d
```diff
@@ -814,7 +814,7 @@ struct Json {
 		return ret.data;
 	}
 
-	private void checkType(TYPES...)(string op = null)
+	private void checkType(TYPES...)(string op = null, in string fileName = "DUMB_FILE_NAME")
```
WebFreak001, line 817: ditto for above

WebFreak001 (changes requested):

Geod24:
@nordlow : We're actually weeding ourselves off the JSON dependency, are you fine with closing this ?

PR #2189, opened 2021-11-08
Windows: Copy .pdb files

### LINKED ISSUE #1962: PDB file not copied to targetPath alongside executable on Windows
### System information
- **dub version**: `DUB version 1.20.1, built on Apr 17 2020`
- **OS Platform and distribution**: `Microsoft Windows [Version 10.0.18363.535]` -- (1909), 64-bit
- **compiler version**: `dmd-2.091.1`

### Bug Description
When building an executable the `.exe` is copied to the `targetPath` however the `.pdb` file with the debugging information is not.

### How to reproduce?
```
> type source\app.d
void main() {}

> type dub.sdl
name "foo"
targetName "foo-bin"
targetType "executable"
targetPath "bin\\64"

> dub build
...

> dir /b bin\64
foo-bin.exe
```

### Expected Behavior
```
> dir /b bin\64
foo-bin.exe
foo-bin.pdb
```

### Use Case
On Windows it's common for in-house software to ship the `.pdb` files along with the `.exe` to make live debugging easier. Alternatively the `.pdb` files are published to a [Symbol Server](https://en.wikipedia.org/wiki/Microsoft_Symbol_Server) where they can be downloaded on-the-fly when debugging. In both scenarios we need easy access to the `.pdb` file rather than spelunking the `.dub\build` folder.

kinke:
Fixes #1962.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -371,12 +371,30 @@ class BuildGenerator : ProjectGenerator {
 
 	private void copyTargetFile(NativePath build_path, BuildSettings buildsettings, GeneratorSettings settings)
 	{
-		auto filename = settings.compiler.getTargetFileName(buildsettings, settings.platform);
-		auto src = build_path ~ filename;
-		logDiagnostic("Copying target from %s to %s", src.toNativeString(), buildsettings.targetPath);
 		if (!existsFile(NativePath(buildsettings.targetPath)))
 			mkdirRecurse(buildsettings.targetPath);
-		hardLinkFile(src, NativePath(buildsettings.targetPath) ~ filename, true);
+
+		string[] filenames = [
+			settings.compiler.getTargetFileName(buildsettings, settings.platform)
+		];
+
+		// Windows: add .pdb if found
+		const tt = buildsettings.targetType;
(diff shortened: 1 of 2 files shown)


source/dub/generators/build.d
```diff
@@ -371,12 +371,30 @@ class BuildGenerator : ProjectGenerator {
...
+		];
+
+		// Windows: add .pdb if found
+		const tt = buildsettings.targetType;
+		if ((tt == TargetType.executable || tt == TargetType.dynamicLibrary) &&
+		    settings.platform.platform.canFind("windows"))
+		{
+			import std.path : setExtension;
+			const pdbFilename = filenames[0].setExtension(".pdb");
+			if (existsFile(build_path ~ pdbFilename))
```
kinke, line 388: Even with `-g` or `-gc`, things such as `/PDB:dir\binary.pdb` as lflags, a presumably common current workaround, prevent the generation of the default pdb, as well as stuff like `-gdwarf` with LDC. So check if the file exists and only copy in that case.

kinke:
[Tested manually on Windows, seems to work fine.]

thewilsonator (approved):

PR #2194, opened 2021-11-27
Also add options when we combine buildsettings

### LINKED ISSUE #2193: Unnable to get import path from sourceLibrary packages
### System information
- dub: 1.27.0
- win10 x64

### Bug Description

VSCode code-d extention rely this to get list of imports: https://github.com/Pure-D/workspace-d/blob/fe22c452ac176cfe4d85bfd9839bf6d4dd8165bc/source/workspaced/com/dub.d#L206

The problem is when it tries to read import about a package "sourceLibrary", it throws an exception, it is due to: https://github.com/dlang/dub/blob/27908126db37c01435110143971904a0b32303b6/source/dub/project.d#L742

Skipping setting properly the project description because it is a "sourceLibrary"

Therefore, when code-d tries to get list of imports, it crashes because its project description is not complete, and therefore fails assert here: https://github.com/dlang/dub/blob/master/source/dub/description.d#L37

### How to reproduce?

- create a new dub project ``dub init`` 
- change the targetType to ``sourceLibrary``
- open vscode
- install code-d extention
- open the project folder
- open the command palette and type: Reload Import Path
- It'll say "import paths are empty"

### Expected Behavior

- it should detect the import paths

ryuukk:
Fixes #2193

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -104,6 +104,7 @@ struct BuildSettings {
 		addPostBuildCommands(bs.postBuildCommands);
 		addPreRunCommands(bs.preRunCommands);
 		addPostRunCommands(bs.postRunCommands);
+		addOptions(bs.options);
 	}
 
 	void addDFlags(in string[] value...) { dflags = chain(dflags, value.dup).uniq.array; }


PetarKirov:
Hi @ryuukk, thanks for the thorough investigation and the fix! Can you add a test that reproduces the issue you described and asserts that different import paths can be properly retrieved?

ryuukk:
I didn't notice this was still open my bad!

I am not sure how to write tests for that kind of stuff..

I'll give another try whenever i got some free time

thewilsonator (approved):

WebFreak001:
sorry this PR was getting lost. Could you rebase or allow contributors from dlang/dub to push to your branch? Then the tests should pass again.

If you don't want to rebase this yourself, I rebased locally, but I can't push to your branch, at the right there is a checkbox to allow others to push to your branch, so click that and tell us if you want me to do it for you.

If you want to rebase yourself you can do so like this:
```
# in your checked out repostiory
# first associate the remote "upstream" with dlang/dub
git remote add upstream https://github.com/dlang/dub.git
# get commits from there
git fetch upstream
# checkout your own master
git checkout origin master
# update with upstream master commits (should just fast-forward)
git pull upstream master
# checkout your feature branch
git checkout patch-1
# rebase (reapplies the patch commits here onto master)
git rebase master
# force push to patch-1 (as we rewrote history, we need to force, so the old actual commit is deleted from history)
git push origin patch-1 -f
```

PR #2197, opened 2021-12-03
fix CLI --print-builds implementation

WebFreak001:
this also serves as base for my next PR where I plan to add the functionality of --print-configs and --print-builds to `dub describe`


maxhaton (approved):
I'm going to hesitantly approve this but my understanding of the code here may betray me.

PR #2198, opened 2021-12-03
Exclude macOS - dmd-latest on GitHub actions

WebFreak001:
Tests are failing for unknown reasons (maybe cache)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,6 +30,9 @@ jobs:
         # Latest stable version, update at will
         os: [ macOS-10.15, ubuntu-18.04, windows-2019 ]
         dc: [ dmd-latest, ldc-latest, dmd-master, ldc-master ]
+        exclude:
+        - os: macOS-10.15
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -30,6 +30,9 @@ jobs:
         # Latest stable version, update at will
         os: [ macOS-10.15, ubuntu-18.04, windows-2019 ]
         dc: [ dmd-latest, ldc-latest, dmd-master, ldc-master ]
+        exclude:
+        - os: macOS-10.15
+          dc: dmd-latest
```
PetarKirov, line 35:
```suggestion
          dc: dmd-mastet
```

PetarKirov (changes requested):

s-ludwig:
I'd suggest to simply make the test not required for merge instead of removing it, so that it at least stays on the radar.

WebFreak001:
ok thanks that works too

PR #2201, opened 2021-12-06
CLI: refactor common GeneratorSettings init

WebFreak001:
I originally did this change to have the DescribeCommand properly filled, but I couldn't find a test case where it mattered, so this is effectively just code style now.

While reviewing myself no side effects from populating m_defaultConfig inside the commandline.d file were found: DustmiteCommand, GenerateCommand, TestCommand, LintCommand and DescribeCommand either don't use it, override it back or have used it before and no longer determine it manually now.

- changes from manually creating GeneratorSettings to using defaultGeneratorSettings:
	- GenerateCommand.execute: no changes
	- TestCommand.execute: no changes
	- DescribeCommand.execute: now set: buildMode, buildSettings, (previously only the nogc flag was copied) single
	- DustmiteCommand.execute: now set: buildMode, filterVersions, single

There is a slim possibility of the m_defaultConfig population changing applications behavior for applications that depend on DUB as a library, have extended `PackageBuildCommand` and relied on m_defaultConfig initializing to `null` with the setupPackage function call. It now initializes to the actual default dub configuration.

I don't think this edge case is blocking this PR though, as before the common use pattern with that m_defaultConfig was for setting GeneratorSettings like this:

```d
gensettings.config = m_buildConfig.length ? m_buildConfig : m_defaultConfig;
```

Where there is now a valid explicitly defined default config instead of implicitly default `null` value.

Workaround for API users if this behavior changes anything significant:
before:
```d
setupPackage(...);
```
after:
```d
setupPackage(...);
m_defaultConfig = null;
```

It has a little performance penalty, but as the setupPackage function is just intended for setting up, it should be fine.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -988,6 +988,20 @@ abstract class PackageBuildCommand : Command {
 		]);
 	}
 
+	protected GeneratorSettings defaultGeneratorSettings()
+	{
+		GeneratorSettings gensettings;
+		gensettings.platform = m_buildPlatform;
+		gensettings.config = m_buildConfig.length ? m_buildConfig : m_defaultConfig;
+		gensettings.buildType = m_buildType;
+		gensettings.buildMode = m_buildMode;
+		gensettings.compiler = m_compiler;
+		gensettings.buildSettings = m_buildSettings;
+		gensettings.filterVersions = m_filterVersions;
+		gensettings.single = m_single;
+		return gensettings;
+	}
+
 	protected void setupVersionPackage(Dub dub, string str_package_info, string default_build_type = "debug")
 	{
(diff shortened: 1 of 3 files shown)


wilzbach (approved):

WebFreak001:
ok after more testing this actually fixed a bug that CLI passed debug version were not included in the output:
```
dub describe -dfoo --data-list --data=debug-versions
```
after this change this includes foo. If an IDE emits the same common arguments for all dub calls this would be relevant, which is exactly the use-case of dub describe.

Gonna add a test case for this.

Geod24:
Would you consider https://github.com/dlang/dub/pull/2346 to have superseded this ?

PR #2203, opened 2021-12-07
Posix: use /etc/dub/settings.json if in /usr

### LINKED ISSUE #1472: Does not read settings from /etc/dub/settings.json
Hi!
I tried to use the recent addition of system-wide dub packages (#838) by adding `customCachePaths` to `/etc/dub/settings.json`. Unfortunately, it looks like dub never reads this file and only considers the local settings in `~/.dub/settings.json`.
Please make dub read the file in /etc, as the docs say, or alternatively a file in /usr/lib/dub/, so distributors can set a system location for dub packages.
Cheers,
    Matthias

WebFreak001:
Fix #1472

/etc/dub/settings.json overrides /usr/etc/dub/settings.json (as
suggested by the `man hier` manpage)

### DIFF
--- a/changelog/etc-dub-settings-json.dd
+++ b/changelog/etc-dub-settings-json.dd
@@ -1,0 +1,9 @@
+Posix: use /etc/dub/settings.json if DUB is installed in /usr
+
+For Linux distributions that put the dub installation in /usr, there is now a
+special case that DUB will load from /etc/dub/settings.json (absolute path) if
+the installation is inside /usr.
+
+Previously settings would have attempted to be loaded from
+`/usr/etc/dub/settings.json` if installed in `/usr/bin/dub`. This is still
+loaded if it exists, but if not `/etc/dub/settings.json` will be loaded.
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -290,7 +290,15 @@ class Dub {
 		m_dirs.temp = NativePath(tempDir);
 
 		m_config = new DubConfig(jsonFromFile(m_dirs.systemSettings ~ "settings.json", true), m_config);
-		m_config = new DubConfig(jsonFromFile(NativePath(thisExePath).parentPath ~ "../etc/dub/settings.json", true), m_config);
+
+		auto dubFolderPath = NativePath(thisExePath).parentPath;
+		m_config = new DubConfig(jsonFromFile(dubFolderPath ~ "../etc/dub/settings.json", true), m_config);
+		version (Posix) {
+			if (dubFolderPath.absolute && dubFolderPath.startsWith(NativePath("usr"))) {
+				m_config = new DubConfig(jsonFromFile(NativePath("/etc/dub/settings.json"), true), m_config);
+			}
+		}
+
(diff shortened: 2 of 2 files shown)


WebFreak001:
another idea would be to have /etc/dub.conf which could be considered better than /etc/dub/settings.json because dub only has a single config file (and probably won't get much more that belongs in /etc)

Geod24:
Yeah I think `/etc/dub.json` is better

PetarKirov:
> I think /etc/dub.json is better

I disagree. `dub.json` sounds just like `dub.sdl` - package metadata. Here we're talking about global dub configuration. If people find it strange that `dub.conf` is a JSON file, then I suggest `dub.config.json`.

WebFreak001:
I would suggest dub.conf - maybe it should be a SDL file too though

WebFreak001:
I asked on the discord and got these preferences by users: (mostly weak preferences)
- `/etc/dub/settings.json` (Wild, Snarwin) - Reasons:
	- consistency with the existing `~/.dub/settings.json`
	- If dub ever needs an additional file for something, already having a folder make that easier
	- Wild: I generally think of `/etc/*.*` of being more system configs, while `/etc/*/*.*` is more program configs. But ofc there are always outlieers
- `/etc/dub.conf` (WebFreak) - Reasons:
	- consistency with a lot of other linux package config files

adr doesn't really care but thinks `/etc/dub/settings.json` is good enough for him, it would be easy to find as /etc/dub<tab> will find it anyway.

So I would be for putting it in `/etc/dub/settings.json` as final decision to avoid stretching this bikeshed for longer.

rikkimax:
I do like ``dub.conf.json/dub.conf.sdl``, but yes let's just go with ``/etc/dub/settings.json``.

This does need documenting though.

WebFreak001:
@rikkimax added changelog entry and made PR to dub-docs (though I doubt that PRs there will be merged any time soon with deployment being broken)

PetarKirov:
I'm on board with `/etc/dub/settings.json` as well.

WebFreak001:
can someone approve the PR so I can merge it?

ljmf00:
Moving towards the XDG standard is preferable tho. `~/.dub/settings.json` is not XDG compliant. See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html .

WebFreak001:
`~/.dub/settings.json` has already been used and is not touched by this PR, I would suggest talking about that in a separate issue or e.g. plan that for a second version of dub.

I think how I PRd it here is best for how the current structure is

ljmf00:
> `~/.dub/settings.json` has already been used and is not touched by this PR, I would suggest talking about that in a separate issue or e.g. plan that for a second version of dub.
> 
> I think how I PRd it here is best for how the current structure is.
> 
> So if anyone could approve the PR that would be nice.

Right, my bad.

ljmf00 (approved):
I agree with the change, but I'm not part of dub development and I may miss something, although LGTM.

WebFreak001:
ok seems like I would need someone of the dub team to approve to be able to merge

thewilsonator (approved):

PR #2204, opened 2021-12-08
Add output-path, configs, builds to dub describe

WebFreak001:
Also contains default-config and default-build

Extends the describe command from only being build options to also custom fields like this. Configs and builds was previously only exposed in "interactive" TUI, now as machine parsable format.

output-path is for building the full path including platform specific extensions for where a file is placed. This can be used to check for example after a build if a file has changed or do some extra work on it.


WebFreak001:
found bugs:
- when running with `--build=unittest`, the output-path is not the unittest executable
- the error message `Target 'arsd-official:eventloop' doesn't exist. Is the target type set to "none" in the package recipe?` is gone - it now segfaults instead

PR #2211, opened 2022-02-08
Removes dynamicLibrary being built as staticLibrary for targets other than OMF.

### LINKED ISSUE #571: Dynamic/shared libraries as dependencies
I looked for a bit and didn't find any issue for this.  When a dependency (in my case, a subpackage) has a targetType of dynamicLibrary, dub gives the error message "Dynamic libraries are not yet supported as dependencies - building as static library."  Is there any blocker to this feature being implemented other than time?  How high is it on the priorities list?

In my case I would need some of the dependencies getting placed in bin along with the root package's executable, but some would need to be placed into an alternative path (at least on Unix), so I would need full manual control over where each package is placed (via targetPath or some such property).  Then I would add something to the rpath of the root package for Linux and use the painful SetDllDirectory functionality in Windows (not an issue with dub).

### LINKED ISSUE #2206: Build subPackages as dynamicLibrary: "Dynamic libraries are not yet supported as dependencies - building as static library".

rikkimax:
Succeeds #2149

Closes #571 and closes #2206

#2149 has a lot of information about why this isn't everything needed to get dynamicLibrary behavior being completely correct or user friendly.
But this is the best we can do right now due to compiler/linker state.

### DIFF
--- a/changelog/actual-dynamiclibrary.dd
+++ b/changelog/actual-dynamiclibrary.dd
@@ -1,0 +1,7 @@
+Builds dynamicLibrary targets as dynamic libraries instead of static libraries.
+
+Dub will no longer build dynamicLibrary targetType's as staticLibrary.
+
+Except for x86_omf. This has been disabled due to numerous issues that will lead this to not doing what is expected of it.
+
+No compiler or linker flags have been added at this time, you will need to specify the relevant flag to get the compiler to link dynamically against Phobos.
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -252,15 +252,21 @@ class ProjectGenerator
 		auto roottarget = &targets[rootPackage.name];
 
 		// 0. do shallow configuration (not including dependencies) of all packages
-		TargetType determineTargetType(const ref TargetInfo ti)
+		TargetType determineTargetType(const ref TargetInfo ti, const ref GeneratorSettings genSettings)
 		{
 			TargetType tt = ti.buildSettings.targetType;
 			if (ti.pack is rootPackage) {
 				if (tt == TargetType.autodetect || tt == TargetType.library) tt = TargetType.staticLibrary;
 			} else {
 				if (tt == TargetType.autodetect || tt == TargetType.library) tt = genSettings.combined ? TargetType.sourceLibrary : TargetType.staticLibrary;
-				else if (tt == TargetType.dynamicLibrary) {
-					logWarn("Dynamic libraries are not yet supported as dependencies - building as static library.");
+				else if (genSettings.platform.architecture.canFind("x86_omf") && tt == TargetType.dynamicLibrary) {
+					// Unfortunately we cannot remove this check for OMF targets,
+					// due to Optlink not producing shared libraries without a lot of user intervention.
+					// For other targets, say MSVC it'll do the right thing for the most part,
+					// export is still a problem as of this writing, which means static libraries cannot have linking to them removed.
+					// But that is for the most part up to the developer, to get it working correctly.
+
+					logWarn("Dynamic libraries are not yet supported as dependencies for Windows target OMF - building as static library.");
 					tt = TargetType.staticLibrary;
 				}
 			}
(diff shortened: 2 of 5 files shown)


thewilsonator (approved):

rikkimax:
I'm very worried about dmd-master failing on a couple of them.

I have no idea how it's thinking it should use ldmd2 there. I think I need help.

rikkimax:
Okay just looked at another PR, and yup it's there also.

I'll remove my "fixes" to it on the next push and I'll end up rebasing so the add/remove doesn't go into commit history.

rikkimax:
@Geod24 I'm not sure what is going on with dmd-master.

The last change was for a bug related to tempDir returning multiple slashes on OSX, but that should not have done this (and it passed CI prior to pulling).

https://github.com/dlang/dub/runs/5077841404?check_suite_focus=true

I'm pinging you specifically because as far as I can tell it's someplace in build.d which you created and I have no idea where ldmd2 is coming from even though it suggests it's rdmd at fault in the stack trace.

https://github.com/dlang/dub/runs/5104069561?check_suite_focus=true

rikkimax:
It could be related to https://github.com/dlang/installer/commit/d081202c87b4078528c284e61de0e86469eb7aea

Geod24:
Yes, it looks like `rdmd` is now built with LDC by default, and so it default to LDC as a compiler.
The `std.process` stack trace points to `rdmd`, which is called from the shebang of `build.d` AFAICS.

rikkimax:
> Yes, it looks like `rdmd` is now built with LDC by default, and so it default to LDC as a compiler. The `std.process` stack trace points to `rdmd`, which is called from the shebang of `build.d` AFAICS.

Yes. rdmd should be getting the compiler from environment variable (which I did confirm previously using export that its been set). So I don't know.

Geod24:
Which variable does `rdmd` read ? Not `DC` I suppose ?

rikkimax:
> Which variable does `rdmd` read ? Not `DC` I suppose ?

Nope, I got that wrong. I was mixing up other dub-related files.

https://github.com/dlang/tools/blob/master/rdmd.d

It'll need $DMD and $DC support.

No wonder this is broken.

WebFreak001 (approved):
I like this, maybe we should add a buildOption+buildType for the link-defaultlib-shared argument for all compilers? I think it's a little bad style to use dflags directly.

WebFreak001:
let's just merge this and see where issues pop up

schveiguy:
I think this is going to break any existing projects. The way the dynamic lib dependencies are linked is completely wrong. On my Linux system, it links against the *exact relative path*, i.e. this is the output I got from ldd after running dub on a test project: 
```
../thelib/.dub/build/library-debug-linux.posix-x86_64-dmd_v2.100.0-rc.1-598A3645AB83B7B50F68A3BA1F437458/libthelib.so
```
This means if you copied this app somewhere, you'd have to reconstruct this exact directory structure to run it.

Dynamic lib deployment is not a trivial build task, which is why I think dub punted on it in the first place.

rikkimax:
So to go over a few points I raised on Discord.

Previous to this change, dub was doing the wrong thing, to the point that it had potential legal ramifications depending upon the dependencies license. This is *significantly* worse than being unable to package & deploy automatically an application built by dub.

Any usage of targetType dynamicLibrary for packages whereupon used as a dependency built was built as a static library. This is not what it says it will do and as such is relying on broken behavior. The workaround if you did this is to change the dependency to be a static library like it should have been originally and is both forwards and backwards compatible.

Any issues surrounding linking need to be directed upstream to the compilers which handle the linking. While dub can abstract over any flags for them, the compilers need to be doing the right thing in the first place for all users of D.

PR #2212, opened 2022-02-08
feat(semver): add support for versions with leading zeros

ljmf00:
DMD version scheme uses leading zeros and it is currently unable to use the
official package manager with the official reference compiler correctly with
versioning. This patch relaxes the semver requirements and allow leading zeros
on versions. This change doesn't affect the dependency resolver nor introduce
any breaking change.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -829,7 +829,7 @@ struct Version {
 	{
 		enforce(vers.length > 1, "Version strings must not be empty.");
 		if (vers[0] != branchPrefix && !vers.isGitHash && vers.ptr !is UNKNOWN_VERS.ptr)
-			enforce(vers.isValidVersion(), "Invalid SemVer format: " ~ vers);
+			enforce(vers.isValidVersion(true), "Invalid SemVer format: " ~ vers);
 		m_version = vers;
 	}
 
--- a/source/dub/semver.d
+++ b/source/dub/semver.d
@@ -24,7 +24,7 @@ import std.conv;
 /**
 	Validates a version string according to the SemVer specification.
 */
-bool isValidVersion(string ver)
+bool isValidVersion(string ver, bool allow_leading_zeros = false)
 pure @nogc {
 	// NOTE: this is not by spec, but to ensure sane input
 	if (ver.length > 256) return false;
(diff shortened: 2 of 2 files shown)


thewilsonator (approved):

CyberShadow:
Do we really need this, considering that we will have D 2.100 in a few months and thus become SemVer compatible? ;) ;) ;)

thewilsonator:
>D 2.100 

Thats 2.0100 IIRC. So no ;(

CyberShadow:
No, we decided to go with D 2.100 at the foundation meeting.

There was a bug on the website that showed it as 2.0100 in the preview changelog, but that was just because the version calculation was doing some very naive zero-padding.

thewilsonator:
Ah, good to hear.

Geod24 (changes requested):
I agree we should not add a DMD-specific special case, especially when we're finally ridding ourself of said special case.

Geod24:
> A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and **MUST NOT** contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.

[Source](https://semver.org/)

And the specs are very clear about leading zeroes.

rikkimax:
Since this PR would make dub no longer Semver complaint and dmd is getting its versions fixed upstream very soon, can we close this PR?

ljmf00:
> Do we really need this, considering that we will have D 2.100 in a few months and thus become SemVer compatible? ;) ;) ;)

Well, sticking with `2.100.0` is a workaround to the current version scheme. What if I want to support older versions for backward compatibility? I can't.

> And the specs are very clear about leading zeroes.

I know. I actually read the specification, and I understand if we don't want to move forward with this, but it is kinda odd to not be able to use older versions of the compiler on a package.

If we diverge here, we should probably think of alternatives. Perhaps, add old tags of DMD with semver compatible alternatives? It is possible to do it without messing around with the current timestamps, by manually setting `Date` header on the git object.

Geod24:
The unfortunate state of things is that it would be akin to putting lipstick on a pig: DMD as a dub package does not have a stable interface, hence it having a major version number > 0 is wrong. Every minor release should be a major release from the PoV of the library.
Until DMD-as-a-library is a reality with a stable API, any argument for backward compatibility just doesn't make sense.

rikkimax:
> Well, sticking with `2.100.0` is a workaround to the current version scheme. What if I want to support older versions for backward compatibility? I can't.

If you have a really good reason to do this, you can specify the exact hash from the repo to use.

https://github.com/dlang/dub/blob/79e312d042f66a0ea56879f13c2167bc66d35146/test/git-dependency/dub.json#L5

ljmf00:
> Until DMD-as-a-library is a reality with a stable API, any argument for backward compatibility just doesn't make sense.

Ok, fair, that makes sense.

> If you have a really good reason to do this, you can specify the exact hash from the repo to use.

One thing that is clear to me is the fact that depending on `~master` is dangerous. It introduces a lot of reproducibility issues. It would be fine, for me, to be able to do some kind of version pinning either on a hash (on the lockfile), or, ideally, not only be able to pin head refs but any git ref, including tags.

Geod24:
@ljmf00 : We do this using `git submodule` and path dependency, completely bypassing `dub` for version resolution.
You can do this with DMD because no other library is going to depend on it in a way that dub can understand.

WebFreak001 (changes requested):
https://semver.org/#semantic-versioning-specification-semver

This is against semver and allowing it in dub projects might introduce inconsistencies with other semver compatible parsers that somehow use the release tags.

ibuclaw:
> Do we really need this, considering that we will have D 2.100 in a few months and thus become SemVer compatible? ;) ;) ;)

In my head, it's still 2.0.99 -> 2.1.00 ;-)

Issue #2192, opened 2021-11-21
documented syntax for environment variables doesn't work in command directives


huglovefan:
### System information
- **dub version**: dub 1.27.0
- **OS Platform and distribution**: windows, linux
- **compiler version**: dmd-2.098.0

### Bug Description

https://dub.pm/package-format-sdl.html#environment-variables  
https://dub.pm/package-format-json.html#environment-variables

> Inside of build setting values, it is possible to use variables using dollar notation. Any variable not matching a predefined name will be taken from the program environment. To denote a literal dollar sign, use `$$`.

so according to this, the variable syntax should be the about the same as in POSIX shell, but trying to use it that way in a command directive like `preRunCommands "echo $DUB_PACKAGE"` results in an error: `Invalid variable: DUB_PACKAGE`

what actually works on posix platforms is adding a second dollar sign like `preRunCommands "echo $$DUB_PACKAGE"`, and on windows you have to use percent signs, like `preRunCommands "echo %DUB_PACKAGE%"` but neither of these are obvious from the documentation (especially the windows one)

if i'm understanding it right, a way to fix this in the docs would be:

- clarify that the variables in the third table are real environment variables and use the syntax for the platform's shell (with posix requiring an extra dollar sign to skip dub's own variable expansion)
- remove the dollar signs from variable names in the third table to hint that they're different from those in the other tables which use `$` on all platforms

### How to reproduce?

1. do `dub init`
2. add this to dub.sdl: `preRunCommands "echo $DUB_PACKAGE"` or this to dub.json: `"preRunCommands": ["echo $DUB_PACKAGE"]`
3. do `dub run`

### Expected Behavior

should print the package name before running it

### Logs

```
% grep echo dub.sdl
preRunCommands "echo $DUB_PACKAGE"
% dub
Invalid variable: DUB_PACKAGE
```

Geod24:
Using two dollars (`$$`) works.

>    "preRunCommands": ["echo $$DUB_PACKAGE"],

The documentation should be better though, and so should the error message.

WebFreak001:
the workaround seems to only work on POSIX systems, presumably it calls spawnShell, which then substitutes `$DUB_PACKAGE` with the value. However this does not work on Windows and will break stuff there.

Sample output when using `$$DC`, while `$DC` works only if DUB is being started with a DC environment variable being set:
```
'$DC' is not recognized as an internal or external command,
operable program or batch file.
[...]

PR #2217, opened 2022-02-13
move command variable substitution to execution

WebFreak001:
This is needed as commands that need to be run need to have generator
settings available before running to avoid inconsistencies with
environment variables passed to the actually ran program and variables
used in substitution.

Fixes https://github.com/dlang/dub/issues/2192

Fixes pre/postGenerateCommands inconsistently using dub environments
(execution was substituted including buildEnvironments, but shell cmd
did not get the build variables for substitution) - now both get
the build variables. Solved by removing code duplication. (env building)

Now exposes some previously private project.d processVars functions, substitution in dub describe is now done there. Exposes pre-run-commands/post-run-commands in dub describe (were implemented already, just missing in validation switch)

Brings in line the DUB behavior with the documentation (seems more sensible than telling people, that they need to make a wrapper script to actually use the environment variables or hope for shell substitution on the current platform using `$$`)

e.g. previously this was bugged:
```json
"preBuildCommands": ["$DC -run foo.d"]
```
which would say invalid variable if it wasn't an existing environment variable or with the workaround
```json
"preBuildCommands": ["$$DC -run foo.d"]
```
it would work on POSIX platforms, but not on Windows because CMD needs %DC% and powershell probably needs different syntax altogether.

Now $DC is actually replaced by DUB and the `$$` hack no longer needs to be used.

### DIFF
--- a/changelog/command-variables.dd
+++ b/changelog/command-variables.dd
@@ -1,0 +1,20 @@
+Command environment variable substitution changed
+
+Now users can use the documented predefined variables inside custom command
+directives without the need for a wrapper shell script.
+
+Before this would have failed:
+```json
+"preBuildCommands": ["$DC -run foo.d"]
+```
+unless DC was defined as environment variable outside DUB.
+
+It was before possible to run a script that used the $DC environment variable or
+on POSIX escape the `$` with `$$DC` to make the shell substitute the variable.
+These workarounds are no longer needed now.
+
+API change: none of the different command directives are no longer substituted
+with the process environment variables. You now access the raw commands as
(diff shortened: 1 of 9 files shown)


thewilsonator (approved):

rikkimax:
Right now dmd-master's is failing due to upstream issue whose patch will hit the CI in the next nightly.

But the -latest failing are an issue that needs to be resolved.

WebFreak001:
will add another test case to validate the dub describe behavior in the future

WebFreak001:
rebased into more concise isolated commits, will rebase & merge once tests pass

PR #2218, opened 2022-02-14
Fix #1856 - enable building the special test runner config

### LINKED ISSUE #1856: It should be possible to build the test target and not run it
During development it makes sense to frequently compile the codebase without running the compiled binary. However `dub build` does not build the test target, `dub test` does. But `dub test` also runs the test binary.

I want the ability to compile the `test` target without running it.

kinke:


source/dub/commandline.d
```diff
@@ -1181,8 +1182,16 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.single = m_single;
 
 		logDiagnostic("Generating using %s", m_generator);
-		dub.generateProject(m_generator, gensettings);
+
+		// With an explicitly requested `unittest` config, switch to the special test
+		// runner config (which doesn't require an existing `unittest` configuration).
+		if (m_buildConfig == "unittest")
+			dub.testProject(m_generator, gensettings, null, NativePath());
+		else
```
WebFreak001, line 1190: this should not be special to the command-line, I think this makes sense to have for API users as well (move to dub.d)

source/dub/dub.d
```diff
@@ -638,11 +638,11 @@ class Dub {
...
 	}
 
-	/** Executes tests on the current project.
+	/** Generate project files using the special test runner (`dub test`) configuration.
 
-		Throws an exception, if unittests failed.
+		Any existing project files will be overridden.
 	*/
-	void testProject(GeneratorSettings settings, string config, NativePath custom_main_file)
+	void testProject(string ide, GeneratorSettings settings, string config, NativePath custom_main_file)
```
WebFreak001, line 645: where does the name `ide` come from?

The only part where you use this is in createProjectConfig like
https://github.com/dlang/dub/blob/a72d2e5acbe2f289800d5eab6d5299a27fdc05ef/source/dub/generators/generator.d#L739

so I think you should rather name this `generator_type` to keep consistent with the other APIs.

Additionally I think you should provide a deprecated overload with the old signature that fills this value with `"build"` to make sure DUB API code stays working.

source/dub/dub.d
```diff
@@ -657,24 +657,24 @@ class Dub {
...
 
 		if (lbuildsettings.targetType == TargetType.none) {
-			logInfo(`Configuration '%s' has target type "none". Skipping test.`, config);
+			logInfo(`Configuration '%s' has target type "none". Skipping test runner build.`, config);
 			return;
 		}
 
 		if (lbuildsettings.targetType == TargetType.executable && config == "unittest") {
-			logInfo("Running custom 'unittest' configuration.", config);
+			logInfo("Building custom 'unittest' configuration.", config);
```
WebFreak001, line 673: targetType is executable, I think keeping this as `Running` makes sense

WebFreak001 (dismissed):
good changes, does this affect `dub describe` in any way?

source/dub/commandline.d
kinke: I was thinking about it, but then chose the safer path, as `generateProject()` is called in some more places. I wasn't thinking about dub-library users at all.

source/dub/dub.d
kinke: I was thinking about `dub build ...` printing `Running` to be confusing.

kinke:
> does this affect dub describe in any way?

Yep; it doesn't complain very early about a potentially missing `unittest` config, but sometime later for a quick test. I haven't tested it with a partial `unittest` config (no `executable` targetType) yet.

kinke:
Okay, I've moved the thing to `Project` in order to use it for `dub describe` as well. It's still a bit of a PITA that `dub describe` doesn't output the final path, but `targetPath` and `targetName` separately.

kinke:
Ping @WebFreak001. We're using a dub fork at Symmetry to include this.

rikkimax:
The tests should be rerun (lots of stuff has been done since then), but otherwise LGTM.

I do want to revisit refactoring addTestRunnerConfiguration further but as part of injectSourceFiles feature, so I like what I'm seeing wrt. that.

source/dub/version_.d
```diff
@@ -1,2 +1,2 @@
 module dub.version_;
-enum dubVersion = "v1.28.0-rc.1";
+enum dubVersion = "v1.28.0";
```
WebFreak001: unrelated change

WebFreak001 (approved):

PR #2220, opened 2022-02-14
fix: parse version from filesystem supplier correctly

Panke:


source/dub/packagesuppliers/filesystem.d
```diff
@@ -61,7 +61,8 @@ class FileSystemPackageSupplier : PackageSupplier {
 		string packageFileContent = packageInfoFileFromZip(filePath, packageFileName);
 		auto recipe = parsePackageRecipe(packageFileContent, packageFileName);
 		Json json = toJson(recipe);
-		json["version"] = filePath.toNativeString().split("-")[$-1].stripExtension();
+		auto basename = filePath.head.name;
+		json["version"] = basename[packageId.length + 1 .. $-4];
```
WebFreak001, line 65:
```suggestion
		enforce(basename.startsWith(packageId), "Malformed package file name " ~ filePath.toNativeString());
		json["version"] = basename[packageId.length + 1 .. $].stripExtension();
```

but you probably need to import std.exception here as well.

WebFreak001 (changes requested):
I think there should be some validation that the basename actually starts with the packageId here

Panke:
Done

source/dub/packagesuppliers/filesystem.d
```diff
@@ -61,7 +62,10 @@ class FileSystemPackageSupplier : PackageSupplier {
 		string packageFileContent = packageInfoFileFromZip(filePath, packageFileName);
 		auto recipe = parsePackageRecipe(packageFileContent, packageFileName);
 		Json json = toJson(recipe);
-		json["version"] = filePath.toNativeString().split("-")[$-1].stripExtension();
+		auto basename = filePath.head.name;
+		enforce(basename[$-4..$] == ".zip", "Malformed package filename: " ~ filePath.toNativeString);
```
WebFreak001, line 66: use endsWith for consistency pls (and to avoid potential range errors, though they cannot really occur now)

WebFreak001 (changes requested):
last change request by me (also needs import)

Panke:
Done, if the tests are still happy.

WebFreak001 (approved):

PR #2227, opened 2022-03-04
Adapt determinePlatform to the new defaults for Win32

MoonlightSentinel:
Starting with DMD 2.099, `-m32` defaults to MsCOFF instead of OMF and
`-m32omf` is required to use the OMF format + OPTLINK.

This commit changes the logic to select the appropriate logic based
on the current compiler version. Potentially using the wrong object file
format during platform probing is fine because the codegen is skipped
(dmd is invoked with `-c -o-`).


thewilsonator (approved):

thewilsonator:
Please merge when you're happy with this.

MoonlightSentinel:
Should be good now. But I've got no power here.

PR #2228, opened 2022-03-07
Define  the $DUB_BUILD_PATH variable.

veelo:
The `$DUB_BUILD_PATH` variable is now defined inside the `postBuildCommands`
section. It contains the absolute path in which the package was built, and can
be used to copy by-products of the build process to their intended locations.

This has been suggested earlier https://github.com/dlang/dub/issues/1962#issuecomment-640602652 by @WebFreak001 .

### DIFF
--- a/changelog/build-path-variable.dd
+++ b/changelog/build-path-variable.dd
@@ -1,0 +1,15 @@
+The `$DUB_BUILD_PATH` variable was added
+
+The `$DUB_BUILD_PATH` variable is now defined inside the `postBuildCommands`
+section. It contains the absolute path in which the package was built, and can
+be used to copy by-products of the build process to their intended locations.
+
+For example, if an executable exports symbols, you will want to make the
+resulting import library and symbols export file available somewhere. That can
+be done with a `dub.json` section like this:
+-------
+    "postBuildCommands-windows": [
+        "copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.lib $PACKAGE_DIR\\lib"
+        "copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.exp $PACKAGE_DIR\\lib"
+    ],
+-------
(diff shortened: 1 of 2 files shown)


changelog/build-path-variable.dd
```diff
@@ -0,0 +1,15 @@
...
+be used to copy by-products of the build process to their intended locations.
+
+For example, if an executable exports symbols, you will want to make the
+resulting import library and symbols export file available somewhere. That can
+be done with a `dub.json` section like this:
+-------
+	"postBuildCommands-windows": [
+		"copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.lib $PACKAGE_DIR\\lib"
+		"copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.exp $PACKAGE_DIR\\lib"
+	],
```
thewilsonator, line 14:
```
Searching for tabs
./changelog/pending.dd:814:	"postBuildCommands-windows": [
./changelog/pending.dd:815:		"copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.lib $PACKAGE_DIR\\lib"
./changelog/pending.dd:816:		"copy /y $DUB_BUILD_PATH\\$DUB_TARGET_NAME.exp $PACKAGE_DIR\\lib"
./changelog/pending.dd:817:	],
```
replace those with spaces

thewilsonator (approved):
Otherwise looks OK

changelog/build-path-variable.dd
veelo: Thanks

MoonlightSentinel:
This should also be available in the other commands to e.g. build other artifacts (as already mentioned in the referenced comment)

veelo:
> This should also be available in the other commands to e.g. build other artifacts (as already mentioned in the referenced comment)

Currently the `targetPath` is only generated inside this function, so making it available in `preBuildCommands` or even earlier requires a bigger reorganization. I tried that by calling `computeBuildID` with the same arguments in `generator.makeCommandEnvironmentVariables`, but that produced a different hash because somewhere in between that call and the call in this function the path was shrinked. So you would have to be very careful and generate that hash only once at one place only, to prevent any future unintended breakage.

Can you give a use case of how you would use the build path before any build would have been done? I mean, you'll have to assume that that folder is empty, so you could only use it for output or temporary files. If the process produces temporary files (`.o` files for example) I think a dependency or sub-package is a better solution. If it only produces output, I would send it to `sourceFiles` instead of the build path, and use `copyFiles` if it needs to be distributed.

PR #2239, opened 2022-04-05
Fix #2238 - Renew source files after preGenerateCommands

### LINKED ISSUE #2238: preGenerateCommands doesn't work in dub dependency
When a source file to be built is generated, using preGenerateCommands works, and it gets included in the build, but not if it's a library dependency of *another* dub package. To reproduce:

```sdl
// oops/dub.sdl
name "oops"
targetType "library"
preGenerateCommands "cd $PACKAGE_DIR/gen && dub --single gen.d"
```

```d
// oops/gen/gen.d
#!/usr/bin/env dub
/+ dub.sdl:
name "gen"
targetType "executable"
+/

void main() {
    import std.file : thisExePath, write;
    import std.path : dirName, buildPath;
    const path = buildPath(thisExePath.dirName, "..", "source", "foo.d");
    write(path, "module foo;\nint answer() { return 42; }\n");
}
```
```d
// oops/source/oops.d
import std.stdio;
import foo;

void stuff() {
    answer.writeln;
}
```

```sdl
// ledep/dub.sdl
name "ledep"
dependency "oops" path="../oops"
```

```d
// ledep/source/app.d
import oops;

void main() {
    stuff;
}
```

When trying to build "ledep", the preGenerateCommands for "oops" is run, but `foo.d` isn't included in the oops build, yielding a linker error about the missing symbol `foo.answer`.

atilaneves:


source/dub/generators/generator.d
```diff
@@ -1051,3 +1053,51 @@ private void storeRecursiveInvokations(ref const(string[string])[] env, string[]
...
+version(Posix) {
+    // https://github.com/dlang/dub/issues/2238
+	unittest {
+		import dub.internal.vibecompat.data.json : parseJsonString;
+		import dub.compilers.gdc : GDCCompiler;
+		import std.algorithm : canFind;
+		import std.path : absolutePath;
+		import std.file : mkdirRecurse, rmdirRecurse, write;
+
+		mkdirRecurse("dubtest/preGen/source");
```
WebFreak001, line 1066: I don't think the dub unittests write to the filesystem. It would probably be better to use the tests folder test system with shell scripts.

WebFreak001 (changes requested):

source/dub/generators/generator.d
atilaneves: Is that a good enough reason? I think it's a lot better to unit test even if it requires writing to the file system, especially as the only way I can get this to manifest in real life is with two dub packages where one depends on the other.

WebFreak001: my main reason would be that all the existing similar tests use the test folder (shell scripts or D scripts, optionally with their own folders where they can do whatever they want) and I don't think dub has such tests inside the unittests.

Additionally the test runner in the tests folder is more easily usable with isolating tests and debugging issues.

Example test script:
https://github.com/dlang/dub/blob/059f7a2a16cf7a50a5c68dc295c796482116ff87/test/environment-variables.script.d

and test folder:
https://github.com/dlang/dub/tree/059f7a2a16cf7a50a5c68dc295c796482116ff87/test/environment-variables

atilaneves: Can we please merge? I don't think "We've always done it this way" is a valid argument.

nordlow:
I presume using inotify on Linux and corresponding mechanism on Windows could be used here to only renew sources files that were actually changed or created.

atilaneves:
> I presume using inotify on Linux and corresponding mechanism on Windows could be used here to only renew sources files that were actually changed or created.

That's a lot more complicated for, I think, little benefit. This works.

WebFreak001 (approved):

nordlow:
Is this about making dub aware of modified and/or added source files?

WebFreak001:
no this is just to fix #2238

PR #2240, opened 2022-04-10
Inject source file (replaces #2207)

rikkimax:
See: https://github.com/dlang/dub/pull/2207

Name changed to be injectSourceFiles.

Supports plural (and should also support globs, along with platform suffixes both haven't been tested as that is not a usecase).

Conversion between json/sdl has been tested.


rikkimax:
@WebFreak001 it has been 5 days, and you're aware of the previous PR so giving you a ping.

SixthDot:
`72hours-no-objection->merge` ? maybe

WebFreak001:
will review in a minute

examples/injected-from-dependency/dub.json
```diff
@@ -0,0 +1,19 @@
...
+
+	"subPackages": [
+		{
+			"name": "toload",
+			"buildOptions": ["betterC"],
+			"sourcePaths": ["toload"],
+			"importPaths": ["toload"],
+			"injectSourceFiles": ["ahook.d"]
+		}
+	]
```
WebFreak001, line 18: I think subPackages with injectSourceFiles sound a little unrealistic in real code. If I understand the purpose of this change correctly a user would depend on something on DUB which has an injectSourceFile set. Having this as an example could cause some confusion. (though for the test it's fine)

I would suggest making a sub-folder with an isolated package, (just move this content here into a dub.json in the toload folder) which is then used as a dependency with `{"path":"./toload"}` which would effectively be the same, but for documentation purposes is closer to what will actually happen on the field.

source/dub/project.d
```diff
@@ -1022,6 +1025,7 @@ class Project {
 		case "libs":                       return listBuildSetting!"libs"(args);
 		case "linker-files":               return listBuildSetting!"linkerFiles"(args);
 		case "source-files":               return listBuildSetting!"sourceFiles"(args);
+		case "injectSourceFiles":          return listBuildSetting!"injectSourceFiles"(args);
```
WebFreak001, line 1028:
```suggestion
		case "inject-source-files":        return listBuildSetting!"injectSourceFiles"(args);
```

for consistency

source/dub/project.d
```diff
@@ -1242,6 +1246,9 @@ void processVars(ref BuildSettings dst, in Project project, in Package pack,
 	dst.addPreRunCommands(settings.preRunCommands);
 	dst.addPostRunCommands(settings.postRunCommands);
 
+	if (!settings.injectSourceFiles.empty)
```
WebFreak001, line 1249: why the check if there is any element? processVars should not do any work if it's empty and the other parts here don't use it either.

WebFreak001 (changes requested):

source/dub/project.d
rikkimax: Okay, that can go.

WebFreak001 (approved):

PR #2253, opened 2022-05-31
Fix #2234 make copies writable

### LINKED ISSUE #2234: copyFiles should overwrite read only files.
### System information

- **dub version**: 1.27.0
- **OS Platform and distribution**: Windows 10
- **compiler version** dmd-2.098.0

### Background

Binary files in a version controlled working copy are typically marked read only, in order to prevent multiple people from making simultaneous changes. Systems such as Git Large File Storage ([Git LFS](https://git-lfs.github.com/)) allow devs to acquire a lock on the file, which temporarily makes it writable, but otherwise these files remain read-only.

### Bug Description

Currently, when copyFiles copies read-only files to the targetPath, the read-only flag is retained, which causes any subsequent copy to fail. So either all copied files should be marked read-write after the copy operation, or any existing read-only files should be removed prior to the copy operation, or a copy operation should be used that ignores the read-only flag.

veelo:
When `copyFiles` copies read-only files, the copy should become writable. Otherwise a subsequent copy will fail.


source/dub/internal/vibecompat/core/file.d
```diff
@@ -156,8 +156,56 @@ private bool sameFile(NativePath a, NativePath b)
...
+		name.setAttributes(name.getAttributes & ~FILE_ATTRIBUTE_READONLY);
+	}
+	else version (Posix)
+	{
+		import core.sys.posix.sys.stat;
+
+		name.setAttributes(name.getAttributes | S_IWUSR);
+	}
+	else
+		static assert("Needs implementation.");
```
WebFreak001, line 197:
```suggestion
		static assert(false, "Needs implementation.");
```

source/dub/internal/vibecompat/core/file.d
```diff
@@ -156,8 +156,56 @@ private bool sameFile(NativePath a, NativePath b)
...
+
+	import std.exception;
+	import std.stdio;
+	assertNotThrown!ErrnoException(File(name, "w"));
```
WebFreak001, line 201:
```suggestion
	version(assert) assertNotThrown!ErrnoException(File(name, "a"));
```

didn't this clear the output file in testing? imo this line is dangerous and might also be removed completely.

WebFreak001 (changes requested):

source/dub/internal/vibecompat/core/file.d
veelo: Well spotted. Yes it probably cleared the output file, but since that was already 0 bytes in length, there was no change. I'll remove the line completely.

veelo:
Should not have force pushed. I messed things up.

source/dub/internal/vibecompat/core/file.d
```diff
@@ -156,8 +156,52 @@ private bool sameFile(NativePath a, NativePath b)
...
+	makeWritable(name.toNativeString);
+}
+
+private void makeWritable(string name)
+{
+	version (Windows)
+	{
+		import core.sys.windows.windows;
+
+		name.setAttributes(name.getAttributes & ~FILE_ATTRIBUTE_READONLY);
```
thewilsonator, line 188: Can this operations fail (should it return a bool)?

veelo: No, although setAttributes will throw if the file does not exist.

veelo: So yes it can fail, but I don't think it needs to return `bool` as `copyFile` also may throw.

thewilsonator: Fail if not exists I think is fine to leave as is.

thewilsonator:
Ping @WebFreak001

veelo:
Many thanks.

PR #2259, opened 2022-06-06
Fix issue 2258 where dependency target binaries are not copied

### LINKED ISSUE #2258: targetType: dynamicLibrary includes DLL causes compiler to error
I know this is a problem for Windows, I'm not sure about posix with .so's.

The dll should not be included, the import library should take over (it has been added).

This applies to both dmd and ldc.

I'll fix this tomorrow.

```
Target 'P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.dll' doesn't exist, need rebuild.
C:\Tools\D\ldc2-1.30.0-beta1-windows-multilib\bin\ldc2.exe -preview=dip1000 -preview=dip25 -shared -ofP:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.dll -d-debug -g -w -betterC --oq -od=P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\/obj -d-version=DynamicSideroBase -d-version=Have_sidero_base -IP:\ProjectSidero\basic_memory\source P:\ProjectSidero\basic_memory\source\sidero\base\allocators\alternatives\allocatorlist.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\alternatives\bucketizer.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\alternatives\fallback.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\alternatives\quantizer.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\alternatives\segregator.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\api.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\buffers\buddylist.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\buffers\defs.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\buffers\freelist.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\buffers\freetree.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\buffers\region.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\gc.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\gc_hook.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\locking.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\mapping\malloc.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\mapping\mmap.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\mapping\package.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\mapping\vars.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\mapping\virtualalloc.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\package.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\predefined.d P:\ProjectSidero\basic_memory\source\sidero\base\allocators\stats.d P:\ProjectSidero\basic_memory\source\sidero\base\bitmanip.d P:\ProjectSidero\basic_memory\source\sidero\base\dllmain.d P:\ProjectSidero\basic_memory\source\sidero\base\hash\adler32.d P:\ProjectSidero\basic_memory\source\sidero\base\hash\blake.d P:\ProjectSidero\basic_memory\source\sidero\base\hash\crc.d P:\ProjectSidero\basic_memory\source\sidero\base\hash\fletcher.d P:\ProjectSidero\basic_memory\source\sidero\base\hash\fnv.d P:\ProjectSidero\basic_memory\source\sidero\base\math\fixednum.d P:\ProjectSidero\basic_memory\source\sidero\base\parallelism\mutualexclusion.d P:\ProjectSidero\basic_memory\source\sidero\base\parallelism\rwmutex.d P:\ProjectSidero\basic_memory\source\sidero\base\parallelism\spincounter.d -relocation-model=pic -vcolumns
   Creating library P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.lib and object P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.exp
spew-dem2:base ~master: building configuration "spew-dem2-base-test-library"...
Copying target from P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.dll to P:\ProjectSidero\basic_memory
Copying target from P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.pdb to P:\ProjectSidero\basic_memory
Target 'P:\spew-dem2\.dub\build\spew-dem2-base-test-library-unittest-windows-x86_64-ldc_v1.30.0-beta1-CDB519E36DE1A2122FBA6F3F04D7EF26\spew-dem2-base-test-library.exe' doesn't exist, need rebuild.
C:\Tools\D\ldc2-1.30.0-beta1-windows-multilib\bin\ldc2.exe -preview=dip1000 -preview=dip25 -preview=dip1000 -preview=dip25 -of.dub\build\spew-dem2-base-test-library-unittest-windows-x86_64-ldc_v1.30.0-beta1-CDB519E36DE1A2122FBA6F3F04D7EF26\spew-dem2-base-test-library.exe -d-debug -g -unittest -w --oq -od=.dub\build\spew-dem2-base-test-library-unittest-windows-x86_64-ldc_v1.30.0-beta1-CDB519E36DE1A2122FBA6F3F04D7EF26\/obj -d-version=Have_spew_dem2_base -d-version=Have_bc_string -d-version=Have_sidero_base -d-version=DynamicSideroBase -Ibase -Igenerated_database\generated\base -IC:\Users\alpha\AppData\Local\dub\packages\bc-string-1.2.3\bc-string\source -IP:\ProjectSidero\basic_memory\source .dub\code\spew-dem2-base-test-library-unittest-windows-x86_64-ldc_v1.30.0-beta1-B0D25370577DDF82BA13CCCA75BEAD98_dub_test_root.d base\spewdem\base\array.d base\spewdem\base\console.d base\spewdem\base\containers\defs.d base\spewdem\base\containers\dynamicarray.d base\spewdem\base\containers\internal\tests.d base\spewdem\base\containers\list\defs.d base\spewdem\base\containers\list\linkedlist.d base\spewdem\base\containers\map\defs.d base\spewdem\base\containers\map\hash.d base\spewdem\base\containers\map\symboltable.d base\spewdem\base\datetime.d base\spewdem\base\encoding\base64.d base\spewdem\base\encoding\uri.d base\spewdem\base\logging.d base\spewdem\base\math\linear\matrix.d base\spewdem\base\math\linear\vector.d base\spewdem\base\meta.d base\spewdem\base\parallelism_old\exceptions.d base\spewdem\base\parallelism_old\mutualexclusion.d base\spewdem\base\parallelism_old\rwmutex.d base\spewdem\base\parallelism_old\spincounter.d base\spewdem\base\random\defs.d base\spewdem\base\random\distribution\basic.d base\spewdem\base\random\distribution\defs.d base\spewdem\base\random\distribution\normal.d base\spewdem\base\random\distribution\package.d base\spewdem\base\random\lfsr113.d base\spewdem\base\random\package.d base\spewdem\base\random\well512.d base\spewdem\base\range.d base\spewdem\base\reference\defs.d base\spewdem\base\reference\errors.d base\spewdem\base\reference\package.d base\spewdem\base\streams.d base\spewdem\base\string.d base\spewdem\base\system.d base\spewdem\base\text\ascii\builder.d base\spewdem\base\text\ascii\codepoints.d base\spewdem\base\text\ascii\readonly.d base\spewdem\base\text\ascii\wordbreak.d base\spewdem\base\text\format.d base\spewdem\base\text\internal\builder.d base\spewdem\base\text\unicode\codepoints.d base\spewdem\base\text\unicode\defs.d base\spewdem\base\text\unicode\normalization.d base\spewdem\base\text\wordbreak.d base\spewdem\base\uri\builder.d base\spewdem\base\uri\checks.d base\spewdem\base\uri\defs.d base\spewdem\base\uri\length_calculation.d base\spewdem\base\uri\mutation.d base\spewdem\base\uri\package.d base\spewdem\base\uri\readonly.d generated_database\generated\base\spewdem\base\text\unicode\database\DerivedNormalizationProps.d generated_database\generated\base\spewdem\base\text\unicode\database\PropList.d generated_database\generated\base\spewdem\base\text\unicode\database\SpecialCasing.d generated_database\generated\base\spewdem\base\text\unicode\database\UnicodeData.d generated_database\generated\base\spewdem\base\text\unicode\database\WordBreakProperty.d generated_database\generated\base\spewdem\base\text\unicode\database\package.d P:\ProjectSidero\basic_memory\source\sidero\base\gc_hook.d C:\Users\alpha\AppData\Local\dub\packages\bc-string-1.2.3\bc-string\.dub\build\default-unittest-windows-x86_64-ldc_v1.30.0-beta1-9CCBD5E0B8EBE55BD01E35397EB62F8F\bc-string.lib P:\ProjectSidero\basic_memory\.dub\build\library-unittest-windows-x86_64-ldc_v1.30.0-beta1-4C91E9CD4FAA51A0FCD48C4B2824049E\sidero_base.dll -vcolumns
Error: unrecognized file extension dll
FAIL .dub\build\spew-dem2-base-test-library-unittest-windows-x86_64-ldc_v1.30.0-beta1-CDB519E36DE1A2122FBA6F3F04D7EF26\ spew-dem2-base-test-library executable
C:\Tools\D\ldc2-1.30.0-beta1-windows-multilib\bin\ldc2.exe failed with exit code 1.
```

rikkimax:
This is very likely to create issues surrounding shared library support in ldc (it will detect regressions), I have entirely left out dmd and gdc support right now.

Fixes #2258.

### DIFF
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -47,9 +47,9 @@ bool isLinkerFile(const scope ref BuildPlatform platform, string f)
 		default:
 			return false;
 		case ".lib", ".obj", ".res", ".def":
-			return platform.platform.canFind("windows");
+			return platform.isWindows();
 		case ".a", ".o", ".so", ".dylib":
-			return !platform.platform.canFind("windows");
+			return !platform.isWindows();
 	}
 }
 
@@ -85,9 +85,9 @@ bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
 		default:
 			return false;
 		case ".lib", ".pdb", ".dll", ".exp":
-			return platform.platform.canFind("windows");
+			return platform.isWindows();
 		case ".so", ".dylib":
-			return !platform.platform.canFind("windows");
+			return !platform.isWindows();
 	}
 }
 
@@ -133,7 +133,7 @@ void resolveLibs(ref BuildSettings settings, const scope ref BuildPlatform platf
 	if (settings.targetType == TargetType.library || settings.targetType == TargetType.staticLibrary) {
 		logDiagnostic("Ignoring all import libraries for static library build.");
 		settings.libs = null;
-		if (platform.platform.canFind("windows"))
+		if (platform.isWindows())
 			settings.sourceFiles = settings.sourceFiles.filter!(f => !f.endsWith(".lib")).array;
 	}
 
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -76,7 +76,7 @@ unittest {
 /**
 	Determines if a specific file name has the extension related to dynamic libraries.
 
-	This includes dynamic libraries and for Windows pdb and import library files.
+	This includes dynamic libraries and for Windows pdb, export and import library files.
 */
 bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
 {
@@ -84,7 +84,7 @@ bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
 	switch (extension(f)) {
 		default:
 			return false;
-		case ".lib", ".pdb", ".dll":
+		case ".lib", ".pdb", ".dll", ".exp":
 			return platform.platform.canFind("windows");
 		case ".so", ".dylib":
 			return !platform.platform.canFind("windows");
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -74,6 +74,50 @@ unittest {
 
 
 /**
+	Determines if a specific file name has the extension related to dynamic libraries.
+
+	This includes dynamic libraries and for Windows pdb and import library files.
+*/
+bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
+{
+	import std.path;
+	switch (extension(f)) {
+		default:
+			return false;
+		case ".lib", ".pdb", ".dll":
+			return platform.platform.canFind("windows");
+		case ".so", ".dylib":
+			return !platform.platform.canFind("windows");
+	}
+}
+
+unittest {
+	BuildPlatform p;
+
+	p.platform = ["windows"];
+	assert(!isDynamicLibraryFile(p, "test.obj"));
+	assert(isDynamicLibraryFile(p, "test.lib"));
+	assert(isDynamicLibraryFile(p, "test.dll"));
+	assert(isDynamicLibraryFile(p, "test.pdb"));
+	assert(!isDynamicLibraryFile(p, "test.res"));
+	assert(!isDynamicLibraryFile(p, "test.o"));
+	assert(!isDynamicLibraryFile(p, "test.d"));
+	assert(!isDynamicLibraryFile(p, "test.dylib"));
+
+	p.platform = ["something else"];
+	assert(!isDynamicLibraryFile(p, "test.o"));
+	assert(!isDynamicLibraryFile(p, "test.a"));
+	assert(isDynamicLibraryFile(p, "test.so"));
+	assert(isDynamicLibraryFile(p, "test.dylib"));
+	assert(!isDynamicLibraryFile(p, "test.obj"));
+	assert(!isDynamicLibraryFile(p, "test.d"));
+	assert(!isDynamicLibraryFile(p, "test.lib"));
+	assert(!isDynamicLibraryFile(p, "test.dll"));
+	assert(!isDynamicLibraryFile(p, "test.pdb"));
+}
+
+
+/**
 	Replaces each referenced import library by the appropriate linker flags.
 
 	This function tries to invoke "pkg-config" if possible and falls back to
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -64,6 +64,7 @@ class BuildGenerator : ProjectGenerator {
 
 	override void generateTargets(GeneratorSettings settings, in TargetInfo[string] targets)
 	{
+		import std.path : setExtension;
 		scope (exit) cleanupTemporaries();
 
 		void checkPkgRequirements(const(Package) pkg)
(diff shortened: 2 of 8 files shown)


rikkimax:
Okay, looks like I have broken CI somewhat for *nix.

I won't fix that tonight.

But it is working locally for my code base. So close!

rikkimax:
After thinking about this, I think I am doing this the wrong way.

For instance the build directory may not even exist by the time I do the copy.

I'll have to start again.

rikkimax:
TIL: determining build artifacts in dub is pretty complicated.

But last build had a couple of green *nix platforms with test running for ldc, so looks like this should be green now.

rikkimax:
Okay this isn't just me: https://github.com/vibe-d/vibe.d/issues/2647

So it's actually green yay.

source/dub/generators/build.d
```diff
@@ -99,18 +100,44 @@ class BuildGenerator : ProjectGenerator {
...
 
-			NativePath[] additional_dep_files;
+			NativePath[] additional_dep_files, dependencyBinariesToCopy;
 			auto bs = ti.buildSettings.dup;
 			foreach (ldep; ti.linkDependencies) {
-				if (bs.targetType != TargetType.staticLibrary && !(bs.options & BuildOption.syntaxOnly)) {
-					bs.addSourceFiles(target_paths[ldep].toNativeString());
+				auto location = target_paths[ldep].toNativeString();
+
+				if (bs.targetType != TargetType.staticLibrary && !(bs.options & BuildOption.syntaxOnly) && isLinkerFile(settings.platform, location)) {
```
WebFreak001, line 108: where does this `isLinkerFile(settings.platform, location)` come from? doesn't seem related to the change and I don't quite see how this makes sense here either.

source/dub/compilers/utils.d
```diff
@@ -73,6 +73,50 @@ unittest {
...
+	This includes dynamic libraries and for Windows pdb, export and import library files.
+*/
+bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
+{
+	import std.path;
+	switch (extension(f)) {
+		default:
+			return false;
+		case ".lib", ".pdb", ".dll", ".exp":
+			return platform.platform.canFind("windows");
```
WebFreak001, line 88: I think we should make a function in BuildPlatform like `isWindows` which checks for this, hardcoding the string everywhere seems kinda bad.

WebFreak001, line 88: I think we should make a function in BuildPlatform like `isWindows` which checks for this, hardcoding the string everywhere seems kinda bad.

source/dub/generators/build.d
```diff
@@ -387,6 +414,14 @@ class BuildGenerator : ProjectGenerator {
...
 			if (existsFile(build_path ~ pdbFilename))
 				filenames ~= pdbFilename;
+
+			const importLibraryLocation = filenames[0].setExtension(".lib");
+			if (existsFile(build_path ~ importLibraryLocation))
+				filenames ~= importLibraryLocation;
+
+			const exportFilesLocation = filenames[0].setExtension(".exp");
+			if (existsFile(build_path ~ exportFilesLocation))
+				filenames ~= exportFilesLocation;
```
WebFreak001, line 424: I think this entire conditional block can be removed now that we have the dependencyBinariesToCopy parameter here?

WebFreak001, line 424: I think this entire conditional block can be removed now that we have the dependencyBinariesToCopy parameter here?

WebFreak001 (changes requested):

source/dub/generators/build.d
rikkimax: It comes from ``dub.compilers.utils``.

```d
bool isLinkerFile(const scope ref BuildPlatform platform, string f)
{
	import std.path;
	switch (extension(f)) {
		default:
			return false;
		case ".lib", ".obj", ".res", ".def":
			return platform.platform.canFind("windows");
		case ".a", ".o", ".so", ".dylib":
			return !platform.platform.canFind("windows");
	}
}
```

Basically it prevents ``.dll`` files from being added as a source file (which errors). This is the original error that triggered me to write this PR.

source/dub/compilers/utils.d
rikkimax: Directly above this is ``isLinkerFile`` which is what I copied to create this function.

This is a refactoring that I'm not sure is worth the potential breaking changes.

source/dub/generators/build.d
rikkimax: Capture

Double checked, nope required.

WebFreak001: I don't quite understand, isn't the array from the buildTarget function filled with this information? Is this method called elsewhere? seems like this work is duplicated

rikkimax: One works on the dependencies for a package that has already been built, and the other runs after building its package.

While they are very similar, the dependency one does have special behavior associated with "linker" files and for switching dll files to be import libraries on Windows.

source/dub/compilers/utils.d
rikkimax: There are 21 instances of ``platform.canFind("windows")``, but ok done.

WebFreak001 (commented):
change LGTM

I don't really have much experience with shared libraries on windows platforms, can you link to some reference why .lib and .exp also need to be copied to the target path? I thought they would be compiled in statically and not need to be copied.

rikkimax:
> I don't really have much experience with shared libraries on windows platforms, can you link to some reference why .lib and .exp also need to be copied to the target path? I thought they would be compiled in statically and not need to be copied.

They are compiled in statically yes.

You may want to distribute them for shared libraries, even if they are also being used as a dependency. In that combination, there is no reason to not copy them. It's common enough.

The copying of executable import and export libraries I am not sure is correct, but I am erring on the side of caution and doing it regardless because the last thing we want is people having to dig into ``.dub`` for these files.

thewilsonator (approved):

Issue #2262, opened 2022-06-08
git-derived package version of `git+https` selections needs work


kinke:
For some `"symmetry-imap": {"repository":"git+https://github.com/symmetryinvestments/imap-d.git","version":"11848bf351814cadd0095b633ea19506774739a6"}` selection, the git-derived version with dub v1.28 is: `0.1.18+commit.26.g11848bf`. This unfortunately seems to match a `0.1.18` selection and can so break another build later on (not compatible anymore 26 commits later). Note that the `git describe --tags` output uses minus, not plus: `v0.1.18-26-g11848bf`. git tags containing a plus might be problematic too.

Maybe we could just use some `0.0.0-git.<hash>` or so?

kinke:
Fixed in #2263.

PR #2263, opened 2022-06-08
Harden package version detection from `git describe`

kinke:
Resolves issue #2262.

### DIFF
--- a/source/dub/internal/git.d
+++ b/source/dub/internal/git.d
@@ -65,9 +65,6 @@ else
 // by invoking the "git" executable
 private string determineVersionWithGitTool(NativePath path)
 {
-	import dub.semver;
-	import std.algorithm : canFind;
-	import std.conv : to;
 	import std.process;
 
 	auto git_dir = path ~ ".git";
@@ -81,17 +78,9 @@ private string determineVersionWithGitTool(NativePath path)
 		return null;
 	}
 
-	auto tag = exec("git", git_dir_param, "describe", "--long", "--tags");
-	if (tag !is null) {
-		auto parts = tag.split("-");
-		auto commit = parts[$-1];
-		auto num = parts[$-2].to!int;
-		tag = parts[0 .. $-2].join("-");
-		if (tag.startsWith("v") && isValidVersion(tag[1 .. $])) {
-			if (num == 0) return tag[1 .. $];
-			else if (tag.canFind("+")) return format("%s.commit.%s.%s", tag[1 .. $], num, commit);
-			else return format("%s+commit.%s.%s", tag[1 .. $], num, commit);
-		}
+	if (const describeOutput = exec("git", git_dir_param, "describe", "--long", "--tags")) {
+		if (const ver = determineVersionFromGitDescribe(describeOutput))
+			return ver;
 	}
 
 	auto branch = exec("git", git_dir_param, "rev-parse", "--abbrev-ref", "HEAD");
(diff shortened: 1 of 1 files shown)


source/dub/internal/git.d
```diff
@@ -102,6 +91,44 @@ private string determineVersionWithGitTool(NativePath path)
...
+	// 1 commit after v1.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0-1-deadbeef") == "1.0.0-commit.1.deadbeef");
+	// tag v1.0.0+2.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0+2.0.0-0-deadbeef") == "1.0.0+2.0.0");
+	// 12 commits after tag v1.0.0+2.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0+2.0.0-12-deadbeef") == "1.0.0-commit.12.deadbeef+2.0.0");
+	// tag v1.0.0-beta.1
+	assert(determineVersionFromGitDescribe("v1.0.0-beta.1-0-deadbeef") == "1.0.0-beta.1");
+	// 2 commits after tag v1.0.0-beta.1
+	assert(determineVersionFromGitDescribe("v1.0.0-beta.1-2-deadbeef") == "1.0.0-beta.1-commit.2.deadbeef");
```
kinke, line 125: @Geod24: Could you please have a look at these and confirm that's all semver-compatible?

thewilsonator (approved):

source/dub/internal/git.d
```diff
@@ -102,6 +91,44 @@ private string determineVersionWithGitTool(NativePath path)
...
+		auto r = format("%s-commit.%s.%s", tag[1 .. (i < 0) ? $ : i], num, commit);
+		if (i > 0) r ~= tag[i .. $];
+		return r;
+	}
+	return null;
+}
+
+unittest {
+	// tag v1.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0-0-deadbeef") == "1.0.0");
```
Geod24, line 116: Note that the commit hash will always start with `g`. Don't know if you wish to rely on that or not.

source/dub/internal/git.d
```diff
@@ -102,6 +91,49 @@ private string determineVersionWithGitTool(NativePath path)
...
+	// 1 commit after v1.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0-1-deadbeef") == "1.0.0-commit.1.deadbeef");
+	// tag v1.0.0+2.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0+2.0.0-0-deadbeef") == "1.0.0+2.0.0");
+	// 12 commits after tag v1.0.0+2.0.0
+	assert(determineVersionFromGitDescribe("v1.0.0+2.0.0-12-deadbeef") == "1.0.0-commit.12.deadbeef+2.0.0");
```
Geod24, line 122: What happens with plain `v1.0.0+2.0.0` ?

Geod24: Likewise, `v1.0.0-alpha+alpn`

Geod24 (commented):
It's better, but I don't think it's the right solution.
[Using a hyphen denotes a pre-release version](https://semver.org/#spec-item-9).
Obviously, those versions are not pre-release. They are, if anything, post-release.

The only benefit we derive from using a pre-release approach is that version comparison will favor the final version over those. But again, it can lead to surprising results.

So something such as `v1.1.1-26-g6caf458` would be more suited as one of:
- `v1.1.2-26-g6caf458`
- `v1.2.0-26-g6caf458`
- `v2.0.0-26-g6caf458`

There are arguments for all three choices. I'd stay clear of the last one, as it's too purist to be of use, and would consider the second one over the first one, because it's the most likely scenario (new features).

The last approach would also to just say "dang it" and use the commit hash and expect no version match to work.

source/dub/internal/git.d
kinke: That's no valid `git describe --long --tags` output.

Geod24: If it's run on a tag, you get whatever the tag is, or am I missing something ?

kinke:
> The only benefit we derive from using a pre-release approach is that version comparison will favor the final version over those.

The (important) thing that is fixed by this is that a `1.0.0` selection isn't satisfied by some existing `1.0.0+…` git+https cache entry anymore. AFAIU, it also shouldn't match a `~>1.0` or `>=1.0.0` anymore. But a `1.0.0-commit.N.<hash>` would probably still match some `>=1.0.0-beta.1`, but I guess that's not very likely to cause trouble.

> The last approach would also to just say "dang it" and use the commit hash and expect no version match to work.

I haven't tested it, but was expecting loads of `selection <bla> doesn't match requirement <blub>` warnings in that case.

source/dub/internal/git.d
kinke: Nope, I've tested it - you get `<tag>-0-<hash>` in that case as well. Hence the `0` special case in the original code as well.

Geod24: Perfect, thanks

Geod24 (approved):
Makes sense, LGTM

s-ludwig:
This is actually introduces a pretty bad regression. Mapping the following commits to the latest version is very much on purpose and an essential part of working with multiple dependent packages side by side. Also, `1.0.0-commit.1fa6345` is not a valid SemVer version, because the last segment would be interpreted as a number due to the leading digit.

If the original behavior causes issues with a new Git cache feature that this needs to be fixed in the logic of the latter.

s-ludwig:
One way to go about this would be to exclude cached git repositories that are not checked out on an exact tag during lookup. That probably makes sense regardless, because unexpectedly getting `1.0.0-commit.1.xxxxxxx` instead of `1.0.0-beta.1` might be just as problematic.

PR #2264, opened 2022-06-10
commandline: Refactor and streamline special syntaxes

Geod24:
All those special syntaxes only trigger if args.length >= 2,
so first we can nest them in a new scope, which also makes
it visually more distinct.
Additionally, the third and second cases were actually inverted:
if the third case was to pass, it means the second case couldn't
have passed (because they check present/absence of an extension,
respectively), so we can swap them and use an else-if.
The result is arguably much more readable.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -391,6 +391,11 @@ unittest {
 */
 int runDubCommandLine(string[] args)
 {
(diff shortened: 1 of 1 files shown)


Geod24:
So all the cases were actually dependent on each other. I hope the new code is clearer.

thewilsonator (approved):

PR #2271, opened 2022-06-16
Dependency: Isolate SemVer version logic in its own struct

Geod24:
```
Currently, Dependency is a mesh of logic, having evolved from a SemVer + branch only
version specification to semver, branch, path, repository.
In order to bring it under control, and attempt to have a smarter dependency matching,
we must first isolate the different ways we specify dependencies.
```

Code has been mostly moved. I assume the CI will yell at me, so I'll likely break this down.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -45,21 +45,22 @@ struct Dependency {
 	private {
 		// Shortcut to create >=0.0.0
 		enum ANY_IDENT = "*";
-		bool m_inclusiveA = true; // A comparison > (true) or >= (false)
-		Version m_versA;
-		bool m_inclusiveB = true; // B comparison < (true) or <= (false)
-		Version m_versB;
+		VersionRange m_range;
 		NativePath m_path;
+		Repository m_repository;
 		bool m_optional = false;
 		bool m_default = false;
-		Repository m_repository;
+
 	}
 
 	/// A Dependency, which matches every valid version.
 	static @property Dependency any() { return Dependency(ANY_IDENT); }
 
 	/// An invalid dependency (with no possible version matches).
-	static @property Dependency invalid() { Dependency ret; ret.m_versA = Version.maxRelease; ret.m_versB = Version.minRelease; return ret; }
+	static @property Dependency invalid()
+	{
+		return Dependency(VersionRange(Version.maxRelease, Version.minRelease));
+	}
 
 	/** Constructs a new dependency specification from a string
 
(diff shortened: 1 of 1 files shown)


s-ludwig:
Same idea, but it got lost on the sideway: https://github.com/dlang/dub/pull/1284

Geod24:
@s-ludwig : Ideally I'd like to move this logic to a single module, namely https://github.com/dcarp/semver/ . I'll vendor it in DUB for the time being, but ideally I'd like to move it to dlang-community as many people would probably have similar needs.

Then the idea would be to make `Dependency` use `SumType` so that there is only one "provider" active at a time (path, repository, or version range).

s-ludwig:
> @s-ludwig : Ideally I'd like to move this logic to a single module, namely https://github.com/dcarp/semver/ . I'll vendor it in DUB for the time being, but ideally I'd like to move it to dlang-community as many people would probably have similar needs.

It would be really nice to switch to a self-hosted workflow where we can just use dub packages. There can still be a bootstrap script, and we could even include copies of external dependencies, instead of downloading them on demand, if someone deems that necessary. Although there should really be no need for that, considering that there is a binary included in every compiler distribution.

> Then the idea would be to make `Dependency` use `SumType` so that there is only one "provider" active at a time (path, repository, or version range).

Sounds good.

s-ludwig (approved):

PetarKirov:
@Geod24 I suggest splitting dub into subpackages, similar to vibe-d, all hosted in this repo. A good example of other package managers whose repo is split like I envision is Yarn: https://github.com/yarnpkg/berry/tree/master/packages

rikkimax:
Doing sub packagers would be a good idea indeed.

Semver, git and repository management could all be candidates for this
treatment.

Geod24:
> It would be really nice to switch to a self-hosted workflow where we can just use dub packages.

I agree. I should restart work on https://github.com/dlang/dub/pull/1972 (Two years, wow!).

> There can still be a bootstrap script, and we could even include copies of external dependencies, instead of downloading them on demand, if someone deems that necessary. Although there should really be no need for that, considering that there is a binary included in every compiler distribution.

The only negative feedback to the above PR was asking for a bootstrap, so I think I'll pick the latest version and add a CI check to make sure we don't break it.

> @Geod24 I suggest splitting dub into subpackages, similar to vibe-d, all hosted in this repo. A good example of other package managers whose repo is split like I envision is Yarn: https://github.com/yarnpkg/berry/tree/master/packages

I'm really not found of subpackages at the moment. They have quite a few quirks, and overlap with configurations. But I'll take a look.

Issue #1881, opened 2020-02-19
Root package dflags does not propagate to subpackages


Geod24:
### System information
- **dub version**: (e.g. dub 1.3.0): v1.19.0

### Bug Description

I might be way off, but when using sub-packages, I expect them to be a component of the root package.
Hence, I expect some settings to propagate from the root package to subpackages.
I recently tried setting:
> dflags "-preview=dip25"

In Vibe.d's `dub.dsl` and could see it was not propagated to subpackages: there was a DIP25 violation in vibe-d:data which wasn't caught until I added the flag there too.

### Expected Behavior

Flags in root package propagate to subpackages.

s-ludwig:
The general idea with sub packages is that they are logically completely separate packages, with the exception of being tied to the same version. Everything in addition to that would lead to another dimension of opt-in/opt-out directives that become necessary to override the default inheritance behavior.

It would also complicate the build logic for sub packages, as they would have to trace back to the root package to determine their build configuration, and it may be counter-intuitive in cases where the dependencies are reversed (`foo:bar` depends on `foo` vs. `foo` depends on `foo:bar`).

Geod24:
@s-ludwig : One problem with this approach is that some flags need to affect the whole dependency tree. Think about flags which modify the ABI used (`-fPIC`), or the semantics / mangling (`-preview=XXX`).
This ties back to https://github.com/dlang/dub/issues/279 . I tried to use `-preview=in` in my project, but since Vibe.d does not use it, I get a linker error.

PR #2272, opened 2022-06-17
Look up `dub.selections.json` files in parent directories too

kinke:
In case the root package directory doesn't contain a `dub.selections.json` file, dub now looks in parent directories too and potentially uses the first (deepest) one it finds - iff that JSON file contains an optional new `"inheritable": true` flag.

This allows using a 'central' `dub.selections.json` file for a repository containing multiple dub projects, making it automatically apply to all builds in that source tree if located in the repository root directory (unless a local `dub.selections.json` overrides it).

Such an inherited selections file is never mutated when running dub for a nested project, i.e., changes are always saved to a *local* `dub.selections.json` file.


kinke:
Ouch, looks like the testsuite fails due to dub's own `dub.selections.json` in its repo root...

WebFreak001:
I think it's probably fine to add the feature, but I'm not sure yet on the semantics. If we just inherit from parent directories that would be nice for sub-packages of projects, which want to read from the dub.selections.json anyway and not need to manually `dub upgrade` all the sub-packages.

I think it might be best to have some kind of `"inheritSelections": true` property in the dub.json/sdl, but I'm not yet fully set on that thought.

What do you think @s-ludwig ?

Geod24:
[To quote](https://github.com/dlang/dub/issues/1881#issuecomment-588397010):

> The general idea with sub packages is that they are logically completely separate packages, with the exception of being tied to the same version. Everything in addition to that would lead to another dimension of opt-in/opt-out directives that become necessary to override the default inheritance behavior.

I think we're ought to either stick with this principle and properly document it, OR re-evaluate what is people's natural expectation / need around subpackages. It's confusing even to maintainers when subpackages should be used vs configurations.

nordlow:
How do other languages solves these issues? Would be interesting to check if Rust have these same issues and implement them in a similar way to what we plan.

nordlow:
Apart from merging with master what remains in this MR, @kinke?

kinke:
> Apart from merging with master what more work needs to happen in this MR?

Well, it's a pretty big breaking change and might not be always desirable. So maybe make it opt-in via cmdline flag or extending the dub.selections.json file (`inheritable: true` or so - but that would still lead to more file lookups).

> What is meant by "changes are always saved to a local dub.selections.json file."?

If you run e.g. `dub upgrade` in a nested project, changes are saved to a dub.selections.json file *in that dir*, not to the inherited file from some parent dir.

source/dub/project.d
```diff
@@ -1742,6 +1750,23 @@ final class SelectedVersions {
...
+		in the root package directory.
+	*/
+	this(Selected data, NativePath relPathPrefix)
+	{
+		this(data);
+		if (relPathPrefix.empty) return;
+		foreach (ref dep; m_selections.versions.byValue) {
+			const depPath = dep.path;
+			if (!depPath.empty && !depPath.absolute)
+				dep.path = relPathPrefix ~ depPath;
```
WebFreak001, line 1766:
```
source/dub/project.d(1766,5): Deprecation: function `dub.dependency.Dependency.path` is deprecated - Construct a new `Dependency` object instead
```

kinke: Yeah thx, seen it via that awesome bot too, cool stuff! :+1:

kinke:
Rebased and made opt-in via a new optional `inheritable: true` in the `dub.selections.json` file to be inherited by nested projects.

I've left the `fileVersion` at 1 for now; I guess we'd have to bump it. But maybe only for files containing `inheritable: true`, so that the whole D ecosystem doesn't have to be adapted.

WebFreak001:
I think keeping fileVersion at 1 is fine here, since it's not backwards incompatible

kinke:
Ping.

source/dub/project.d
```diff
@@ -1957,3 +1979,39 @@ private immutable DefaultTestRunnerCode = q{
...
+			// The default is to error, but as the previous parser wasn't
+			// complaining, we should first warn the user.
+			auto selected = parseConfigFileSimple!Selected(selverfile.toNativeString(), StrictMode.Warn);
+			const isValid = !selected.isNull() && (dir == absRootPackagePath || selected.get().inheritable);
+			if (isValid) {
+				return N(SelverLookupResult(
+					selverfile,
+					new SelectedVersions(selected.get(), dir.relativeTo(absRootPackagePath))
+				));
+			}
```
WebFreak001, line 2011: in case of invalid selections files, shouldn't we rather error than silently ignoring the file?

WebFreak001 (changes requested):

changelog/selections_from_parent_dir.dd
```diff
@@ -0,0 +1,15 @@
+`dub.selections.json` files are now looked up in parent directories too
+
+In case the root package directory doesn't contain a `dub.selections.json`
+file, dub now looks in parent directories too and potentially uses the
+first (deepest) one it finds - iff that JSON file contains an optional
+new `inheritable: true` flag.
```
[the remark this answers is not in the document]
kinke: It wouldn't make it clear that for some project in `root/dir1/dir2`, an inheritable `root/dub.selections.json` wouldn't apply if there's a non-inheritable `root/dir1/dub.selections.json`. `iff` should be pretty common in IT specs. :)

source/dub/project.d
kinke: I'm pretty sure I kept the behavior as-is (at that time - merge conflicts now obviously after all that time) - an empty `Selections` object if there's no file or the file is invalid.

atilaneves:
Given the opt-in nature now with having to set inheritability in the parent selections file, I don't see any reason to not merget his. Objections?

kinke:
Updated yet again to resolve merge conflicts. The change is better integrated now; I got rid of the earlier new extra helper (`lookupAndParseSelectionsFile()` or so), and instead adapted the new (in master) `PackageManager.readSelections()` method, incl. making that `public`. The reason is that reggae needs to know the path to the used `dub.selections.json`, so that it can add a proper dependency in the generated build scripts, leading to a reggae re-run whenever a used `dub.selections.json` has changed.

test/selections-from-parent-dir.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+cd "${CURR_DIR}/selections-from-parent-dir/pkg2" || die "Could not cd."
+
+"$DUB" build --nodeps
+
+if [ -f dub.selections.json ]; then
+    die "Shouldn't create a dub.selections.json in pkg2 subdir."
+fi
```
Geod24, line 10: Any chance you could convert this to the new test framework ? E.g. see `source/dub/test/other.d`.

changelog/selections_from_parent_dir.dd
```diff
@@ -0,0 +1,16 @@
...
+
+This allows using a 'central' `dub.selections.json` file for a repository
+containing multiple dub projects, making it automatically apply to all
+builds in that source tree if located in the repository root directory
+(unless a local `dub.selections.json` overrides it).
+
+Such an inherited selections file is never mutated when running dub for a
+nested project, i.e., changes are always saved to a *local*
+`dub.selections.json` file. E.g., when running `dub upgrade` for a nested
+project.
```
Geod24, line 16: Can we add a test for this ?

source/dub/packagemanager.d
```diff
@@ -60,6 +60,11 @@ public string toString (PlacementLocation loc) @safe pure nothrow @nogc
     }
 }
 
+struct SelectionsFileLookupResult {
+	NativePath absolutePath; // to dub.selections.json
+	SelectionsFile selectionsFile;
+}
+
```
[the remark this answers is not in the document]
kinke: 2. would be doable after a quick glance, but would need a changed signature for the public `SelectionsFile.fromYAML()` factory function (unused in dub AFAICT), adding a `path` parameter.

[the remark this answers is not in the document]
kinke: Yeah I figured, so I went with option 1.

source/dub/packagemanager.d
```diff
@@ -1151,32 +1156,40 @@ symlink_exit:
...
+	 *                    which to load the selections file.
 	 *
 	 * Returns:
-	 *   The parsed `SelectionsFile`, or a null instance if none exists.
+	 *   Either `null` (if no selections file exists or parsing encountered an error),
+	 *   or a `SelectionsFileLookupResult`. Note that the nested `SelectionsFile`
+	 *   might use an unsupported version (see `SelectionsFile` documentation).
 	 */
-	package final Nullable!SelectionsFile readSelections(in Package project) {
+	final Nullable!SelectionsFileLookupResult readSelections(in NativePath absProjectPath)
```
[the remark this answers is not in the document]
kinke: Non-final now.

test/selections-from-parent-dir.sh
kinke: I did; pretty cool. :+1: - `dub test -c library-nonet` fails for some rather bizarre reason (the CI problem), `dub test` alone works fine.

kinke: Solved; was a problem of the mocked file system (no entries for `.`  and `..`, so needs extra normalization; might be required in other places too).

source/dub/packagemanager.d
```diff
@@ -1151,32 +1160,40 @@ symlink_exit:
...
-		return wrapException(parseConfigString!SelectionsFile(
-			content, path.toNativeString(), StrictMode.Warn));
+		alias N = typeof(return);
+
+		// check for dub.selections.json in root project dir first, then walk up its
+		// parent directories and look for inheritable dub.selections.json files
+		for (NativePath dir = absProjectPath; !dir.empty; dir = dir.hasParentPath ? dir.parentPath : NativePath.init) {
+			const path = dir ~ "dub.selections.json";
+			if (!this.existsFile(path))
+				continue;
```
Geod24, line 1183: Logically this would be better represented as something that returns a path that is then acted on. The code as-is does not mean to be changed (it works, that's what matters) but a loop that ends up with an unconditional `break` is not a loop (or IOW, contains too much code).

source/dub/project.d
```diff
@@ -108,22 +108,24 @@ class Project {
 		import dub.version_;
 		import dub.internal.dyaml.stdsumtype;
 
-		auto selected = mgr.readSelections(pack);
-		// Parsing error that will be displayed to the user or just no selections
-		if (selected.isNull())
+		auto lookupResult = mgr.readSelections(pack.path);
```
Geod24, line 111: How do we know here that the path is absolute ?
Loops like we should just change the signature of the function ?

source/dub/recipe/selection.d
```diff
@@ -44,6 +44,18 @@ public struct SelectionsFile
...
+     * without local dub.selections.json
+     */
+    public bool inheritable () const @safe pure nothrow @nogc
+    {
+        return this.content.match!(
+            (const Selections!0 _) => false,
+            (const Selections!1 s) => s.inheritable,
+        );
+    }
+
```
Geod24, line 58: I don't know if we should expose this convenience function here. The intent of `SelectionFile` is just to wrap a specific `Selections!uint`. However, the implementation makes sense logically. IMO the question boils down to: "Do we ever find ourselves in a situation where we manipulate an invalid `Selections!uint`?". I would be tempted to say no, as trying to make a generic `Dependency` instead of the discriminated union it is now is what made the code a mess.

source/dub/packagemanager.d
```diff
@@ -60,6 +60,15 @@ public string toString (PlacementLocation loc) @safe pure nothrow @nogc
...
 
+/// A SelectionsFile associated with its file-system path.
+struct SelectionsFileLookupResult {
+	/// The absolute path to the dub.selections.json file
+	/// (potentially inherited from a parent directory of the root package).
+	NativePath absolutePath;
+	/// The parsed dub.selections.json file.
+	SelectionsFile selectionsFile;
+}
+
```
Geod24, line 71: I would like to fold this into `SelectionsFile`, but I know it's not currently possible. All things considered, it might be a good thing to add to `Configy`, so I'll have a look at it myself. It should not block this being merged.

source/dub/test/selections_from_parent_dir.d
```diff
@@ -0,0 +1,113 @@
+module dub.test.selections_from_parent_dir;
```
Geod24, line 1: Module doc please

source/dub/test/selections_from_parent_dir.d
```diff
@@ -0,0 +1,113 @@
...
+    assert(dub.project.selections.getSelectedVersion(PackageName("pkg1")).path == NativePath("../pkg1"));
+
+    // invoking `dub upgrade` for the pkg2 root package should generate a local dub.selections.json,
+    // leaving the inherited one untouched
+    dub.upgrade(UpgradeOptions.select);
+    const nestedPath = pkg2Dir ~ "dub.selections.json";
+    assert(dub.fs.existsFile(nestedPath));
+    assert(dub.fs.readFile(path) == dubSelectionsJsonContent,
+        "Inherited dub.selections.json modified after dub upgrade!");
+    const nestedContent = cast(string) dub.fs.readFile(nestedPath);
```
Geod24, line 48: `readText`

source/dub/test/selections_from_parent_dir.d
```diff
@@ -0,0 +1,113 @@
...
+
+    // after removing non-inheritable root/a/dub.selections.json: inherited root selections for root/a/b/
+    {
+        dub.fs.removeFile(root_a ~ "dub.selections.json");
+
+        const result = dub.packageManager.readSelections(root_a_b);
+        assert(!result.isNull());
+        assert(result.get().absolutePath == root ~ "dub.selections.json");
+        assert(result.get().selectionsFile.inheritable);
+    }
```
Geod24, line 112: Maybe those tests should check the version too ? IIUC before the `rm` `hasAlldependencies` should be `false`, and it should be `true` afterwards. You can use `TestDub.newTest` to easily test this.

Geod24 (approved):
Looks good! Made a few suggestions. I thing there's nothing blocking the merge (although I would like to see the points addressed, especially the ones about `inheritable` property and absolute path, as that could trigger an assert easily) but can do it myself if you just want to go agead.

Geod24:
@kinke : I took the liberty to push the requested fixes, as I know you are away and I'm sure you'd like this fixed sooner than later.

Geod24:
> This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.
 
 Well...

Geod24:
```
../../../.dub/packages/vibe-core/2.7.1/vibe-core/source/vibe/core/path.d(72,8): Error: none of the overloads of `endsWithSlash` are callable using argument types `(bool) const`
../../../.dub/packages/vibe-core/2.7.1/vibe-core/source/vibe/core/path.d(510,17):        Candidates are: `vibe.core.path.GenericPath!(PosixPathFormat).GenericPath.endsWithSlash() const`
../../../.dub/packages/vibe-core/2.7.1/vibe-core/source/vibe/core/path.d(512,17):                        `vibe.core.path.GenericPath!(PosixPathFormat).GenericPath.endsWithSlash(bool v)`
source/dub/project.d(127,60): Error: template instance `vibe.core.path.relativeTo!(const(GenericPath!(PosixPathFormat)))` error instantiating
```

Well looks like vibe-core has a bug too, but we can work around that.

Geod24:
None of the workaround is pretty though, so I opened https://github.com/vibe-d/vibe-core/pull/400

kinke:
Thanks Mathias!

PR #2276, opened 2022-06-20
sdl/json: Do not create dummy dependencyBuildSettings entry

Geod24:
Because if a user provide an object as a dependency,
such as `{ "path": "foo/bar", "optional": true }`,
we would create a dummy entry in the AA.

### DIFF
--- a/source/dub/recipe/json.d
+++ b/source/dub/recipe/json.d
@@ -182,7 +182,10 @@ private void parseJson(ref BuildSettingsTemplate bs, Json json, string package_n
 					{
 						BuildSettingsTemplate dbs;
 						dbs.parseJson(verspec, package_name);
-						bs.dependencyBuildSettings[pkg] = dbs;
+						// Only create an entry if there's an actual BuildSetting
+						// defined by the user.
+						if (dbs !is BuildSettingsTemplate.init)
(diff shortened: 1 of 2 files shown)


Geod24:
CC @omerfirmak

WebFreak001:
does this fix any issue? In the codebase there are sometimes AAs that have simply no values, just to act as look-up for the keys to see if they are contained quickly and I think this could affect such usage.

Geod24:
> does this fix any issue?

Context: I'm revamping the `dub` recipe parsing logic. I have a library which was made for configuration parsing written in a declarative style, and ideally with the idea of targeting this library. It worked surprisingly well with `dub`, but some minor differences came up, this being one of them.

> In the codebase there are sometimes AAs that have simply no values, just to act as look-up for the keys to see if they are contained quickly and I think this could affect such usage.

This is not the case here. Things would crash with `"vibe-d": "~>0.9.5"` otherwise (because it wouldn't add that dummy entry).
Also the code originally didn't take SDL into account, and that would have been caught by any sdl recipe using a dependency.

omerfirmak:
LGTM. 

AFAICS, dummy entry is not assumed to be present anywhere in the code.

WebFreak001 (approved):

PR #2279, opened 2022-06-20
Remove leading hyphen in platform specs

Geod24:
It doesn't serve any purpose. I was tripped by it when iterating over the AA.

### DIFF
--- a/source/dub/platform.d
+++ b/source/dub/platform.d
@@ -188,16 +188,14 @@ string determineCompiler()
 */
 bool matchesSpecification(in BuildPlatform platform, const(char)[] specification)
 {
-	import std.string : format;
+	import std.string : chompPrefix, format;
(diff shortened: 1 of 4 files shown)


WebFreak001 (changes requested):
this is a public method that's assumed to be simple enough that it might be used elsewhere in other projects, please move the `-` stripping into the matchesSpecification method.

WebFreak001:
no I meant make it so the leading hyphen is optional (stripped if present, otherwise left as is)

e.g. do
```d
specification = specification.chompPrefix("-");
```

as input sanitizing

Geod24:
Right. I'll keep it as is until I flush out all uses, then I'll make the change you suggested.

Geod24:
Okay, we're rid of them, I added the `chompPrefix`

kinke (approved):

PR #2281, opened 2022-06-21
add $DPATH environment variable for file locations

### LINKED ISSUE #229: Support custom paths for the default installation folder
Currently all packages are installed to `~/.dub/packages/*` by default, or `%AppData%\dub\packages\*` on Windows. A setting in the `~/dub/settings.json` fie should be made available to change that default.

WebFreak001:
fix #229

I chose not to add a setting for this to make it easier for other apps
in the ecosystem to understand how to locate and change the dub folder,
to parse packages, manipulate them, maintain them, etc.

Having a setting would require other users to potentially locate the
config files (which has a bunch of exceptions), parse a JSON and find
an optional value. Limiting it to an environment variable means that
also tools like scripts can easily find the location of this.


Geod24:
But an environment variable is prone to other kind of issues. For one it's not persistent unless you edit your shell config, which looses locality. Also names are prone to collision. IMO if one wants to get this value in script, we should have a `dub config` command.

WebFreak001:
imo if you set your environment variable only locally then you probably only want DUB to store its data where you specified in that session. (maybe set to `/tmp/dub` to quickly try out some package without downloading it, or different stores for every different user)

WebFreak001:
if we don't plan to add other tools, it might also make sense to define a `DUB_HOME` environment variable instead, which defaults to `~/.dub` / `%APPDATA%/dub`, so the folder can be overwritten in whole.

source/dub/dub.d
```diff
@@ -1931,4 +1954,12 @@ private class DubConfig {
 		if (m_parentConfig) return m_parentConfig.defaultPostRunEnvironments;
 		return null;
 	}
+
+	@property string dubHome()
+	const {
+		if(auto pv = "dubHome" in m_data)
```
WebFreak001, line 1960: should the setting for this be called `dubHome` or something else?

WebFreak001:
after discussing with @Geod24 I changed this to also include a setting instead of only an environment variable. This is so that system-wide changes can be made, e.g. by linux package maintainers, but also kept an environment variable so that it's possible to create a quick temporary clean local dub package storage or in general if it might be needed for some specific user's projects on the system.

To support changing the path of the packages (and also the config), the config in /etc/dub/settings.json (or /usr/...) will be read first and if the `dubHome` property is set in there, loading of further configs will be redirected to there, as well as the packages being stored there.

dk949:
Hi 👋

I've been working on something similar (I guess now I'll just wait for this to merge), and just wanted to share some thoughts from my implementation.

I also considered something like `DUB_HOME` or `DUB_PATH`, but came to the conclusion that it's far too generic, which means users may have it set to control some other part of their system (`DPATH` would suffer the most from this). 

I am not sure how common this is on Windows or Mac, but a lot of Linux programs use environment variables to control the location of their config/cache/etc files (see [this page](https://wiki.archlinux.org/title/XDG_Base_Directory#Partial) for some examples).

In the end I settled for `DUB_SYSTEM_SETTINGS_PATH`, `DUB_USER_SETTINGS_PATH` and `DUB_LOCAL_REPOSITORY_PATH`. As I understand you intend other D tooling to use this directory, so maybe something like `DLANG_HOME` (in line with `JDK_HOME`)?

As for "...it's not persistent unless you edit your shell config", well yes, but users are already used to doing this. zsh even has a special config file pretty much for this purpose (`.zshenv`).

[My impl](https://github.com/dk949/dub/tree/user_specified_directories) for reference.

WebFreak001:
I made it now so that environment variables can only change both folders at once, but you can use the settings for finer control of user settings and repository path, or use the general "dubHome" setting to set both.

Additionally for package maintainers and sysadmins it's probably interesting to have environment variables in the settings, for example to have a different user store for each user, so I added that as well.

See the changelog entry for details https://github.com/dlang/dub/blob/5addac0e22036cf6b328bda72ac87c47d55d0331/changelog/dpath.dd

source/dub/dub.d
```diff
@@ -285,11 +311,48 @@ class Dub {
...
+			if (m_config.userSettings.length) {
+				m_dirs.userSettings = NativePath(m_config.userSettings.expandEnvironmentVariables);
+
+				overrideUserSettings = true;
+			} else if (m_config.dubHome.length) {
+				m_dirs.userSettings = NativePath(m_config.dubHome.expandEnvironmentVariables);
+
+				overrideUserSettings = true;
+			}
+		}
```
Geod24, line 329: The circular dependency is IMO weird. I need a bit of time to think about it.

Geod24 (commented):
I just realized we read `/etc/` before `/var/lib`...

The idea of a "dub home" is a bit weird: We don't have a unified "root location". We have the badly-named `Repository` (see the `private struct` in `packagemanager.d`) and we tackled on config files on top of that.

Also what's the relationship with `customCachePaths` ?

Geod24:
Any chance we could get https://github.com/dlang/dub/pull/2343 before, since it overhaul this part of the code and simplifies the logic ? I can then rebase your branch on top of it, if you'd like.

changelog/dpath.dd
```diff
@@ -0,0 +1,41 @@
...
+Additionally, these config paths will have environment variables using the `$VARIABLE` syntax resolved.
+
+The following list describes which path is going to be picked, from top to bottom, stopping whenever one is found:
+
+- `$DUB_HOME` environment variable
+- `$DPATH` environment variable
+- system-wide settings.json: `"userSettings"` property
+- system-wide settings.json: `"dubHome"` property (only for userSettings)
+- through all settings.json: `"localRepository"` property
+- through all settings.json: `"dubHome"` property (only for localRepository)
```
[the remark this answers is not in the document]
WebFreak001: this overrides the built-in dub location, so it would never even attempt to read anything from the built-in path, while for customCachePaths it's just an addition of paths.

source/dub/dub.d
WebFreak001: it's a little weird in the code, but I think it's great for user flexibility.

A use case here: package maintainer creates global settings.json in `/var/lib/dub` which then specifies where the user-overridable config is located. On linux distros that have read-only filesystems with some special paths for writable things that might for example be useful. Or for users using non-standard paths like storing all dub things on a separate flash drive or something that just isn't mounted in /etc.

Additionally it's possible to make the config only per-user and use an environment variable like `$HOME` inside the settings file so each user can have different dub settings.

changelog/dpath.dd
```diff
@@ -0,0 +1,41 @@
...
+
+The `DPATH` environment variable is intended to be used by all D tooling related things doing user-space installation of things. It can be used to avoid cluttering the home folder.
+
+Additionally to environment variables it is possible to configure the package placement path + settings.json path through DUB's settings.json file. To configure where the user-editable settings.json is placed you need to adjust the system-wide dub configuration.
+
+In the settings.json you can set the following fields:
+
+```json
+{
+    "dubHome": "/path/to/dub", // sets both package store and config location
```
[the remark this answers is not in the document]
WebFreak001: this overrides the user settings (e.g. what /etc/dub is for, but with a custom path)

`/var/lib/dub` is still used for reading the first base settings.

WebFreak001:
> I just realized we read /etc/ before /var/lib...

I don't think that's the case, the systemSettings path is the very first config that is read, everything after that overrides the configs.

changelog/dpath.dd
```diff
@@ -0,0 +1,41 @@
+DUB settings & packages directory placement overhauled
+
+You can now configure where DUB places its downloaded packages and where the user configuration is stored through environment variables or through the dub configuration. You need to use an environment variable or the system-wide dub configuration to specify where the user configuration is stored.
+
+By default DUB stores the packages on
+- Windows: `%APPDATA%/dub/{packages/,settings.json}`
```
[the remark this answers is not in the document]
WebFreak001: that is for localRepository, but I think the more used thing is the userRepository. I don't think localRepository is ever used unless the user explicitly wants to store things in the local repository.

For simplification in the config, the config overrides both the user and the local repository.

source/dub/dub.d
```diff
@@ -271,10 +271,36 @@ class Dub {
...
+
+			}
+			if (dpathOverride.length) {
+				overrideUserSettings = true;
+				overrideLocalRepository = true;
+
+				m_dirs.userSettings = NativePath(dpathOverride);
+				m_dirs.localRepository = m_dirs.userSettings;
+			}
+		}
```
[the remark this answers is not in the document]
WebFreak001: I think that's difficult as the state variables overrideUserSettings and overrideLocalRepository are later used to determine if custom path logic should be applied.

changelog/dpath.dd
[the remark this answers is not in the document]
Geod24: `dub fetch` (and related `fetch` commands) use `localRepository`:
https://github.com/dlang/dub/blob/17d8c9467b81cf2b2ac57c338f01f71c2c71b013/source/dub/dub.d#L768-L773

The logic behind it, as I very recently understood and documented, is that using APPDATA for settings is good (you want your settings to roam), but not for packages (because it might use a lot of space / lead to complications with stale binaries):
https://github.com/dlang/dub/blob/17d8c9467b81cf2b2ac57c338f01f71c2c71b013/source/dub/dub.d#L1748-L1750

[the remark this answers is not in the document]
Geod24: IOW:
```diff
-- Windows: `%APPDATA%/dub/{packages/,settings.json}`
+- Windows: `%APPDATA%/dub/settings.json` 
+- Windows: `%LOCALAPPDATA%/dub/packages/` 
```

I don't know if that should affect the functionality too.

Geod24:
> I don't think that's the case, the systemSettings path is the very first config that is read, everything after that overrides the configs.

My bad, I meant the other way around :facepalm: 
I think we should read `/etc/` first, but we currently don't.

source/dub/dub.d
```diff
@@ -285,11 +307,48 @@ class Dub {
...
 		}
 
+		// Override user + local package path from /etc/dub/settings.json
+		// Then continues loading local settings from these folders. (keeping
+		// global /etc/dub/settings.json settings intact)
+		//
+		// Don't use it if either $DPATH or $DUB_HOME are set.
+		if (!overrideUserSettings) {
+			if (m_config.userSettings.length) {
+				m_dirs.userSettings = NativePath(m_config.userSettings.expandEnvironmentVariables);
```
Geod24, line 317: What happens if the environment variable can't get expanded ?

changelog/dpath.dd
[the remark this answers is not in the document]
WebFreak001: oh I see now, I thought localSettings correspond to PlacementLocation.local and userSettings to PlacementLocation.user, but it turns out to be a bit different there, which is confusing. Will update the changelog then.

source/dub/dub.d
Geod24: Can we make things a bit more hierarchical, now that we have the ability to (easily) do so ?
Like `core.author` in `git`, you know.

source/dub/dub.d
WebFreak001: nothing is inserted in that case, (variable replaced with empty string) probably worth putting into dub-docs

WebFreak001: nothing is inserted in that case, (variable replaced with empty string) probably worth putting into dub-docs

Geod24: I think it should error out.

WebFreak001:
> > I don't think that's the case, the systemSettings path is the very first config that is read, everything after that overrides the configs.
> 
> My bad, I meant the other way around facepalm I think we should read `/etc/` first, but we currently don't.

I strongly feel that we should not read /etc first, but /var/lib, like currently already done. I think the idea behind that is:

1. consistency with other programs (e.g. systemd)
2. the `/var/lib/...` path is maintained by the distro package maintainers and should never be updated by the user, this implies:
    - any setting in /var/lib should not override the user-set settings, thus it must not be loaded after the user settings in /etc

source/dub/dub.d
WebFreak001: done

Geod24:
One use-case that might be important is the default placement path.
Currently the the `localRepository` (user HOME on POSIX) is the default placement.
Some users might want to use `system`, and some might want to use `local` (e.g. in the package itself). Currently the always have to pass an argument to DUB.

If you prefer we can tackle it in a separate PR, but for me it makes much more sense to think of them together (I thought the scope of this PR was originally bigger).

WebFreak001:
I think a new setting to set the default placement location should be part of a separate PR.

dk949:
I though you were planning to tie the location of `localRrpository` (aka `$HOME/.dub`, aka default placement) to `$DPATH`. If this is the case, wouldn't it make it somewhat inseparable from this PR?

(I too have possibly misunderstood the scope of these changes).

WebFreak001:
rebased to new yaml loader

WebFreak001:
> I though you were planning to tie the location of `localRrpository` (aka `$HOME/.dub`, aka default placement) to `$DPATH`. If this is the case, wouldn't it make it somewhat inseparable from this PR?
> 
> (I too have possibly misunderstood the scope of these changes).

no the default placement location setting I was talking about would set a default for the `dub --cache=*` setting (can set it to local, system or user)

dk949:
Ohh, that makes sense. My bad.

WebFreak001:
@Geod24 what's your opinion on this?

source/dub/dub.d
```diff
@@ -275,16 +275,77 @@ class Dub {
...
+		bool overrideUserSettings;
+		bool overrideLocalRepository;
+		{
+			string dpathOverride = environment.get("DUB_HOME");
+			if (!dpathOverride.length) {
+				dpathOverride = environment.get("DPATH");
+				if (dpathOverride.length)
+					dpathOverride = (NativePath(dpathOverride) ~ "dub/").toNativeString();
+
+			}
```
Geod24, line 289:
```suggestion
			string dubHome = environment.get("DUB_HOME");
			if (!dubHome.length) {
				if (auto dpath = environment.get("DPATH"))
					dubHome = (NativePath(dpath) ~ "dub/").toNativeString();
			}
```

IMO that makes more sense, no?

source/dub/dub.d
```diff
@@ -350,6 +411,10 @@ class Dub {
 	@property const(string[string]) defaultPreRunEnvironments() const { return this.m_config.defaultPreRunEnvironments; }
 	@property const(string[string]) defaultPostRunEnvironments() const { return this.m_config.defaultPostRunEnvironments; }
 
+	private @property const(string) dubHome() const { return this.m_config.dubHome; }
+	private @property const(string) userSettingsOverride() const { return this.m_config.userSettings; }
```
Geod24, line 415: Why the `override` ? Also wouldn't we be better simply exposing `SpecialDir` ?
But again, why expose it ? I don't see any user, and YAGNI.

source/dub/dub.d
```diff
@@ -275,16 +275,77 @@ class Dub {
...
 		version (Posix) {
 			if (dubFolderPath.absolute && dubFolderPath.startsWith(NativePath("usr")))
 				readSettingsFile(NativePath("/etc/dub/settings.json"));
 		}
+
+		// Override user + local package path from /etc/dub/settings.json
+		// Then continues loading local settings from these folders. (keeping
+		// global /etc/dub/settings.json settings intact)
+		//
+		// Don't use it if either $DPATH or $DUB_HOME are set.
```
Geod24, line 311:
```suggestion
		// Don't use it if either $DPATH or $DUB_HOME are set, as environment
		// variables usually take precedence over configuration.
```

source/dub/dub.d
```diff
@@ -275,16 +275,77 @@ class Dub {
...
+
+		// load system-wide configs:
 		readSettingsFile(m_dirs.systemSettings ~ "settings.json");
 		readSettingsFile(dubFolderPath ~ "../etc/dub/settings.json");
 		version (Posix) {
 			if (dubFolderPath.absolute && dubFolderPath.startsWith(NativePath("usr")))
 				readSettingsFile(NativePath("/etc/dub/settings.json"));
 		}
+
+		// Override user + local package path from /etc/dub/settings.json
```
Geod24, line 307:
```suggestion
		// Override user + local package path from system / binary settings
```

Since we already loaded the config from 3 locations

source/dub/dub.d
Geod24: Oh damn this is returning the config, not even the actual values! A big no from me.

source/dub/dub.d
```diff
@@ -275,16 +275,77 @@ class Dub {
...
+			if (this.userSettingsOverride.length) {
+				m_dirs.userSettings = NativePath(this.userSettingsOverride.expandEnvironmentVariables);
+
+				overrideUserSettings = true;
+			} else if (this.dubHome.length) {
+				m_dirs.userSettings = NativePath(this.dubHome.expandEnvironmentVariables);
+
+				overrideUserSettings = true;
+			}
+		}
```
Geod24, line 322: I'm a bit lost... Why do we have `dubHome` AND `userSettings` ?
Also use `this.m_config.userSettings.set` for condition, instead of checking length.

test/dpath-variable/dub.json
```diff
@@ -0,0 +1,6 @@
+{
+	"name": "dpath-variable",
+	"dependencies": {
+		"gitcompatibledubpackage": "1.0.1"
+	}
+}
```
Geod24, line 6: Newline

test/dpath-variable/source/app.d
```diff
@@ -0,0 +1,3 @@
+void main()
+{
+}
```
Geod24, line 3: Newline

source/dub/dub.d
```diff
@@ -275,16 +275,77 @@ class Dub {
...
+				m_dirs.localRepository = NativePath(this.localRepository.expandEnvironmentVariables);
+
+				overrideLocalRepository = true;
+			} else if (this.dubHome.length) {
+				m_dirs.localRepository = NativePath(this.dubHome.expandEnvironmentVariables);
+
+				overrideLocalRepository = true;
+			}
+		}
+	}
```
Geod24, line 343: Ah, that's why we have `dubHome`. And `userSettings`. And `localRepository`.
Honestly, just expose `dubHome`. YAGN `userSettings` / `localRepository` and it's not consistent with the env approach.

Geod24 (commented):
The env approach is sound, the settings code still makes me uneasy but makes much more sense now, made a few suggestions.

source/dub/dub.d
WebFreak001: we can't do the `if (auto ...` thing here because we need to check for length, otherwise someone running `env DPATH= dub` would wrongly get dub home set to `NativePath("") ~ "dub/"`

WebFreak001:
adjusted to review, can rebase if it's fine now

source/dub/dub.d
```diff
@@ -1828,6 +1878,9 @@ private struct UserConfiguration {
 	SetInfo!(string[string]) defaultPostBuildEnvironments;
 	SetInfo!(string[string]) defaultPreRunEnvironments;
 	SetInfo!(string[string]) defaultPostRunEnvironments;
+	SetInfo!(string) dubHome;
+	SetInfo!(string) userSettings;
+	SetInfo!(string) localRepository;
```
Geod24, line 1883: So those can be removed ?

source/dub/dub.d
```diff
@@ -213,17 +220,60 @@ class Dub {
 
 		const dubFolderPath = NativePath(thisExePath).parentPath;
 
+		// override default userSettings + localRepository if a $DPATH or
+		// $DUB_HOME environment variable is set.
+		bool overrideUserSettings;
```
Geod24, line 225: This only needs to be one boolean now, right ?

source/dub/dub.d
```diff
@@ -213,17 +220,60 @@ class Dub {
...
+		// Override user + local package path from system / binary settings
+		// Then continues loading local settings from these folders. (keeping
+		// global /etc/dub/settings.json settings intact)
+		//
+		// Don't use it if either $DPATH or $DUB_HOME are set, as environment
+		// variables usually take precedence over configuration.
+		if (!overrideUserSettings && this.m_config.dubHome.set) {
+			m_dirs.userSettings = NativePath(this.m_config.dubHome.expandEnvironmentVariables);
+
+			overrideUserSettings = true;
```
Geod24, line 260: No need to set it, right ?

source/dub/dub.d
```diff
@@ -213,17 +220,60 @@ class Dub {
...
 		if (!root_path.empty)
 			readSettingsFile(root_path ~ "dub.settings.json");
 
-		determineDefaultCompiler();
+		// same as overrideUserSettings above, but taking into account the
+		// config loaded from user settings and per-package config as well.
+		if (!overrideLocalRepository && this.m_config.dubHome.set) {
+			m_dirs.localRepository = NativePath(this.m_config.dubHome.expandEnvironmentVariables);
+
+			overrideLocalRepository = true;
```
Geod24, line 275: Ditto superfluous assign

changelog/dpath.dd
```diff
@@ -0,0 +1,41 @@
...
+In the settings.json you can set the following fields:
+
+```json
+{
+    "dubHome": "/path/to/dub", // sets both package store and config location
+
+    // OR
+
+    "userSettings": "/path/to/dub/settings/folder",
+    "localRepository": "/path/to/dub/repository/folder"
```
Geod24, line 28: Changelog needs an update

WebFreak001:
fixed left-overs

Geod24 (approved):

WebFreak001:
thanks!

PR #2283, opened 2022-06-27
Allow "dub :subpackage" shortcut syntax to run a certain sub package

s-ludwig:
Currently, an explicit `dub run :sub` is required to perform the equivalent of a simple `dub` for a sub package. Since commands never have a colon in their name, this can be safely used to distinguish sub package names from commands, enabling just `dub :sub` as a short syntax.

### DIFF
--- a/changelog/subpackage_short_syntax.dd
+++ b/changelog/subpackage_short_syntax.dd
@@ -1,0 +1,4 @@
+The shortcut syntax for "dub run" is now also available for sub packages.
+
+Invoking dub as "dub :subpackage" is now equivalent to "dub run :subpackage",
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):
Changelog entry?

s-ludwig:
Added

thewilsonator:
LGTM, I assume this will end up in the docs

s-ludwig:
I had a look at the existing documentation and it's kind of already implied by the first sentence of the command line docs: https://dub.pm/commandline.html

So I'd say that the previous behavior could be considered more of an oversight/bug.

Geod24 (approved):

Issue #1245, opened 2017-09-06
Don't create dub.selections.json for library and test targets by default


s-ludwig:
Since libraries are usually built as dependencies of other (executable )packages, dub.selections.json usually doesn't serve a useful purpose. There may be exceptions when the generated library is intended to be distributed or is used as a plugin (i.e. is itself meant as the final output of the build process), so there should be a command line switch to force generating a selections file.

Whenever a selections file already exists, it should however be used and updated as usual. Apart from library targets, a selections file should also not be created for test runs (`dub test`).

See #1153 for the previous discussion.

ljmf00:
I have very strong opinions about this and we should discuss this carefully.

I don't think this should be implemented at all. Let's please not follow other
package managers just because they do the same.

Lockfiles serves a clear purpose, reproduce an environment. Encouraging someone
to not use them, in libraries, leads to a lot of issues being created to
intentionally remove lock files, making people confused about the purpose of it.
On libraries, we should both test in a locked and a suitable different
environment (more commonly the latest, but supposedly any supported version,
according to the specified version range).

See my and other cited articles [here](https://lsferreira.net/posts/lockfile-lib-misconception/).

Geod24:
From my experience, some experienced users really care about having `dub.selections.json` checked in, so that tests don't randomly break, so I think for the time being we should hold off on that.

PR #2284, opened 2022-06-27
Don't generate a dub.selections.json file for "dub test"

s-ludwig:
See #1245 - @ljmf00, I think for for library targets writing selections is a debatable choice, but for "dub test" I can't really find any pro argument, while writing selections is very frequently annoying in practice.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -956,6 +956,7 @@ abstract class PackageBuildCommand : Command {
 		bool m_nodeps;
 		bool m_forceRemove = false;
 		bool m_single;
+		bool m_test = false;
 		bool m_filterVersions = false;
 	}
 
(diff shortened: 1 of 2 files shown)


WebFreak001:
is it intended that now `dub test` and `dub --config=unittest` behave differently again?

s-ludwig:
Hm, do you have a reference for that?

WebFreak001:
since https://github.com/dlang/dub/pull/2218 the --config=unittest --build=unittest invocation is intended to work the same as `dub test`, so the code is using the same test code now. So I would assume `dub --config=unittest --build=unittest` would still generate a dub.selections.json now.

PetarKirov:
@s-ludwig I completely agree with @ljmf00 (https://lsferreira.net/posts/lockfile-lib-misconception/).
There is no good reason to avoid lock files in any case, especially when testing, where you most definitely want reproducibility.
Anyone who really wants to avoid reproducibility can simply run `dub <command>
 && rm -f dub.selections.json` or create a wrapper script around `dub` that does the same.

Btw, we really should add sha256 hashes for each dependency in dub.selections.json.

WebFreak001:
If we go with a hashes addition I would be for using [sha3 or shake256](https://code.dlang.org/packages/sha3-d) to add hashes to dub for it being advertised as more collision resistant and secure than sha2, so we can keep the implementation the same for a while. Not sure if that's a property we need though, on wikipedia sha3 is claimed to be slower or on par with sha2. Unsure what the actual performance difference is between stdlib and the sha3 package.

PR #2285, opened 2022-06-28
Add a command line option for upgrading all sub packages.

s-ludwig:
This mitigates the issue of `dub test` (or even `dub build`) generating dub.selection.json for library projects by going in the opposite direction, facilitating the use of selection files for all sub packages. Added as an optional command line flag for the sake of backwards compatibility.

See also #1245 and #2284.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1758,6 +1758,7 @@ class AddCommand : Command {
 class UpgradeCommand : Command {
 	private {
 		bool m_prerelease = false;
+		bool m_includeSubPackages = false;
 		bool m_forceRemove = false;
 		bool m_missingOnly = false;
 		bool m_verify = false;
@@ -1782,6 +1783,9 @@ class UpgradeCommand : Command {
 	{
 		args.getopt("prerelease", &m_prerelease, [
 			"Uses the latest pre-release version, even if release versions are available"
+		]);
+		args.getopt("s|sub-packages", &m_includeSubPackages, [
+			"Also upgrades dependencies of all directory based sub packages"
 		]);
 		args.getopt("verify", &m_verify, [
 			"Updates the project and performs a build. If successful, rewrites the selected versions file <to be implemented>."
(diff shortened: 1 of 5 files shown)


s-ludwig:
I think that the test failures are unrelated to the changes here. @Geod24, is the "pipefail" in 3e5a5ba18efa73a87a4e88d422de75507056f115 required?

source/dub/commandline.d
```diff
@@ -1809,6 +1813,31 @@ class UpgradeCommand : Command {
...
+
+			auto spacks = dub.project.rootPackage
+				.subPackages
+				.filter!(sp => sp.path.length);
+
+			foreach (sp; spacks) {
+				try {
+					auto fullpath = (dub.projectPath ~ sp.path).toNativeString();
+					logInfo("Upgrading sub package in %s", fullpath);
+					auto sdub = new Dub(fullpath, dub.packageSuppliers, SkipPackageSuppliers.all);
```
WebFreak001, line 1828: why do we need to construct an extra dub instance? If it's not possible to easily make it work with the regular upgrade function a comment about this would be appreciated here for future readers.

source/dub/commandline.d
```diff
@@ -1783,6 +1784,9 @@ class UpgradeCommand : Command {
 		args.getopt("prerelease", &m_prerelease, [
 			"Uses the latest pre-release version, even if release versions are available"
 		]);
+		args.getopt("s|sub-packages", &m_includeSubPackages, [
```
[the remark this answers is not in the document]
s-ludwig: Anyone who has sub packages is expected to use this frequently. I don't think it matters much whether 1% or 50% of the users are using sub packages, if you do, you'll have to frequently type this out.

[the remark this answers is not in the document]
WebFreak001: oh sorry, I just realized this is the new behavior, I thought this was going to be the default behavior and the switch was to turn it off again.

WebFreak001 (commented):
do you think it might be possible to enable this flag in the future by default? I often had problems with subpackages being out-of-date, causing weird build/test glitches that were fixed by simply running `dub upgrade` in the subpackage folders.

I can see the point that this would break backwards compatibility, but I can't really think of a real case where you wouldn't want to upgrade your subpackages through `dub upgrade`.

s-ludwig:
I'm really not sure whether making this the default would actually interfere with anyone's work, except if there are people who want to explicitly avoid creating dub.selection.json for sub packages and never `dub test`ed them before. Maybe we could make a poll somewhere? I'd certainly *like* to have it as the default.

But you could also argue that the CLI is also part of the SemVer scheme, in which case this would be a change for 2.0.0.

rikkimax:
> should we add warnings for when using `dub upgrade` and subpackages aren't upgraded (that would be upgraded with `-s` though)?

Yes, please!

Helpful hints that tell you something isn't right and how to rectify it are always good.

test/timeout.sh
```diff
@@ -31,7 +31,7 @@ Content-Length: 2\r
 \r
 {}')
     for i in $(seq 0 $((${#res} - 1))); do
-        echo -n "${res:$i:1}"
+        echo -n "${res:$i:1}" || true
```
WebFreak001, line 34: how exactly does _echo_ fail? output buffer full?

WebFreak001: ah right, makes sense

WebFreak001 (approved):

WebFreak001 (approved):
changelog entry? Also would be nice to put the same content onto https://github.com/dlang/dub-docs

s-ludwig:
> changelog entry? Also would be nice to put the same content onto https://github.com/dlang/dub-docs

Why put it there redundantly? It's already part of the command line documentation.

WebFreak001 (approved):

thewilsonator (approved):

PR #2286, opened 2022-06-28
Add support for unit-threaded in auto-generated unittest main

kinke:
The existing auto-generated main module for unittest runners without explicit `mainSourceFile` already imports all source modules and exposes them as an `AliasSeq` - for the `tested` framework, which has been updated the last time in 2013, has 0 downloads this month, and 2,140 total downloads according to `code.dlang.org`.

`unit-threaded` had some ~2,700 downloads this month alone, and about 600k downloads in total. So reuse the existing infrastructure to generate a `main()` that forwards all modules (of the root project being tested) to `unit-threaded`.

### DIFF
--- a/changelog/unit_threaded-main.dd
+++ b/changelog/unit_threaded-main.dd
@@ -1,0 +1,17 @@
+Auto-generated main module for unittest configurations now supports unit-threaded
+
+If there's a dependency on `unit-threaded`, the auto-generated main module now
+forwards all modules (of the root project being tested) to its `runTests()`.
+
+So the following `dub.sdl` skeleton now suffices to run all tests of the project,
+without having to maintain a list of modules or resort to other auto-gen solutions:
+```
+name "project"
+
+configuration "default" {
+}
+
+configuration "unittest" {
+    dependency "unit-threaded" version="~>2.0"
+}
+```
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -345,17 +345,21 @@ class Project {
 				fil.write(format("import %s;\n", custommodname));
 			} else {
 				fil.write(q{
-import core.runtime;
-
-void main() {
-	version (D_Coverage) {
+int main(string[] args) {
+	version (Have_unit_threaded) {
+		import unit_threaded : runTests;
+		return args.runTests!allModules;
+	} else version (D_Coverage) {
+		return 0;
 	} else {
 		import std.stdio : writeln;
 		writeln("All unit tests have been run successfully.");
+		return 0;
 	}
 }
-shared static this() {
-	version (Have_tested) {
+
+version (Have_tested) {
+	shared static this() {
 		import tested;
 		import core.runtime;
 		import std.exception;
--- a/test/unit_threaded-main.sh
+++ b/test/unit_threaded-main.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+. $(dirname "${BASH_SOURCE[0]}")/common.sh
+DIR=$(dirname "${BASH_SOURCE[0]}")
+
+cd "$DIR/unit_threaded-main"
+
+# build and run the tests
+$DUB test
+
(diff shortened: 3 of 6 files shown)


WebFreak001:
I think it's certainly better to have unit-threaded here, however I'm not a fan of this change to hardcode unit-threaded.

We already have `dub_test_root`, can't unit-threaded make use of that?

rikkimax:
> We already have `dub_test_root`, can't unit-threaded make use of that?

It is doing that in the hardcoded support.

The solution I came up with to account for shared libraries was to use injectSourceFiles to inject registration but that requires the companion feature which handles the list of modules akin to dub_test_root and some way to override the test runner using directives not cli.

This would be a lot more flexible solution.

rikkimax:
> do the Runtime unittest runner methods not work?
> 
> https://gitlab.com/AntonMeep/silly/-/blob/master/silly.d#L29

Yes and no.

It works fine if you have druntime and don't use UDA's.

So far I haven't needed the companion feature to injectSourceFiles (hence why I haven't implemented it), but I think this is a good example of when we should talk about it as an alternative.

Basically, the approach I described is for the registration of symbols in a given binary using CT information. This is why it could work for test runners, ORM's, GC hooking ext.

s-ludwig:
Why not simply generate a separate module containing just that alias sequence and make it available for import by the user specified custom test main file? I'd rather just remove the "tested" support than introduce more or different special cases here.

rikkimax:
That is the companion feature I've been talking about more or less.

kinke:
> Why not simply generate a separate module containing just that alias sequence and make it available for import by the user specified custom test main file?

Simply? :D - This ~3 lines diff is the simplest thing I could think of, and would get rid of basically all test main files at Symmetry. The only problem I see is for projects depending on both unit-threaded *and* silly.

WebFreak001:
I would be fine with this for now to help big projects using unit-threaded right now to more easily use it (both for old and new versions)

I would really prefer if we can make unit-threaded make use of dub_test_root instead though, even if it needs updates by the users then.

With this change also you are calling `runTests!allModules`, but allModules is aliases to the imported packages, does unit-threaded even allow that? (I always saw it use module names as strings in all the examples)

WebFreak001:
(sorry did not mean to close, ctrl-enter did that)

s-ludwig:
Yeah, I meant "simply" just as in still pretty straight forward, although admittedly the diff would be a few lines more. But the goal should be to let any test runner get access to the list. The generic test main would then of course be provided by unit-threaded itself, instead of having one in each repository.

I don't even use `tested` myself, because I never got around to integrating its metrics capabilities with our CI infrastructure, so I think that we can safely drop that instead of supporting it alongside a more general solution. A unit-threaded specific approach on the other hand would probably result in some breakage if we'd ever decide to remove support later down the line.

kinke:
> I would really prefer if we can make unit-threaded make use of dub_test_root instead though, even if it needs updates by the users then.

I have a prototype for a new `unit-threaded` subpackage, importing `dub_test_root` like silly. It works - but only with all-at-once compilation, `--build-mode=singleFile` fails because the module path is something like `dub/code/unit_threaded-main-test-unittest-unittest-linux.posix-x86_64-dmd_v2.100.0-2F30A7DA2FB5F7C6116EA05A469F1735_dub_test_root.d`.

And it wouldn't work with reggae's default per-package compilation either - but reggae apparently doesn't support the automated `dub_test_root` generation at all anyway.

> With this change also you are calling runTests!allModules, but allModules is aliases to the imported packages, does unit-threaded even allow that? (I always saw it use module names as strings in all the examples)

That's obviously covered by the test; yes, it supports both string and symbol variants.

kinke:
Okay, proposed alternative via explicit `unit-threaded:autorunner` subpackage dependency working like `silly`: https://github.com/atilaneves/unit-threaded/pull/270

PR #2287, opened 2022-06-29
Revert determineVersionWithGitTool to the behavior before #2263

s-ludwig:
See https://github.com/dlang/dub/pull/2263#issuecomment-1170129677

### DIFF
--- a/source/dub/internal/git.d
+++ b/source/dub/internal/git.d
@@ -103,10 +103,7 @@ private string determineVersionFromGitDescribe(string describeOutput)
 	if (tag.startsWith("v") && isValidVersion(tag[1 .. $])) {
 		if (num == 0) return tag[1 .. $];
 		const i = tag.indexOf('+');
-		auto r = format("%s%scommit.%s.%s", tag[1 .. (i < 0) ? $ : i],
-			parts.length > 3 ? "." : "-", num, commit);
-		if (i > 0) r ~= tag[i .. $];
-		return r;
+		return format("%s%scommit.%s.%s", tag[1 .. $], i >= 0 ? '.' : '+', num, commit);
 	}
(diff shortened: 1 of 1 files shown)


WebFreak001:
you posted a sample for an invalid semver version, however as Mathias wrote, commits always start with a g, so is it even possible to ever have such an invalid version?

s-ludwig:
The main issue is that the "+" suffix is defined by the SemVer spec to not have any effect in version comparisons and acts just like a comment. Changing the "commit" suffix to a kind of pre-release version is IMO just a hack and won't help at all for projects that use pre-release versions a lot. I guess it wouldn't hurt to change the *cache logic* to *prefer* an exact match anyway (possibly emitting a warning if an exact match isn't possible), which should solve the issue AFAICS.

However, where this property of SemVer is actually very useful is the scenario where you work on a local fork of some library - lets say you have the following structure:

```
~/dev/lib1     (~master, v1.0.0)
~/dev/lib2     (~feature1, v1.2.0+commit.2.g12312312)
~/dev/app      (selects lib1@1.0.0 and lib2@1.2.0)
```

If you now have set up `~/dev` to be scanned for packages with `dub add-path ~/dev`, packages in `~/dev` will have precedence over cached packages during lookup. This means that you can now freely change the code in the libraries, while building "app" unmodified to test the changes. When working on multiple dependencies, this is a huge time saver by reducing friction for dependency version management.

kinke:
Thx for the rationale. My current workflow in such cases is to make the app selections path-based (`{"path":"../<name>}"`). A bit more work but locally contained and, AFAICT, working.

Geod24:
Done in https://github.com/dlang/dub/pull/2293

PR #2288, opened 2022-06-29
Fix dub_test_root imports for non-all-at-once compilations

kinke:
Which requires a regular `dub_test_root.d` filename *and* adding the containing directory to the importPaths.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -329,11 +329,8 @@ class Project {
 		tcinfo.sourceFiles[""] ~= escapedMainFile;
 		tcinfo.mainSourceFile = escapedMainFile;
 		if (!settings.tempBuild) {
-			// Add the directory containing dub_test_root.d to the import paths.
-			// The directory needs to exist, e.g., for `dub describe`, so don't add if it doesn't
-			// exist *and* we are not about to generate the module.
-			if (generate_main || existsFile(mainfile.parentPath))
-				tcinfo.importPaths[""] ~= NativePath(escapedMainFile).parentPath.toNativeString();
+			// add the directory containing dub_test_root.d to the import paths
+			tcinfo.importPaths[""] ~= NativePath(escapedMainFile).parentPath.toNativeString();
 		}
 
 		if (generate_main && (settings.force || !existsFile(mainfile))) {
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -322,12 +322,14 @@ class Project {
 			mainfile = getTempFile("dub_test_root", ".d");
 		else {
 			import dub.generators.build : computeBuildName;
-			mainfile = rootPackage.path ~ format(".dub/code/%s_dub_test_root.d", computeBuildName(config, settings, import_modules));
+			mainfile = rootPackage.path ~ format(".dub/code/%s/dub_test_root.d", computeBuildName(config, settings, import_modules));
 		}
 
 		auto escapedMainFile = mainfile.toNativeString().replace("$", "$$");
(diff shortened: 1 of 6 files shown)


source/dub/project.d
```diff
@@ -319,12 +319,14 @@ class Project {
...
 			import dub.generators.build : computeBuildName;
-			mainfile = rootPackage.path ~ format(".dub/code/%s_dub_test_root.d", computeBuildName(config, settings, import_modules));
+			mainfile = rootPackage.path ~ format(".dub/code/%s/dub_test_root.d", computeBuildName(config, settings, import_modules));
 		}
 
 		auto escapedMainFile = mainfile.toNativeString().replace("$", "$$");
 		tcinfo.sourceFiles[""] ~= escapedMainFile;
 		tcinfo.mainSourceFile = escapedMainFile;
+		if (!settings.tempBuild)
+			tcinfo.importPaths[""] ~= NativePath(escapedMainFile).parentPath.toNativeString();
```
kinke, line 329: No idea whether the importPath actually needs to be escaped.

source/dub/project.d
```diff
@@ -319,12 +319,19 @@ class Project {
...
 		}
 
 		auto escapedMainFile = mainfile.toNativeString().replace("$", "$$");
 		tcinfo.sourceFiles[""] ~= escapedMainFile;
 		tcinfo.mainSourceFile = escapedMainFile;
+		if (!settings.tempBuild) {
+			// Add the directory containing dub_test_root.d to the import paths.
+			// The directory needs to exist, e.g., for `dub describe`, so don't add if it doesn't
+			// exist *and* we are not about to generate the module.
+			if (generate_main || existsFile(mainfile.parentPath))
```
kinke, line 332: Having to exclude it for some initial `dub describe` run is obviously ugly. The problem is that `generate_main` is false
* with `--annotate` ('dry run', where we definitely don't want to generate directories and `dub_test_root.d`), or
* for `dub describe` in general. And there's no `--annotate` for `dub describe`.

So maybe `dub describe` should get an `--annotate`.

WebFreak001: it looks like `generate_main` was previously used solely as dry run indicator:

https://github.com/dlang/dub/blob/609b9806c16b19c6f31ef2a50e9b59a933be3161/source/dub/dub.d#L666

so we probably don't even want to check on that boolean at all here and always add it as import path.

kinke: Nope, there's also https://github.com/dlang/dub/blob/609b9806c16b19c6f31ef2a50e9b59a933be3161/source/dub/commandline.d#L1637 for `dub describe`, and that's the problem.

WebFreak001: maybe add another parameter for dry run and handle the dry-run logic in this function, before writing to a file?

kinke: I missed that `--annotate` is a *general* option, so I'm using it now for a side-effect-free `dub describe -c unittest --annotate` (possibly warning about the non-existing import dir), while `dub describe -c unittest` makes sure the file is generated & up-to-date (for non-`executable` unittest configs).

But now hitting unexpected broken-pipe errors with `grep -q` and `head -n 1` etc...

kinke: Nope, I rebased this to get that 'fix', but it didn't affect the pipe errors.

I could understand them for `dub describe` output, where the extra importPath might push the output above some block threshold. But a regular `dub test` output should definitely not be affected (just a few lines), e.g., failing https://github.com/dlang/dub/blob/609b9806c16b19c6f31ef2a50e9b59a933be3161/test/version-spec.sh#L19.

PR #2297, opened 2022-07-13
Project: Remove dead m_packageSettings var and related func

Geod24:
git blame points to 2014, and it seems to be the ancestor of dub.selections.json


thewilsonator (approved):
(from a position of complete ignorance) does this deserve a changelog entry?

rikkimax:
Nah, looks to be entirely internal.

Geod24:
Only `setUpgradeCache` is a public function, but it's highly unlikely to be called given it's a no-op

PR #2301, opened 2022-07-13
Deprecate once-called redundant getLatestPackage

Geod24:
getBestPackage present a more generic interface,
as long as we feed it Dependency.any.


Geod24:
Context:
We have this:
https://github.com/dlang/dub/blob/09b35bcf212f84e9fbb65b247a25daa10471bd82/source/dub/dependency.d#L24-L32

And this:
https://github.com/dlang/dub/blob/09b35bcf212f84e9fbb65b247a25daa10471bd82/source/dub/commandline.d#L2818-L2822

I want to kill the latter, and use the former as much as possible, to reduce the use of string as version, and be able to report user error as soon as possible.

thewilsonator (approved):

PR #2303, opened 2022-07-14
Use dip1000 in dub.sdl too

Geod24:
We're already using it in the CI with the bootstrap script (build.d),
but users will be more likely to locally test using dub.


Geod24:
This needs a new diet-ng release, in progress...

Geod24:
Right, Vibe.d is not DIP1000 compatible...

PR #2305, opened 2022-07-14
Move some version-only constants and helpers in VersionRange

Geod24:
Context: https://github.com/dlang/dub/pull/2302


Geod24:
Had to add a third commit as it identified some missing attributes, should be G2G

thewilsonator (approved):

PR #2309, opened 2022-07-14
CI: Fix script for issue1505

Geod24:
The syntax was incorrect but the script was happily completing.


maxhaton (approved):

Geod24:
I rebased not too long ago, I think the test is actually broken... I'll look into it when I have a bit of time.

PR #2324, opened 2022-07-18
CI: Replace old GDC script with matrix job

Geod24:
Since COVERAGE is always false, this never ran.
With an up-to-date GDC now available, we can re-enable it.


Geod24:
Needs https://github.com/dlang-community/setup-dlang/pull/57 merged & released

Geod24:
So OpenSSL doesn't support `gdc`. This can be worked around with something like:
```diff
+// Used by the CI
+configuration "library-nonet-gdc" {
+       dependency "vibe-d:http" version=">=0.9.0 <0.10.0"
+    dependency "openssl" version="~>3.0"
+    subConfiguration "openssl" "library-manual-version"
+       targetType "library"
+       excludedSourceFiles "source/app.d"
+}
```

but then, I found out that `-Werror` turns on `-Werror=deprecated`, so essentially `-de`, which makes GDC a bit unusable (!).
CC @ibuclaw . Will see what we can do on `dub` side when I get a bit of free time.

PetarKirov (approved):
The changes to the CI scripts looks good to me, so feel free to merge when ready.

WebFreak001:
issue is here: https://github.com/dlang/dub/runs/7418736496?check_suite_focus=true#step:8:76

it tries to run `gdc -run` which doesn't exist, probably need to adjust a test here.

PR #2326, opened 2022-07-19
Fix missing sanitizing of subpackage targetTypes

kinke:
A regression introduced in #2277.

### DIFF
--- a/source/dub/recipe/io.d
+++ b/source/dub/recipe/io.d
@@ -77,13 +77,21 @@ PackageRecipe parsePackageRecipe(string contents, string filename, string parent
 	else assert(false, "readPackageRecipe called with filename with unknown extension: "~filename);
 
 	// Fix for issue #711: `targetType` should be inherited, or default to library
-	TargetType defaultTT = (ret.buildSettings.targetType == TargetType.autodetect) ?
-		TargetType.library : ret.buildSettings.targetType;
-	foreach (ref conf; ret.configurations)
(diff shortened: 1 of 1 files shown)


source/dub/recipe/io.d
```diff
@@ -132,6 +140,23 @@ unittest { // issue #711 - configuration default target type not correct for SDL
 	}
 }
 
+unittest { // make sure targetType of sub packages are sanitized too
+	import dub.compilers.buildsettings : TargetType;
+	auto inputs = [
+		"dub.sdl": "name \"test\"\nsubPackage {\nname \"sub\"\ntargetType \"sourceLibrary\"\nconfiguration \"a\" {\n}\n}",
+		"dub.json": "{\"name\": \"test\", \"subPackages\": [ { \"name\": \"sub\", \"targetType\": \"sourceLibrary\", \"configurations\": [{\"name\": \"a\"}] } ] }"
```
Geod24, line 147: You might want to use backtip-delimited strings here ?

kinke: I just copy&pasted it from the other unittests above. Backticks would be usable for the JSON, yes, but not for the sdl (in a single line), so for consistency...

kinke:
> Well, thinking about this more, this goes against what subpackages are meant to be

? - Each subpackage recipe is sanitized separately, not inheriting anything from the parent package if that's what you meant.

Geod24 (approved):
Thanks for tracking this. Failure looks spurious, retried the CI.

PR #2327, opened 2022-07-19
test/version-spec: Fix broken pipe issue

Geod24:
Short explanation: Unix doesn't always make great decisions
Long explanation: https://superuser.com/questions/554855/how-can-i-fix-a-broken-pipe-error


Geod24:
Finally the CI logs are somewhat readable

PR #2329, opened 2022-07-20
Replace `Dependency.merge(Dependency)` with `Dependency.matches(Dependency)`

kinke:


source/dub/dub.d
```diff
@@ -1668,8 +1668,7 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 
 	protected override bool matches(Dependency configs, Dependency config)
 	{
-		if (!configs.path.empty) return configs.path == config.path;
-		return configs.merge(config).valid;
+		return configs.matches(config);
```
kinke, line 1671: I haven't checked what `configs` and `config` actually are.

source/dub/dependency.d
```diff
@@ -462,13 +462,36 @@ struct Dependency {
...
+			(const Repository a,       any         ) => true,
+			(      any         , const Repository b) => false,
+
+			(const VersionRange a, const VersionRange b) {
+				if (a.matchesAny()) return b.isValid();
+				if (b.matchesAny()) return a.isValid();
+
+				VersionRange copy = a;
+				copy.merge(b);
+				return copy.isValid();
```
kinke, line 481: Checking whether an actual version *range* matches another range is probably something else (I assume `VersionRange.merge()` returns the intersection). So this might need special-casing or asserting that `a` isn't a *range*, but a single version.

PetarKirov:
@kinke @Geod24 What's the status of this PR?

Geod24:
Looks pretty good to me, was waiting until @kinke takes it out of draft, but the idea is sound.

kinke:
TODO:
- Decide whether to keep the old `.merge()` (for potential API users - reggae doesn't need it AFAICT)
- Decide if treating all paths as equivalent is good enough for now (we get a bunch of new warnings at Symmetry with dub master regarding `../pkg1` vs. `../../pkg1` etc., which would be eliminated by doing so)
- Decide if the changed priorities (path-based dominating repos) are a benign 'breaking' change
- Look into the unittest failure wrt. `1.0.0` *not* matching `>=1.0.0 <2.0.0`, though not really related to this PR
- Decide how to handle 2 actual VersionRanges, i.e., whether a non-empty intersection is good enough, or whether the lhs should be guaranteed to be a single version (the *selected* version). This requires looking into how `DependencyResolver.matches()` is used (WTF are `configs` and `config`).

PR #2330, opened 2022-07-20
Fail fast on Windows.

veelo:
Ref https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference


Geod24:
Makes sense, I've retrigered the build

PetarKirov (approved):

PR #2339, opened 2022-07-28
Do not call Project.reinit from commandline

Geod24:
reinit is called directly when Project is instantiated. It calls PackageManager.refresh,
which re-reads all packages stored in the package directory.
If we want to make dub smarter and faster, we need to stop scanning all the packages
on disk when we have a resolved dependency tree, and move to a lazy/on-demand approach.
The first step would be to remove this call. See also #2388.


Geod24:
I'm not 100% sure this won't break something, let's see what the CI says.

thewilsonator (approved):

PR #2340, opened 2022-07-28
Remove long-deprecated 'install' and 'uninstall' command

Geod24:
There were deprecated before dub v1.
I stopped going up the history after https://github.com/dlang/dub/commit/4d8f81b21e314ee9cf0833d5074f9efe68a2c8ea,
on 2013-11-13, where they were already deprecated.


thewilsonator (approved):
Does this warrant a changelog entry?

Geod24:
Since it's been deprecated for 10 years, I don't think anyone has kept it around, but might as well.

Issue #2338, opened 2022-07-27
A single `dub build` refreshes the local package cache 3 times


Geod24:
`dub build -v` inside of a previously build `dub` folder:

```
Using dub registry url 'https://code.dlang.org/'
Refreshing local packages (refresh existing: true)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/geod24/.dub/packages/local-packages.json
Looking for local package map at /home/geod24/projects/dlang/dub/.dub/packages/local-packages.json
Note: Failed to determine version of package msgpack-d at .. Assuming ~master.
Determined package version using GIT: zmqd 1.2.0+commit.8.gbf88663
Ignoring version specification (>=0.0.0) for path based dependency ./
Ignoring version specification (>=0.0.0) for path based dependency ./
Ignoring version specification (>=0.0.0) for path based dependency ./
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
Note: Failed to determine version of package msgpack-d at .. Assuming ~master.
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Ignoring version specification (>=0.0.0) for path based dependency .
Determined package version using GIT: windows-headers 1.0.4+commit.2.g4922dbd
Determined package version using GIT: windows-headers 1.0.4+commit.2.g4922dbd
Determined package version using GIT: jwtd 0.5.0
Determined package version using GIT: dub 1.29.1+commit.104.ge95f403e
Refreshing local packages (refresh existing: false)...
Looking for local package map at /var/lib/dub/packages/local-packages.json
Looking for local package map at /home/geod24/.dub/packages/local-packages.json
Looking for local package map at /home/geod24/projects/dlang/dub/.dub/packages/local-packages.json
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
Ignoring version specification (*) for path based dependency ../..
[...]

PR #2341, opened 2022-07-28
Move some package manager duties to `PackageManager`

Geod24:
Context: https://github.com/dlang/dub/issues/2338#issuecomment-1197960109
TL;DR: I want to change the way we store package. First step is to clean up the code.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -318,8 +318,12 @@ class PackageManager {
 		}
 
 		string gitReference = versionSpec.chompPrefix("~");
-		const destination = m_repositories[LocalPackageType.user].packagePath ~
-			NativePath(name ~ "-" ~ gitReference) ~ (name~"/");
+		NativePath destination = getPackagePath(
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

source/dub/packagemanager.d
```diff
@@ -305,8 +318,12 @@ class PackageManager {
 		}
 
 		string gitReference = versionSpec.chompPrefix("~");
-		const destination = m_repositories[LocalPackageType.user].packagePath ~
-			NativePath(name ~ "-" ~ gitReference) ~ (name~"/");
+		NativePath destination = getPackagePath(
+			m_repositories[LocalPackageType.user].packagePath,
+			name, versionSpec);
```
WebFreak001, line 323: destination changed here, is that right? (using versionSpec directly instead of gitReference here)

Geod24: `getPackagePath` also `chompPrefix`, so that didn't change.
However `getPackagePath` also replaces `+` with `_`, but git commit hash shouldn't have any `+`.

PR #2343, opened 2022-07-28
Use Configy to read user settings (`settings.json`) too

Geod24:
~Depends on https://github.com/dlang/dub/pull/2310~
First step towards https://github.com/dlang/dub/issues/1832

Only the last two commits belong to this PR.

Once #2310 and this are merged, adding settings will be massively simpler.
I  am thinking of holding back on #1832 so that we can come up with a "clean" format for the YAML settings, and it will make writing conversion code (from `settings.json` to `settings.yaml`) much simpler.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -126,23 +126,11 @@ class Dub {
 		PackageSupplier[] m_packageSuppliers;
 		NativePath m_rootPath;
 		SpecialDirs m_dirs;
-		DubConfig m_config;
+		UserConfiguration m_config;
 		NativePath m_projectPath;
 		Project m_project;
 		NativePath m_overrideSearchPath;
 		string m_defaultCompiler;
-		string m_defaultArchitecture;
-		bool m_defaultLowMemory;
-		string[string] m_defaultEnvironments;
-		string[string] m_defaultBuildEnvironments;
-		string[string] m_defaultRunEnvironments;
-		string[string] m_defaultPreGenerateEnvironments;
-		string[string] m_defaultPostGenerateEnvironments;
-		string[string] m_defaultPreBuildEnvironments;
-		string[string] m_defaultPostBuildEnvironments;
-		string[string] m_defaultPreRunEnvironments;
-		string[string] m_defaultPostRunEnvironments;
-
 	}
 
 	/** The default placement location of fetched packages.
@@ -223,7 +211,7 @@ class Dub {
 
 		if (skip_registry < SkipPackageSuppliers.configured)
 		{
-			ps ~= m_config.registryURLs
+			ps ~= m_config.registryUrls
 				.map!(url => getRegistryPackageSupplier(url))
 				.array;
 		}
@@ -245,13 +233,13 @@ class Dub {
 		scope (exit) environment.remove("DUB_REGISTRY");
 		auto dub = new Dub(".", null, SkipPackageSuppliers.none);
 
-		dub.m_config = new DubConfig(Json(["skipRegistry": Json("none")]), null);
+		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.none);
 		assert(dub.getPackageSuppliers(null).length == 1);
 
-		dub.m_config = new DubConfig(Json(["skipRegistry": Json("configured")]), null);
+		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.configured);
 		assert(dub.getPackageSuppliers(null).length == 0);
 
-		dub.m_config = new DubConfig(Json(["skipRegistry": Json("standard")]), null);
+		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.standard);
 		assert(dub.getPackageSuppliers(null).length == 0);
 
 		environment["DUB_REGISTRY"] = "http://example.com/";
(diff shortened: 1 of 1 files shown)


source/dub/dub.d
```diff
@@ -273,36 +277,51 @@ class Dub {
...
-		m_defaultPostRunEnvironments = m_config.defaultPostRunEnvironments;
+		m_defaultEnvironments = cast(string[string]) m_config.defaultEnvironments;
+		m_defaultBuildEnvironments = cast(string[string]) m_config.defaultBuildEnvironments;
+		m_defaultRunEnvironments = cast(string[string]) m_config.defaultRunEnvironments;
+		m_defaultPreGenerateEnvironments = cast(string[string]) m_config.defaultPreGenerateEnvironments;
+		m_defaultPostGenerateEnvironments = cast(string[string]) m_config.defaultPostGenerateEnvironments;
+		m_defaultPreBuildEnvironments = cast(string[string]) m_config.defaultPreBuildEnvironments;
+		m_defaultPostBuildEnvironments = cast(string[string]) m_config.defaultPostBuildEnvironments;
+		m_defaultPreRunEnvironments = cast(string[string]) m_config.defaultPreRunEnvironments;
+		m_defaultPostRunEnvironments = cast(string[string]) m_config.defaultPostRunEnvironments;
```
Geod24: This is a temporary hack removed in the next commit to make things compile. I probably will squash both commits.

Geod24:
Rebased, should be G2G, let's see what the CI says.

Geod24:
Hold on, there's one thing I forgot to tackle: Error handling.
This will error out on incorrectly formatted settings. I think that's a good thing, but we should add it to the changelog.
Additionally, the error message will not be properly (with colors) printed ATM, I'll fix this.

source/dub/dub.d
```diff
@@ -245,13 +233,13 @@ class Dub {
 		scope (exit) environment.remove("DUB_REGISTRY");
 		auto dub = new Dub(".", null, SkipPackageSuppliers.none);
 
-		dub.m_config = new DubConfig(Json(["skipRegistry": Json("none")]), null);
+		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.none);
```
WebFreak001, line 236: does SetInfo not override opAssign?

source/dub/dub.d
```diff
@@ -273,36 +261,44 @@ class Dub {
...
 	{
+		import configy.Read;
+
 		this.m_dirs = SpecialDirs.make();
 
-		m_config = new DubConfig(jsonFromFile(m_dirs.systemSettings ~ "settings.json", true), m_config);
+		const systemSettings = (m_dirs.systemSettings ~ "settings.json").toNativeString();
+		if (systemSettings.exists)
+			this.m_config = this.m_config.merge(
+				parseConfigFile!UserConfiguration(CLIArgs(systemSettings)));
```
WebFreak001, line 271: please add a helper function to call for each additional config check + load

source/dub/dub.d
```diff
@@ -1260,17 +1256,28 @@ class Dub {
...
-		if (m_defaultBuildEnvironments) settings.buildSettings.addBuildEnvironments(m_defaultBuildEnvironments);
-		if (m_defaultRunEnvironments) settings.buildSettings.addRunEnvironments(m_defaultRunEnvironments);
-		if (m_defaultPreGenerateEnvironments) settings.buildSettings.addPreGenerateEnvironments(m_defaultPreGenerateEnvironments);
-		if (m_defaultPostGenerateEnvironments) settings.buildSettings.addPostGenerateEnvironments(m_defaultPostGenerateEnvironments);
-		if (m_defaultPreBuildEnvironments) settings.buildSettings.addPreBuildEnvironments(m_defaultPreBuildEnvironments);
-		if (m_defaultPostBuildEnvironments) settings.buildSettings.addPostBuildEnvironments(m_defaultPostBuildEnvironments);
-		if (m_defaultPreRunEnvironments) settings.buildSettings.addPreRunEnvironments(m_defaultPreRunEnvironments);
-		if (m_defaultPostRunEnvironments) settings.buildSettings.addPostRunEnvironments(m_defaultPostRunEnvironments);
+		settings.platform = settings.compiler.determinePlatform(
+			settings.buildSettings, compiler_binary, this.m_config.defaultArchitecture);
```
WebFreak001, line 1260: all of these reads should probably use the public getters and not directly access m_config.

WebFreak001 (changes requested):

source/dub/dub.d
Geod24: Only the constructor is overriden. That's on purpose, because people usually don't change those values after they are read.

source/dub/dub.d
Geod24: Done

source/dub/dub.d
Geod24: Done

source/dub/dub.d
```diff
@@ -1260,17 +1243,28 @@ class Dub {
...
-		if (m_defaultRunEnvironments) settings.buildSettings.addRunEnvironments(m_defaultRunEnvironments);
-		if (m_defaultPreGenerateEnvironments) settings.buildSettings.addPreGenerateEnvironments(m_defaultPreGenerateEnvironments);
-		if (m_defaultPostGenerateEnvironments) settings.buildSettings.addPostGenerateEnvironments(m_defaultPostGenerateEnvironments);
-		if (m_defaultPreBuildEnvironments) settings.buildSettings.addPreBuildEnvironments(m_defaultPreBuildEnvironments);
-		if (m_defaultPostBuildEnvironments) settings.buildSettings.addPostBuildEnvironments(m_defaultPostBuildEnvironments);
-		if (m_defaultPreRunEnvironments) settings.buildSettings.addPreRunEnvironments(m_defaultPreRunEnvironments);
-		if (m_defaultPostRunEnvironments) settings.buildSettings.addPostRunEnvironments(m_defaultPostRunEnvironments);
+		settings.platform = settings.compiler.determinePlatform(
+			settings.buildSettings, compiler_binary, this.defaultArchitecture);
+		if (this.m_config.defaultLowMemory)
```
WebFreak001, line 1248: shouldn't we also make the `if`s use the public getters?

Geod24: Tunnel vision. Done.

WebFreak001 (commented):
LGTM, but feel free to first add the error handling you wanted to add

Geod24:
> LGTM, but feel free to first add the error handling you wanted to add

I actually tested it yesterday and the error handling LGTM. It's not colored but all the information is here.
Since coloring the error handling will need to depend on whether we have a TTY or not, and we now have the colored output, I think I'll integrate things in a single PR separately.

WebFreak001 (approved):

PR #2346, opened 2022-07-28
Commandline: Do not duplicate GeneratorSettings fields

Geod24:
Instead of adding fields individually to the Command, just store a GeneratorSettings,
and getopt it as needed.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -944,31 +944,24 @@ class InitCommand : Command {
 
 abstract class PackageBuildCommand : Command {
 	protected {
-		string m_buildType;
-		BuildMode m_buildMode;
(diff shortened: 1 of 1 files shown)


Geod24:
@WebFreak: Those fields are `protected` so *technically* part of the "public" API, but I hope we can agree that `dub.commandline` is not part of the `dub-as-library` API...

WebFreak001 (approved):

WebFreak001:
yeah I think the API changes are fine here

Geod24:
> Tests currently failing though.

Turns out I forgot to remove one duplicated field :grin:

PetarKirov (commented):
Nice!

PR #2349, opened 2022-07-29
Support escaped forward slashes in YAML parser

Geod24:
This provides better compatibility with JSON, and is what Vibe.d does by default:
https://github.com/vibe-d/vibe.d/blob/7604eea772b1a733f7bfd16591ddc5bd37850bd9/data/vibe/data/json.d#L2555-L2563

Will need to submit the fix to D-YAML with a proper test there.
https://github.com/dlang-community/D-YAML/issues/302

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1843,7 +1843,7 @@ unittest {
 	import configy.Read;
 
     const str1 = `{
-  "registryUrls": [ "http://foo.bar" ],
+  "registryUrls": [ "http://foo.bar\/optional\/escape" ],
   "customCachePaths": [ "foo/bar", "foo/foo" ],
 
   "skipRegistry": "all",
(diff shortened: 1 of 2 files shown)


source/dyaml/scanner.d
```diff
@@ -1407,6 +1407,12 @@ struct Scanner
                         char[2] escapeSequence = ['\\', cast(char)c];
                         reader_.sliceBuilder.write(escapeSequence);
                     }
+                    // YAML 1.2 supports for JSON optional escape of fwd slashes
+                    else if (c == '/')
+                    {
+                        reader_.sliceBuilder.write(c);
+                        reader_.forward();
```
WebFreak001, line 1414: what does forward do? in the backslash branch it's called before write

Geod24: It just advance the scanner to the next char. I can swap them if you'd like.

Geod24:
Let's wait for https://github.com/dlang-community/D-YAML/pull/303

Geod24:
Once the above PR is in, I will strip this PR to be simply about the fix. I removed D-YAML tests from the vendored files because it introduced complications when trying to support `build.d` and `dub build` at the same time. I think I'm starting to understand why :)

W.r.t. stripping the test I wrote in this PR, I think it's important that we keep tests separated, otherwise we'll end up with every layer testing the layers below. In this instance, we need to assume that D-YAML will parse JSON files correctly. If it doesn't, the fix and test needs to go to D-YAML.

WebFreak001:
I think the concrete test case for the settings json with escaped forward slashes is more like an integration test of the yaml component. I think it'd be a good idea to keep the json test that was in before in dub. That way we can also add more tests in the future with concrete issues happening to users.

Geod24:
Updated

WebFreak001 (approved):

Issue #2347, opened 2022-07-28
Improve CLI output


WebFreak001:
with the new colored output PR merged (#2268) a few things now look a little weird, so opening an issue to track them:

- [ ] `dub list` and `dub list-overrides` now has a lot of whitespace to the left
- [ ] `dub search` is not colored/bolded
- [ ] some single-line error messages look a little confusing with the extra whitespace, such as running `dub convert --format=sdl` twice
- [ ] `dub lint` dscanner output is not colored / does not follow dub output
- [ ] `dub dustmite` does not have any coloring

Geod24:
@WebFreak001 : Want to finish this before next release ?

PR #2350, opened 2022-07-30
improve some output coloring

WebFreak001:
See #2347

### DIFF
--- a/source/dub/internal/logging.d
+++ b/source/dub/internal/logging.d
@@ -76,8 +76,38 @@ public alias Color = fg;
 */
 public alias Mode = mode;
 
-/// The tag width in chars, defined as a constant here
-private const int TAG_WIDTH = 12;
+/// Defines the current width of logging tags for justifying in chars.
+/// Can be manipulated through push and pop.
+struct TagWidth {
+	import core.atomic;
+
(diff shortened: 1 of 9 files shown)


thewilsonator (approved):

source/dub/internal/logging.d
```diff
@@ -380,3 +380,19 @@ string color(const string str, const Mode m = Mode.init)
...
+	tagWidthStack[tagWidthIndex] = tagWidth;
+	tagWidthIndex++;
+	assert(tagWidthIndex < tagWidthStack.length);
+	tagWidth = w;
+}
+
+void popTagWidth()
+{
+	tagWidth = tagWidthStack[--tagWidthIndex];
+}
```
Geod24, line 398: Move it into its own struct please, instead of 3 globals

WebFreak001: done

Geod24:
I think the commits can be squashed BTW. The CI is erroring because of a minor difference in output, it seems

WebFreak001:
squashed commits, fixed CI (issue was output format change caused the grep to be an invalid / non-matching regex)

Geod24 (approved):

Issue #1064, opened 2017-02-04
option to tell which dependencies added a particular library dependency


timotheecour:
in a large program I'm trying to figure out which dependency produced `-L-lsasl2` which is in the final linking stage (that fails);

ideally it would show a dependency path:
`current=>dependency1=>dependency2=>sasl2`
but at least for starters showing `dependency2=>sasl2` would help.

in the meantime, how would i modify the code to show this

WebFreak001:
Since users are most likely interested in the dependency tree, that's what we will focus on the most. Showing linked in libraries / system libraries is a good idea to add into that tree as well though.

so this plus requirements that came from #2359:

- idea of PR is that `dub describe` shows full tree of dependencies
- e.g. `dub describe eventcore` lets me know that my app depends on vibe-d which depends on vibe-core which depends on eventcore.
- show linker / system dependencies for each package inside the tree, or allow the option to do this

PR #2359, opened 2022-08-03
`dub select`

WebFreak001:
initial implementation, still bad code quality, depends on #2350 

don't merge yet.


PetarKirov:
I'm not sure that `select` is right name, as it sounds like an action.

Examples from other package managers:

* `cargo tree`: https://doc.rust-lang.org/stable/cargo/commands/cargo-tree.html
* `yarn why`: https://yarnpkg.com/cli/why
* `yarn set resolution`: https://yarnpkg.com/cli/set/resolution (with `dub select` I would expect to change which versions are selected / resolved - e.g. to update `dub.selections.json`
* `npm explain`: https://docs.npmjs.com/cli/v7/commands/npm-explain
* `pipenv graph` https://pipenv.pypa.io/en/latest/cli/#pipenv-graph (I'm not sure about this one)

WebFreak001:
yes `dub select <pkg> <version>` can be used to select packages in dub.selections.json

I'm thinking that doing `dub select` without any arguments should also perform some useful action, so I was thinking showing the full dependency graph would be a good addition.

maxhaton:
I guess dub describe already exists so it could be a sub-whatever of that

source/dub/commandline.d
```diff
@@ -2117,13 +2127,124 @@ class ListCommand : Command {
...
 		logInfo("");
 		return 0;
 	}
 }
 
+class SelectCommand : PackageBuildCommand {
+	this() @safe pure nothrow
+	{
+		this.name = "select";
+		this.argumentsPattern = "(no args)|<package> <version|path>|<package> <repository> <commit>";
```
Geod24, line 2141:
```suggestion
		this.argumentsPattern = "[<package> <version|path>|<package> <repository> <commit>]";
```

source/dub/commandline.d
```diff
@@ -2117,13 +2127,124 @@ class ListCommand : Command {
...
+		this.description = "Management for dependency selections";
+		this.helpText = [
+			"Shows the full dependency tree if executed without arguments, highlighting "~
+			"dependencies that are conflicting or causing other errors."
```
Geod24, line 2145: It would be great to have a way to know where the dependency to `<package-name>` comes from.
So e.g. `dub select eventcore` lets me know that my app depends on `vibe-d` which depends on `vibe-core` which depends on `eventcore`.

source/dub/commandline.d
```diff
@@ -2117,13 +2127,124 @@ class ListCommand : Command {
...
+			// TODO: add subcommand to select specific versions of dependencies
+			// (write into dub.selections.json)
+		];
+	}
+	override int execute(Dub dub, string[] free_args, string[] app_args)
+	{
+		enforceUsage(free_args.length == 0
+			|| free_args.length == 2
+			|| free_args.length == 3, "Expecting zero or two or three extra arguments.");
```
Geod24, line 2154:
```suggestion
			|| free_args.length == 3, "Expecting either zero, two, or three arguments.");
```

WebFreak001:
maybe it's better to separate `dub select` and `dub explain` into two PRs.

maxhaton:
> maybe it's better to separate `dub select` and `dub explain` into two PRs.

"Everything should be made as simple as possible, but no simpler" - (Not) Einstein

Geod24:
Suggestion: `dub explain` => `dub why` to better reflect my state of mind every time I need this command (`/s`).

WebFreak001:
I moved the `dub explain` part out of here into a separate branch for now, as the select interface is quite a lot easier to add, show and test.

`dub upgrade -l` now shows this kind of output and does an implicit dry run: (can be made a non-dry run)

image

`dub select pkg ver` can be used to insert or change a dependency in dub.selections.json

`dub deselect pkg` can be used to remove a dependency from dub.selections.json

source/dub/commandline.d
WebFreak001: we can move this into requirements for `dub explain`, see #1064

WebFreak001:
test failure seems unrelated? (dmd-master only)

WebFreak001:
need an approve to merge @Geod24

changelog/selections.dd
```diff
@@ -0,0 +1,28 @@
...
+dub.selections.json.
+
+`dub deselect` can be used to remove a dependency from dub.selections.json.
+
+`dub upgrade --list` (or `-l`) now does a dry-run and lists all the existing
+packages + packages to upgrade in an easy-to-read interface. This can be
+combined with `--missing-only` (now available as `-m` as well) to show what is
+being done internally when running `dub build`. Both an upgrade as well as
+listing can be done by running `dub upgrade --list --dry-run=false` or
+`dub upgrade -l -dfalse`.
```
Geod24, line 15: Are you sure `-dfalse` works ? That would be odd - and not something we should recommend. Or perhaps you mean `-d false` ?

changelog/selections.dd
```diff
@@ -0,0 +1,28 @@
...
+`dub select` can be used to select or change a dependency inside
+dub.selections.json.
+
+`dub deselect` can be used to remove a dependency from dub.selections.json.
+
+`dub upgrade --list` (or `-l`) now does a dry-run and lists all the existing
+packages + packages to upgrade in an easy-to-read interface. This can be
+combined with `--missing-only` (now available as `-m` as well) to show what is
+being done internally when running `dub build`. Both an upgrade as well as
+listing can be done by running `dub upgrade --list --dry-run=false` or
```
Geod24, line 14: I'm a bit bothered by the fact we now have more switches to do the same thing, slightly differently.
We have `dub -v` and `dub upgrade -l` and `-l` implies dry run but we can also use `-d false`...
How about simplifying it and always having `dub upgrade -l` behavior, and just keeping dry-run for `dub upgrade -l` ?
Note that `dub upgrade -l` is not IMO very consistent CLI behavior: options (things starting with dash) shouldn't be used to implement subcommands, they should only slightly modify command behavior. So if you want to keep the `dub upgrade -l` behavior I would have gone with `dub upgrade list`. I'm not attached to the specifics, so happy to hash this out.

source/dub/commandline.d
```diff
@@ -1815,7 +1820,19 @@ class UpgradeCommand : Command {
 			"",
 			"This will update the versions stored in the selections file ("~SelectedVersions.defaultFile~") accordingly.",
 			"",
-			"If one or more package names are specified, only those dependencies will be upgraded. Otherwise all direct and indirect dependencies of the root package will get upgraded."
+			"If one or more package names are specified, only those dependencies will be upgraded. Otherwise all direct and indirect dependencies of the root package will get upgraded.",
+			"",
+			"List output format: (not script-safe)",
```
Geod24, line 1825: I wouldn't write "(not script-safe)". Except for `dub describe`, nothing is script safe.

source/dub/commandline.d
```diff
@@ -2183,6 +2275,118 @@ class ListCommand : Command {
...
+			return selectPackageScm(dub, free_args[0], free_args[1], free_args[2]);
+		else
+			throw new UsageException("Expecting either two or three free arguments.");
+	}
+
+	protected int selectPackage(Dub dub, string pack, string versionSpec)
+	{
+		bool assumeFile;
+		if (versionSpec.startsWith("~/")) {
+			versionSpec = versionSpec.expandTilde;
```
Geod24, line 2312: Good job Github...
image

source/dub/dependency.d
```diff
@@ -225,6 +225,28 @@ struct Dependency {
 		);
 	}
 
+	/** Returns a modified dependency that's relative to a given path, if the
+		path was absolute.
```
Geod24, line 229: I think we already have a function to do this no ?

source/dub/project.d
```diff
@@ -1761,38 +1771,33 @@ final class SelectedVersions {
 		m_dirty = true;
 	}
 
-	/// Selects a certain version for a specific package.
-	void selectVersion(string package_id, Version version_)
+	/// Selects the exact dependency for a specific package.
+	void selectVersion(string package_id, Dependency d)
```
Geod24, line 1775: I was trying to remove any use of `Dependency` from the API because it's almost always wrong. Why re-introduce it?

Geod24:
I'm a bit confused, I thought this was only for optional dependencies, originally, but there's no mention of it ?

Geod24:
@WebFreak001 : Do you want to pursue this ? It would be a useful addition.

WebFreak001:
I think this would need a rewrite since the underlying architecture changed quite a lot, which I'm currently not able to work on. I'd leave this open for anyone who wants to tackle it though

Issue #1681, opened 2019-04-22
Regression: Range violation


jacob-carlborg:
### System information

- Dub version: 1.15.0-beta.1+2-g3936dbf
- Platform: macOS
- DMD version: 2.085.1

### Bug Description

When executing Dub in a certain way using `spawnProcess` I get a Range violation error.

### How to reproduce?

1. Create a Dub project with the following content:

dub.json:

```json
{
    "name": "foo",
    "targetType": "executable",

    "configurations": [
        {
            "name": "unittest",
            "mainSourceFile": "test.d"
        },
    ],
}
```

test.d:

```d
void main()
{
    import std.process : spawnProcess, wait;
    spawnProcess(["/full/path/to/dub", "build", "--vverbose", ""]).wait();
}
```

2. Execute `dub test`

Replace `/full/path/to/dub` with the full path to the Dub executable. The issue here is the third empty argument. This is what's causing the range violation. I assume when Dub is normally executed from the shell no trailing empty argument will be passed through, even if trailing whitespace is added when executing the Dub command.

### Logs

Verbose output of Dub executed through `spawnProcess`:

```
Range violation
Full exception: core.exception.RangeError@source/dub/commandline.d(2225): Range violation
----------------
??:? _d_arrayboundsp [0x533717d]
~/.dvm/compilers/dmd-2.085.1/osx/bin/../../src/druntime/import/object.d:342 pure nothrow @safe dub.commandline.splitPackageName(immutable(char)[]).PackageAndVersion dub.commandline.splitPackageName(immutable(char)[]) [0x512bdf6]
~/.dvm/compilers/dmd-2.085.1/osx/bin/../../src/druntime/import/object.d:342 int dub.commandline.BuildCommand.execute(dub.dub.Dub, immutable(char)[][], immutable(char)[][]) [0x5123e1f]
~/.dvm/compilers/dmd-2.085.1/osx/bin/../../src/druntime/import/object.d:342 int dub.commandline.runDubCommandLine(immutable(char)[][]) [0x511fba2]
source/app.d:14 _Dmain [0x5113b6f]
```

John-Colvin:
Just running `dub build ""` from bash is enough to trigger this for me.

John-Colvin:
For people looking for a workaround in their bash scripts with optional arguments to dub, use the `${var:+"$var"}` syntax to truly expand to nothing if `var` is not defined.

jacob-carlborg:
[...]

PR #2360, opened 2022-08-04
Use PackageName for stronger typing and automatic coloring via PackageName.toString

nordlow:
Also standardizes variables naming as

- `package_id` to `package_name`
- `parent_name` to `parent_package_name`
- `pack` to `package_name`
- `package_` to `name`

This change adds no new features but instead acts as a strategic move to reduce logic bugs in the future.

It might be motivated to make sub-typing more detailed via, for instance, `QualifiedPackageName`, `SubPackageName`, etc.

Will also add `Config` sub-type either in this same PR or in a separate PR.

FYI, @Geod24.


source/dub/commandline.d
```diff
@@ -1943,15 +1943,15 @@ class RemoveCommand : FetchRemoveCommand {
...
+		auto package_name = PackageName(free_args[0]);
 		auto location = dub.defaultPlacementLocation;
 
 		size_t resolveVersion(in Package[] packages) {
 			// just remove only package version
 			if (packages.length == 1)
 				return 0;
 
-			writeln("Select version of '", package_id, "' to remove from location '", location, "':");
+			writeln("Select version of '", package_name, "' to remove from location '", location, "':");
```
WebFreak001, line 1954: will this print properly? I think it might add unneccessary `PackageName(...)` around it

source/dub/dependency.d
```diff
@@ -21,12 +21,16 @@ import std.exception;
 import std.string;
 import std.sumtype;
 
+struct PackageName {
+    string value;
+    alias value this;
```
nordlow, line 26: I feel dubious about whether we should use alias this here or not.

source/dub/dependency.d
```diff
@@ -21,12 +21,16 @@ import std.exception;
 import std.string;
 import std.sumtype;
 
+struct PackageName {
+    string value;
```
nordlow, line 25: Feel free to propose a better naming for this field.

source/dub/commandline.d
nordlow: Ahh, this is solved by adding a toString member, right?

source/dub/dependency.d
nordlow: Now that I've added toString member should we explicit use that and make value private and remove alias this?

source/dub/commandline.d
WebFreak001: yes that should work

source/dub/commandline.d
```diff
@@ -1279,13 +1279,13 @@ class BuildCommand : GenerateCommand {
...
 		} else {
-			if (packageParts.name.startsWith(":") ||
+			if (packageParts.name.value.startsWith(":") ||
 				dub.packageManager.getFirstPackage(packageParts.name))
 				// found locally
 				return 0;
 
 			// search for the package and filter versions for exact matches
-			auto basePackageName = getBasePackageName(packageParts.name);
+			auto basePackageName = getBasePackageName(PackageName(packageParts.name));
```
WebFreak001, line 1288: packageParts.name should already be of PackageName type

source/dub/commandline.d
```diff
@@ -2745,7 +2745,7 @@ private bool addDependency(Dub dub, ref PackageRecipe recipe, string depspec)
 			const ver = dub.getLatestVersion(depname);
 			dep = ver.isBranch ? Dependency(ver) : Dependency("~>" ~ ver.toString());
 		} catch (Exception e) {
-			logError("Could not find package '%s'.", depname);
+			logError("Could not find package '%s'.", depname.value);
```
WebFreak001, line 2748:
```suggestion
			logError("Could not find package '%s'.", depname);
```
overload toString instead

source/dub/dependency.d
```diff
@@ -21,12 +21,27 @@ import std.exception;
 import std.string;
 import std.sumtype;
 
+/** Encapsulates the name of a package.
+*/
+struct PackageName {
+    string value;
+	this(string value) @safe pure nothrow @nogc {
+        this.value = value;
```
WebFreak001, line 29: mixed tabs and spaces

source/dub/commandline.d
```diff
@@ -2736,7 +2736,7 @@ private bool addDependency(Dub dub, ref PackageRecipe recipe, string depspec)
 {
 	Dependency dep;
 	const parts = splitPackageName(depspec);
-	const depname = parts.name;
+	const depname = PackageName(parts.name); // TODO: const
```
WebFreak001, line 2739: `parts.name` should already be PackageName

WebFreak001:
for colored output you should add a member function on PackageName that uniformly formats the value with `.color(Mode.bold)`. I think this is also good for consistency so that package names are formatted the same everywhere

nordlow:
`dub test` now compiles but fails as

```
core.exception.AssertError@source/dub/dub.d(239): unittest failure
----------------
??:? _d_unittestp [0x55b0a7c73e89]
source/dub/dub.d:239 void dub.dub.Dub.__unittest_L231_C2() [0x55b0a7a482ce]
??:? void dub.dub.__modtest() [0x55b0a7a5af02]
??:? int core.runtime.runModuleUnitTests().__foreachbody6(object.ModuleInfo*) [0x55b0a7c980c2]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0x55b0a7c69deb]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_elf_shared.DSO) [0x55b0a7c9f037]
??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x55b0a7ca3c4f]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))) [0x55b0a7c9efc5]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)) [0x55b0a7c69dbd]
??:? runModuleUnitTests [0x55b0a7c97f0e]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x55b0a7c97648]
??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x55b0a7c975d5]
??:? _d_run_main2 [0x55b0a7c9753e]
??:? _d_run_main [0x55b0a7c972bf]
/usr/include/dmd/druntime/import/core/internal/entrypoint.d:29 main [0x55b0a77649d5]
??:? [0x7f63fe063d8f]
??:? __libc_start_main [0x7f63fe063e3f]
??:? _start [0x55b0a775f784]
    Starting Performing "debug" build using test for x86.
    Building test ~master: building configuration [library]
     Pre-gen Running commands for test
1/44 modules FAILED unittests
       Error Program exited with code 1

Compilation exited abnormally with code 2 at Thu Aug  4 13:20:17
```

First call to getPackageSuppliers in

```d
	unittest
	{
		scope (exit) environment.remove("DUB_REGISTRY");
		auto dub = new Dub(".", null, SkipPackageSuppliers.none);

		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.none);
		assert(dub.getPackageSuppliers(null).length == 1);
         }
```

returns 2. Interesting. Need to investigate. Assistance very much appreciated.

nordlow:
> for colored output you should add a member function on PackageName that uniformly formats the value with `.color(Mode.bold)`. I think this is also good for consistency so that package names are formatted the same everywhere

Very nice idea! Thanks!

Geod24:
So interestingly I started doing something similar recently:
https://github.com/dlang/dub/compare/master...Geod24:dub:pkg-name

The main thing I noticed is that the `PackageName` cannot really be used in `commandline`, as we haven't figured out the real name by that time. Likewise there is some magic in the SDL / JSON parser...

Geod24:
@nordlow : It's because dub picks up your supplier (the work registry). It shouldn't. Ignore the issue.

nordlow:
> @nordlow : It's because dub picks up your supplier (the work registry). It shouldn't. Ignore the issue.

What issue? Are you talking about https://github.com/dlang/dub/pull/2360#issuecomment-1205258644?

nordlow:
assert(e.msg ==
				"Unresolvable dependencies to package d:"
				~ "\n  b 1 depends on d [1]"
				~ "\n  c 1 depends on d [2]");

now fail because e.msg contains ANSI coloring. Shall I disable coloring in e.msg or insert coloring in expected string message?

Geod24:
> What issue? Are you talking about [#2360 (comment)](https://github.com/dlang/dub/pull/2360#issuecomment-1205258644)?

I'm talking about https://github.com/dlang/dub/pull/2360#issuecomment-1205123065

nordlow:
> > What issue? Are you talking about [#2360 (comment)](https://github.com/dlang/dub/pull/2360#issuecomment-1205258644)?
> 
> I'm talking about [#2360 (comment)](https://github.com/dlang/dub/pull/2360#issuecomment-1205123065)

Great. When I disable 

	unittest
	{
		scope (exit) environment.remove("DUB_REGISTRY");
		auto dub = new Dub(".", null, SkipPackageSuppliers.none);

		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.none);
		assert(dub.getPackageSuppliers(null).length == 1);

		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.configured);
		assert(dub.getPackageSuppliers(null).length == 0);

		dub.m_config.skipRegistry = typeof(dub.m_config.skipRegistry)(SkipPackageSuppliers.standard);
		assert(dub.getPackageSuppliers(null).length == 0);

		environment["DUB_REGISTRY"] = "http://example.com/";
		assert(dub.getPackageSuppliers(null).length == 1);
	}

dub test passes locally. Great.

source/dub/commandline.d
nordlow: Fixed via adding of `PackageName.toString`.

source/dub/dependency.d
nordlow: I'm keeping the alias this for now making the code smaller.

source/dub/commandline.d
nordlow: Resolved via b97f546a.

source/dub/commandline.d
nordlow: Resolved via 6b336e3d.

source/dub/dependency.d
nordlow: Resolved via aa38f4d4.

source/dub/commandline.d
nordlow: Resolved via 168a9831.

Geod24:
Ping me here and/or on Slack when ready for review.

nordlow:
Shall I rename all `package_id` with `package_name` for consistency?

nordlow:
> Ping me here and/or on Slack when ready for review.

Thanks! Things seem to be going green now. Ready for review!

source/dub/commandline.d
```diff
@@ -931,11 +931,11 @@ class InitCommand : Command {
 				// Tries getting the name until a valid one is given.
 				import std.regex;
 				auto nameRegex = regex(`^[a-z0-9\-_]+$`);
-				string triedName = input("Name", p.name);
+				string triedName = input("Name", p.name.toString);
```
nordlow, line 934: I'm unsure if there should be a call to `.toString` here or not, @Geod24.

Geod24:
> Btw, shall I rename remaining `package_id` variables to `package_name` for consistency?

I would just use `name`. Having a variable typed `PackageName` and named `package_name` is repetitively redundant :upside_down_face:

source/dub/dependency.d
```diff
@@ -21,12 +21,28 @@ import std.exception;
 import std.string;
 import std.sumtype;
 
+/** Encapsulates the name of a package.
+*/
+struct PackageName {
+	string value;
+	@disable this(PackageName name) @safe pure nothrow @nogc;
```
Geod24, line 28: Why ?

source/dub/dependency.d
```diff
@@ -21,12 +21,28 @@ import std.exception;
...
+	this(string value) @safe pure nothrow @nogc {
+		this.value = value;
+	}
+	static typeof(this) fromString(string value) @safe pure nothrow @nogc {
+		return typeof(return)(value);
+	}
+	string toString() @property @trusted const scope {
+		return value.color(Mode.bold);
+	}
```
Geod24, line 37: I think one should not be able to construct a `PackageName` with an invalid string. You might also want to implement `opAssign` and move the splitting logic (and whether or not the name is that of a subpackage) here.
See the linked commit as an example.

nordlow:
> > Btw, shall I rename remaining `package_id` variables to `package_name` for consistency?
> 
> I would just use `name`. Having a variable typed `PackageName` and named `package_name` is repetitively redundant upside_down_face

Done at 119f1237.

source/dub/dub.d
```diff
@@ -730,9 +730,9 @@ class Dub {
 	}
 
 	/// Fetches the package matching the dependency and places it in the specified location.
-	Package fetch(string packageId, const Dependency dep, PlacementLocation location, FetchOptions options, string reason = "")
+	Package fetch(PackageName package_id, const Dependency dep, PlacementLocation location, FetchOptions options, string reason = "")
```
Geod24: I think most changes in this file are breaking the public interface, which means they break reggae.

source/dyaml/node.d
```diff
@@ -610,8 +610,15 @@ struct Node
...
+                else static if(is(typeof(T.fromString(string.init)) == T))
+                {
+                    enforce(type == NodeType.string, new NodeException(
+                                "Node stores unexpected type: " ~ text(type) ~
+                                ". Expected: " ~ typeid(string).toString(), startMark_));
+                    return T.fromString(getValue!string);
+                }
                 else throw new NodeException("Node stores unexpected type: " ~ text(type) ~
-                    ". Expected: " ~ typeid(T).toString, startMark_);
+                                             ". Expected: " ~ typeid(T).toString, startMark_);
```
Geod24, line 621: ?

source/dyaml/node.d
```diff
@@ -610,8 +610,15 @@ struct Node
                             " out of range. Value: " ~ to!string(temp), startMark_));
                     return temp.to!T;
                 }
+                else static if(is(typeof(T.fromString(string.init)) == T))
+                {
+                    enforce(type == NodeType.string, new NodeException(
+                                "Node stores unexpected type: " ~ text(type) ~
+                                ". Expected: " ~ typeid(string).toString(), startMark_));
+                    return T.fromString(getValue!string);
+                }
```
Geod24, line 619: We shouldn't have to modify D-YAML locally.

source/dub/recipe/sdl.d
```diff
@@ -505,9 +505,9 @@ lflags "lf1" "lf2"
 lflags "lf3"
 `;
 	PackageRecipe rec1;
-	parseSDL(rec1, sdl, null, "testfile");
+	parseSDL(rec1, sdl, PackageName(null), "testfile");
```
Geod24, line 508: Use `PackageName.init`

source/dub/recipe/packagerecipe.d
```diff
@@ -175,15 +175,15 @@ struct ConfigurationInfo {
...
+	Dependency[PackageName] dependencies;
 	BuildSettingsTemplate[string] dependencyBuildSettings;
 	string systemDependencies;
 	TargetType targetType = TargetType.autodetect;
 	string targetPath;
 	string targetName;
 	string workingDirectory;
 	string mainSourceFile;
-	string[string] subConfigurations;
+	string[PackageName] subConfigurations; // TODO: Config[PackageName]
```
Geod24, line 186: ?

source/dub/recipe/json.d
```diff
@@ -165,10 +165,11 @@ private void parseJson(ref BuildSettingsTemplate bs, Json json, string package_n
 		switch(basename){
 			default: break;
 			case "dependencies":
-				foreach (string pkg, verspec; value) {
+				foreach (string pkgString, verspec; value) {
+                    auto pkg = PackageName(pkgString);
```
Geod24, line 169: Indent

source/dub/recipe/io.d
```diff
@@ -17,17 +18,17 @@ import dub.internal.vibecompat.inet.path;
...
 	Params:
 		filename = NativePath of the package recipe file
-		parent_name = Optional name of the parent package (if this is a sub package)
+		parent_package_name = Optional name of the parent package (if this is a sub package)
 
 	Returns: Returns the package recipe contents
 	Throws: Throws an exception if an I/O or syntax error occurs
 */
-PackageRecipe readPackageRecipe(string filename, string parent_name = null)
+PackageRecipe readPackageRecipe(string filename, PackageName parent_package_name = null)
```
Geod24, line 26: E.g. here `PackageName parent` is much more readable

source/dub/project.d
```diff
@@ -573,7 +573,7 @@ shared static this() {
 	}
 
 	/// Returns the name of the root package.
-	@property string name() const { return m_rootPackage ? m_rootPackage.name : "app"; }
+	@property PackageName name() const { return m_rootPackage ? m_rootPackage.name : PackageName("app"); }
```
Geod24, line 576: Breaks the API

Geod24 (changes requested):
The changes are needed, but we can't just change all the public API like this.
Additionally I think we should be much stricter about what is accepted in a `PackageName`.

Geod24:
@nordlow : I don't think `PackageName` belongs to dependency, but at the same time, I reckon it's the most logical place to put it.
I was hoping to gradually move all the light public data types to `dub.recipe` (and probably rename it once done).

As for implementation, here's what I had:
```D
/**
 * Holds a package name, including its possible parent(s)
 */
public struct PackageName {
    import std.ascii, std.format;

	private string[] pkgs;

	/// Instantiate and validate a name
	public this (string value) @safe pure {
		assert(value.length);

		foreach (size_t idx, char c; value)
			enforce(c.isLower || c.isDigit || c == '-' || c == '_' || c == ':',
					format("Character '%c' at index %s is not allowed in package name (full name: '%s')",
						   idx, c, value));

		this.pkgs = value.split(':').array;
		assert(this.pkgs.length >= 1);

		version (none) {
            // Disabled because it's used in the package recipe
            // Allowing it means that a `PackageName` needs a context
            enforce(this.pkgs[0].length,
                    format("Package name '%s' is missing parent package name", value));
        }
		foreach (val; this.pkgs[1 .. $])
			enforce(val.length, format("Package name '%s' should not contain empty subpackage(s)", value));
	}

	/// Writes a string representation to the sink
	public void toString (scope void delegate (in char[]) @safe sink)
		const scope @safe
	{
		sink(this.pkgs[0]);
		foreach (part; this.pkgs[1 .. $]) {
			sink(":");
			sink(part);
		}
	}

	/// Returns: A string representation
	public string toString () const scope @safe
	{
		string result;
		this.toString((in char[] data) { result ~= data; });
		return result;
	}
}
```

It's missing the color fixes which are useful. I was looking into moving some of the subpackage logic there, but it's not an easy job.

source/dub/dependency.d
nordlow: I used it detect unneccesary wrappings `PackageName(PackageName)` because of the alias this. Maybe I should remove the alias instead and add the necessary member (property) member function so PackageName to mimic a string. What do you say?

source/dub/dependency.d
nordlow: I don't understand the part about the opAssign. Can you elaborate on why we need `PackageName.opAssign`?

source/dub/dub.d
nordlow: Shall I just revert all the parameters in this package from `PackageName` to `string`?

nordlow: Resolved at d2c30ec9.

source/dyaml/node.d
nordlow: Resolved via d8f29134.

source/dyaml/node.d
nordlow: Are you saying we should upstream this change before we merged this?

source/dub/recipe/sdl.d
nordlow: Resolved via 7bfa49f3.

source/dub/recipe/packagerecipe.d
nordlow: Resolved via 5eae24ab.

source/dub/dependency.d
Geod24: Yes. Otherwise we bypass the check as we have `alias this`.

source/dub/dependency.d
Geod24: Removing the `alias this` might be good, but not sure how much the code depends on strings at the moment ?

source/dub/dub.d
Geod24: `private` is okay. For `public` function, you can just change them, but add a wrapper function that is deprecated.
In this example:
```D
deprecated("Use the overload that accepts a `PackageName` instead of a string as first argument")
Package fetch(PackageName id, const Dependency dep, PlacementLocation location, FetchOptions options, string reason = "") {
    return this.fetch(PackageName(id), dep, location, options, reason);
}

/// Ditto
Package fetch(PackageName id, const Dependency dep, PlacementLocation location, FetchOptions options, string reason = "") {
    // Real code
}
```

source/dyaml/node.d
Geod24: Yes, but also, this change shouldn't be necessary, Configy is the one that recognize `fromString` / `this(string)`. What error did you see ?

source/dub/commandline.d
nordlow: Resolving this.

nordlow:
Should we use ANSI coloring in exception strings, @Geod24?

Geod24:
@nordlow : I don't think so. The approach I took with configy, because I didn't want to assume the output was going to a TTY, was to expose a different format string: https://github.com/Geod24/configy/blob/b48b4635a5b654af18dad967e94be3f4b73d6e52/source/configy/Exceptions.d#L87-L89

nordlow:
Turns out removing alias this at https://github.com/dlang/dub/pull/2360/commits/99a513e6018f919011b5763778f2a89413ea6183

exposed more pending conversion from string to PackageName. Now

dub test --compiler=/usr/bin/dmd

fails as

```
source/dyaml/node.d(521,13): Error: static assert:  `allowed!(PackageName) || hasNodeConstructor!(inout(PackageName)) || !hasIndirections!(PackageName) && hasNodeConstructor!(Unqual!T)` is false
source/configy/Read.d(764,33):        instantiated from here: `get!(PackageName, Flag.yes)`
source/configy/Read.d(544,46):        instantiated from here: `parseFieldImpl!(__T8FieldRefTS3dub6recipe9selection8SelectedVAyaa8_76657273696f6e73Vbi0Z)`
source/configy/Read.d(643,20):        instantiated from here: `convertField!(__T8FieldRefTS3dub6recipe9selection8SelectedVAyaa8_76657273696f6e73Vbi0Z)`
/usr/include/dmd/phobos/std/meta.d(652,50):        ... (4 instantiations, -v to show) ...
source/configy/Read.d(286,37):        instantiated from here: `parseConfigFileSimple!(Selected)`
source/dub/project.d(85,50):        instantiated from here: `parseConfigFileSimple!(Selected)`
/usr/bin/dmd failed with exit code 1.
```

. Do you have any clues to why, @Geod24?

nordlow:
> 

Ok, thanks. Gonna use `opSlice` for those cases if it's ok with you.

source/dub/recipe/selection.d
```diff
@@ -16,7 +16,7 @@ public struct Selected
     public uint fileVersion;
 
     /// The selected package and their matching versions
-    public YAMLSelectedDependency[string] versions;
+    public YAMLSelectedDependency[PackageName] versions;
```
Geod24, line 19: @nordlow : Looks like this is where the error originate. Looks like a Configy bug, will look into it.

nordlow: Thank you!

source/dub/recipe/io.d
nordlow: See 4f47bff4.

source/dub/dependency.d
nordlow: I've removed the alias this in favor of more member functions so resolving this.

source/dub/dub.d
nordlow: Ok, I'm gonna use the pattern:

```	/// Compatibility overload. Use the version without a `force_remove` argument instead.
	void remove(PackageName name, PlacementLocation location, bool force_remove,
				scope size_t delegate(in Package[] packages) resolve_version)
	{
		remove(name, location, resolve_version);
	}
    /// ditto
	deprecated("Use the overload taking a `PackageName` as first parameter instead ")
	void remove(string name, PlacementLocation location, bool force_remove,
				scope size_t delegate(in Package[] packages) resolve_version)
    {
        remove(PackageName(name), location, force_remove, resolve_version);
    }
```

Geod24:
@nordlow : I suggest you squash the commits, because as it stands, the commits are a log of your development iterations, when they should really be a log of atomic changes.

nordlow:
> @nordlow : I suggest you squash the commits, because as it stands, the commits are a log of your development iterations, when they should really be a log of atomic changes.

Ok if squash at the end?

source/dub/recipe/selection.d
nordlow: Have you made any progress on tracking down the Configy bug? If not what's the easiest way to circumvent this issue, @Geod24?

Geod24: You can use `string` for the moment yes.

nordlow: I've revert back to

source/dub/recipe/json.d
nordlow: Fixed.

source/dub/project.d
nordlow: Resolved via e8321eb6.

source/dub/dub.d
nordlow: Resolved via 6468bb2f

nordlow:
Seems like the tests are creating PackageNames from paths at

https://github.com/dlang/dub/runs/7711089700?check_suite_focus=true#step:6:34

. What to do about this, @Geod24?

Geod24:
@nordlow : The point of the check is that some of those strings are not actually package names.
For example, from commandline, we get package name + version spec.

source/dub/dependency.d
```diff
@@ -21,12 +21,97 @@ import std.exception;
...
+            foreach (const idx, char c; value)
+                enforce(c.isLower || c.isDigit || c == '-' || c == '_' || c == ':' || c == '.',
+                        format("Character '%c' at index %s is not allowed in full package name '%s'",
+                               c, idx, value));
+        }
+		this._value = value;
+
+		version (none) {
+            /* Disabled because it's used in the package recipe
+               Allowing it means that a `PackageName` needs a context */
```
Geod24, line 54: Looking into this now

nordlow:
> ```d
> ```d
> auto recipe_default_package_name = PackageName(path.toString.baseName.stripExtension.strip);
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> ```

Are you saying we should disable the PackageName sanity check in calls such as

```
auto recipe_default_package_name = PackageName(path.toString.baseName.stripExtension.strip);
```

via an optional parameter to PackageName.this(string, bool sanityCheck = true)

?

Geod24:
No, I think the caller should do the transformation. Converts anything that's not a valid package name to a package name, e.g. call `toLower`, `replace(".", "-")`, etc...

nordlow:
> @nordlow : The point of the check is that some of those strings are not actually package names. For example, from commandline, we get package name + version spec.

Yes, and in this above it's a file path to a .d file containing a dub build spec in the top-comment.

nordlow:
> ```
> auto recipe_default_package_name = PackageName(path.toString.baseName.stripExtension.strip);
> ```

So this needs to replace '/' with character then? There might be unforeseen characters in the path that needs to converted. How do adjust this to cover all the characters allowed in a path?

nordlow:
Working on merging with upstream/master right now.

nordlow:
Shall we disallow

unittest
{
	// https://github.com/dlang/dub/issues/1681
	assert(splitPackageName("") == PackageAndVersion(PackageName(""), null));
}


currently defined in `commandline.d` unittest

?

nordlow:
Is https://github.com/dlang/dub/pull/2360/commits/ac325eacdead692536cc8353c174ddd30a7b2567 ok, @Geod24?

nordlow:
On second thought, from an end-user perspective I don't think it's much use in adding warnings (nor enforces) to `PackageName.this(string)` because the diagnostics can give no information about what dub.sdl/json file to modify to adjust the naming. What do you say, @Geod24?

nordlow:
`dub test` passes locally now again. Let's what the CI says.

nordlow:
I don't understand why https://github.com/dlang/dub/runs/7712137904?check_suite_focus=true#step:6:79 fails and I don't know how run it locally? 

I'm running it locally as

```d
dub build --compiler=/usr/bin/dmd && bin/dub run --single test/issue2051_running_unittests_from_dub_single_file_packages_fails.d
```

Do you, @Geod24?

nordlow:
I tried removing

nordlow:
Found the error at https://github.com/dlang/dub/runs/7716164874?check_suite_focus=true#step:6:119 caused by missing qualifiers on PackageName.fromString. Resolved via https://github.com/dlang/dub/pull/2360/commits/390369a84d2a98103baccbdaabc37f145cff761f.

Geod24:
> On second thought, from an end-user perspective I don't think it's much use in adding warnings (nor enforces) to `PackageName.this(string)` because the diagnostics can give no information about what build file, typically `dub.sdl` or `dub.json`, to edit to adjust the naming. What do you say, @Geod24?
> 
> I don't think asserts should be added either.

Adding `enforce` will give information about which `dub.json` field failed parsing. The new configuration parsing framework will catch any exception thrown from a constructor or `fromString` and wrap it with file/line/field information. So `enforce` really is the way to go here.

Geod24:
> Is [e9ab013](https://github.com/dlang/dub/commit/e9ab0134de41b145780e8141305ee5f9a5d90d51) ok despite it triggering 2890 warnings during `dub test`, @Geod24?

No, the constructor / `fromString` really should not trigger warnings, because one expect to be able to construct it from anywhere in the codebase. It would be okay to take that approach if we had, for example, `fromUserString` or similar, which would then have very clear expectation. But as explained above, Configy was built to handle this with exceptions.

Geod24:
> I don't understand why https://github.com/dlang/dub/runs/7712137904?check_suite_focus=true#step:6:79 fails. I'm running it locally as
> 
> ```d
> dub build --compiler=/usr/bin/dmd && bin/dub run -f --single test/issue2051_running_unittests_from_dub_single_file_packages_fails.d
> ```
> 
> Do you have any clues, @Geod24?

It doesn't fail. The logs say "Ok. Unittests failed.", which is a bit misleading, but it essentially test that tests are run (and does this by running a test that fails and making sure it did).
Real error have a red line on the CI logs, because they have return code != 0.
In this case the error was https://github.com/dlang/dub/runs/7712137904?check_suite_focus=true#step:6:793 as you found out.

source/dub/dependencyresolver.d
```diff
@@ -392,136 +392,141 @@ unittest {
...
+		assert(res.resolve(TreeNode(PackageName("a"), ic(0))) == [PackageName("b"):ic(2), PackageName("c"):ic(3), PackageName("d"):ic(1), PackageName("e"):ic(2)], format("%s", res.resolve(TreeNode(PackageName("a"), ic(0)))));
 	}
 
 	// handle cyclic dependencies gracefully
 	with (TestResolver) {
 		auto res = new TestResolver([
-			"a:0": [TreeNodes("b", ics([ic(1)]))],
-			"b:1": [TreeNodes("b", ics([ic(1)]))]
+            "a:0": [TreeNodes(PackageName("b"), ics([ic(1)]))],
+            "b:1": [TreeNodes(PackageName("b"), ics([ic(1)]))]
```
Geod24, line 417: Looks like you changed the indent in quite a few places too

source/dub/dependency.d
```diff
@@ -21,12 +21,116 @@ import std.exception;
 import std.string;
 import std.sumtype;
 
+/** Encapsulates the name of a package.
+*/
+struct PackageName {
+    const enum separator = ":"; ///< Separates sub-package names.
```
atilaneves, line 27: The indentation is off.

source/dub/dependency.d
```diff
@@ -21,12 +21,116 @@ import std.exception;
...
+
+	private string _value;
+
+	@disable this(PackageName name) @safe pure nothrow @nogc; // TODO: maybe remove when adding of `PackageName` has been merged
```
atilaneves, line 31: ?

source/dub/dependency.d
```diff
@@ -21,12 +21,116 @@ import std.exception;
...
+
+    version(none)
```
atilaneves, line 33: The indentation is off. Also, why introduce this?

atilaneves:
I don't understand all the `version(none)` in the diff, nor why `PackageName` had to replace `string` everywhere all at once instead of per module to result in smaller PRs.

The word "also" in the description is a red flag.

s-ludwig:
This looks like a hard breaking change and will break builds of external projects if this goes into 1.x.x. We need at least implicit conversion to `string`, an `opAssign(string)` and `string` based compatibility overloads for all public functions that now take `PackageName`.

Geod24:
Yes, there is no plan to merge this as-is, but getting the full diff to work is IMO worthwhile.
Then a good chunk of the changes can be cherry-picked into smaller commits though.

nordlow:
> > On second thought, from an end-user perspective I don't think it's much use in adding warnings (nor enforces) to `PackageName.this(string)` because the diagnostics can give no information about what build file, typically `dub.sdl` or `dub.json`, to edit to adjust the naming. What do you say, @Geod24?
> > I don't think asserts should be added either.
> 
> Adding `enforce` will give information about which `dub.json` field failed parsing. The new configuration parsing framework will catch any exception thrown from a constructor or `fromString` and wrap it with file/line/field information. So `enforce` really is the way to go here.

Which checks in do you want to add enforces for in `PackageName.this(string)`?

nordlow:
Merged with master. Will look at the other comments now.

nordlow:
> This looks like a hard breaking change and will break builds of external projects if this goes into 1.x.x. We need at least implicit conversion to `string`, an `opAssign(string)` and `string` based compatibility overloads for all public functions that now take `PackageName`.

Is adding

```d
struct PackageName
private string value;
alias value this;
```

the only way to realize implicit conversion from `PackageName` to `string`, @s-ludwig?

And should

`private string value;`

be made public?

nordlow:
> This looks like a hard breaking change and will break builds of external projects if this goes into 1.x.x. We need at least implicit conversion to `string`, an `opAssign(string)`

I've added opAssign(string) aswell now. 

>  and `string` based compatibility overloads for all public functions that now take `PackageName`.

Is it ok if I tag the existing string-based compatibility overloads as deprecated?

Geod24:
> Is it ok if I tag the existing string-based compatibility overloads as deprecated?

I think so, except for `PackageSupplier`.

nordlow:
I wonder if `PackageName` is the ideal naming considering it include sub-package (paths) and versions. What about `PackageId`? The term is already currentl used as variables names (`package_id` and `packageId`). Shorter and more indicating that we are not solely talking about a named here.

nordlow:
When a suitable naming has been setteld on shall we start by branching of changes into a separate PR that only includes the definition of `PackageName/Id/Path` along with a single module making use of it? Ping, @Geod24 @thewilsonator.

Geod24:
> I wonder if `PackageName` is the ideal naming considering it include sub-package (paths) and versions.

It should not include the version, only the (fully qualified) package name.

> What about `PackageId` which is already being in the variables namings `package_id` and `packageId`. Fewer characters to type and more indicating that we are not solely talking about the name part here. A possible alternate naming is `PackagePath`. Ping, @Geod24.

We **are** talking solely about the name. A subpackage name is just qualified / nested.

Geod24:
@nordlow : Could you rebase on master and remove merge commits ?

nordlow:
Ok let's stick to PackageName then. Will do rebase soon. Hopefully today.

Issue #789, opened 2016-03-15
Deprecation plan for dub.json/package.json


greenify:
Hey,

I am quite new to dub, so I don't know whether this has been discussed before, but from that I can tell by looking at older projects was dub in the beginning inspired by npm and used `package.json` in its first version. I guess that people pretty fast discovered that it doesn't work well with npm/web projects, so it was renamed to `dub.json`. At some point you realized that `json` is a horrible configuration format
and added support for sdl.

=> That's great - you learned from your mistakes and I like sdl too. However I (as a newcomer) think you should move forward. I saw that you already have a [tool to convert](https://github.com/D-Programming-Language/dub/pull/732) old json project to sdl, so there's nothing that stops you from deprecation it!

As a first action I recommend to make `sdl` the default package format on code.dlang.org and add a deprecation warning to `json`. This is very important for newcomers - I myself bumped into this problem of not knowing that the support for json is deprecated.

And imho for your next major release:

1) you should deprecate `dub.json` and `package.json`
2) you should remove the documentation for the `json` format from your website

s-ludwig:
Unfortunately (IMO) it was ultimately decided to take a step back and make JSON the prominent default format again. However, while I very much favor the SDLang format, removing support for JSON would be far too disrupting at this point. Instead, it is supposed to stay as a fully supported alternative. But I'd at least agree with going one step further to deprecate "package.json", it should probably result in a warning for both, the CLI and on code.dlang.org).

greenify:
> Unfortunately (IMO) it was ultimately decided to take a step back and make JSON the prominent default format again.

Wow that's sad to hear - do you recall the reasons for this decisions?

> Instead, it is supposed to stay as a fully supported alternative

I don't know any other package manager that supports two formats. It just makes stuff more complicated to document and keep up-to-date. Imho once you decided for which format you want to go, you should make that one the prominent one and let the other at least slip into the background.
One good example is the installation section at the front of every package page - currently it's for json - even though new projects are created in sdl.

Dicebot:
> Wow that's sad to hear - do you recall the reasons for this decisions?

Relevant NG archive (link to Andrei opinion on topic) http://forum.dlang.org/post/n3517b$9gv$1@digitalmars.com

AFAIR Walter and Andrei are pushing for throwing SDL away and sticking to JSON. I personally hope that will never happen :)

s-ludwig:
> Wow that's sad to hear - do you recall the reasons for this decisions?

[...]

Issue #724, opened 2015-11-26
Walter and Andrei want dub to stick with JSON and ditch SDL


jmdavis:
Since Sonke hasn't yet jumped in on the discussion in the D newsgroup on this, I'm creating this bug report / enhancement request to highlight the issue. But per this newsgroup discussion

http://forum.dlang.org/post/gnfctbuhiemidetngrzi@forum.dlang.org

Walter and Andrei are both firmly against dub using SDL and seem to want it to use JSON exclusively rather than use SDL at all (let alone as the default). Andrei in particular seems to think that using SDL is a huge mistake.

s-ludwig:
We've resolved this case by deciding to keep SDLang, but to also keep JSON as the default format for now.

bubnenkoff:
Please switch default view of http://code.dlang.org/package-format to JSON

PR #2361, opened 2022-08-04
Default `dub init` to SDL

WebFreak001:
Átila wants to throw out dub.json, so this is the first step to discouraging dub.json

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -863,7 +863,7 @@ class HelpCommand : Command {
 class InitCommand : Command {
 	private{
 		string m_templateType = "minimal";
-		PackageFormat m_format = PackageFormat.json;
+		PackageFormat m_format = PackageFormat.sdl;
 		bool m_nonInteractive;
 	}
 	this() @safe pure nothrow


thewilsonator (approved):

Geod24:
:/
https://github.com/dlang/dub/issues/724

Geod24:
Also, can we *not* merge such controversial changes without a discussion, pretty please ?

WebFreak001:
I agree this is a pretty controversial change. I had a chat with Átila about the decision to drop dub.json, I'm not fully happy with that decision, but I think it will certainly offer benefits:

- the SDL format is easier to read and write
- it supports comments
- it supports `` `grave accent` `` style strings, which is good for path strings

Átila suggested roughly the following plan:
- deprecate dub.json
- automatically open issues on all dub packages that are using dub.json
- PR with `dub convert` to all packages
  - side-effect: dead repositories will be identified like this
  - repositories that are dead, but actually used, should be moved to the dlang-community organization in that case
- disallow building root packages with `dub.json` (but keep dependencies working)

now there is still time to discuss things and what to do exactly. I think another idea would be simply disallowing packages in dub.json format to be published to the registry anymore, but keep everything as-is otherwise.

There had been the failed attempt #789, but I think with Átila's backing here we can actually make it through.

I think just changing the `dub init` default for now is going to be better for users, we could maybe add a dub.settings.json setting for that for those that _really_ don't want to change. Additionally I'm currently improving the dub docs (e.g. also merge sdl and json into single pages, with tabbing) and would like to have SDL as the default there.

Now I just merged in this PR without discussion as Átila strongly expressed his feelings about this to me and that we definitely want to drop JSON eventually. Having this dub release be a big change in many different aspects (colors, json parsing improvements, possibly `dub explain` in it) I think it's additionally beneficial to make such a change now, when a lot is changing, rather than later.

Geod24:
> but I think it will certainly offer benefits: [...]

I don't find `sdl` easy to write. Each time I need to write it, I have to refer to the documentation.
And `sdl` has one major drawback which IMO is a showstopper: Almost no one else supports it. We want more people, not less.

> Átila suggested roughly the following plan: [...]

What's the timeline here ? Because if we are to deprecate `dub.json`, that should have a looooooong deprecation period.
And we should be consistent and not rely on JSON everywhere in the dub code before we even think about this, otherwise it's just an arbitrary limitation.

> now there is still time to discuss things and what to do exactly.

We can start by reverting this.

> I think another idea would be simply disallowing packages in dub.json format to be published to the registry anymore, but keep everything as-is otherwise.

Don't see it as a good idea. It will just drive people not to use the registry.

>  Having this dub release be a big change in many different aspects (colors, json parsing improvements, possibly dub explain in it) I think it's additionally beneficial to make such a change now, when a lot is changing, rather than later.

First: I don't think those things are related. We might have a lot of improvements coming over the next few months, because that's how long it will take to get dub in a good state. Whether the change comes in v1.30.0 or v1.33.0 matters little.
Second: I have very little interest in contributing to dub if that's the direction we're going. Deprecating JSON means that a large chunk of the contributions I made recently are, in fact, useless. And as you know quite well, it also goes in the opposite direction of my long term goals, so this really just feels like sabotage.

WebFreak001:
ok gonna revert then, we will probably want to raise these points to Átila as well and figure out if it's even really a problem to begin with.

Herringway:
Trying to remember what keys and values dub expects is enough of a cognitive load. don't make me learn yet another configuration language with its own quirks and edge cases without a real good reason for it. comments and backtick strings are not compelling.

PR #2373, opened 2022-08-08
PackageManager: Move some `Location`-specific routines to `Location`

Geod24:
I think most of the `PackageManager` routines could be moved to `Location`, but not sure they should.
Those one are no-brainers tho.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -482,27 +482,27 @@ class PackageManager {
 	void addOverride(PlacementLocation scope_, string package_, Dependency version_spec, Version target)
 	{
 		m_repositories[scope_].overrides ~= PackageOverride(package_, version_spec, target);
-		writeLocalPackageOverridesFile(scope_);
+		m_repositories[scope_].writeOverrides();
 	}
 	/// ditto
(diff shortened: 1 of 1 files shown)


source/dub/packagemanager.d
```diff
@@ -523,7 +523,7 @@ class PackageManager {
 			if (ovr.package_ != package_ || ovr.source != src)
 				continue;
 			rep.overrides = rep.overrides[0 .. i] ~ rep.overrides[i+1 .. $];
-			writeLocalPackageOverridesFile(scope_);
+			(*rep).writeOverrides();
```
WebFreak001, line 526:
```suggestion
			rep.writeOverrides();
```
should work, no?

Geod24: Should work, but I don't like not being explicit with pointers, as it can break in non-obvious ways (e.g. when having a pointer to an array...).

WebFreak001 (approved):

WebFreak001:
huh would have thought dlang-bot would wait until the CI finished, but it's ok when most of them succeed anyway

Geod24:
It only waits for required ones :(

PR #2374, opened 2022-08-08
Dub.fetch: Deprecate Dependency overload, only accept Version[range]

Geod24:
Because fetching a path-based or repository dependency doesn't
make much sense.

I checked what reggae is doing, and indeed:
https://github.com/atilaneves/reggae/blob/4048813c1893fe70c424f6974d037063f8e98ffa/src/reggae/dub/interop/fetch.d#L47

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -586,7 +586,7 @@ class Dub {
 
 			FetchOptions fetchOpts;
 			fetchOpts |= (options & UpgradeOptions.preRelease) != 0 ? FetchOptions.usePrerelease : FetchOptions.none;
-			if (!pack) fetch(p, ver, defaultPlacementLocation, fetchOpts, "getting selected version");
+			if (!pack) fetch(p, ver.version_, defaultPlacementLocation, fetchOpts, "getting selected version");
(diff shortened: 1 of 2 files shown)


source/dub/dub.d
```diff
@@ -813,7 +830,7 @@ class Dub {
 			import std.zip : ZipException;
 
 			auto path = getTempFile(basePackageName, ".zip");
-			supplier.fetchPackage(path, basePackageName, dep, (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
+			supplier.fetchPackage(path, basePackageName, Dependency(range), (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
```
WebFreak001, line 833: shouldn't the suppliers also be updated?

Geod24: I can't because it's an `interface`, so it would potentially break library users.
The `PackageSupplier` has many issues (tied to Json, this, and generally too tailored to the registry) and I plan to overhaul it in the future, but I'd rather do it in one pass.

WebFreak001 (approved):

PR #2375, opened 2022-08-08
PackageManager.getBestPackage: Introduce Version[Range] overloads

Geod24:
Continuing on my crusade to have correctly-typed interfaces when it comes to dependencies.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -660,7 +660,7 @@ class Dub {
 
 		auto tool = "dscanner";
 
-		auto tool_pack = m_packageManager.getBestPackage(tool, ">=0.0.0");
+		auto tool_pack = m_packageManager.getBestPackage(tool);
 		if (!tool_pack) tool_pack = m_packageManager.getBestPackage(tool, "~master");
 		if (!tool_pack) {
 			logInfo("Hint", Color.light_blue, "%s is not present, getting and storing it user wide", tool);
@@ -1155,7 +1155,7 @@ class Dub {
 	private void runCustomInitialization(NativePath path, string type, string[] runArgs)
 	{
 		string packageName = type;
-		auto template_pack = m_packageManager.getBestPackage(packageName, ">=0.0.0");
+		auto template_pack = m_packageManager.getBestPackage(packageName);
 		if (!template_pack) template_pack = m_packageManager.getBestPackage(packageName, "~master");
 		if (!template_pack) {
 			logInfo("%s is not present, getting and storing it user wide", packageName);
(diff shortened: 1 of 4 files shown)


source/dub/packagemanager.d
```diff
@@ -224,6 +224,7 @@ class PackageManager {
 
 	/** Looks up the first package matching the given name.
 	*/
+	deprecated("The semantic of this function are unhelpful, prefer `getBestPackage` with `name, Dependency.any` instead")
```
WebFreak001, line 227:
```suggestion
	deprecated("The semantic of this function are unhelpful, prefer `getBestPackage` with `name, VersionRange.any` instead")
```

source/dub/dub.d
```diff
@@ -1666,10 +1666,13 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 					logDiagnostic("Sub package %s doesn't exist in %s %s.", name, basename, dep.version_);
 					return null;
 				}
-			} else if (auto ret = m_dub.m_packageManager.getBestPackage(name, dep)) {
-				return ret;
 			} else {
-				logDiagnostic("External sub package %s %s not found.", name, dep.version_);
+				auto ret = dep.visit!(
+					(VersionRange rng) => m_dub.m_packageManager.getBestPackage(name, rng),
+					(any) => assert(0, "Cannot call getPackageRaw with a non-version dependency on subpackages"),
```
WebFreak001, line 1672: it looks like it'd be possible that this `assert(false)` branch is called from

getSpecificConfigs -> getPackage -> getPackageRaw

WebFreak001 (changes requested):

source/dub/dub.d
Geod24: It shouldn't be, because otherwise the `assert(0)` in `Dependency.version_` would get triggered.
But I'll look at the CI failure.

WebFreak001: the dependency resolution logic is tricky to understand, so I don't really know where to start searching for such a case, but in general these non-version cases are not that well tested, so an edge case is unlikely to trigger in CI here

Geod24:
@WebFreak001 : Removed deprecation, this should be a no-brainer now (and fix the "TODO").

WebFreak001 (approved):
want to deprecate the `Dependency` overload or do you see any value in keeping it?

Geod24:
I plan to deprecate it, but the previous iteration didn't pass the CI for unknown reason, so I'll do the deprecation later (it was a side product, not the main target, of the refactoring).

PR #2376, opened 2022-08-10
Project: Use Dependency.visit instead of if/else

Geod24:
A smaller chunk of #2375 to reduce the diff (and bisect the failure).

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -510,17 +510,20 @@ shared static this() {
 					}
 				} else if (m_selections.hasSelectedVersion(basename)) {
 					vspec = m_selections.getSelectedVersion(basename);
-					if (!vspec.path.empty) {
-						auto path = vspec.path;
-						if (!path.absolute) path = m_rootPackage.path ~ path;
-						p = m_packageManager.getOrLoadPackage(path, NativePath.init, true);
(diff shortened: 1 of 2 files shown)


source/dub/project.d
```diff
@@ -510,17 +510,20 @@ shared static this() {
...
+							auto path = path_.absolute ? path_ : m_rootPackage.path ~ path_;
+							auto tmp = m_packageManager.getOrLoadPackage(path, NativePath.init, true);
+							return resolveSubPackage(tmp, subname, true);
+						},
+						(Repository repo) {
+							auto tmp = m_packageManager.loadSCMPackage(basename, repo);
+							return resolveSubPackage(tmp, subname, true);
+						},
+						(VersionRange range) {
+							return m_packageManager.getBestPackage(dep.name, vspec);
```
Geod24, line 524: Note that until the `VersionRange` overload is introduced, this needs to use `vspec`.

WebFreak001: let's add a TODO?

Geod24: Well it's going to be fixed in the next PR.

Geod24:
Self merging as it's pretty straightforward and the comment is addressed in the follow-up.

PR #2382, opened 2022-08-14
Forward flag -color to compiler calls

nordlow:
Need help determining how to forward `options.colors_mode` to setting the instance of `BuildOption.color` which should, afaict, be forwarded automatically to `DMDCompiler.s_options` and `LDCCompiler.s_options`.

Update: I think I nailed it.

Can you have a look, @Geod24?

FYI, @kinke.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -173,9 +173,17 @@ struct CommandLineHandler
 				// Use default determined in internal.logging.initLogging().
 				break;
 			case on:
+				foreach (ref grp; commandGroups)
+					foreach (ref cmd; grp.commands)
+						if (auto pc = cast(PackageBuildCommand)cmd)
+							pc.baseSettings.buildSettings.options |= BuildOption.color;
 				setLoggingColorsEnabled(true);  // enable colors, no matter what
 				break;
 			case off:
+				foreach (ref grp; commandGroups)
+					foreach (ref cmd; grp.commands)
+						if (auto pc = cast(PackageBuildCommand)cmd)
+							pc.baseSettings.buildSettings.options &= ~BuildOption.color;
 				setLoggingColorsEnabled(false); // disable colors, no matter what
 				break;
 		}
--- a/test/colored-output.sh
+++ b/test/colored-output.sh
@@ -15,3 +15,14 @@ ${DUB} build --compiler=${DC} 2>&1 | { ! \grep $'^\x1b\[' -c; }
 
 # Test that --color=on enables colors in any case
 ${DUB} build --color=on --compiler=${DC} 2>&1 | \grep $'^\x1b\[' -c
+
+# Test forwarding to dmd flag -color
+
+# Test that --color=on set dmd flag -color
+${DUB} build -v --color=on --compiler=${DC} -f 2>&1 | \grep '\-color' -c
+
+# Test that --color=off set no dmd flag
+${DUB} build -v --color=off --compiler=${DC} -f 2>&1 | { ! \grep '\-color' -c; }
+
+# Test that --color=automatic set no dmd flag
+${DUB} build -v --color=automatic --compiler=${DC} -f 2>&1 | { ! \grep '\-color' -c; }
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -390,10 +390,11 @@ enum BuildOption {
 	betterC = 1<<23,              /// Compile in betterC mode (-betterC)
 	lowmem = 1<<24,               /// Compile in lowmem mode (-lowmem)
 	coverageCTFE = 1<<25,         /// Enable code coverage analysis including at compile-time (-cov=ctfe)
+	color = 1<<26,                /// Colorize output (-color)
 
 	// for internal usage
-	_docs = 1<<26,                // Write ddoc to docs
-	_ddox = 1<<27,                // Compile docs.json
+	_docs = 1<<27,                // Write ddoc to docs
+	_ddox = 1<<28,                // Compile docs.json
 }
 
 struct Flags (T) {
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -65,6 +65,7 @@ class DMDCompiler : Compiler {
 		tuple(BuildOption.profileGC, ["-profile=gc"]),
 		tuple(BuildOption.betterC, ["-betterC"]),
 		tuple(BuildOption.lowmem, ["-lowmem"]),
+		tuple(BuildOption.color, ["-color"]),
 
 		tuple(BuildOption._docs, ["-Dddocs"]),
 		tuple(BuildOption._ddox, ["-Xfdocs.json", "-Df__dummy.html"]),
--- a/source/dub/compilers/gdc.d
+++ b/source/dub/compilers/gdc.d
@@ -46,6 +46,7 @@ class GDCCompiler : Compiler {
 		tuple(BuildOption.property, ["-fproperty"]),
 		//tuple(BuildOption.profileGC, ["-?"]),
 		tuple(BuildOption.betterC, ["-fno-druntime"]),
+		tuple(BuildOption.color, ["-fdiagnostics-color=always"]),
 
 		tuple(BuildOption._docs, ["-fdoc-dir=docs"]),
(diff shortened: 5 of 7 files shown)


source/dub/commandline.d
```diff
@@ -173,9 +173,29 @@ struct CommandLineHandler
 				// Use default determined in internal.logging.initLogging().
 				break;
 			case on:
+                        	foreach (ref commandGroup; commandGroups)
+                                {
+                                	foreach (ref commands; commandGroup.commands)
+                                        {
```
thewilsonator, line 179: you're mixing tabs and spaces here. Also please dedent by 3 tabs worth and align the braces with the foreach (or drop them entirely). Ditto throughout

WebFreak001:
should we only pass explicit color argument to dmd or also implicit as set by NO_COLOR env (or other external ways)?

kinke:
What's important to me is that the color switch (`-fdiagnostics-color=always` for GDC btw) isn't used for the build hash - enforcing colors should not lead to cached packages being rebuilt. Otherwise colors could already be enforced via DFLAGS env var.

source/dub/commandline.d
nordlow: Resolved at 69ac9438.

nordlow:
> `-fdiagnostics-color=always` for GDC btw

Added via 50fe225b.

nordlow:
Should BuildOption.color be part of `inheritedBuildOptions` in `buildsettings.d`?

nordlow:
> What's important to me is that the color switch (-fdiagnostics-color=always for GDC btw) isn't used for the build hash - enforcing colors should not lead to cached packages being rebuilt.

Can anybody point me to the location in dub where this hash is calculated?

nordlow:
While I'm at it shall I replace MD5 with SHA1 or SHA256 in

```d
string computeBuildName(string config, in GeneratorSettings settings, const string[][] hashing...)
{
	import std.digest;
	import std.digest.md;

	MD5 hash;
	hash.start();
	void addHash(in string[] strings...) { foreach (s; strings) { hash.put(cast(ubyte[])s); hash.put(0); } hash.put(0); }
	foreach(strings; hashing)
		addHash(strings);
	auto hashstr = hash.finish().toHexString().idup;

    return format("%s-%s-%s-%s-%s_v%s-%s", config, settings.buildType,
			settings.platform.platform.join("."),
			settings.platform.architecture.join("."),
			settings.platform.compiler, settings.platform.compilerVersion, hashstr);
}
```

? If so, shall I truncate the result hash hex string to the same number of digits as MD5? Ping, @kinke @Geod24.

nordlow:
Anything else?

nordlow:
> should we only pass explicit color argument to dmd or also implicit as set by NO_COLOR env (or other external ways)?

I don't know.

nordlow:
> Otherwise colors could already be enforced via DFLAGS env var.

Are you saying we shall exclude color flags from DFLAGS env var aswell, @kinke?

kinke:
[Nope, just saying those rebuilds prevented me from working around this via DFLAGS for now.]

nordlow:
> [Nope, just saying those rebuilds prevented me from working around this via DFLAGS for now.]

So is this good to go, @thewilsonator @Geod24?

Geod24:
I'm on vacation this week so on mobile most of the time but I'll try to take a look one evening.

Geod24:
Changes look good but please add a test.

nordlow:
> Changes look good but please add a test.

Is it ok if I extend the existing tests for the `--color(s)` flag?

nordlow:
> Changes look good but please add a test.

Do you have a suggestion for how to add a test for this? Grep for the forwarded dflag for dmd only?

veelo:
> > Changes look good but please add a test.
> 
> Do you have a suggestion for how to add a test for this? I vote for grepping for the forwarded compiler flag in the output of a call to `dub -v ...` for both dmd, ldc and gdc? This prevents use from having to build with all 3 compilers.

I think that will be good enough. If you really would want to check the existence of color in the compiler output you could probably check for the presence of ANSI escape codes, but that requires you to know how each compiler formats its output, which can change between releases.

nordlow:
I'm trying to run the test colored-output.sh as

```sh
DC=/usr/bin/dmd DUB=../bin/dub CURR_DIR=. ./colored-output.sh 
```

under test sub-dir but that fails

as

```
DC=/usr/bin/dmd DUB=../bin/dub CURR_DIR=. ./colored-output.sh 
0
[ERROR] ./colored-output.sh:8 command failed
```

. What am I doing wrong? We really need a trivial way to run individual tests along with a documentation for it.

Need help. Ping, @Geod24 @thewilsonator.

veelo:
Just to rule out the obvious: did you try to run the command at line 8 by hand (after `cd 1-exec-simple`)?

nordlow:
Figured it out. The command dub build and test command I use is now

```sh
dub build && cd test && DC=dmd DUB=~/Work/dub/bin/dub CURR_DIR=. ./colored-output.sh 
```

. Do you have a better way of calling the tests?

nordlow:
Going green now!

veelo:
IMNSHO all shell scripts should be eliminated from all repositories. D is great for scripting, and cross platform.

Geod24:
> Questions. Why isn't variable
> 
> 1. `CURR_DIR` defaulted to `${BASH_SOURCE[0]}`?
> 2. `DC` defaulted to `dmd`?
> 3. `DUB` defaulted to `../bin/dub`?

2 and 3, because that would make it way too easy to accidentally make things work when we fail to pass the right compiler.
1 I don't remember but I think there's a reason. In any case, this test suite infrastructure was started almost 10 years ago by a college student that had no idea what he was doing at the time, so yeah we should rewrite it in D. I'm on a (side) quest of moving environment reads / IO out of some deeply nested functions so that we can actually unittest things (much faster than integration tests).

Geod24 (approved):
Great, thanks!

Geod24:
@nordlow : Whenever you want to update a PR, please do `git fetch upstream && git rebase upstream/master`, it makes life much easier.

test/colored-output.sh
```diff
@@ -13,5 +13,19 @@ ${DUB} build --color=automatic --compiler=${DC} 2>&1 | { ! \grep $'^\x1b\[' -c;
 # Test that no --color= has same behaviour as --color=automatic
 ${DUB} build --compiler=${DC} 2>&1 | { ! \grep $'^\x1b\[' -c; }
 
-# Test that --color=on enables colors in any case
+# Test that --color=on enables color in any case
 ${DUB} build --color=on --compiler=${DC} 2>&1 | \grep $'^\x1b\[' -c
+
+# Test --color=on enabling color in any case
+${DUB} build --color=on --compiler=${DC} 2>&1 | \grep $'^\x1b\[' -c
```
Geod24, line 20: Bad rebase

Geod24:
Rebased, G2G

nordlow:
How long does `buildkite/dub` normally take?

nordlow:
Hmm, I might have forgotten to forward coloring to compilers in the case where `--color=automatic`...

Geod24:
Hum, probably :)

nordlow:
> Hum, probably :)

Shall we do anything about this?

To be strict about this we need to add to change type of flag BuildOption.color from bool to `enum color { automatic, on, off }enum color { automatic, on, off }. Do we need to bother?

veelo:
Don't the compilers already display colors automatically (i.e., if output is a TTY)? In that case not forcing anything will already do the right thing in the default case.

>  change the type of the flag BuildOption.color from bool to enum color { automatic, on, off } which will require significant refactor of the type of build options as it currently assumes all build options boolean.

Which leads me to question how this is currently implemented.
```d
color = 1<<26, /// Colorize output (-color)
```
This adds a flag to force color on, no? And we don't have a flag to force color off. Thus from the tristate we only forward one state to the compiler currently.

I don't know if this is needed in any practical situation, but you could add another flag `nocolor` to force color off. Then if both flags are false it would imply automatic. Both flags true would be illegal.

nordlow:
> I don't know if this is needed in any practical situation, but you could add another flag `nocolor` to force color off. Then if both flags are false it would imply automatic. Both flags true would be illegal.

We could add that but that is bad design because it introduces error states. I vote for refactoring build options to be able to store flags other boolean such as, in this case, `enum color`.

Issue #2385, opened 2022-08-15
Replace MD5 with more secure hash in computeBuildName


nordlow:
Shall we replace MD5 with SHA1 or SHA256 or some other hash in

```d
string computeBuildName(string config, in GeneratorSettings settings, const string[][] hashing...)
{
	import std.digest;
	import std.digest.md;

	MD5 hash;
	hash.start();
	void addHash(in string[] strings...) { foreach (s; strings) { hash.put(cast(ubyte[])s); hash.put(0); } hash.put(0); }
	foreach(strings; hashing)
		addHash(strings);
	auto hashstr = hash.finish().toHexString().idup;

    return format("%s-%s-%s-%s-%s_v%s-%s", config, settings.buildType,
			settings.platform.platform.join("."),
			settings.platform.architecture.join("."),
			settings.platform.compiler, settings.platform.compilerVersion, hashstr);
}
```

? If so, shall I truncate the result hash hex string to the same number of digits as MD5?

Phobos' module `std.digest.sha` currently defines

```d
alias SHA1 = SHA!(512, 160);  /// SHA alias for SHA-1, hash is ubyte[20]
alias SHA224 = SHA!(512, 224);  /// SHA alias for SHA-224, hash is ubyte[28]
alias SHA256 = SHA!(512, 256);  /// SHA alias for SHA-256, hash is ubyte[32]
alias SHA384 = SHA!(1024, 384); /// SHA alias for SHA-384, hash is ubyte[48]
alias SHA512 = SHA!(1024, 512); /// SHA alias for SHA-512, hash is ubyte[64]
alias SHA512_224 = SHA!(1024, 224); /// SHA alias for SHA-512/224, hash is ubyte[28]
alias SHA512_256 = SHA!(1024, 256); /// SHA alias for SHA-512/256, hash is ubyte[32]
```

Which one shall we pick?

Ping, @kinke @Geod24 @John-Colvin.

Geod24:
I have a preference for Blake2b. I think we should decide on a unified function to use for the build ID and the dependency checksum.

Geod24:
> Blake2b is not part of Phobos. I vote for SHA-256 in the mean time to get this in. We can always switch to Blake later on.

I don't see the urgency here though? The current hash function is good enough, the only reason I'm considering changing it (and this is just my personal opinion) is that we want to add checksums and it's a good time to revisit the hash function.

But if we're putting more focus on security, there's a lot of places where we can have a higher impact than changing the hash function.

[...]

PR #2386, opened 2022-08-15
Use SHA256 in place of MD5 in computeBuildName

nordlow:
Closes https://github.com/dlang/dub/issues/2385.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -34,15 +34,14 @@ string getObjSuffix(const scope ref BuildPlatform platform)
 
 string computeBuildName(string config, in GeneratorSettings settings, const string[][] hashing...)
 {
-	import std.digest;
-	import std.digest.md;
+	import std.digest.sha : SHA256, toHexString;
 
-	MD5 hash;
+	SHA256 hash;
 	hash.start();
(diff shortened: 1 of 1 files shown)


nordlow:
Searching `import std.digest` gave hit in `pathToObjName`. So, btw, shall we change `crc32Of` to `SHA256` at

```d
	static string pathToObjName(const scope ref BuildPlatform platform, string path)
	{
		import std.digest.crc : crc32Of;
		import std.path : buildNormalizedPath, dirSeparator, relativePath, stripDrive;
		if (path.endsWith(".d")) path = path[0 .. $-2];
		auto ret = buildNormalizedPath(getcwd(), path).replace(dirSeparator, ".");
		auto idx = ret.lastIndexOf('.');
		const objSuffix = getObjSuffix(platform);
		return idx < 0 ? ret ~ objSuffix : format("%s_%(%02x%)%s", ret[idx+1 .. $], crc32Of(ret[0 .. idx]), objSuffix);
	}
```

aswell?

thewilsonator:
Will leave that for input from others who are more awake than I.

Geod24:
That is probably going to invalidate a lot of build caches, which means it will inflate people's local cache, but that's another problem. Let's go ahead with this.

PR #2388, opened 2022-08-15
Use SHA256 and better namings in hashPackage

nordlow:


Geod24 (approved):
LGTM, although this function is completely unused AFAIK.

Issue #2377, opened 2022-08-12
Windows regression: dynamicLibrary targets wrongly copy all static .lib files they depend on


kinke:
As per title. In our case at work, standalone `dynamicLibrary` targets linking all dub deps statically now get dozens of static .lib files copied to their output directory. This didn't happen with dub v1.28 but does for recent dub master.

rikkimax:
This one is almost certainly on me. The problem is import libraries have
the same extension as static libraries and it'll be copying them all by
extension a bit too eagerly.

rikkimax:
I have a hunch that this should not be a regression.

If you convert it to executable, using the older dub, does it still exhibit
this behavior?

kinke:
Yep, this was almost certainly introduced by #2259.

> If you convert it to executable, using the older dub, does it still exhibit this behavior?

I assume you meant the *newer* dub - yes, loads of useless files ending up in the output dir too now. And an import lib for the executable (it has a few exports).

---

Copying the import .lib and .exp files is fine - but IMO only for the `dynamicLibrary` target itself. So not for an executable target even if it has exports, and definitely no .lib/exp files from dependencies - regardless of whether they are static libs or dynamic libs.

rikkimax:
> Yep, this was almost certainly introduced by #2259.
> 
> > If you convert it to executable, using the older dub, does it still exhibit this behavior?
> 
> I assume you meant the _newer_ dub - yes, loads of useless files ending up in the output dir too now. And an import lib for the executable (it has a few exports).

No, I did mean the older dub.

For executables that should be an easy change. Just move a couple of the extension handling into its own dynamic library check.

https://github.com/dlang/dub/blob/master/source/dub/generators/build.d#L423

However, as for dynamic libraries as dependencies, this is a capability that I do want. This makes picking and choosing for distributing (say frameworks) as binaries much easier. The alternative is to go looking into build directories or do it step by step which are both more work than what should be required for common outputs.

My changes should not have affected static libraries.

Right now things are consistent, but yeah I can see why a large project like that might have a bit of a problem. We may want to split binary copying into "default" and "full" directives for target directory support.

kinke:
> However, as for dynamic libraries as dependencies, this is a capability that I do want. This makes picking and choosing for distributing (say frameworks) as binaries much easier. The alternative is to go looking into build directories or do it step by step which are both more work than what should be required for common outputs.

[...]

PR #2396, opened 2022-08-17
Windows: Copy .lib files from DLL deps only

kinke:
One aspect of issue #2377 - don't copy static libs. And only copy if the dependee is a DLL too.

### DIFF
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -74,50 +74,6 @@ unittest {
 
 
 /**
-	Determines if a specific file name has the extension related to dynamic libraries.
-
-	This includes dynamic libraries and for Windows pdb, export and import library files.
-*/
-bool isDynamicLibraryFile(const scope ref BuildPlatform platform, string f)
-{
-	import std.path;
-	switch (extension(f)) {
-		default:
-			return false;
-		case ".lib", ".pdb", ".dll", ".exp":
-			return platform.isWindows();
-		case ".so", ".dylib":
-			return !platform.isWindows();
-	}
-}
-
-unittest {
-	BuildPlatform p;
-
-	p.platform = ["windows"];
-	assert(!isDynamicLibraryFile(p, "test.obj"));
-	assert(isDynamicLibraryFile(p, "test.lib"));
-	assert(isDynamicLibraryFile(p, "test.dll"));
-	assert(isDynamicLibraryFile(p, "test.pdb"));
-	assert(!isDynamicLibraryFile(p, "test.res"));
-	assert(!isDynamicLibraryFile(p, "test.o"));
-	assert(!isDynamicLibraryFile(p, "test.d"));
-	assert(!isDynamicLibraryFile(p, "test.dylib"));
-
-	p.platform = ["something else"];
-	assert(!isDynamicLibraryFile(p, "test.o"));
-	assert(!isDynamicLibraryFile(p, "test.a"));
-	assert(isDynamicLibraryFile(p, "test.so"));
-	assert(isDynamicLibraryFile(p, "test.dylib"));
-	assert(!isDynamicLibraryFile(p, "test.obj"));
-	assert(!isDynamicLibraryFile(p, "test.d"));
-	assert(!isDynamicLibraryFile(p, "test.lib"));
-	assert(!isDynamicLibraryFile(p, "test.dll"));
-	assert(!isDynamicLibraryFile(p, "test.pdb"));
-}
-
-
-/**
 	Adds a default DT_SONAME (ELF) / 'install name' (Mach-O) when linking a dynamic library.
 	This makes dependees reference their dynamic-lib deps by filename only (DT_NEEDED etc.)
 	instead of by the path used in the dependee linker cmdline, and enables loading the
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -103,41 +103,48 @@ class BuildGenerator : ProjectGenerator {
 
 			NativePath[] additional_dep_files, dependencyBinariesToCopy;
 			auto bs = ti.buildSettings.dup;
+			const dependeeTT = bs.targetType;
 			foreach (ldep; ti.linkDependencies) {
-				auto location = target_paths[ldep].toNativeString();
+				const ldepPath = target_paths[ldep].toNativeString();
+				const doLink = dependeeTT != TargetType.staticLibrary && !(bs.options & BuildOption.syntaxOnly);
 
-				if (bs.targetType != TargetType.staticLibrary && !(bs.options & BuildOption.syntaxOnly) && isLinkerFile(settings.platform, location)) {
-					bs.addSourceFiles(location);
-				} else if (settings.platform.isWindows() && location.endsWith(".dll")) {
-					// switch from linking against the dll to against the import library,
-					// and copy any dependent build artifacts if found too
+				if (doLink && isLinkerFile(settings.platform, ldepPath))
+					bs.addSourceFiles(ldepPath);
+				else
+					additional_dep_files ~= target_paths[ldep];
 
-					const pdbFilename = location.setExtension(".pdb");
-					if (existsFile(pdbFilename))
-						dependencyBinariesToCopy ~= NativePath(pdbFilename);
+				if (targets[ldep].buildSettings.targetType == TargetType.dynamicLibrary) {
+					// copy the .{dll,so,dylib} if the dependee is an executable or dynamic lib
+					if (dependeeTT == TargetType.executable || dependeeTT == TargetType.dynamicLibrary)
+						dependencyBinariesToCopy ~= NativePath(ldepPath);
 
-					const importLibraryLocation = location.setExtension(".lib");
-					if (existsFile(importLibraryLocation)) {
-						bs.addSourceFiles(importLibraryLocation);
-						dependencyBinariesToCopy ~= NativePath(importLibraryLocation);
+					if (settings.platform.isWindows()) {
+						if (dependeeTT == TargetType.executable || dependeeTT == TargetType.dynamicLibrary) {
+							// copy the accompanying .pdb if found
+							const pdb = ldepPath.setExtension(".pdb");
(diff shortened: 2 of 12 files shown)


rikkimax:
LGTM

kinke:
Sorry, had to fix some stuff after manual tests (e.g., forgot to copy the .{dll,so,dylib} itself).

And I've seen another little problem - it's only the *direct* dynamicLib deps that get copied to the dependee. E.g., with 3 dynamicLibs, `lib1` depending on `lib2` depending on `lib3`, only `liblib2.so` is copied to the `lib1` output dir, `liblib3.so` is missing (but in the lib2 output dir).

kinke:
Okay, I've changed it to
* handle indirect dynamicLib deps too, and
* copy the extra files to the *root* package output dir only.

This means that building `lib1` now ends up with all 3 libs in its output dir, while `liblib3.so` isn't copied to the `lib2` output dir - only when building `lib2` as the root package.

kinke:
@rikkimax: We might have to adapt the logic for your frameworks use case, e.g, copying all extra files for a root package with `none` targetType too.

rikkimax:
For my use case I'm happy with using a different target type for the root.
However that does seem like something other people would do and would
complain if it didn't work.

kinke:
Okay, added according support for `none` root packages too. That required some more changes, as such packages aren't actually built [and their `targetPath` is apparently ignored].

rikkimax:
CI is a bit red, it may be on GH's end. I.e.

``git clone -b v2.100.1 --depth=1 https://github.com/dlang/druntime .generated/druntime-2.100.1`` failing.

kinke:
Oh boy, there's still a couple of issues in general.

1. Bash test isn't run on Windows.
2. On Posix, the generated .so libs have no `SONAME`, so that the binaries end up referencing .so deps like `../dep2/.dub/build/dynlib-debug-…/libdep2.so` instead of `libdep2.so`. Which leads to linker errors (CI) or runtime loading errors (my box). This can be worked around via some `lflags "-soname=libdep2.so"` but should probably be added by dub implicitly.
3. *Indirect* dynamicLib deps apparently don't make it to the command line. This works for simple cases, but probably breaks very soon when templates enter the picture.

rikkimax:
> 2\. On Posix, the generated .so libs have no `SONAME`, so that the binaries end up referencing .so deps like `../dep2/.dub/build/dynlib-debug-…/[libdep2.so](http://libdep2.so)` instead of `[libdep2.so](http://libdep2.so)`. Which leads to linker errors (CI) or runtime loading errors (my box). This can be worked around via some `lflags "-soname=[libdep2.so](http://libdep2.so)"` but should probably be added by dub implicitly.

This sounds related to RPATH.

Quite a while back I did discuss that on Discord briefly, and no good simple solution was come up with.

Unfortunately, these are good examples of shared library support in D that need some serious rethinking. We don't have solid stories to tell people on how to use them without expecting errors.

kinke:
With `SONAME`, running the dynamically linked exe works fine by extending `LD_LIBRARY_PATH` (to include the exe dir) here on my Ubuntu 22 box. I've checked the bfd, gold, lld and mold linkers - the latter 3 all work, it's only ancient bfd (surely used by CI too) that complains about not finding indirect (and in this case, not directly needed) `libdep2.so` when linking the `parent` executable dynamically. Indirect deps should be linked by dub too, but the trivial testcase ought to work.

I'm not sure whether dub should mess with `RUNPATH` automatically, this might be a security issue (setting it to `$ORIGIN` makes it load .so libs from the same dir as on Windows). At work, we simply postprocess all binaries (executables and bundled .so deps) to be deployed via `patchelf --set-rpath '$ORIGIN/../lib' <path>`.

rikkimax:
Yeah you came to the same conclusion I did.

RPATH shouldn't be set automatically, but we may want to look at ways to make it easier to work with new directives or something along these lines. I just don't feel like I'm the right person to design this experience, hence it's kind of been left.

kinke:
Besides Windows (still need to look into it), it seems to be working now as I would expect it. The soname/install_name thing requires fiddling with RUNPATH for the executables run for existing test packages (`test/{2-dynLib-dep,issue2258-dynLib-exe-dep}`) - which works 'fine' on Linux with `$ORIGIN`, but `@executable_path` doesn't work on Mac. No idea what the 'right' setting for Mac is.

AFAICT, we need the so/install names in order to be able to use the copied dynlib deps in the same directory as the executable - either via LD_LIBRARY_PATH env var or RUNPATH. Otherwise the loader tries to load it from the original location specified in the linking cmdline, which is some `<path to dep pkg>/.dub/…` path (and according to previous errors, not absolute, but relative, and thus even more brittle).

> I just don't feel like I'm the right person to design this experience

Me neither. ;)

source/dub/compilers/ldc.d
```diff
@@ -150,8 +150,15 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 			settings.lflags = null;
 		}
 
-		if (settings.options & BuildOption.pic)
-			settings.addDFlags("-relocation-model=pic");
+		if (settings.options & BuildOption.pic) {
+			if (platform.isWindows()) {
+				// This has nothing to do with PIC, but as it's implicitly/commonly used to
+				// compile a shared library, explicitly specify what `-shared` defaults to.
+				settings.addDFlags("-fvisibility=public", "-dllimport=all");
```
rikkimax, line 157: If we go through with this change, it might be about time I did a DIP to replace export. Might as well make the change at the language level if this is what is actually required to make shared libraries work correctly.

@ibuclaw do any changes need to be made to match this with GDC?

kinke: Replace it with what? `export` is IMO still very useful to explicitly export a symbol when compiling with (implicit or explicit) `-fvisibility=hidden`. And there's a rather new `@ldc.attributes.hidden` UDA to complement it on a by-symbol basis, which should probably be standardized in `core.attribute`, at least with GDC too.

rikkimax: By default everything is exported if it's visible outside a module (so non-private). If you want fine-grained control you use a pragma (although a UDA works too).

This would match these changes and how people actually use it in practice.

kinke: IDK, to me the status quo (with LDC) in this regard seems totally fine - with `-fvisibility`, `export` and (exotic) `@hidden`.

kinke:
Static libs linked into Windows DLLs now also get the required `-fvisibility=public -dllimport=all` compile flags [with LDC]. And a DLL can override its own visibility via `dflags "-fvisibility=hidden"` - well, it's not passed down to its deps at least, but might be inherited by parents (haven't checked that).

So from the LDC side, the only real missing thing seems to be a default `-link-defaultlib-shared` for executables with direct/indirect dynamic-lib deps (required, otherwise the program aborts at runtime startup). Would make sense for DMD on Linux too (`-defaultlib=libphobos2.so`).

kinke:
> And a DLL can override its own visibility via dflags "-fvisibility=hidden" - well, it's not passed down to its deps at least, but might be inherited by parents (haven't checked that).

Oh, for our use case at work (few fat plugin DLLs with selective exports), we'd need to override these to `-fvisibility=hidden -dllimport=defaultLibsOnly`, for the DLL and all of its (static) deps. Grml.

kinke:
Okay, one (ugly) way to achieve that is setting env var `DFLAGS` to `-fvisibility=hidden -dllimport=defaultLibsOnly` - these flags are added *after* the implicit ones.

kinke:
@JohanEngelen @thewilsonator Do you guys know how to get the rpath to work on Mac? Quick recap for context: the approach is to copy all (direct and indirect) .dynlib deps to the root package's output directory (if it's a DLL or executable). For ELF, the SONAME of the generated .so is set implicitly to the filename; for Mach-O, it's the `-install_name` ld64 switch. Then for ELF, all we (apparently) need is either setting the LD_LIBRARY_PATH env var, or baking a suitable RUNPATH (e.g. `$ORIGIN`) into the executable (NOT automatically by dub!). Now for Mach-O, I've tried `-rpath @executable_path` for the executable, but that doesn't work. So the question is what it needs to be, and/or whether the .dylibs need further linker flags.

kinke:
Oh: https://github.com/ldc-developers/ldc/blob/1c3b55d19ed53776d63bb2c5df1ff71dca974cf9/CMakeLists.txt#L777

kinke:
I'll clean this up once the derived PRs above are in.

kinke:
Okay, ready now.

JohanEngelen:
> @JohanEngelen @thewilsonator Do you guys know how to get the rpath to work on Mac? Quick recap for context: the approach is to copy all (direct and indirect) .dynlib deps to the root package's output directory (if it's a DLL or executable). For ELF, the SONAME of the generated .so is set implicitly to the filename; for Mach-O, it's the `-install_name` ld64 switch. Then for ELF, all we (apparently) need is either setting the LD_LIBRARY_PATH env var, or baking a suitable RUNPATH (e.g. `$ORIGIN`) into the executable (NOT automatically by dub!). Now for Mach-O, I've tried `-rpath @executable_path` for the executable (**edit**: [5cf98d5](https://github.com/dlang/dub/commit/5cf98d5fd085ac2abfd32e8153dbbc9096b57daa)), but that doesn't work. So the question is what it needs to be, and/or whether the .dylibs need further linker flags.

Sorry I missed this, will try to take a look this Sunday.
In the meanwhile, perhaps this helps: https://github.com/ldc-developers/ldc/blob/b0eab59b1baf8d2618d47efaf39aa52082808a52/driver/linker-gcc.cpp#L322-L331

kinke:
@JohanEngelen: Oh, don't worry, it's been long solved. ;)

thewilsonator (approved):

kinke:
ping @Geod24 @WebFreak001

thewilsonator:
added `72h no objection -> merge` if someone doesn't merge it before then.

rikkimax:
At least you have a chance of not getting the implementation stuff wrong ;)

Issue #2398, opened 2022-08-17
Rename --colors to --color before next release?


nordlow:
At https://github.com/dlang/dub/pull/2268, what is the reason for picking the naming `--colors` instead of `--color`?

IRC, --color is more established naming in other (GNU) CLI tools such as `ls` and `grep`.

Ping, @kinke @Geod24 @veelo.

Geod24:
That's a good point. I'd be in favor of the rename. CC @WebFreak001 .

veelo:
Indeed, it should be `--color`.

WebFreak001:
sure, sounds good

nordlow:
See https://github.com/dlang/dub/pull/2399.

PR #2399, opened 2022-08-18
Rename CLI flag --colors to --color any D code symbols accordingly

nordlow:
Closes https://github.com/dlang/dub/issues/2398.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -167,7 +167,7 @@ struct CommandLineHandler
 			options.root_path = options.root_path.expandTilde.absolutePath.buildNormalizedPath;
 		}
 
-		final switch (options.colors_mode) with (options.colors)
+		final switch (options.color_mode) with (options.color)
 		{
 			case automatic:
 				// Use default determined in internal.logging.initLogging().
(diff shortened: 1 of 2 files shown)


source/dub/commandline.d
```diff
@@ -562,8 +562,8 @@ struct CommonOptions {
 	bool help, annotate, bare;
 	string[] registry_urls;
 	string root_path;
-	enum colors { automatic, on, off } // Style violation in support of invalid option error formatting.
-	colors colors_mode = colors.automatic;
+	enum Color { automatic, on, off } // Style violation in support of invalid option error formatting.
```
veelo, line 565: By changing the case from `color` to `Color` it is no longer a style violation ;-)

The reason I chose to violate the style and use lower case is so that if a user selects an unsupported value (like `--color=blue`) a corresponding error message is produced to the effect of "`blue` is not a valid value for `color`". With your change the error will become "`blue` is not a valid value for `Color`", which is not consistent with the option name.

veelo (changes requested):

source/dub/commandline.d
veelo: Perhaps the comment should read `// Lower case "color" in support of invalid option error formatting.` instead...

nordlow: Adjusted via https://github.com/dlang/dub/compare/a9a740c0d0985ed86e77dae7c06175d576927f78..53b01ca3d6e10a742c5c7790f648474120dccb82.

veelo (approved):

Geod24 (approved):
LGTM.

PR #2407, opened 2022-08-18
Qualify NativePath members

nordlow:
Trivial. Ready for merge.

### DIFF
--- a/source/dub/internal/vibecompat/inet/path.d
+++ b/source/dub/internal/vibecompat/inet/path.d
@@ -7,9 +7,6 @@
 */
 module dub.internal.vibecompat.inet.path;
 
-version (Have_vibe_core) public import vibe.core.path;
-else:
-
 import std.algorithm;
 import std.array;
 import std.conv;
--- a/source/dub/internal/vibecompat/inet/path.d
+++ b/source/dub/internal/vibecompat/inet/path.d
@@ -36,7 +36,7 @@ struct NativePath {
 	alias bySegment = nodes;
 
 	/// Constructs a NativePath object by parsing a path string.
-	this(string pathstr)
+	this(string pathstr) pure
 	{
 		m_nodes = splitPath(pathstr);
 		m_absolute = (pathstr.startsWith("/") || m_nodes.length > 0 && (m_nodes[0].toString().countUntil(':')>0 || m_nodes[0] == "\\"));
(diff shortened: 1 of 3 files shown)


Geod24:
Doesn't compile because Vibe.d has different qualifiers.

nordlow:
> Doesn't compile because Vibe.d has different qualifiers.

Oops, thanks. Forgot to compile all tests.

nordlow:
Ehh, I'm clueless about the CI errors. How can a simple adding of qualifiers break CI in any way?

Geod24:
How do you compile?
Test what the CI is doing. Building optionally with vibe-core is probably what breaks it.

nordlow:
> Building optionally with vibe-core is probably what breaks it.

How do I do that? I find no reference to vibe-core in dub.sdl.

nordlow:
> How do you compile? Test what the CI is doing. Building optionally with vibe-core is probably what breaks it.

The problem is that the travis.sh script fails locally in on master.

nordlow:
> How do you compile? Test what the CI is doing. Building optionally with vibe-core is probably what breaks it.

Grepping for vibe-core gives no hits. I don't understand what you mean.

Geod24:
Modified and rebased but surprisingly fails in UT...

nordlow:
Any clues why some targets are passing and some are not?

Geod24:
Older compilers don't run test, they just compile

source/dub/internal/vibecompat/inet/path.d
```diff
@@ -39,22 +36,22 @@ struct NativePath {
...
-	this(immutable(PathEntry)[] nodes, bool absolute)
+	this(immutable(PathEntry)[] nodes, bool absolute) pure nothrow
 	{
 		m_nodes = nodes;
 		m_absolute = absolute;
 	}
 
 	/// Constructs a relative path with one path entry.
-	this(PathEntry entry){
+	this(PathEntry entry) { pure nothrow
```
WebFreak001, line 54: broken attributes

source/dub/internal/vibecompat/inet/path.d
```diff
@@ -7,9 +7,6 @@
 */
 module dub.internal.vibecompat.inet.path;
 
-version (Have_vibe_core) public import vibe.core.path;
```
WebFreak001, line 10: unrelated removal?

WebFreak001 (changes requested):

source/dub/internal/vibecompat/inet/path.d
Geod24: No it's related, but as you can see in a separate commit. Otherwise we have inconsistent attributes.

Geod24:
I recommend to go with something like https://github.com/dlang/dub/pull/2821 instead.
Closing as stalled.

PR #2408, opened 2022-08-18
InitCommand Add version parsing for type arg

Tynukua:


Geod24:
What's the goal here ?

Tynukua:
> What's the goal here ?

parse version in command init
f.e.
dub init -t dagon@0.12.0

source/dub/commandline.d
```diff
@@ -960,12 +960,15 @@ class InitCommand : Command {
...
-
+		auto pkg = splitPackageName(m_templateType);
+		string[string]depVers;
+		if(pkg.version_ !is null)
+			depVers[pkg.name] = pkg.version_;
 		if (!["vibe.d", "deimos", "minimal"].canFind(m_templateType))
 		{
-			free_args ~= m_templateType;
+			free_args ~= pkg.name;
 		}
```
WebFreak001, line 970: I think this new code fits better inside the createEmptyPackage method, not needing to change anything in the InitCommand and also avoiding to move the depVers logic from `createEmptyPackage` into here.

The if block here should also be moved into there imo, modifying `free_args` because we want to pass in the custom template type as dependency seems more like a hack. The dependency version determination logic is done inside `createEmptyPackage`, so I think it's better to determine the init template dependency version there anyway.

source/dub/dub.d
```diff
@@ -1145,21 +1145,24 @@ class Dub {
...
 
-	private void runCustomInitialization(NativePath path, string type, string[] runArgs)
+	private void runCustomInitialization(NativePath path, string type, string typeVersion, string[] runArgs)
 	{
 		string packageName = type;
-		auto template_pack = m_packageManager.getBestPackage(packageName);
-		if (!template_pack) template_pack = m_packageManager.getBestPackage(packageName, "~master");
+		Package template_pack;
+		VersionRange versionRange = VersionRange.Any;
+		if (typeVersion)
```
WebFreak001, line 1160:
```suggestion
		if (typeVersion.length)
```

because typeVersion comes from the public facing API where an empty string might be passed in (which evaluates to true)

source/dub/dub.d
```diff
@@ -1113,17 +1113,17 @@ class Dub {
 			recipe_callback = Optional callback that can be used to
 				customize the recipe before it gets written.
 	*/
-	void createEmptyPackage(NativePath path, string[] deps, string type,
+	void createEmptyPackage(NativePath path, string[] deps, string[string] depVers, string type, string typeVersion,
```
WebFreak001, line 1116: I think it's a good idea to split type and typeVersion for the caller to explicitly put in a version, however this needs another overload only taking the type for backwards-compatibility purposes. (e.g. would pass `>=0.0.0` in that overload there)

WebFreak001 (changes requested):

Issue #278, opened 2014-03-12
[linux] lflags should include -rpath


Geod24:
Currently dub doesn't automatically includes the -rpath flags of the shared library it linked to.
This is an issue for several reasons:
- Security : One can link against a library in ~/bin/dmd/lib/libphobos.2, but at runtime it will only pick the library in the system path.
- Usability : One would expect the program to run after it's linked. For programs, that is less of an issue, but it's really a PITA when you do a PAM module, for example.
- Portability : To work around that, I had to put:

```
"lflags": [ "-rpath=/home/geod/bin/dmd-2.065.0/linux/lib64/" ]
```

in my dub.json. Luckily I'm the only one working on that project right now, but it's obviously a problem.

So the idea is to "just" add it to the rpath, as there shouldn't be any issue with duplicate libraries in the rpath.

PR #2413, opened 2022-08-19
Posix: Add default SONAME/install_name for dynamic libraries

kinke:
Setting it to the filename (ELF) or `@rpath/<file name>` (Mach-O).

This seems to be required in order to use the libraries copied to the output directory of dependees - either by setting the `LD_LIBRARY_PATH `env var when starting the executable, or baking a `RUNPATH` into the executable (e.g., via `patchelf --set-rpath` for deployment).

Otherwise the loader tries to load the library from the original location specified in the linking cmdline, which is some `<path to dep pkg>/.dub/…` path (and apparently potentially relative, and thus even more brittle). For ELF, I would assume that all `SONAME` + `DT_NEEDED` entries (of executable + all .so libs) would need to be post-processed to enable loading the libs from the executable's directory (well, any non-original location).

[extracted from #2396]


Geod24:
That fixes https://github.com/dlang/dub/issues/278 doesn't it ? :upside_down_face:

source/dub/compilers/utils.d
```diff
@@ -117,6 +117,18 @@ unittest {
 }
 
 
+package void addDynamicLibName(ref BuildSettings settings, in BuildPlatform platform, string fileName)
```
Geod24, line 120: Can you add a command explaining why this function is needed so that the next guy doesn't have to `git blame` ?
Don't need to do the whole `Params` thing but a simple description is useful.

test/2-dynLib-dep/dub.json
```diff
@@ -3,5 +3,7 @@
     "dependencies": {
         "dynlib-simple": { "path": "../1-dynLib-simple/" }
     },
+    "lflags-linux": ["-rpath", "$$ORIGIN"],
+    "lflags-darwin": ["-rpath", "@executable_path"],
```
Geod24, line 7: I'm not sure I understand where `$ORIGIN` comes from ?

source/dub/compilers/utils.d
kinke: Will do.

test/2-dynLib-dep/dub.json
kinke: It's a special placeholder for the executable's directory, and is added as-is as `DT_RUNPATH` to the executable.

kinke:
> That fixes https://github.com/dlang/dub/issues/278 doesn't it ?

Hmm, that seems to be about shared druntime/Phobos only? - LDC adds an implicit RUNPATH to its lib dir when linking with `-link-defaultlib-shared`; DMD AFAIK 'handles' this by setting `LD_LIBRARY_PATH` when activating the compiler.

This just paves the way for adding a RUNPATH to load the libs copied to the output dir, by making sure the executable/dynamic-lib dependee reference their dynamic lib deps by filename only, not some (relative) path used during linking, leading to linker and/or loader startup errors.

source/dub/compilers/utils.d
kinke: Added.

kinke:
As a follow-up to this, adding an implicit RUNPATH for executables with (direct or indirect) dynamic-lib deps might be nice - and setting it to the absolute path to the executable output directory. Such executables should also get `-link-defaultlib-shared` automatically with LDC (and `-defaultlib=libphobos2.so` with DMD for ELF platforms) for max convenience.

Geod24 (approved):

PR #2415, opened 2022-08-21
PackageManager: Simplify removeLocalPackage by using algorithm

Geod24:
Instead of doing (potentially) multiple append operations of two slices,
just do a filter + array which is much easier to read and prove correct.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -18,13 +18,14 @@ import dub.recipe.io;
 import configy.Exceptions;
 public import configy.Read : StrictMode;
 
-import std.algorithm : countUntil, filter, sort, canFind, remove;
+import std.algorithm : countUntil, filter, map, sort, canFind, remove;
 import std.array;
 import std.conv;
 import std.digest.sha;
 import std.encoding : sanitize;
 import std.exception;
 import std.file;
+import std.range;
 import std.string;
 import std.sumtype;
 import std.zip;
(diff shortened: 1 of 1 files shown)


thewilsonator (approved):

source/dub/packagemanager.d
```diff
@@ -713,10 +714,12 @@ class PackageManager {
...
 
 		string[Version] removed;
-		foreach_reverse( i; to_remove ) {
+		foreach (i; to_remove)
 			removed[(*packs)[i].version_] = (*packs)[i].name;
-			*packs = (*packs)[0 .. i] ~ (*packs)[i+1 .. $];
-		}
+
+		*packs = (*packs).enumerate
+			.filter!(en => to_remove.canFind(en.index))
```
WebFreak001, line 721:
```suggestion
			.filter!(en => !to_remove.canFind(en.index))
```

should this not be inverted?

Geod24: `filter` always confused me.

WebFreak001: can we add a test case for this? before the CI failed on something which didn't exactly look like a direct cause by this (only really through a lot of other things)

the whole pointer usage here makes this code kinda more complicated to read and reason about as there might be slices of the data used elsewhere in other code.

Geod24: > can we add a test case for this? before the CI failed on something which didn't exactly look like a direct cause by this (only really through a lot of other things)

I tried writing a test for it, but ended up with the same test code, because one will just do `add-local`, something idempotent (describe, list), and then `remove-local`, which is what `version-spec.sh` does. If anything, we should probably rename `version-spec.sh` because it does test a lot of things.

Geod24: > the whole pointer usage here makes this code kinda more complicated to read and reason about as there might be slices of the data used elsewhere in other code.

Agreed, will try to remove it in the future.

Geod24:
@WebFreak001 : As mentioned, the test that covered this is actually quite a good fit, can we move ahead with this ?

WebFreak001 (approved):

PR #2419, opened 2022-08-21
Deprecate Dub.getPackageSuppliers, remove one env modification from UT

Geod24:
Having UT modify environment variable defeats the point of UT.
Also, in this case, I want to move towards having a single point where the environment is read, which would make the logic much easier to follow, and can't do that with a function that reads it on every call (even if it's unlikely that someone relies on this specific behavior). Luckily the function itself is a breach of abstraction, so I think that deprecating it was the best way forward.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -163,7 +163,7 @@ class Dub {
 
 		init(m_rootPath);
 
-		m_packageSuppliers = getPackageSuppliers(additional_package_suppliers, skip_registry);
+		m_packageSuppliers = this.computePkgSuppliers(additional_package_suppliers, skip_registry);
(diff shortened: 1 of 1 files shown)


source/dub/dub.d
```diff
@@ -204,13 +205,24 @@ class Dub {
 			skip_registry = Can be used to skip using the configured package
 				suppliers, as well as the default suppliers.
 	*/
+	deprecated("This is an implementation detail. " ~
+		"Use `packageSuppliers` to get the computer list of package " ~
```
thewilsonator, line 209: `s/computer/computed`

Geod24:
Fixed both instances of the typo, rebased.

thewilsonator (approved):

PR #2280, opened 2022-06-20
Use configy library to parse dub.json file

Geod24:
Over the last few years, I've been working on a project that had growing configuration needs.
We started by manually parsing YAML files and using dyaml, then one day, when the application was sizeable, all the manual work was turned into a library.

The goal was simple: Using a declarative approach (using D native value types, so no `class`es) and minimizing YAML dependency. This project was separated in its own library, [configy](https://github.com/bosagora/configy/).

One of the major advantages of Configy is that it produces **good error messages**, even with user-defined types.
User-defined types with a `fromString`, string ctor, etc... can throw Exceptions, which the library then catches and wrap in an Exception that contains context information. It also contains a "strict mode" which will warn users if they use a key that isn't valid. I understand that the original intent of not erroring out on keys to preserve forward compatibility, but experience has shown that's causing more trouble than it's worth. Plus, we have `ToolchainRequirements` now.

PR #2422, opened 2022-08-22
Dependency: Trivial fixes around `opEquals` / `matches`

Geod24:
Was working on #2280 and found a couple minor things to improve, turns out they aren't needed for that PR but still useful.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -374,8 +374,9 @@ struct Dependency {
 		using `Dependency` as a key in hash or tree maps.
 	*/
 	bool opEquals(in Dependency o) const scope @safe {
-		return this.m_value == o.m_value
-			&& o.m_optional == m_optional && o.m_default == m_default;
+		if (o.m_optional != this.m_optional) return false;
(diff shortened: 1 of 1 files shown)


source/dub/dependency.d
```diff
@@ -374,8 +374,9 @@ struct Dependency {
 		using `Dependency` as a key in hash or tree maps.
 	*/
 	bool opEquals(in Dependency o) const scope @safe {
-		return this.m_value == o.m_value
-			&& o.m_optional == m_optional && o.m_default == m_default;
+		if (o.m_optional != this.m_optional) return false;
+		if (o.m_default	 != this.m_default)  return false;
```
WebFreak001, line 378:
```suggestion
		if (o.m_optional != this.m_optional) return false;
		if (o.m_default  != this.m_default)  return false;
```

(mixed tab inside the expression)

Geod24: Ah my `M-x tabify` issues... Fixed

thewilsonator (approved):

WebFreak001 (approved):

PR #2428, opened 2022-08-23
Support --temp-build also in `dub build`

gizmomogwai:
This allows for complete out of git builds. The link command now also
prints the resulting filename (although that might be too much for
normal usage).

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -678,7 +678,6 @@ class CommandArgs {
 	{
 		foreach (ref arg; m_recognizedArgs)
 			if (names == arg.names) {
-				assert(help_text is null);
 				*var = arg.value.get!T;
 				return;
 			}
@@ -1239,6 +1238,10 @@ class BuildCommand : GenerateCommand {
 
 	override void prepare(scope CommandArgs args)
 	{
+		args.getopt("temp-build", &this.baseSettings.tempBuild, [
+			"Builds the project in the temp folder if possible."
+		]);
+
 		args.getopt("rdmd", &this.baseSettings.rdmd, [
 			"Use rdmd instead of directly invoking the compiler"
 		]);
@@ -1344,10 +1347,6 @@ class RunCommand : BuildCommand {
 
 	override void prepare(scope CommandArgs args)
 	{
-		args.getopt("temp-build", &this.baseSettings.tempBuild, [
-			"Builds the project in the temp folder if possible."
-		]);
-
 		super.prepare(args);
 		this.baseSettings.run = true;
 	}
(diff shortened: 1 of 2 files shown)


source/dub/commandline.d
```diff
@@ -678,7 +678,6 @@ class CommandArgs {
 	{
 		foreach (ref arg; m_recognizedArgs)
 			if (names == arg.names) {
-				assert(help_text is null);
```
WebFreak001, line 681: don't remove this assert, it checks if there are duplicate recognized args, and if there are the duplicates must not change the help text.

Instead you want to remove the `temp-build` getopt from RunCommand, as you now added it in BuildCommand, which RunCommand inherits from

WebFreak001 (changes requested):

source/dub/commandline.d
gizmomogwai: thanks for the explanation. i reworked the commits and added a perhaps clearer message to the error.

source/dub/commandline.d
```diff
@@ -678,7 +678,7 @@ class CommandArgs {
...
+				enforce(help_text is null, format!("duplicated argument '%s' must not change helptext, consider to remove the duplication")(names));
```
WebFreak001, line 681: I think this is a good addition now, but it should stay an assert (it's a programming bug, not a runtime exception that may occur)

gizmomogwai: true ...

WebFreak001:
```suggestion
				assert(help_text is null, format!("duplicated argument '%s' must not change helptext, consider to remove the duplication")(names));
```

source/dub/generators/build.d
```diff
@@ -579,7 +579,7 @@ class BuildGenerator : ProjectGenerator {
 			settings.compiler.invoke(buildsettings, settings.platform, settings.compileCallback);
 
 			if (generate_binary) {
-				logInfo("Linking", Color.light_green, "%s", buildsettings.targetName.color(Mode.bold));
+				logInfo("Linking", Color.light_green, "%s => %s", buildsettings.targetName.color(Mode.bold), buildsettings.getTargetPath(settings));
```
WebFreak001, line 582: what do others think about this CLI output change? @Geod24 @thewilsonator ?

The output path is for example `/tmp/.dub/build/issue2051-test-application-unittest-linux.posix-x86_64-dmd_v2.100.0-0AC041139053DBA3B189AED31C7F8A27221CA401DEE41BA7B05C3F5F74C4D4BE/issue2051-test-application`, possibly even longer for longer project folder names in local temp cache.

I don't think this is really useful most of the time, should we only show this when having the --temp-build option used?

gizmomogwai: yes ... that would be my preference as well! normally its not really needed!

WebFreak001 (approved):

PR #2430, opened 2022-08-24
Integrate `dependencyBuildSettings` into dependencies

Geod24:
Because those are specified in `dependencies`,
they should be stored alongside them.
Unfortunately this will slihtly increase memory usage in the common case,
but unlikely to a point where it would be problematic.

This is technically a breaking change as a field is getting removed,
however this struct has been manipulated a few times in breaking ways,
and it's unlikely `dependencyBuildSettings` is being used outside of dub,
given how recent it is and how narrow its usage is.
However, as `dependencies` is likely to be widely used, a compatibility
type and an `alias this` are provided not to break client code.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -599,7 +599,7 @@ class ProjectGenerator
 		}
 
 		// apply both top level and configuration level forced dependency build settings
-		void applyDependencyBuildSettings (const BuildSettingsTemplate[string] configured_dbs)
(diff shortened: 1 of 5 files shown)


Geod24:
CC @omerfirmak ;)

WebFreak001 (changes requested):
overall this is a nice improvement, can you make the RecipeDependency struct public? (or why don't we just put BuildSettingsTemplate inside Dependency?)

Geod24:
Done

WebFreak001 (approved):

PR #2432, opened 2022-08-24
Rename PlacementLocation & co to 'Scope'

Geod24:
Because the name is overly verbose.
While looking for a better name, I come across the use of the 'Scope'
term in NPM documentation, which seems to be a good fit, even if not
specific to packages.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1300,8 +1300,8 @@ enum ListBuildSettingsFormat {
 	commandLineNul, /// NUL character separated list entries (unescaped, data lists separated by two NUL characters)
 }
 
-deprecated("Use `dub.packagemanager : PlacementLocation` instead")
-public alias PlacementLocation = dub.packagemanager.PlacementLocation;
+deprecated("Use `dub.packagemanager : Scope` instead")
+public alias PlacementLocation = dub.packagemanager.Scope;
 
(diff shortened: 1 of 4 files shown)


source/dub/project.d
```diff
@@ -1300,8 +1300,8 @@ enum ListBuildSettingsFormat {
 	commandLineNul, /// NUL character separated list entries (unescaped, data lists separated by two NUL characters)
 }
 
-deprecated("Use `dub.packagemanager : PlacementLocation` instead")
-public alias PlacementLocation = dub.packagemanager.PlacementLocation;
+deprecated("Use `dub.packagemanager : Scope` instead")
+public alias PlacementLocation = dub.packagemanager.Scope;
```
Geod24, line 1304: This hasn't been released yet hence the simple rename

WebFreak001 (changes requested):
not a fan, Scope is more abstract and tells the reader less about what it is really (what kind of scope is it?) - PlacementLocation is much more concrete and at least conveys some information about what it is.

If you really want to change the name we could go for something like `PlacementScope`, but just `Scope` is a no-go from me. Also in the NPM docs, at least on https://docs.npmjs.com/about-scopes the NPM scopes mean a whole different thing (basically unique prefixes that you can use to manage access and use already registered global package names with)

Geod24:
Yeah, let's leave it as is.

PR #2433, opened 2022-08-24
`PackageManager`: Remove double duty as a `Location`

Geod24:
`PackageManager` currently has a double duty, as a `Location` manager where each `Location`, or `Scope` as I'd like to call them, has a certain priority, but also it does some of the job that a `Location` does, by keeping track of two (!!) lists of packages, and having a `searchPaths`. The two lists, as I understand it, are motivated by the special case of bare mode, which doesn't need to be a special case.

Ideally, I'd like to move things so that `m_disableDefaultSearchPaths` doesn't exists, and instead we have an array of `Location` that is 1 or 4 entries long. Then many operations will just iterate over that array and be done with it. However that might not work out to well if there is a need to expose an API with `PlacementLocation`, but let's see.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -48,10 +48,16 @@ public enum PlacementLocation {
 class PackageManager {
 	private {
 		Location[] m_repositories;
-		/// The extra 'internal' location, used by the PackageManager
-		/// to store path and repository-based dependencies or in bare mode.
+		/**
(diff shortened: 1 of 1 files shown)


Geod24:
Okay, this finally works. I can't move `add` to the location yet, because we now have two arrays. Working on that...

thewilsonator (approved):

PR #2446, opened 2022-08-30
Revert "Temporarily disable ldc-master as the dustmite test is broken"

Geod24:
This partially reverts commit b1f3b0333d762027a4533d5389487797abec88c1.
So that it doesn't get forgotten.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,7 +33,7 @@ jobs:
           - dmd-latest
           - ldc-latest
           - dmd-master
-#          - ldc-master
+          - ldc-master
 
     runs-on: ${{ matrix.os }}
     steps:


Geod24:
Rebased, still an issue as of today (CC @CyberShadow if you have some bandwidth).

CyberShadow:
ldc-master seems to fail to compile the test program:

https://github.com/dlang/dub/blob/master/test/dustmite-no-redirect-test/project/source/app.d

Logs say:

```
dustmite-no-redirect-test/source/app.d(3,1): Error: declaration expected, not `{`
dustmite-no-redirect-test/source/app.d(5,1): Error: unmatched closing brace
```

"Initial test fails" means that DustMite did not even try to modify the program. However, it still reads it in, parses, and writes it back out before invoking the compiler, so it's possible that DustMite is miscompiled and produces a garbled program.

Any advice on reproducing this locally?

Geod24:
We can reproduce with:
```shell
~/dlang/install.sh install ldc-latest-ci
source ~/dlang/ldc-b0eab59b/activate
cd $DUB_REPO
cd test/
DUB=dub CURR_DIR=$(pwd) ./dustmite-no-redirect.sh
```

CyberShadow:
> We can reproduce with:

Thanks, that helps a lot! I reproduced it too.

Filed an LDC bug: https://github.com/ldc-developers/ldc/issues/4134

thewilsonator (approved):

Geod24:
Thanks everyone !

PR #2461, opened 2022-09-06
Deprecate overrides: local-overrides.json, add-override, remove-override

Geod24:
As explained in the changelog, overrides were a solution to a system
with version-only dependencies, which is long gone.

### DIFF
--- a/changelog/overrides_deprecated.dd
+++ b/changelog/overrides_deprecated.dd
@@ -1,0 +1,9 @@
+The override system is deprecated
+
+Dub had an "override" system, allowing a specific version or version range
+to be overriden by a specific package.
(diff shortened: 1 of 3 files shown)


Geod24:
This was suggested by @jacob-carlborg as part of DUB v2.0: https://github.com/dlang/dub/discussions/2069#discussioncomment-254373

@s-ludwig : Any opinion on this ? I think the approach we should recommend is:
- For a single package, edit `dub.selections.json` or add the package to `.dub/packages/`
- For a system-wide override, change the search path;

WebFreak001 (approved):

Geod24:
@s-ludwig : Any opinion on this ?

s-ludwig (approved):

Issue #2345, opened 2022-07-28
MacOS 11/dmd-master: Spurious CI failure on issue1416-maven-repo-pkg-supplier.sh


Geod24:
```
[INFO] Running /Users/runner/work/dub/dub/test/issue1416-maven-repo-pkg-supplier.sh...
[...]
Trying to download maven-dubpackage (1.0.5)
Package maven-dubpackage not found for maven repository at http://localhost:10936/maven/release/dubpackages: Failed to download http://localhost:10936/maven/release/dubpackages/maven-dubpackage/maven-metadata.xml
No package maven-dubpackage was found matching the dependency 1.0.5
Error:  :17 command failed
[ERROR] Script failure.
```

Only saw it on dmd-master + macOS 11 so far. Retrying usually make it go away.

Geod24:
Don't see this happening anymore.

PR #2481, opened 2022-09-15
Master regression fixes

s-ludwig:
- Fixes the scanning order of packages, so that add-path packages have precedence over cached packages
- Fixes compilation on LDC 1.30.0

Why was it decided to drop all compiler support earlier than "ldc-latest" and "dmd-latest" (which doesn't explain why it didn't even compile on ldc-latest)? Having a buffer range of a few compiler versions greatly reduces friction during version transitions, not only in the build environment for dub itself, but also for dependencies of dub that might move slower. And IMHO using the latest shiny features, such as throw expressions in an infrastructure project like this is just utterly unnecessary. I think this really was a bad decision and should be revised, if possible. BTW, compiling with DMD 1.29.0 on Windows still produces backend errors due to some SumType.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -1138,10 +1138,11 @@ private struct Location {
 			// loaded by the user (e.g. the project and its subpackages),
 			// so don't clean it.
 			this.fromPath = null;
-			this.scanPackageFolder(this.packagePath, mgr, existing);
 		}
 		foreach (path; this.searchPath)
 			this.scanPackageFolder(path, mgr, existing);
+		if (this.packagePath !is NativePath.init)
+			this.scanPackageFolder(this.packagePath, mgr, existing);
 	}
 
     /**
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -785,7 +785,7 @@ class Dub {
 	{
 		const vrange = dep.visit!(
 			(VersionRange range) => range,
-			(any)                => throw new Exception("Cannot call `dub.fetch` with a " ~ typeof(any).stringof ~ " dependency"),
+			function VersionRange (any) { throw new Exception("Cannot call `dub.fetch` with a " ~ typeof(any).stringof ~ " dependency"); }
 		);
 		return this.fetch(packageId, vrange, location, options, reason);
 	}


Geod24:
Hum that was done in https://github.com/dlang/dub/commit/3c1c005ffa005acaa9173efe311be4b5e4464ba9 and I agree it wasn't a great idea. I'd be very happy with supporting more versions. The usage of the `throw` expression is, I think, more accidental than something I wanted absolutely (IOW: I wrote the code, it compiled).

source/dub/packagemanager.d
```diff
@@ -1138,10 +1138,11 @@ private struct Location {
 			// loaded by the user (e.g. the project and its subpackages),
 			// so don't clean it.
 			this.fromPath = null;
-			this.scanPackageFolder(this.packagePath, mgr, existing);
 		}
 		foreach (path; this.searchPath)
 			this.scanPackageFolder(path, mgr, existing);
+		if (this.packagePath !is NativePath.init)
+			this.scanPackageFolder(this.packagePath, mgr, existing);
```
Geod24, line 1145: The scan order shouldn't matter, the iteration order is done in `getPackageIterator`, which IIUC is where you want to shift things.

s-ludwig: The problem is that both, search path packages and package path packages end up in the same array (`fromPath`) and the order within this array matters.

s-ludwig:
Regarding the compile error, I'm not sure what exactly happens here, it obviously also works for the CI jobs, but not with the Windows version of LDC that I have installed (from the standard distribution ZIP file).

s-ludwig:
> Warning Package maven-dubpackage not found for maven repository at http://localhost:13733/maven/release/dubpackages: Failed to download http://localhost:13733/maven/release/dubpackages/maven-dubpackage/maven-metadata.xml

Not sure what to do about this one, doesn't seem to be related to the changes.

Geod24:
There's a spurious failure for this: https://github.com/dlang/dub/issues/2345

s-ludwig:
Okay, all tests have passed after a re-run.

Geod24 (approved):
LGTM but the `PackageManager` change could use a test to prevent future regression.

PR #2484, opened 2022-09-16
Remove commandline arguments that do not work for lint (dscanner)

### LINKED ISSUE #1940: dub lint fails when --config argument is used for both dub and dscanner
### System information
- **dub version**: `DUB version 1.20.1, built on Apr 17 2020`
- **OS Platform and distribution**: Tried on Windows 10 and Ubuntu
- **compiler version**: Not relevant

### Bug Description

When using `dub lint --config <path to dub config file> -- --config <path to dscanner config file>`,  `dscanner` is called with both dub config and dscanner config arguments, and results in `dscanner` usage/help message being displayed (i.e: equivalent to `dscanner --help`).

Could that be caused by the fact that both argument have the same name `--config`?

### How to reproduce?

First create a dub project and a `dscanner.ini` file, just to be sure.
```
$ dub init
$ touch dscanner.ini
```

Then run `dub lint` with config arguments used twice, once for `dub` and once for `dscanner`.
```powershell
$ dub lint --config .\dub.sdl -- --config .\dscanner.ini
Running pre-generate commands for dscanner...
# ...more logs...
Running ..\..\..\..\AppData\Local\dub\packages\dscanner-0.9.0\dscanner\bin\dscanner.exe --config .\dub.sdl --config .\dscanner.ini C:\Users\Sam\Development\D\action-dscanner\testdata\

    Usage: ..\..\..\..\AppData\Local\dub\packages\dscanner-0.9.0\dscanner\bin\dscanner.exe <options>

Options:
    --help, -h
        Prints this help message
# ...rest of the usage/help message...
Program exited with code 1
```

Notice how `dscanner.exe` is called with arguments `--config .\dub.sdl --config .\dscanner.ini`.

### Expected Behavior

I would instead expect `dub.exe` to be called with the argument `--config .\dub.sdl` and `dscanner.exe` to be called with the argument `--config .\dscanner.ini`.

gizmomogwai:
`dub lint` (which in turn just calls dscanner) does not build and does not look into dub.{sdl|json}. So it does not need the parameters that a regular build requires.

The helptext now hints, that the config is a dscanner config (in contrast to a dub config).

Fixes #1940

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1499,8 +1499,8 @@ class LintCommand : PackageBuildCommand {
 			"Import paths"
 		]);
 
-		args.getopt("config", &m_config, [
-			"Use the given configuration file."
+		args.getopt("dscanner-config", &m_config, [
+			"Use the given d-scanner configuration file."
 		]);
 
 		super.prepare(args);


thewilsonator (approved):

Geod24:
I would rather we change `config` to `--dscanner-config`. Yes currently `dub lint` is very limited, but we'd like to keep the ability to extend it to actually draw information from `dub.sdl` / `dub.json`.

gizmomogwai:
That's would also been an idea of mine, but the current help text is completely misleading.   I can prepare a patch as you suggested, keeping the defunct options in place and just renaming the scanner config option

Geod24 (approved):

WebFreak001:
no deprecation period?

Geod24:
Conceptually, this is just adding a new option. `c|config` is still recognized, but now it means the same thing as everywhere else. To me that's a bug fix.

gizmomogwai:
> no deprecation period?

Good point:
If someone is using `dub lint --config` at the moment, then the new version will behave differently (unfortunately).
ATM --config selects the dscanner config, in the future with this patch it would be ignored, as the dub config is not taken in by dscanner.

PR #2485, opened 2022-09-16
Use a typed PackageRecipe for packagesuppliers

skoppe:
Instead of dealing with raw json, it is better to parse it at first sight.


source/dub/packagemanager.d
```diff
@@ -508,6 +508,14 @@ class PackageManager {
 		throw new Exception(format("No override exists for %s %s", package_, version_spec));
 	}
 
+	/// Extracts the package supplied as a path to it's zip file to the
+	/// destination and sets a version field in the package description.
+	Package storeFetchedPackage(NativePath zip_file_path, PackageRecipe recipe, NativePath destination)
+	{
+		import dub.recipe.json : toJson;
+		return storeFetchedPackage(zip_file_path, recipe.toJson, destination);
+	}
```
Geod24, line 517: Remove this, it's been deprecated

Geod24:
Yes! I've been wanting this for a while, but the problem is that it's a breaking change for library users.
Instead adding a `final` function that forwards to the implementation might do the trick ?

skoppe:
> but the problem is that it's a breaking change for library users.

Ah, you mean people who inherit from `AbstractFallbackPackageSupplier`.

Hmm yes, likely not many, but still.

Geod24:
> Ah, you mean people who inherit from `AbstractFallbackPackageSupplier`.
> 
> Hmm yes, likely not many, but still.

Actually, changing the return type of `fetchPackageRecipe` means anyone implementing the `PackageSupplier` interface.
Still, likely not many, but there aren't many users of Dub as a library. Adding a `final` function that does the job would help in the short term, and we can just rework this interface in a future major version.

source/dub/dub.d
```diff
@@ -1732,9 +1733,9 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 			if (rootpack == name) {
 				try {
 					auto desc = ps.fetchPackageRecipe(name, dep, prerelease);
-					if (desc.type == Json.Type.null_)
+					if (desc.isNull)
 						continue;
-					auto ret = new Package(desc);
+					auto ret = new Package(desc.get);
```
Geod24, line 1738: I think that was also the last use of the `Package` constructor that takes a JSON, so feel free to deprecate it.

source/dub/packagesuppliers/filesystem.d
```diff
@@ -59,15 +62,14 @@ class FileSystemPackageSupplier : PackageSupplier {
...
 		auto basename = filePath.head.name;
 		enforce(basename.endsWith(".zip"), "Malformed package filename: " ~ filePath.toNativeString);
 		enforce(basename.startsWith(packageId), "Malformed package filename: " ~ filePath.toNativeString);
-		json["version"] = basename[packageId.length + 1 .. $-4];
-		return json;
+
+		string packageFileName;
+		string packageFileContent = packageInfoFileFromZip(filePath, packageFileName);
+		return typeof(return)(parsePackageRecipe(packageFileContent, packageFileName));
+		// recipe.version_ = basename[packageId.length + 1 .. $-4];
```
Geod24, line 72: Setting the `version_` is important, why was it removed ?

source/dub/packagesuppliers/maven.d
```diff
@@ -36,11 +38,11 @@ class MavenRegistryPackageSupplier : PackageSupplier {
 	{
 		import std.algorithm.sorting : sort;
 		auto md = getMetadata(package_id);
-		if (md.type == Json.Type.null_)
-			return null;
+		if (md.isNull)
+			return [];
```
Geod24, line 42: `null` is better IMO

Geod24:
@skoppe : If you can find a small change to make (e.g. the `[]` => `null` changes) and PR it separately, that would whitelist you for future CI runs of Github, otherwise you'll need someone to keep on approving it every time you update this PR.

source/dub/packagesuppliers/maven.d
```diff
@@ -114,18 +116,19 @@ class MavenRegistryPackageSupplier : PackageSupplier {
...
+		return typeof(return)(entry.data);
 	}
 
 	SearchResult[] searchPackages(string query)
 	{
 		// Only exact search is supported
 		// This enables retrival of dub packages on dub run
 		auto md = getMetadata(query);
-		if (md.type == Json.Type.null_)
+		auto recipe = getBestPackage(md, query, Dependency.any, true);
```
WebFreak001, line 129: why remove the check if metadata is null? logically would want to have it everywhere before testing getBestPackage, maybe we can merge the two functions together?

WebFreak001: nvm it's done in the getBestPackage function, however there are a few other cases in the code where it's still checked against before getBestPackage.

source/dub/packagesuppliers/filesystem.d
WebFreak001: seems like we don't have any `version_` field inside PackageRecipe

skoppe:
@Geod24 I don't quite understand how `final` is going to help, maybe I am missing something.

The way I see it is that I can't just change the `PackageSupplier` interface, since it is public, and people might have inherited from it.

This means we need to keep the `Json fetchPackageRecipe(string packageId, Dependency dep, bool pre_release)` function for a while, and deprecate it.

Additionally I can't just add a `Nullable!PackageRecipe fetchPackageRecipe(string packageId, Dependency dep, bool pre_release)` function, without providing a base implementation. Not to mention that its name conflicts with the deprecated one.

---

I think I need to add a `fetchRecipe` function with a base implementation that just calls the original `fetchPackageRecipe`, while deprecating the latter.

Using that I can implement both `fetchRecipe` and `fetchPackageRecipe` in the current package suppliers.

source/dub/packagesuppliers/filesystem.d
Geod24: Last I checked we did. It's used when packages are downloaded from the registry and stored locally. That's how `dub` knows what version a package is at (it doesn't rely on the path).

Geod24:
> I think I need to add a [...]

Yes, that's what I was recommending. `final` because "function body only allowed in `final` functions in interface".

PR #2328, opened 2022-07-19
Make 'bare' option honor the 'root' option & deprecate changing `rootPath`

Geod24:
As explained in the first commit, having `--root` behave any different than `make -C` or `git -C` is IMO a mistake.

PR #2487, opened 2022-09-19
Dub: Deprecate changing the rootPath

Geod24:
It's an essential part of the setup, and too many properties are derived from it that we don't want to let the library user the ability to mess with it.


Geod24:
Extracted from https://github.com/dlang/dub/pull/2328

thewilsonator (approved):

John-Colvin:
Very happy to see someone tackling path madness in dub.

PR #2492, opened 2022-10-15
Return null instead of empty array

skoppe:
Silly change to get my account verified


Geod24:
Damn it I broke the nightlies again...

WebFreak001 (approved):
code change looks reasonable, but how does this help you verify your account?

WebFreak001:
that doesn't answer my question though :p

Geod24:
Because I approved it.

Geod24:
> that doesn't answer my question though

@skoppe has another PR which is substantial. Everytime he does a change and push, CI doesn't run, because he doesn't have a commit in master yet. Now he has a commit in master, so CI will run.

WebFreak001:
oh right that makes sense then, I thought this had something to do with meson package search somewhere explicitly requiring null, which would have been weird.

Issue #2391, opened 2022-08-15
dub master cannot be built by older compilers


ibuclaw:
The stable branch can be built by compilers based on 2.076.x (i.e: gdc-11).  But master fails to build with anything younger than 2.099.x.

dmd 2.098:
```
source/dub/dub.d(746): Error: expression expected, not `throw`
source/dub/dub.d(746): Error: found `new` when expecting `)` following template argument list
source/dub/dub.d(746): Error: semicolon expected following auto declaration, not `Exception`
source/dub/dub.d(747): Error: expression expected, not `)`
```

Geod24:
I don't think we should support v2.076, but we should support the last 10 versions. That looks like a rather simple fix.

Geod24:
It's not a simple fix. Since we are using `std.sumtype`, which is in Phobos only since v2.097.0, it creates a lot of issues.
Using `dyaml.stdsumtype` leads to other problems.

maxhaton:
Hack it with `-mv`?

Geod24:
> Using `dyaml.stdsumtype` leads to other problems.

`-mv` is never the answer. And as I mentioned:

> Using `dyaml.stdsumtype` leads to other problems.

SingingBush:
2.097 and above is still better than 2.099 and above. Probably worth just going with that and putting

```
frontend=">=2.097"
```

In the dub.sdl

ibuclaw:
DMD nightly builds are the latest victim of this issue.

https://github.com/dlang/dmd/actions/workflows/nightlies.yml

PR #2494, opened 2022-10-17
Restore compatibility with v2.093

Geod24:
~Fixes https://github.com/dlang/dub/issues/2391~
CC @ibuclaw

### DIFF
--- a/source/dyaml/stdsumtype.d
+++ b/source/dyaml/stdsumtype.d
@@ -392,38 +392,45 @@ public:
             tag = tid;
         }
 
-        static if (isCopyable!(const(T)))
+        // DUB: Those traits compile work around bugs in < v2.098
+        static if (!__traits(compiles, { T c = const(T).init; }))
         {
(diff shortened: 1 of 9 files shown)


source/dyaml/stdsumtype.d
```diff
@@ -251,6 +251,12 @@ private template DeducedParameterType(T)
         alias DeducedParameterType = T;
 }
 
+/// Compatibility with < v2.095.0
+private struct __InoutWorkaroundStruct{}
+private @property T rvalueOf(T)(inout __InoutWorkaroundStruct = __InoutWorkaroundStruct.init);
+private @property ref T lvalueOf(T)(inout __InoutWorkaroundStruct = __InoutWorkaroundStruct.init);
+private enum isRvalueAssignable(Lhs, Rhs = Lhs) = __traits(compiles, { lvalueOf!Lhs = rvalueOf!Rhs; });
```
ibuclaw, line 258: Should be wrapped in a static if condition?
```
static if (__VERSION__ < 2095)
```

Geod24:
We should be G2G.

PR #2496, opened 2022-10-18
Don't redundantly call generateProject for DDOX build

Geod24:
It's already being called from runDdox.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1506,8 +1506,10 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.buildDeep = m_deep;
 
 		logDiagnostic("Generating using %s", m_generator);
-		dub.generateProject(m_generator, gensettings);
-		if (this.baseSettings.buildType == "ddox") dub.runDdox(gensettings.run, app_args);
+		if (this.baseSettings.buildType == "ddox")
+			dub.runDdox(gensettings.run, app_args);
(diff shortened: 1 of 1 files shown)


source/dub/commandline.d
```diff
@@ -1209,8 +1209,10 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.runArgs = app_args;
 
 		logDiagnostic("Generating using %s", m_generator);
-		dub.generateProject(m_generator, gensettings);
-		if (this.baseSettings.buildType == "ddox") dub.runDdox(gensettings.run, app_args);
+		if (this.baseSettings.buildType == "ddox")
```
WebFreak001, line 1212: the log line above may not be accurate anymore if we don't use m_generator, no?

Also it looks like m_generator is checked if it exists and set to the first free arg of the command line, so the user probably expects that generator to run.

The only time it looks like runDdox calls generateProject the same was as here is when calling `dub generate --build=ddox build` - I think we should probably check that m_generator == "build" in this if?

WebFreak001 (changes requested):

source/dub/commandline.d
Geod24: Since the CI fails it looks like I was wrong. I'll look more into it soon-ish.

Geod24:
This was incorrect.

PR #2499, opened 2022-10-19
Document PackageManager.getBestPackage, deprecate Dependency overload

Geod24:
The Dependency overload is not needed, as calling getBestPackage with an SCM / path dependency does not make sense.


source/dub/dub.d
```diff
@@ -1709,8 +1709,6 @@ private class DependencyVersionResolver : DependencyResolver!(Dependency, Depend
 					logDiagnostic("Sub package %s doesn't exist in %s %s.", name, basename, dep.version_);
 					return null;
 				}
-			} else if (auto ret = m_dub.m_packageManager.getBestPackage(name, dep)) {
-				return ret;
```
Geod24, line 1713: I think this is unreachable. Let's see what the CI says.

thewilsonator (approved):

PR #2502, opened 2022-10-20
Fix #1752: Use 'no' as string representation of Dependency.Invalid

### LINKED ISSUE #1752: docs says toolchainRequirements' values can be "no", but causes issue for dependent projects
in `spasm:0.2.0-beta.1` I use `toolchainRequirements dmd="no" gdc="no" ldc=">=1.15.0"`, which works great for local development, but when I create a project that depends on it it gives me errors about invalid SemVer after downloading the spasm package.

When I look into the downloaded `.dub/packages/spasm` folder I see a `dub.json` where the values of the toolchainRequirements say `invalid` instead of the `no` I wrote earlier.

Is `no` a valid value?

Geod24:
Pretty simple and straightforward fix. No test because whatever idea I have to test it will probably break soon, as I'm working on not having the files rewritten in JSON. But tested locally, works.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -1031,7 +1031,7 @@ public struct VersionRange
 
 		string r;
 
-		if (this == Invalid) return "invalid";
+		if (this == Invalid) return "no";
 		if (this.isExactVersion() && m_inclusiveA && m_inclusiveB) {
 			// Special "==" case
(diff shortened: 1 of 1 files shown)


Geod24:
Targeting stable so it makes it to the next release.

PR #2506, opened 2022-10-27
Fix #2425: Do not select path / repository based dependency

### LINKED ISSUE #2425: `dub.selections.json`: We should not store info about non-optional path/repository dependency
Currently, we store information about non-version dependencies in `dub.selections.json`.
I think we should not, unless it's to mention the dependency is selected (although I'm a bit fuzzy on the usefulness of such a case).
Instead, we should consider `dub.{sdl,json}` with a repo / path dependency to be an implicit `dub.selections.json` mention.
This would avoid the risk of inconsistency, and make pure path-based dependencies easier. It would probably also avoid a few minor issues we're seeing about download.

I had to work around this behavior in the past, e.g. in https://github.com/bosagora/agora/blob/v0.x.x/dub.selections.json

Geod24:
Draft for now, just seeing what the CI says. And yes this needs tests.


WebFreak001:
aren't path based dependencies in dub.json ignored if they are found with a version inside dub.selections.json? That would be a complete breaking change.

Geod24:
Yes, they are, and to me that's a bug.

source/dub/project.d
```diff
@@ -1756,6 +1756,7 @@ final class SelectedVersions {
 	}
 
 	/// Selects a certain path for a specific package.
+	deprecated("Selecting path dependency is deprecated")
```
WebFreak001, line 1759: should these be deprecated? If we add `dub select [dependency] [path]`, it would call this function

Geod24: It would only make sense if the dependency was `optional`, correct ?

WebFreak001: also if you locally develop with a different test version you have locally checked out in your project, but don't want to register it globally across all your projects

Geod24: I don't follow ? If you develop locally, you just change your `dub.json`, don't you ?

WebFreak001: you are right, it's probably better to put it inside `dub.json`/`dub.sdl` then

Issue #1558, opened 2018-09-07
Configuration "library-nonet" is broken


0xEAB:
### System information

- **dub version**: v1.11.0
- **OS Platform and distribution**: GNU/Linux 4.18, Ubuntu 18.04
- **compiler version:** DMD64 D Compiler v2.081.2

### Bug Description
```
dub 1.11.0: building configuration "library-nonet"...
../../.dub/packages/dub-1.11.0/dub/source/dub/packagesuppliers/maven.d(12,9): Error: module `dub.internal.utils` import HTTPStatusException not found
../../.dub/packages/dub-1.11.0/dub/source/dub/packagesuppliers/registry.d(14,9): Error: module `dub.internal.utils` import HTTPStatusException not found
dmd failed with exit code 1.
```
or
```
dub 1.11.0: building configuration "dynamic-library-nonet"...
../../.dub/packages/dub-1.11.0/dub/source/dub/packagesuppliers/maven.d(12,9): Error: module `dub.internal.utils` import HTTPStatusException not found
../../.dub/packages/dub-1.11.0/dub/source/dub/packagesuppliers/registry.d(14,9): Error: module `dub.internal.utils` import HTTPStatusException not found
/usr/bin/dmd failed with exit code 1.
```

### How to reproduce?
```json
    "dependencies": {
        "dub": "~>1.11.0"
    },
     "subConfigurations": {
         "dub": "library-nonet"
     }
```

or
```json
    "dependencies": {
        "dub": "~>1.11.0"
    },
     "subConfigurations": {
         "dub": "dynamic-library-nonet"
     }
```

### Expected Behavior
The said configuration can be used without breaking the package.

0xEAB:
### Workaround
```json
"subConfigurations": {
        "dub": "library"
     }
```

andre2007:
@s-ludwig I am not really sure how to solve this. 
HTTPStatusException  is either introduced by version DubUseCurl or Have_vibe_d_http.

Is library-nonet still relevant? What should be the proposed solution?
- HTTPStatusException is also always introduced 
- Package supplier maven & registry are not available for library-nonet
- Coding using HTTPStatusException is deactivated if library-nonet is used
- Remove library-nonet
- ...

andre2007:
I tried it with dub-docs (https://github.com/dlang/dub-docs) which also uses "library-nonet" and set dub version to "~master". There is no syntax error.
Is the issue still reproducible for you with master?

[...]

PR #2509, opened 2022-10-27
Remove Vibe.d HTTP support / library-nonet

Geod24:
library-nonet has been broken for years, and is not used. Most importantly, it has no support for timeout.
Lastly, the likely reason it was originally introduced was to avoid a hard dependency on cURL, something that has been fixed in Phobos directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ __dummy.html
 
 # Ignore build files.
 /bin/dub
-/bin/__test__library-nonet__
 /bin/__test__library__
 /bin/dub-test-library
 /bin/libdub.a
--- a/dub.sdl
+++ b/dub.sdl
@@ -11,7 +11,7 @@ configuration "application" {
 	targetType "executable"
 	mainSourceFile "source/app.d"
 	libs "curl"
-	versions "DubUseCurl" "DubApplication"
+	versions "DubApplication"
 	// Uncomment to get rich output about the file parsing and json <-> YAML
 	// integrity checks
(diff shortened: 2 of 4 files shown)


Geod24:
Would fix:
- https://github.com/dlang/dub/issues/1923 (asking for application-nonet because of bad coverage)
- https://github.com/dlang/dub/issues/1558

WebFreak001 (approved):
ok it has been broken anyway, although we would probably like to use some library like `requests` instead of curl in the future, which abstracts this away. (and doesn't use curl in the first place)

s-ludwig:
https://github.com/dlang/dub-registry/blob/master/dub.sdl#L13

Geod24:
Will fix.

s-ludwig:
Just to be clear, this code path is required for the registry to work properly. Using curl or some other library will cause the main thread to hang during requests and consequently will make the service highly unresponsive.

Also, removing a build configuration is a hard breaking change. Honestly, I think we should uphold a higher standard w.r.t. backwards compatibility. The gain of such a clean-up is not nearly making up for the cost of even a single person or project depending on this (which we don't know of course, apart from dub-registry).

WebFreak001:
I did not see that that was the case and thought the configuration was completely broken, thus non-usable, since #1558 was open and referenced here.

@Geod24 should we revert this PR? imo until we have a better replacement for curl here (that uses the vibe sockets, like the requests library) we should not remove that functionality if it actually works.

PR #2511, opened 2022-10-28
Correctly handle UTF-32 / UTF-16 files

Geod24:
Our previous implementation was simply "strip some BOM". This re-uses DMD battle-tested implementation to re-encode files in UTF-8 if needed.


source/dub/internal/utf.d
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * Params:
+ *  src = Content of the text file to process
+ *  assumeASCII = Assume the the first char must be ASCII (true by default)
+ *
+ * Returns:
+ *   UTF-8 encoded variant of `src`, stripped of any BOM,
+ *   or `null` if an error happened.
+ */
+package(dub) string processText (immutable(ubyte)[] src, bool assumeASCII = true)
```
WebFreak001, line 29: why do we have our own implementation / copy-paste from dmd instead of using std.utf or std.encoding from phobos?

Geod24: `std.utf` doesn't deal with BOM, but I didn't think of `std.encoding`. I'll see if I can use it.

PR #2515, opened 2022-10-28
Do not fall back to Vibe.d in vibecompat

Geod24:
This remove the fallback to Vibe.d, as there is no plan in sight to remove the compatibility layer, hence we always end up depending on something out of date, or rely on a contributor upgrading the code. In other words, having the fallback complicates development.

As a result, this removes:
- The outdated, now empty `dub.selections.json`;
- The code in `Have` versions;
- Some old methods that import vibe and couldn't possibly compile;


s-ludwig:
I think we should rather integrate this with CI for now, for example by building dub-registry. At least dub-registry depends on this (e.g. assuming that the returned `Json` in some places is `vibe.data.json.Json`). Generally, this will be a hard breaking API change for any type that is used in the API.

PR #2517, opened 2022-10-28
Unify reading and writing files under a single API

Geod24:
This removes the range API and instead uses Appender, while using a limited and centralized set of functions to read/write from files.
The end goal is to permit dependency injection,
which is a pre-requisite to extending the test-suite.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -110,7 +110,7 @@ bool isWritableDir(NativePath p, bool create_if_missing = false)
 	import std.random;
 	auto fname = p ~ format("__dub_write_test_%08X", uniform(0, uint.max));
 	if (create_if_missing && !exists(p.toNativeString())) mkdirRecurse(p.toNativeString());
-	try openFile(fname, FileMode.createTrunc).close();
+	try writeFile(fname, "Canary");
 	catch (Exception) return false;
 	remove(fname.toNativeString());
 	return true;
@@ -118,9 +118,7 @@ bool isWritableDir(NativePath p, bool create_if_missing = false)
 
 Json jsonFromFile(NativePath file, bool silent_fail = false) {
 	if( silent_fail && !existsFile(file) ) return Json.emptyObject;
-	auto f = openFile(file.toNativeString(), FileMode.read);
-	scope(exit) f.close();
-	auto text = stripUTF8Bom(cast(string)f.readAll());
+	auto text = stripUTF8Bom(cast(string)readFile(file));
 	return parseJsonString(text, file.toNativeString());
 }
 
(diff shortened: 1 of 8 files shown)


source/dub/internal/vibecompat/core/file.d
```diff
@@ -22,58 +22,18 @@ import std.string;
...
-	auto ret = std.stdio.File(path.toNativeString(), fmode);
-	assert(ret.isOpen);
-	return RangeFile(ret);
+	std.file.write(path.toNativeString(), buffer);
 }
-/// ditto
-RangeFile openFile(string path, FileMode mode = FileMode.read)
+
+/// Returns the content of a file
+public void[] readFile(NativePath path)
```
s-ludwig, line 32: Should probably return `ubyte[]`, even if that makes no difference to the  GC, but it would match vibe.d's `readFile`. Or do you want to implement vibe.d compatibility with the planned dependency injection interface?

Geod24: I originally used `void[]` because that's what `std.file` returns, but I agree that's probably better.

source/dub/recipe/io.d
```diff
@@ -209,16 +202,16 @@ unittest { // make sure targetType of sub packages are sanitized too
...
-	serializePackageRecipe(f, recipe, filename);
+	writePackageRecipe(NativePath(filename), recipe);
 }
 
 /// ditto
 void writePackageRecipe(NativePath filename, const scope ref PackageRecipe recipe)
 {
-	writePackageRecipe(filename.toNativeString, recipe);
+	import std.array;
+	Appender!string app;
```
WebFreak001, line 212:
```suggestion
	auto app = appender!string();
```

source/dub/internal/utils.d
```diff
@@ -159,23 +154,19 @@ string packageInfoFileFromZip(NativePath zip, out string fileName) {
 
 void writeJsonFile(NativePath path, Json json)
 {
-	auto f = openFile(path, FileMode.createTrunc);
-	scope(exit) f.close();
-	f.writePrettyJsonString(json);
+	Appender!string app;
```
WebFreak001, line 157:
```suggestion
	auto app = appender!string();
```

source/dub/internal/utils.d
```diff
@@ -159,23 +154,19 @@ string packageInfoFileFromZip(NativePath zip, out string fileName) {
...
 	import std.random : uniform;
 	auto tmppath = path.parentPath ~ format("%s.%s.tmp", path.head, uniform(0, int.max));
-	auto f = openFile(tmppath, FileMode.createTrunc);
-	scope (failure) {
-		f.close();
-		removeFile(tmppath);
-	}
-	f.writePrettyJsonString(json);
-	f.close();
+	Appender!string app;
```
WebFreak001, line 167:
```suggestion
	auto app = appender!string();
```

source/dub/project.d
```diff
@@ -1820,27 +1818,27 @@ final class SelectedVersions {
 	void save(NativePath path)
 	{
 		Json json = serialize();
-		auto file = openFile(path, FileMode.createTrunc);
-		scope(exit) file.close();
+		Appender!string result;
```
WebFreak001, line 1821:
```suggestion
		auto result = appender!string();
```

WebFreak001 (changes requested):
Appender docs say you should construct appenders like this.

Geod24:
Rebased and changed appender

PR #2522, opened 2022-10-30
Add support for C sources and headers to dub

cschlote:
Add changes to support C source and header files with Dub

- This is a work-in-project PR as basis for discussions on scope changes and impacts involved with adding C source and header file support to dub.
- The goal is to be able to use C files (.c, .h and .i) in Dub projects the same way as usual D files (.d and .di). Dub shall be able to pick up these files automatically and pass them to the D compiler.
- We need some switch to enable the support in JSON/SDL files (backward compatibility, test staging). For this 'cSourcePaths' and 'cImportPaths' were added. If specified, dub will pick up C source and pass extra paths to the compiler.
- We need some way to detect existing ImportC support in D compilers (separate issue, WIP)
- We must generalize some code by using isSourceFile(), isHeaderFile() and isImportCFile() instead of just checking for ".d" extension. (separate issue, WIP)

### DIFF
--- a/changelog/collect_c_source_and_headers_for_ImportC.dd
+++ b/changelog/collect_c_source_and_headers_for_ImportC.dd
@@ -1,0 +1,10 @@
+
+Add new keywords 'cSourcePaths' and 'cImportPaths'
+
+The first keyword is added to enable collection of C source files in all 'directories' specified with 'cSourcePaths'.
+All C sources found in the given directories for 'cImportPaths' are passed to the D compiler.
+This ensures backward compatible behaviour for projects that stored C sources aside of D source file, while porting them to D.
+
+The second keyword 'cImportPaths' will add additional search paths for C headers. These directories are passed to the D compilers
+as addition include paths. The feature might need additional tweaking, because the include paths are currently joint with the
+D import paths. This might change in future to support independant search paths for C to be passed to the D compilers.
--- a/changelog/collect_c_source_and_headers_for_ImportC.dd
+++ b/changelog/collect_c_source_and_headers_for_ImportC.dd
@@ -1,7 +1,7 @@
 
 Add new keywords 'cSourcePaths' and 'cImportPaths'
 
-The first keyword is added to enable collection of C source files in all 'directories' specified with 'cSourcePaths'.
+`cSourcePaths` passes the C source files in all specified directories to the compiler.
 All C sources found in the given directories for 'cImportPaths' are passed to the D compiler.
 This ensures backward compatible behaviour for projects that stored C sources aside of D source file, while porting them to D.
 
--- a/changelog/collect_c_source_and_headers_for_ImportC.dd
+++ b/changelog/collect_c_source_and_headers_for_ImportC.dd
@@ -1,5 +1,4 @@
-
-Add new keywords 'cSourcePaths' and 'cImportPaths'
+Add new properties 'cSourcePaths' and 'cImportPaths' to SDL/JSON
 
 `cSourcePaths` passes the C source files in all specified directories to the compiler.
 All C sources found in the given directories for 'cImportPaths' are passed to the D compiler.
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -733,6 +733,7 @@ class ProjectGenerator
 		parent.addVersionFilters(child.versionFilters);
 		parent.addDebugVersionFilters(child.debugVersionFilters);
 		parent.addImportPaths(child.importPaths);
+		parent.addCImportPaths(child.cImportPaths);
 		parent.addStringImportPaths(child.stringImportPaths);
 		parent.addInjectSourceFiles(child.injectSourceFiles);
 		// linker stuff propagates up from static *and* dynamic library deps
@@ -999,6 +1000,7 @@ const(string[string])[] makeCommandEnvironmentVariables(CommandType type,
 	env["LIBS"]                  = join(build_settings.libs, " ");
 	env["SOURCE_FILES"]          = join(build_settings.sourceFiles, " ");
 	env["IMPORT_PATHS"]          = join(build_settings.importPaths, " ");
+	env["C_IMPORT_PATHS"]        = join(build_settings.cImportPaths, " ");
 	env["STRING_IMPORT_PATHS"]   = join(build_settings.stringImportPaths, " ");
 
 	env["DC"]                    = settings.platform.compilerBinary;
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -999,6 +999,7 @@ const(string[string])[] makeCommandEnvironmentVariables(CommandType type,
 	env["VERSIONS"]              = join(build_settings.versions, " ");
 	env["LIBS"]                  = join(build_settings.libs, " ");
 	env["SOURCE_FILES"]          = join(build_settings.sourceFiles, " ");
+	env["C_SOURCE_FILES"]        = join(build_settings.cSourceFiles, " ");
 	env["IMPORT_PATHS"]          = join(build_settings.importPaths, " ");
 	env["C_IMPORT_PATHS"]        = join(build_settings.cImportPaths, " ");
 	env["STRING_IMPORT_PATHS"]   = join(build_settings.stringImportPaths, " ");
--- a/source/dub/generators/visuald.d
+++ b/source/dub/generators/visuald.d
@@ -304,7 +304,8 @@ class VisualDGenerator : ProjectGenerator {
 				string imports = join(getPathSettings!"importPaths"(), " ");
 				string cimports = join(getPathSettings!"cImportPaths"(), " ");
 				string stringImports = join(getPathSettings!"stringImportPaths"(), " ");
-				ret.formattedWrite("    <imppath>%s %s</imppath>\n", imports, cimports);
+				string combinedImports = join([imports, cimports], " ");
+				ret.formattedWrite("    <imppath>%s</imppath>\n", combinedImports);
 				ret.formattedWrite("    <fileImppath>%s</fileImppath>\n", stringImports);
 
 				ret.formattedWrite("    <program>%s</program>\n", "$(DMDInstallDir)windows\\bin\\dmd.exe"); // FIXME: use the actually selected compiler!
--- a/source/dub/generators/visuald.d
+++ b/source/dub/generators/visuald.d
@@ -302,8 +302,9 @@ class VisualDGenerator : ProjectGenerator {
 
 				// include paths and string imports
 				string imports = join(getPathSettings!"importPaths"(), " ");
+				string cimports = join(getPathSettings!"cImportPaths"(), " ");
 				string stringImports = join(getPathSettings!"stringImportPaths"(), " ");
-				ret.formattedWrite("    <imppath>%s</imppath>\n", imports);
+				ret.formattedWrite("    <imppath>%s %s</imppath>\n", imports, cimports);
 				ret.formattedWrite("    <fileImppath>%s</fileImppath>\n", stringImports);
 
 				ret.formattedWrite("    <program>%s</program>\n", "$(DMDInstallDir)windows\\bin\\dmd.exe"); // FIXME: use the actually selected compiler!
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -640,6 +640,7 @@ class Package {
 		ret.versions = bs.versions;
 		ret.debugVersions = bs.debugVersions;
 		ret.importPaths = bs.importPaths;
+		ret.cImportPaths = bs.cImportPaths;
 		ret.stringImportPaths = bs.stringImportPaths;
 		ret.preGenerateCommands = bs.preGenerateCommands;
 		ret.postGenerateCommands = bs.postGenerateCommands;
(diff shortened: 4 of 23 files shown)


rikkimax:
This is going to need tests before going in.

However, I have no idea how that can happen without a build of at least one compiler with the preprocessor support working in the CI.

cschlote:
Yes for sure it needs testing in the CI. But nobody is hindering us to work out things in advance. In the meantime we can use and test the patched dub with our own projects.

I will continue with this patch stack, rebase it from time to time. 

@rikkimax Is there already some code to determine the compiler version / D support level in dub? if we implement that first, we can test the patched dub in the CI - it should disable ImportC support in this case as some kind of negative test. And when the next DMD/LDC/GDC is released with ImportC support, it should switch to ImportC support.

rikkimax:
> @rikkimax Is there already some code to determine the compiler version / D support level in dub?

Yes.

https://github.com/dlang/dub/blob/master/source/dub/compilers/compiler.d#L150

https://github.com/dlang/dub/blob/master/source/dub/platform.d#L252

(note ignore the determine platform stuff in that module, as that only applies to the dub executable).

There is code to compare compiler versions, however, you'll need to dig a bit into it I think.

https://github.com/dlang/dub/blob/a284d97636f746129d47a594115a8ec12986d954/source/dub/recipe/packagerecipe.d#L607

source/dub/generators/visuald.d
```diff
@@ -141,7 +141,8 @@ class VisualDGenerator : ProjectGenerator {
 		bool isHeaderOnlyPackage(string pack, in TargetInfo[string] targets)
 		const {
 			auto buildsettings = targets[pack].buildSettings;
-			if (!buildsettings.sourceFiles.any!(f => f.endsWith(".d"))())
+			if (!buildsettings.sourceFiles.any!(f => f.endsWith(".d"))() &&
+				!buildsettings.sourceFiles.any!(f => f.endsWith(".c"))() )
```
nordlow, line 145: Replace with

```d
if (!buildsettings.sourceFiles.any!(f => f.endsWith(".d", ".c")))
```

.

WebFreak001: can we move the list of hardcoded file endings somewhere else entirely? It's repeated many times in the code.

cschlote:
I tested the current state of the PR with many of my projects and dub packages. I found other bugs, but nothing related to this PR. 
I created some test code for ImportC here: https://gitlab.vahanus.net/dlang/examples/importc-tests.git

Although I agree, that there is much more potential to refactor the dub sources, I'm also a friend of agile development moving in small steps to a final goal. As this PR adds initial support for Import C with the next upcoming DMD release, we should focus on this current scope and proceed with other refactoring and improvements in other PRs. I would be nice to have working support for ImportC in dub before the next DMD release.

ibuclaw:
> The goal is to be able to use C file (.c, .h and .i) in Dub projects the same way as usual D files (.d and .di). Dub shall be able to pick up these files automatically and pass them to the D compiler.

FYI the GDC compiler has been able to compile C/C++/Go/Fortran/Rust... sources since 2004.  It's all delegated by the D compiler driver when you invoke it normally: `gdc file1.d file2.c file3.rs -o combined.exe`

redthing1:
I tried the PR and seems to work quite well! Hope to see merge soon.

cschlote:
> > The goal is to be able to use C file (.c, .h and .i) in Dub projects the same way as usual D files (.d and .di). Dub shall be able to pick up these files automatically and pass them to the D compiler.
> 
> FYI the GDC compiler has been able to compile C/C++/Go/Fortran/Rust... sources since 2004. It's all delegated by the D compiler driver when you invoke it normally: `gdc file1.d file2.cpp file3.go file4.rs -o combined.exe`.
> 
> Calling the gate `haveImportCSupport` is could be misleading, as don't really _need_ importC in order for the D compiler to compile foreign language sources.

Interesting. I have never tried to feed other language sources to GDC, yet.  

But the goal of this PR is to teach 'dub' to pass C files to the DMD, LDC2 and GDC compilers. Otherwise no code is generated and the linker will emit errors on the missing stuff.  So although GDC might be able to compile these files without any addition effort, without this PR the 'dub' project won't build as expected.

WebFreak001:
maybe instead of trying to apply this to sourcePaths/sourceFiles, should we add a new key to the dub.json/sdl like `"cPaths"` or `"sourceCPaths"` or `"cSourcePaths"`?

cschlote:
> maybe instead of trying to apply this to sourcePaths/sourceFiles, should we add a new key to the dub.json/sdl like `"cPaths"` or `"sourceCPaths"` or `"cSourcePaths"`?
> Or a more generic `"foreignSourcePaths"`?

I kept my C/foreign in separate directories and used pre-build/gen Hooks to convert them to object files. The object ist then added to the linker flags.

Starting with ImportC I tend to add my C sources inside of the usuall D directories. Simply because C is then no longer foreign. For more complex C projects, which can't be compiled with ImportC, I would use external hooks and linklibs as before.

> For C paths we could maybe also consider implementing ctod or similar in DUB, so compilers without importC will work as well, or if you want to manually enable it because of importC restrictions.

I wouldn't mix up things here. We have pre Hooks to handle the complicated stuff, or Import for the simpler code. CtoD never worked for me, because most old C code uses the PrePro for funny things.

> Advantages: we don't suddenly start compiling C source files inside D projects, which might have been there incrementally while porting to D. (a case Walter brings up often)

Ok, that would be a point. A switch like 'enable-importc' in the dub file would be a good workaround. if enabled, dub would pass C files to the compiler silently. Otherwise I will gibe a warning instead that C files are found, but not yet sent to the compiler until the 'enable-importc' property ist set. Later this could be removed when most Dub projects are adapted.

> Additionally tooling can recognize when people want to explicitly enable ImportC.
How?

cschlote:
One of my own old dub projects from 2019 uses C files, which are compiled to object files. This PR passes the C file now directly to the D compiler. Unfortunatelly this doesn't work. So we need a configuration switch/option as proposed by @WebFreak001 .

cschlote:
I added new keyword to dub.json/sdl: 'cSourcePaths' and 'cImportPaths' 
They allow to pickup C sources and headers in the given paths.
Tested with https://gitlab.vahanus.net/dlang/examples/importc-tests.git (for 'cSourcePaths').
For 'cImportPaths' more testing and likely some changes are needed. 
Please review and comment.

source/dub/project.d
```diff
@@ -294,7 +294,7 @@ class Project {
 			lbuildsettings.importPaths ~= NativePath(mainfil).parentPath.toNativeString;
 		bool firstTimePackage = true;
 		foreach (file; lbuildsettings.sourceFiles) {
-			if (file.endsWith(".d")) {
+			if (file.isSourceFile) {
```
WebFreak001, line 297: this should stay `.endsWith(".d")`, because this adds the source files / modules to the unittest module list. `.c` and `.i`files don't have any unittests / module name for this to be relevant.

source/dub/project.d
```diff
@@ -975,7 +979,9 @@ class Project {
 			case "importFiles":
 			case "stringImportFiles":
 			case "sourceFiles":
+			case "cSourceFiles":
 			case "importPaths":
+			case "CImportPaths":
```
WebFreak001, line 984:
```suggestion
			case "cImportPaths":
```

source/dub/recipe/packagerecipe.d
```diff
@@ -524,15 +526,18 @@ struct BuildSettingsTemplate {
...
+		dst.addSourceFiles(collectFiles(cSourcePaths, "*.{c,i}"));
 		auto sourceFiles = dst.sourceFiles.sort();
 
  		// collect import files and remove sources
 		import std.algorithm : copy, setDifference;
 
-		auto importFiles = collectFiles(importPaths, "*.{d,di}").sort();
+		auto importFiles =
+            chain(collectFiles(importPaths, "*.{d,di}"), collectFiles(cImportPaths, "*.h"))
+            .sort();
```
WebFreak001, line 537: mixed tabs and spaces + put the `.array` before `.sort()` because it needs memory to operate on (also makes the .release.array changes below obsolete)

source/dub/recipe/json.d
```diff
@@ -287,7 +290,7 @@ private Json toJson(const scope ref BuildSettingsTemplate bs)
 	foreach (suffix, arr; bs.debugVersions) ret[withSuffix("debugVersions", suffix)] = serializeToJson(arr);
 	foreach (suffix, arr; bs.versionFilters) ret[withSuffix("-versionFilters", suffix)] = serializeToJson(arr);
 	foreach (suffix, arr; bs.debugVersionFilters) ret[withSuffix("-debugVersionFilters", suffix)] = serializeToJson(arr);
-	foreach (suffix, arr; bs.importPaths) ret[withSuffix("importPaths", suffix)] = serializeToJson(arr);
+	foreach (suffix, arr; bs.cImportPaths) ret[withSuffix("cImportPaths", suffix)] = serializeToJson(arr);
```
WebFreak001, line 293: you deleted importPaths here

source/dub/generators/generator.d
```diff
@@ -999,6 +1000,7 @@ const(string[string])[] makeCommandEnvironmentVariables(CommandType type,
 	env["LIBS"]                  = join(build_settings.libs, " ");
 	env["SOURCE_FILES"]          = join(build_settings.sourceFiles, " ");
```
WebFreak001, line 1001: also want to add cSourceFiles as environment variable?

source/dub/package_.d
```diff
@@ -721,12 +722,16 @@ class Package {
...
+		immutable hasCSP = ("" in bs.cSourcePaths) !is null;
 		immutable hasIP = ("" in bs.importPaths) !is null;
+		immutable hasCIP = ("" in bs.cImportPaths) !is null;
 		if (!hasSP || !hasIP) {
 			foreach (defsf; ["source/", "src/"]) {
 				if (existsFile(m_path ~ defsf)) {
 					if (!hasSP) bs.sourcePaths[""] ~= defsf;
+					if (!hasCSP) bs.cSourcePaths[""] ~= defsf;
 					if (!hasIP) bs.importPaths[""] ~= defsf;
+					if (!hasCIP) bs.cImportPaths[""] ~= defsf;
```
WebFreak001, line 734: don't auto initialize cSourcePaths and cImportPaths to `source/`

source/dub/project.d
```diff
@@ -1319,6 +1325,7 @@ void processVars(ref BuildSettings dst, in Project project, in Package pack,
 	dst.addVersionFilters(processVars(project, pack, gsettings, settings.versionFilters, false, buildEnvs));
 	dst.addDebugVersionFilters(processVars(project, pack, gsettings, settings.debugVersionFilters, false, buildEnvs));
 	dst.addImportPaths(processVars(project, pack, gsettings, settings.importPaths, true, buildEnvs));
+	dst.addCImportPaths(processVars(project, pack, gsettings, settings.cImportPaths, true, buildEnvs));
```
WebFreak001, line 1328: this is missing addCSourceFiles

WebFreak001 (changes requested):
also add your test project to the tests folder

WebFreak001:
please don't force push while still reviewing, it makes it hard to understand what has changed, especially on a change with this many changed lines of code.

changelog/collect_c_source_and_headers_for_ImportC.dd
```diff
@@ -0,0 +1,10 @@
+
+Add new keywords 'cSourcePaths' and 'cImportPaths'
```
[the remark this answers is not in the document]
cschlote: Good point.

source/dub/generators/visuald.d
```diff
@@ -302,8 +302,9 @@ class VisualDGenerator : ProjectGenerator {
 
 				// include paths and string imports
 				string imports = join(getPathSettings!"importPaths"(), " ");
+				string cimports = join(getPathSettings!"cImportPaths"(), " ");
 				string stringImports = join(getPathSettings!"stringImportPaths"(), " ");
-				ret.formattedWrite("    <imppath>%s</imppath>\n", imports);
+				ret.formattedWrite("    <imppath>%s %s</imppath>\n", imports, cimports);
```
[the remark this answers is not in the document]
cschlote: Reworked code. There should be no trailing space anymore.

source/dub/project.d
cschlote: Reverted

source/dub/recipe/packagerecipe.d
cschlote: Reworked my code.

source/dub/recipe/json.d
cschlote: Oops. Thank. Fixed.

source/dub/generators/generator.d
cschlote: Good point. Added.

source/dub/package_.d
cschlote: Removed.

source/dub/project.d
cschlote: Added.

cschlote:
> please don't force push while still reviewing, it makes it hard to understand what has changed, especially on a change with this many changed lines of code.
> 
> I think for this we will need to squash it all at the end anyway, so we can do this after the reviews pass I think.

Ok, I tend to rebase my changes to the latest master, especially when other changes happen on master, which conflict with the PR. So I can resolve them now. I added the requested changes as commits to the current patch stack.

And yes, I will rework the patch stack at the end and split up the changes into reasonable commits. Some functions in utils are currently unused - probably we want them in a separate PR and with active use in the source.

cschlote:
It seem that some sources are indended with spaces and others with tabstops. Wouldn't it be a good idea to use either space OR tabstops?

cschlote:
`./run-unittest.sh` completes with DMD 2.101.0.
With `DMD64 D Compiler v2.101.0-92-g59c4ad3024` two tests will fail:
```
[INFO] Building /home/cschlote/build/dlang/dub/test/depen-build-settings/...
    Starting Performing "debug" build using dmd-beta for x86_64.
    Building depend2 ~master: building configuration [library]
Error: unrecognized switch '--d-version=must_be_defined'
       run `dmd` to print the compiler manual
       run `dmd -man` to open browser on manual
Error dmd-beta failed with exit code 1.
[ERROR] Build failure.
```
With current LDC2 1.30.0 it won't complete the `use-c-sources` tests, because LDC uses runtime 2.100.1 and not 2.101.0 or newer.

cschlote:
Cleanup up the patch stack. I re-submitted the PR as  #2544.

PR #2525, opened 2022-11-02
Move foreign code from global scope to dub.internal

s-ludwig:
Avoids conflicting with the actual dub packages when working with dub as a library. Also, even if just by convention, it ensures that dependent code doesn't rely on this code as a public API.


s-ludwig:
BTW, I think the cost/benefit ratio of the whole dyaml/Configy based refactoring is highly debatable. Even the complexity of the dub code itself doesn't really seem to have become lower, but these libraries add a substantial amount of code for maybe a few lines less, if any.

WebFreak001 (approved):

WebFreak001:
I think nice advantages of dyaml/configy over the manual parsing code is that
- we don't have to maintain custom deserialization logic anymore
- we get much better automatic parse and deserialization errors

Personally I see a future in mir / mir-ion as a generic (de)serialization framework, so I would like to push that much more for general purpose use, but we have configy working here in DUB + it gives us much nicer errors right now. Also having YAML as format allows us to extend the settings format, which has long been an open issue.

Geod24:
@s-ludwig : The main motivation for using configy / dyaml was scalability. While the upfront cost was high (introducing the libraries themselves), switching individual files has been almost painless. Adding support for YAML is trivial.

On the code quality side, introducing a library gives us the power of abstraction. It allowed me to isolate places where abstraction was breached (e.g. https://github.com/dlang/dub/pull/2277 ) and fix them. This could have also been done by building a library inside of dub, but what I would have ended up with would have been, in spirit, quite similar to Configy. So using it just saves us a bunch of time.

Finally, on the code size front, the reason why we only saved "a few lines" is that I kept the JSON parsing code in for now, as a fallback in case there is a bug with Configy (so far I haven't found one). When the transition period ends, then we will remove large portions of code.

PR #2526, opened 2022-11-04
Fix add-path based development workflow

s-ludwig:
The logic that locally checked out packages added using `add-path` take precedence and don't follow the `strict` matching rules got broken again due to a recent change. It should be noted that this fix requires a call to `PackageManager.refresh()`, which will, in order to get the anticipated performance improvement, probably need to be the target for a more fine-grained approach, where only the cache directory is skipped during package scanning instead of everything.

### DIFF
--- a/build.d
+++ b/build.d
@@ -35,7 +35,11 @@ immutable VersionFilePath = RootPath.buildPath("source", "dub", "version_.d");
 /// Path to the file containing the files to be built
 immutable SourceListPath = RootPath.buildPath("build-files.txt");
 /// Path at which the newly built `dub` binary will be
-immutable DubBinPath = RootPath.buildPath("bin", "dub");
+version (Windows) {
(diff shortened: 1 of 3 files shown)


Geod24 (approved):
It's pretty bad that we need to scan all possible packages on the system for a specific feature, but at least the next time I try it a test will break.

nordlow:
Will this regress the performance?

s-ludwig:
@nordlow There was a functional regression introduced by the performance optimization, not the other way around. Should be back to where it was before now (unfortunately).

@Geod24 We should really look into splitting up the add-path scanning and the cache scanning anyway, as both kinds of packages currently share the same array kind of awkwardly. Then it would at least be straight forward to skip the cache scanning part. The add-path part is then basically "pay for what you actually use".

PR #2539, opened 2022-11-30
Rename localRepository to userPackages

Geod24:
The 'repository' terminology is confusing, and needs to be interpreted from the PoV of the PackageManager to make sense. It is likely that a naive read of this name would lead one to think of git repository, as opposed to the repository where packages are stored.

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -163,7 +163,7 @@ class Dub {
 
 		m_packageSuppliers = this.computePkgSuppliers(additional_package_suppliers,
 			skip_registry, environment.get("DUB_REGISTRY", null));
-		m_packageManager = new PackageManager(m_rootPath, m_dirs.localRepository, m_dirs.systemSettings, false);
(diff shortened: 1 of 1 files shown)


source/dub/dub.d
```diff
@@ -1812,16 +1812,17 @@ private struct SpecialDirs {
...
+	 * so an account on a company network will not save the content of this data,
 	 * unlike `userSettings`.
-	 * On Posix, this is equivalent to `userSettings`.
+	 *
+	 * On Posix, this is currently equivalent to `userSettings`.
 	 *
 	 * See_Also: https://docs.microsoft.com/en-us/windows/win32/shell/knownfolderid
 	 */
-	NativePath localRepository;
+	NativePath userPackages;
```
WebFreak001, line 1825: add a deprecated alias pls

Geod24: The structure is `private`.

Issue #1473, opened 2018-05-16
Fails to use system-wide packages due to missing permissions


ximion:
Hi!
I played around with the new system-wide dub packages (#838) feature recently, and unfortunately it doesn't seem to work right yet.
Dub tries to write into paths like `/usr/include/d/test-1.0/test/.dub/`, which of course fails because this global path is read-only.
Ideally, dub would use a different temporary location for storing the build artifacts. Ideally, this would be in the .dub directory of the currently-built package, so we avoid writing to $HOME, or some other configurable location.

In other news, I am working on integrating the Meson build system with dub, so Meson can publish packages usable by dub (easy, just install them as packages into a system path and have Meson generate a dub.json) and use packages that exclusively use dub as well.
For that, I have dub commands run as build targets by ninja. I use commands like this:
`dub build test -b plain -v -c test-static --root build-dir`
I would expect dub to build the `test`package that is installed in `/usr/include/d/test-1.0/test/` and output its build artifacts to the `build-dir` directory, but that doesn't seem to be what happens here.
Instead I get:
```
Performing "plain" build using /usr/bin/ldc2 for x86_64.
Target '/usr/include/d/test-1.0/test/.dub/build/test-plain-linux.posix-x86_64-ldc2_2078-DAB625B69498E2F13B2B69D41EDFFDE0/libTest.a' doesn't exist, need rebuild.
/usr/include/d/test-1.0/test/.dub: Permission denied
```
It would be pretty cool to pass dub a global package path in `/usr/include/d/<pkg>`, have it build the package and publish the result in a custom directory as well. That would avoid the permission error, and also make it easy to integrate dub as part of other build systems, by having it take care of building selected dependencies.
Cheers,
    Matthias

grillo-delmal:
Any update on this issue? I tried to do something similar on the latest dub commit and got the same result.

s-ludwig:
https://github.com/dlang/dub/pull/1650 was supposed to solve the main issue, but unfortunately has become outdated and I need to find the time to adjust it properly.

grillo-delmal:
I've been toying around with the --cache=* and --tempbuild flags to see how viable is to package dub libraries and software using the latest commit of dub. 

* When building an app with the --tempbuild flag, the local dependencies and the built project end up in tmp... but for some reason, this doesn't happen with libs installed in a system path, 
[...]

Issue #1256, opened 2017-09-21
dub clean --all-packages doesn't remove all build artifacts


MartinNowak:
### System information
- dub 1.5.0
- Linux Fedora 26
- dmd-2.076.0

### Bug Description

Build targets are still left in `~/.dub/package` folders after running `dub clean --all-packages`.

### How to reproduce?

```sh
dub fetch vibe-d
dub build vibe-d
ls -lh ~/.dub/packages/vibe-d-0.8.1/vibe-d/libvibed.a
dub clean --all-packages
ls -lh ~/.dub/packages/vibe-d-0.8.1/vibe-d/libvibed.a # still there
dub build vibe-d:http
ls -lh ~/.dub/packages/vibe-d-0.8.1/vibe-d/http/libvibe-d_http.a
dub clean --all-packages
ls -lh ~/.dub/packages/vibe-d-0.8.1/vibe-d/http/libvibe-d_http.a # stil there
```

### Expected Behavior

All target files of all build configurations, and all copyFiles should be removed.
Even just removing the targets for the default configuration would be a good improvement.

MartinNowak:
Also subproject build caches aren't cleaned.
e.g.

    data/.dub/build/library-debug-linux.posix-x86_64-dmd_2078-1D306B4CDF1D8892F7C502EA0C6AC840/libvibe-d_data.a
    http/.dub/build/library-debug-linux.posix-x86_64-dmd_2078-44E4D7BEF4C82829B478CFE692097C72/libvibe-d_http.a

Issue #631, opened 2015-07-25
build caches should evict LRU cached builds to limit dub folder size


MartinNowak:
Things would be a lot better with shared libraries, but unless that becomes default static D libraries are somewhat big.

693M	/home/dawg/.dub/packages/dub-0.9.23-rc.1
729M	/home/dawg/.dub/packages/userman-0.2.4
1.2G	/home/dawg/.dub/packages/vibe-d-0.7.23
5.8G	/home/dawg/.dub/packages/

PR #2542, opened 2022-12-02
Use a central build cache for all build artifacts

Geod24:
```
Currently the central build cache is not configurable, but as most build artifacts end up in `~/.dub/packages/` anyway, this will be a positive change for users.
```

This is essentially https://github.com/dlang/dub/pull/1650 without configurability, because there's a few things needed before we can make it configurable "nicely".

Fix https://github.com/dlang/dub/issues/1473 and would make adding a GC much easier (e.g. https://github.com/dlang/dub/issues/1256 and https://github.com/dlang/dub/issues/631).

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -255,6 +255,7 @@ class Dub {
 
 				m_dirs.userSettings = NativePath(dubHome);
 				m_dirs.userPackages = m_dirs.userSettings;
+				m_dirs.cache = m_dirs.userPackages ~ "cache";
 			}
 		}
 
@@ -286,6 +287,7 @@ class Dub {
 		// config loaded from user settings and per-package config as well.
 		if (!overrideDubHomeFromEnv && this.m_config.dubHome.set) {
 			m_dirs.userPackages = NativePath(this.m_config.dubHome.expandEnvironmentVariables);
+			m_dirs.cache = m_dirs.userPackages ~ "cache";
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


source/dub/dub.d
```diff
@@ -1824,6 +1828,20 @@ private struct SpecialDirs {
 	 */
 	NativePath userPackages;
 
+	/**
+	 * Location at which build/generation artifact will be written
+	 *
+	 * All build artifacts are stored under a single build cache,
+	 * which is usually located under `$HOME/.dub/cache/` on POSIX,
```
nordlow, line 1835: If we want to follow https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html then ~/.cache/dub could be an alternative directory to place cache files.

Geod24: That will be for another issue.

Geod24:
This passes now. The scheme is a bit more complex to avoid name collision. subpackages are really a PITA.

Geod24:
Any opinion on the new format ?

Possible problem with this: two packages with the same name will end up in the same folder. That's something that we will have to solve, eventually.

We have `$ROOT_PACKAGE_NAME/$PACKAGE_VERSION/$SUB_PACKAGE_NAME` and not `$ROOT_PACKAGE_NAME/$SUB_PACKAGE_NAME/$PACKAGE_VERSION` because, while the latter is easier, the very definition of a subpackage is that it's a collection of independent packages tied to the same version.

Also, I used `+` before `$SUB_PACKAGE_NAME` to ensure that no one can introduce a subpackage name that conflicts with a version. Even if the name would be invalid (because `.` is currently allowed in package name), it would probably not be recognized.

Now I need to look into the metadata_cache and the `code/` cache.

Geod24:
Metadata cache and `code/` handled, we're G2G

Geod24:
Going to merge when it's green.

PR #2548, opened 2022-12-10
enable GDC 12 in alpine CI

WebFreak001:
see #1998


ibuclaw:
ICE fix has been backported to 12 release branch, can't say when the next minor release will be.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108050

ibuclaw:
Linker error has been fixed and backported to the 12 release.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108055

WebFreak001:
awesome! let's see when it will be adopted by alpine

ibuclaw:
> awesome! let's see when it will be adopted by alpine

Having a look at previous release history, the RM for GCC typically does the 3rd point release around April.

PR #2550, opened 2022-12-11
replace std.variant with std.sumtype

WebFreak001:
Variant is quite big and runtime heavy, with sumtype we already cover wrong type cases at compile time.

I was thinking this could be useful in fixing the GDC linker error that's caused by Variant.toString failing with to!string(float) code emission, however while I could fix it with a simple toString replacement for writeln, I couldn't fix it in all the different cases where floats are converted to strings here. Finding the linker error causes is quite difficult and fixing them is very ugly, so I didn't attempt to fix GDC compatibility here. (but using sumtype with explicit types rather than the all-encompassing Variant can certainly help in fixing it)

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -167,7 +167,7 @@ struct CommandLineHandler
 			options.root_path = options.root_path.expandTilde.absolutePath.buildNormalizedPath;
 		}
 
-		final switch (options.color_mode) with (options.color)
+		final switch (options.colorMode) with (options.Color)
 		{
 			case automatic:
 				// Use default determined in internal.logging.initLogging().
@@ -523,10 +523,33 @@ struct CommonOptions {
 	bool help, annotate, bare;
 	string[] registry_urls;
 	string root_path;
-	enum color { automatic, on, off } // Lower case "color" in support of invalid option error formatting.
-	color color_mode = color.automatic;
+	enum Color { automatic, on, off }
+	Color colorMode = Color.automatic;
 	SkipPackageSuppliers skipRegistry = SkipPackageSuppliers.none;
 	PlacementLocation placementLocation = PlacementLocation.user;
+
+	deprecated("Use `Color` instead, the previous naming was a limitation of error message formatting")
+	alias color = Color;
+	deprecated("Use `colorMode` instead")
+	alias color_mode = colorMode;
+
+	private void parseColor(string option, string value) @safe
+	{
+		// `automatic`, `on`, `off` are there for backwards compatibility
+		// `auto`, `always`, `never` is being used for compatibility with most
+		// other development and linux tools, after evaluating what other tools
+		// are doing, to help users intuitively pick correct values.
+		// See https://github.com/dlang/dub/issues/2410 for discussion
+		if (!value.length || value == "auto" || value == "automatic")
+			colorMode = Color.automatic;
+		else if (value == "always" || value == "on")
+			colorMode = Color.on;
+		else if (value == "never" || value == "off")
+			colorMode = Color.off;
+		else
+			throw new ConvException("Unable to parse argument '--" ~ option ~ "=" ~ value
+				~ "', supported values: --color[=auto], --color=always, --color=never");
+	}
 
 	/// Parses all common options and stores the result in the struct instance.
 	void prepare(CommandArgs args)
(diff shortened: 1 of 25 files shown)


ibuclaw:
@WebFreak001 which versions of gdc have you tried?  gdc-10, gdc-11, gdc-12?  As far as my (attempted) reduction is concerned, only gdc-12 is affected, to which I have a fix prepped for it, and will backport for the next 12 minor release + 13.

WebFreak001:
only GDC 12 and above would be supported by DUB because of syntax changes in the language

source/dub/commandline.d
```diff
@@ -21,7 +21,7 @@ import dub.packagesuppliers;
 import dub.project;
 import dub.internal.utils : getDUBVersion, getClosestMatch, getTempFile;
 
-import dub.internal.dyaml.stdsumtype;
```
Geod24, line 24: I think moving this out of the dyaml package is unnecessary. I'd rather keep the diff with upstream `dyaml` low than have a more "correct" placement. There's no tangible benefit in the move, but some tangible downsides (when I sync things, I usually just cp files over and check the diff to see if some adaptations still make sense).

source/dub/commandline.d
```diff
@@ -630,20 +631,25 @@ class CommandArgs {
...
 	{
+		import std.traits : OriginalType;
+
 		foreach (ref arg; m_recognizedArgs)
 			if (names == arg.names) {
 				assert(help_text is null, format!("Duplicated argument '%s' must not change helptext, consider to remove the duplication")(names));
-				*var = arg.value.get!T;
+				*var = arg.value.match!(
+					(OriginalType!T v) => cast(T)v,
+					(_) => assert(false, "value from previous getopt has different type than the current getopt call"))
```
Geod24, line 641: Wouldn't that return incorrect results if the previous getopt call set a value outside of T's bounds ?

source/dub/commandline.d
```diff
@@ -630,20 +631,25 @@ class CommandArgs {
...
+				;
 				return;
 			}
 		assert(help_text.length > 0);
 		Arg arg;
-		arg.defaultValue = *var;
+		arg.defaultValue = Arg.Value(cast(OriginalType!T)*var);
```
Geod24, line 647: Shouldn't implicit conversion take care of this, ditto below ?

source/dub/recipe/sdl.d
```diff
@@ -343,64 +342,79 @@ private string expandPackageName(string name, string parent_name, Tag tag)
...
+)
+{
+	return value.match!(
+		(T v) => v,
+		(fallback)
+		{
+			enforceSDL(false, format("Expected value of type " ~ T.stringof ~ " for '%s', but got %s.",
+				customFieldName.length ? customFieldName : errorInfo.fullName,
+				typeof(fallback).stringof),
+				errorInfo, file, line);
```
Geod24, line 368: We don't pass file/line info for the actual parsed file anywhere ?

Geod24 (commented):
Did a first round of review

Geod24:
Failure is due to the unnecessary move BTW. If we were to proceed with the move, you'd have to update the txt file to change the path of `sumtype.d`. But as I mentioned, I think that's superfluous.

PetarKirov:
Hey @WebFreak001 @Geod24 what's the status of this PR?

source/dub/commandline.d
WebFreak001: yes that could happen, but I think the previous code would have done the same. (just doing it inside the standard library and not in DUB) - It's just an issue for when the programmer makes a mistake anyway and can't usually be caused by the user doing something.

source/dub/recipe/sdl.d
WebFreak001: no

source/dub/commandline.d
WebFreak001: `cast(OriginalType!T)` is important because SumType does not do implicit type conversions. I can leave out the constructor call though and rely on opAssign.

WebFreak001:
adjusted to review, ready for another round

source/dub/recipe/sdl.d
```diff
@@ -343,64 +342,79 @@ private string expandPackageName(string name, string parent_name, Tag tag)
...
 
-private void parsePlatformStringArray(Tag t, ref string[][string] dst)
+private string getPlatformSuffix(Tag t, string file = __FILE__, int line = __LINE__)
 {
 	string platform;
 	if ("platform" in t.attributes)
-		platform = t.attributes["platform"][0].value.get!string;
-	dst[platform] ~= t.values.map!(v => v.get!string).array;
+		platform = t.attributes["platform"][0].value.expect!string(t, t.fullName ~ " platform", file, line);
+	return platform;
```
Geod24, line 392: In the future I'd recommend writing this as:
```D
if (auto ptr = "platform" in t.attributes)
    return (*ptr)[0].value...
return null;
```
To make it more obvious what is happening, but you are just matching the old style here.

Geod24 (approved):

PR #2552, opened 2022-12-15
Implement --color=auto|always|never, fix #2410

### LINKED ISSUE #2410: Support more values passed to --color(s)
Rust's default build tool, `cargo` allows the following values passed to `--color`

```
        --color <WHEN>          Coloring: auto, always, never
```

. I find those namings a bit more natural. Shall we support those as aliases aswell?

WebFreak001:
Also improves error message output for unsupported --color values.

Before:
before

After:
after

The line gets a bit long now, I could improve this, but would need to adjust the exception handler a bit for that.

### DIFF
--- a/changelog/colors.dd
+++ b/changelog/colors.dd
@@ -1,0 +1,6 @@
+The `--color` argument now accepts values `auto`, `never`, `always`
+
+The previous `automatic`, `on`, `off` values are still supported, but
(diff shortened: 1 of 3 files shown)


Geod24 (approved):

nordlow:
Awesome!

PR #2553, opened 2022-12-15
make deprecationWarnings/Errors consistent

WebFreak001:
There was different behavior in GDC and deprecationWarnings did not perform expected results with DMD and LDC.

I added `-Wextra` to GDC, but haven't had a response from @ibuclaw yet if we should or should not include it here.

GDC emits deprecation warnings when importing deprecated modules (which makes sense I think), but this diverges from the other compilers. This was the initial reason why I noticed that deprecations break builds, which didn't happen with other compilers. (dub_test_root imports all modules, even if they are deprecated)

We should probably also fix `dub_test_roots` to emit `deprecated static import` for deprecated symbols, as well as handling them properly. But I think that's for another PR.


ibuclaw:
> I added -Wextra to GDC, but haven't had a response from @ibuclaw yet if we should or should not include it here.

`-Wextra` should turn on these codegen-level warnings.
```
extern(C++) class CPPClass { int a; }
extern(D) class DClass { int a; }
void *ptr;

void test()
{
    if (&ptr == null)  // Waddress
        return;
    if (&ptr)  // Waddress
        return;
    auto cpptod = cast(DClass)new CPPClass;  // Wcast-result
    auto dtocpp = cast(CPPClass)new DClass; // Wcast-result
    ifloat i;
    float f = cast(float)i;  // Wcast-result
}
```

> GDC emits deprecation warnings when importing deprecated modules (which makes sense I think), but this diverges from the other compilers. 

I can't think of why it would converge, unless this is to do with not having a separation between deprecation and warnings.

WebFreak001:
(accidentally closed)

> I can't think of why it would converge, unless this is to do with not having a separation between deprecation and warnings.

yes that's exactly what converged, because DUB treats warnings as errors by default, but in other compilers deprecations are not warnings. So in GDC deprecations used to error, but in other compilers they didn't. This PR changes that.

Geod24:
@ibuclaw : IMHO the lack of separation between deprecations and warnings is quite problematic. The approach taken by DMD here is, I think, the more sensible one. I assume you just hooked into GCC's machinery and that's what it does by default ?

ibuclaw:
> @ibuclaw : IMHO the lack of separation between deprecations and warnings is quite problematic. The approach taken by DMD here is, I think, the more sensible one. I assume you just hooked into GCC's machinery and that's what it does by default ?

Yes. And the problem is that dmd (the front-end) has warnings.

Geod24:
Yes, but it's unlikely to change anytime soon.

WebFreak001:
I think having warnings that you can individually turn on and off, making them error, log or ignore, is a better approach than what DMD has. So I think what GDC does is actually quite useful and this PR makes DUB use that functionality to make GDC behave like the other compilers. (and the user can still restrict and extend it further)

ibuclaw:
> I think having warnings that you can individually turn on and off, making them error, log or ignore, is a better approach than what DMD has. So I think what GDC does is actually quite useful and this PR makes DUB use that functionality to make GDC behave like the other compilers. (and the user can still restrict and extend it further)

Oh yes. :-)

```
gdc -Wunreachable-code -Werror=vector-operation-performance -Wno-cast-result
```

PR #2555, opened 2023-01-01
update version to v1.31.0-beta.1

ibuclaw:


PetarKirov (approved):

PetarKirov:
It looks like the CI is failing on Windows because Dub creates a path that's very long.

@Geod24 do you recall if you encountered such problems when working on the separate build cache feature?

ibuclaw:
> It looks like the CI is failing on Windows because Dub creates a path that's very long.
> 
> @Geod24 do you recall if you encountered such problems when working on the separate build cache feature?

First failure that I see was on: https://github.com/dlang/dub/commit/8ad1d431f1da6a73d43e13424be253e25cb08586
This change was untested by the CI: https://github.com/dlang/dub/commit/98a355e722458410477e62ea8a7346ffdceca46d

It could be there has been a regression in dmd-master, someone with a windows machines can bisect.

Geod24:
The issue would come from https://github.com/dlang/dub/pull/2542 but the CI was green when I tested.

ibuclaw:
> The issue would come from #2542 but the CI was green when I tested.

I've also seen what looks like the same error in dlang/installer too.

https://github.com/dlang/installer/actions/runs/3825037333/jobs/6507952110

Geod24:
> I've also seen what looks like the same error in dlang/installer too.

That's different. We're doing `rm -r` of a directly that does not exists (because it is now created elsewhere).
I don't know if the lack of `-f` is accidental or on purpose (to ensure something was build).

Geod24:
That job just passed on master, so re-triggering...

Issue #1026, opened 2016-12-20
"dun run <package>" fails if it finds "Object.d" in cwd


sprinkle131313:
It seems to be confusing the Object.d with the druntime. 

```
Failed to invoke the compiler dmd to determine the build platform: object.d(12): Error: "module X is in file Y which cannot be read" (error specific to my object.d file)
import path[0] = C:\D\dmd2\windows\bin\..\..\src\phobos
import path[1] = C:\D\dmd2\windows\bin\..\..\src\druntime\import
```

If I rename the Object.d file then it runs no problem.

rikkimax:
What is happening is that the compiler has an implicit include on the cwd, the fix for this is to pass in the appropriate package directory as the CWD when invoking tools.

https://github.com/dlang/dub/blob/17d8c9467b81cf2b2ac57c338f01f71c2c71b013/source/dub/compilers/compiler.d#L127

WebFreak001:
should be fixed / easily fixable with #2562

PR #2562, opened 2023-01-08
Reduce dependency on the current working directory

WebFreak001:
first commit was needed, because introducing the getWorkingDirectory function call in some module here started actually using the symbols from dub.internal.utils it seems. I'm not sure why it didn't trigger before, but it was an issue for me. I just moved the module destructor that cleans up temporary files into a separate module to work around this.

I changed all path concatenation with getcwd and getWorkingDirectory, all spawnProcess/execute calls and some affected dirEntries calls to use a "toolWorkingDirectory" from `GeneratorSettings`.

This affects the compiler and the linker, but also minor things such as the ddox tool or `dub lint`.

This change allows serve-d to always build correctly without needing to chdir the entire application's working directory. (which I can't do, because I have multiple dub instances that may even build in parallel + there were issues that relative paths weren't generated properly)

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -12,6 +12,7 @@ deprecated("Please `import dub.dependency : Dependency` instead") public import 
 public import dub.platform : BuildPlatform, matchesSpecification;
 
 import dub.internal.vibecompat.inet.path;
+import dub.internal.vibecompat.core.file;
 
 import dub.internal.logging;
 
@@ -108,10 +109,24 @@ interface Compiler {
 	void setTarget(ref BuildSettings settings, in BuildPlatform platform, string targetPath = null) const;
 
 	/// Invokes the compiler using the given flags
-	void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback);
+	deprecated("specify the working directory")
+	final void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback)
+	{
+		invoke(settings, platform, output_callback, getWorkingDirectory());
+	}
+
+	/// ditto
+	void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback, NativePath cwd);
 
 	/// Invokes the underlying linker directly
-	void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback);
+	deprecated("specify the working directory")
+	final void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback)
+	{
+		invokeLinker(settings, platform, objects, output_callback, getWorkingDirectory());
+	}
+
+	/// ditto
+	void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback, NativePath cwd);
 
 	/// Convert linker flags to compiler format
 	string[] lflagsToDFlags(const string[] lflags) const;
@@ -124,17 +139,26 @@ interface Compiler {
 		This method should be used by `Compiler` implementations to invoke the
 		compiler or linker binary.
 	*/
+	deprecated("specify the working directory")
 	protected final void invokeTool(string[] args, void delegate(int, string) output_callback, string[string] env = null)
+	{
+		invokeTool(args, output_callback, getWorkingDirectory(), env);
+	}
+
+	/// ditto
+	protected final void invokeTool(string[] args, void delegate(int, string) output_callback, NativePath cwd, string[string] env = null)
 	{
 		import std.string;
 
 		int status;
 		if (output_callback) {
-			auto result = executeShell(escapeShellCommand(args), env);
+			auto result = executeShell(escapeShellCommand(args),
+				env, Config.none, size_t.max, cwd.toNativeString());
 			output_callback(result.status, result.output);
 			status = result.status;
 		} else {
-			auto compiler_pid = spawnShell(escapeShellCommand(args), env);
+			auto compiler_pid = spawnShell(escapeShellCommand(args),
+				env, Config.none, cwd.toNativeString());
 			status = compiler_pid.wait();
 		}
 
--- a/source/dub/internal/temp_files.d
+++ b/source/dub/internal/temp_files.d
@@ -1,0 +1,45 @@
+/**
+	Provides methods to generate temporary file names and folders and
+	automatically clean them up on program exit.
+
+	Copyright: © 2012 Matthias Dondorff, © 2012-2023 Sönke Ludwig
+	License: Subject to the terms of the MIT license, as written in the included LICENSE.txt file.
+	Authors: Matthias Dondorff, Sönke Ludwig, Jan Jurzitza
+*/
+module dub.internal.temp_files;
+
+import std.file;
+
+import dub.internal.vibecompat.core.file;
+
+NativePath getTempDir()
+{
+	return NativePath(std.file.tempDir());
+}
+
+NativePath getTempFile(string prefix, string extension = null)
+{
+	import std.uuid : randomUUID;
+	import std.array: replace;
+
+	string fileName = prefix ~ "-" ~ randomUUID.toString() ~ extension;
+
+	if (extension !is null && extension == ".d")
+		fileName = fileName.replace("-", "_");
+
+	auto path = getTempDir() ~ fileName;
+	temporary_files ~= path;
+	return path;
+}
+
+private NativePath[] temporary_files;
+
+static ~this()
+{
+	foreach (path; temporary_files)
+	{
+		auto spath = path.toNativeString();
+		if (spath.exists)
+			std.file.remove(spath);
+	}
+}
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -30,28 +30,7 @@ version(DubUseCurl)
 	public import std.net.curl : HTTPStatusException;
 }
 
-
-private NativePath[] temporary_files;
-
-NativePath getTempDir()
-{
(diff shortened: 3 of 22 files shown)


source/dub/internal/temp_files.d
```diff
@@ -0,0 +1,37 @@
+module dub.internal.temp_files;
```
Geod24, line 1: Module DDOC missing

source/dub/internal/temp_files.d
```diff
@@ -0,0 +1,37 @@
...
+
+static ~this()
+{
+	foreach (path; temporary_files)
+	{
+		auto spath = path.toNativeString();
+		if (spath.exists)
+			std.file.remove(spath);
+	}
+}
```
Geod24, line 37: Missing newline at end of file

source/dub/internal/utils.d
```diff
@@ -30,28 +30,7 @@ version(DubUseCurl)
...
-	string fileName = prefix ~ "-" ~ randomUUID.toString() ~ extension;
-
-	if (extension !is null && extension == ".d")
-		fileName = fileName.replace("-", "_");
-
-	auto path = getTempDir() ~ fileName;
-	temporary_files ~= path;
-	return path;
-}
+public import dub.internal.temp_files;
```
Geod24, line 33: What was the cycle ?

source/dub/compilers/compiler.d
```diff
@@ -108,10 +109,24 @@ interface Compiler {
...
+		invoke(settings, platform, output_callback, getWorkingDirectory());
+	}
+
+	/// ditto
+	void invoke(in BuildSettings settings, in BuildPlatform platform, void delegate(int, string) output_callback, NativePath cwd);
 
 	/// Invokes the underlying linker directly
-	void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback);
+	deprecated("specify the working directory")
+	final void invokeLinker(in BuildSettings settings, in BuildPlatform platform, string[] objects, void delegate(int, string) output_callback)
```
Geod24, line 123: Technically this is a breaking change but let's roll with it.

source/dub/internal/temp_files.d
Geod24: Also it needs to be added to the txt file.

Geod24 (commented):
Yes on the concept. See comments.

Also:
```
source/dub/generators/build.d(550): Deprecation: function `dub.compilers.compiler.Compiler.invokeLinker` is deprecated - specify the working directory
source/dub/generators/build.d(562): Deprecation: function `dub.compilers.compiler.Compiler.invoke` is deprecated - specify the working directory
source/dub/generators/build.d(582): Deprecation: function `dub.compilers.compiler.Compiler.invoke` is deprecated - specify the working directory
source/dub/generators/build.d(590): Deprecation: function `dub.compilers.compiler.Compiler.invokeLinker` is deprecated - specify the working directory
```

source/dub/internal/utils.d
WebFreak001: I don't have the output anymore, but it was about 6 imports deep, the deepest one before the cycle was dub.d which imported back on dub.internal.utils

source/dub/compilers/compiler.d
WebFreak001: yeah I'm not too keen on this change, but it's unlikely third party apps have inherited the compiler class to add their own compilers. If they did, it's an easy fix at least.

source/dub/internal/utils.d
Geod24: I'd like to know how `dub.internal.utils` can end up importing `dub.d`, as that shouldn't happen.

WebFreak001: it originates from the `import dub.package_ : packageInfoFiles;` inside utils.d

Geod24: Yeah that probably shouldn't be there...

WebFreak001: I think it's fine where that variable is defined, it's just the compiler being silly here and insisting on module constructor ordering.

WebFreak001:
ready for review

Geod24:
Could you clean up the commit history ? Or should I just squash merge ?

WebFreak001:
squash merge should be fine

WebFreak001:
rebased into 3 neat commits

source/dub/internal/temp_files.d
```diff
@@ -0,0 +1,45 @@
...
+
+NativePath getTempFile(string prefix, string extension = null)
+{
+	import std.uuid : randomUUID;
+	import std.array: replace;
+
+	string fileName = prefix ~ "-" ~ randomUUID.toString() ~ extension;
+
+	if (extension !is null && extension == ".d")
+		fileName = fileName.replace("-", "_");
```
PetarKirov, line 28: I think this has the effect of modifying the `prefix` in case it continues `-`, which could lead to weird behavior (especially if the callers of this function are not prepared to handle this case and perform other commands relative to the prefix, without substituting `-` for `_`.
How about either:
- only modifying the baseName component, and not the dirName?
- always using `prefix ~ '_' ~ ...` no matter the file extension

source/dub/internal/temp_files.d
```diff
@@ -0,0 +1,45 @@
...
+
+	auto path = getTempDir() ~ fileName;
```
PetarKirov, line 30: I think it's worth documenting that the prefix is not absolute, but relative to the temp dir, in case someone expects to get this function like this: 

```D
getTempFile("/build/project/some/folder/");
```

source/dub/internal/temp_files.d
```diff
@@ -0,0 +1,45 @@
...
+	temporary_files ~= path;
+	return path;
+}
+
+private NativePath[] temporary_files;
+
+static ~this()
```
PetarKirov, line 37: Note for the future (not this PR): from an architectural point of view, we should refactor the code to use dependency injection instead of relying on global variables.

source/dub/internal/temp_files.d
WebFreak001: This file is just moved from the utils module into its own to fix cyclic dependency with module dtors. No behavior is changed and this function has already been widely used and tested. If you have some problem with its behavior, please change it in a separate PR

source/dub/internal/temp_files.d
WebFreak001: see above

rikkimax:
No bug report, adding a test for #1026 would solve that. Either way looks fine.

WebFreak001:
@Geod24 can we get this in? I need this for serve-d DUB linting to work properly in multi-workspace and temporary directory cases (like shown in the test)

WebFreak001:
@Geod24 @PetarKirov ping, the test here shows what exactly this fixes: https://github.com/dlang/dub/pull/2562/files#diff-9c31e8df9093081ce880575b76f96e115f1778f08c1b4cc51bc9eb111da7b227 (that code didn't work before)

WebFreak001:
rebased to target stable now

WebFreak001:
not sure what's up with DAutoTest and why it tries to merge into master, but merging into stable seems to work fine. GitHub actions tests also run through successfully now.

WebFreak001:
@CyberShadow how can I make the DAutoTest target the stable branch instead of master? (like I'm doing in the PR, which makes the auto tester fail)

CyberShadow:
Enabled Dub branches to the metarepo builder here: https://github.com/CyberShadow/D-dot-git/commit/06dc0a54eb38719835fb7e07d58497c8c670080d

New metarepo is pushed, should be picked up on the next rebuild, so just rebase.

CyberShadow:
Um, should this really target `stable` though? Or is Dub's `stable` not the same as DMD/Phobos' `stable` after all?

WebFreak001:
we wanted to get this change in for the new release because it's purely an API change outside the DUB CLI and I wanted to use it with serve-d (and, well, it should have been merged before the freeze for the tag already ^^)

WebFreak001:
@Geod24 can I have a review? Tests seem to pass now, just don't know what these required pending tests are that never started.

Geod24 (approved):
Sorry for the delay. This is great! Just a question.

source/dub/project.d
```diff
@@ -1405,10 +1405,21 @@ private string[] processVarsWithGlob(Project, Package)(string var, in Project pr
 	}
 	if (i == res.length) //no globbing found in the path
 		return [res];
-	import std.path : globMatch;
 	import std.file : dirEntries, SpanMode;
-	return dirEntries(res[0 .. sepIdx], SpanMode.depth)
-		.map!(de => de.name)
+	import std.path : buildNormalizedPath, globMatch, isAbsolute, relativePath;
+	auto cwd = gsettings.toolWorkingDirectory.toNativeString;
```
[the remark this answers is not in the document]
WebFreak001: dirEntries always returns paths relative to the queried directory, what this code does is:

1) if path is not absolute, prepend the toolWorkingDirectory (in regular DUB exe usage this is just the cwd)
2) iterate directory
3) for each item, if path wasn't absolute in step 1, make each item relative to toolWorkingDirectory

so the paths are still always all relative to the cwd.

What is a behavior change is that I used buildNormalizedPath here though, which normalizes also the parts inside the paths, potentially changing a prefix length one might expect. I can change that to just buildPath, but shorter paths are usually better for the compiler invocations, especially on windows, where there is a path limit, but also for logging to the user in verbose mode + some parts of DUB deduplicate based on normalized paths (which is also a reason why I opened this PR, because with changed cwd it didn't do a correct job when compiling)

source/dub/internal/temp_files.d
```diff
@@ -34,4 +42,4 @@ static ~this()
 		if (spath.exists)
 			std.file.remove(spath);
 	}
-}
```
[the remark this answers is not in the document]
WebFreak001: fixed

PR #2583, opened 2023-02-09
Make `dub run --root=x` default app CWD in dub CWD

### LINKED ISSUE #2582: App's CWD changed when running DUB with `--root=xyz`
Since #2562 running applications using `dub run --root=xyz` will now run them in a different working directory, which could be considered a breaking change. Should we add a workaround to make it behave like the old way again? @Geod24 

Sample setup:
```
~/app/
    dub.sdl
```

the source code just printing the working directory.

before:
```
cd ~
dub run --root=app
# prints ~
```

now:
```
cd ~
dub run --root=app
# prints ~/app/
```

WebFreak001:
Fix #2582

cc @Geod24 

@ibuclaw should I target this on stable, so we can make a hotfix release of DUB with this fix? Last release contained a regression / change in behavior as described in that issue, which this PR fixes and tests.

Note: I'm not fully confident in this solution until the CI has passed, but new tests pass, I tried some CWD tests and general usage seems to work.

### DIFF
--- a/changelog/cwd-fix.dd
+++ b/changelog/cwd-fix.dd
@@ -1,0 +1,12 @@
+Reverted v1.31.0 working directory change when using `dub run --root=<value>`
+
+DUB < v1.31.0 would run applications in the working directory it was invoked in,
+ignoring the `--root` argument.
+
+In v1.31.0, `dub` started to respect the `--root` argument and run programs in
+the requested directory.
+
+While sometimes desirable, this change was not intended, and has now been
+reverted. To restore the 1.31.0 behaivor, set `"workingDirectory"` to `"."` in
+the dub.json/dub.sdl file. This works both with older DUB versions and with
+future DUB versions.
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1211,6 +1211,8 @@ class GenerateCommand : PackageBuildCommand {
 		if (!gensettings.config.length)
 			gensettings.config = m_defaultConfig;
 		gensettings.runArgs = app_args;
+		// legacy compatibility, default working directory is always CWD
+		gensettings.overrideToolWorkingDirectory = getWorkingDirectory();
 
 		logDiagnostic("Generating using %s", m_generator);
 		dub.generateProject(m_generator, gensettings);
--- a/test/dub-as-a-library-cwd/dub.json
+++ b/test/dub-as-a-library-cwd/dub.json
@@ -1,5 +1,6 @@
 {
 	"name": "dub-as-a-library-cwd",
+	"workingDirectory": ".",
 	"dependencies": {
 		"dub": {
 			"path": "../.."
--- a/test/dub-custom-root-2/.gitignore
+++ b/test/dub-custom-root-2/.gitignore
@@ -1,0 +1,6 @@
+.dub
+docs.json
+__dummy.html
+*.o
+*.obj
+/target-exe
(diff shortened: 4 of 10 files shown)


ibuclaw:
@WebFreak001 bug and regression fixes should always target stable really.

WebFreak001:
targetting stable, added changelog entry

WebFreak001:
forgot to push the actual fix in the rebase, but CI seems to pass? Need to fix my test in this case

source/dub/commandline.d
```diff
@@ -1211,6 +1211,8 @@ class GenerateCommand : PackageBuildCommand {
 		if (!gensettings.config.length)
 			gensettings.config = m_defaultConfig;
 		gensettings.runArgs = app_args;
+		// legacy compatibility, default working directory is always CWD
+		gensettings.overrideToolWorkingDirectory = getWorkingDirectory();
```
Geod24, line 1215: I think `--root` should be respected (in the long run). So reverting the behavior is just to avoid breaking code. What is our path forward ?

changelog/cwd-fix.dd
```diff
@@ -0,0 +1,14 @@
...
+in the `dub` command line tool the previous behavior was when using
+`dub run --root=path`, the built app that is being run would be run from dubs
+current cwd, unless `workingDirectory` is specified in the recipe. Now since the
+last release this changed to the package directory (e.g. like defaulting
+`"workingDirectory": "."` in dub.json), which is considered a breaking change /
+regression.
+
+So this release reverts back the behavior that was accidentally changed, so that
+the `dub` CLI working directory is always being used as program cwd when no
+working directory is set in the package recipe.
```
Geod24, line 14:
```
Reverted v1.31.0 working directory change when using `dub run --root=<value>`

dub < v1.31.0 would run applications in the working directory it was invoked in, ignoring the `--root` argument.
In v1.31.0, `dub` started to respect the `--root` argument and run programs in the requested directory.
While desirable, this change was not intended, and has now been reverted. It will be re-introduced as part of a major version upgrade.
```

Let me know if that works for you. I tried to rephrase the text keeping the important informations you already put in, but doing it more succinctly.

Geod24 (commented):
So does `workingDirectory: "."` give the right (respects `--root`) or wrong behavior ?

source/dub/commandline.d
WebFreak001: I think it's sane to keep it like this forever.

Using `workingDirectory "."` you get the behavior that workingDirectory == package directory, which you had with the breaking behavior here. So I think it's fine really

changelog/cwd-fix.dd
WebFreak001: I have adjusted it to mostly that, but omitted that it would be re-introduced. I think the workingDirectory: "." setting is a good universal solution to cover all use cases. (those who want the dub cwd as part of some CI run and those who want to control it to their package)

ibuclaw:
> should we merge this then?
> 
> @ibuclaw I don't know how the release cycle on DUB works and you have usually done it in the past, would we just tag a new minor release with this fix?

I'll tag the current head of stable. How soon should the fix go out?

WebFreak001:
it's a regression fix, so as soon as possible would be nice. I know it breaks the CI in serve-d as well, but it's an easy workaround for me. It may not be such an easy fix for other people and in general not a nice experience if you suddenly need to change your tooling code that has worked for years.

Geod24 (approved):

Issue #2577, opened 2023-02-01
Build fails with gdc 12.2.x


ximion:
### System information

- **dub version**: 1.30 & master
- **OS Platform and distribution**: Linux (Debian)
- **compiler version** gdc 12.2.0

### Bug Description

When trying to compile dub using GDC, the build fails while the previous 1.29 version built fine.
Error message:
```
Command `["gdmd", "-of/tmp/dub/bin/dub", "-I/tmp/dub/source", "-version=DubUseCurl", "-version=DubApplication", "-g", "-O", "-w", "@build-files.txt"]` failed, output was:
Serializing composite type Flags!(BuildRequirement) which has no serializable fields
Serializing composite type Flags!(BuildOption) which has no serializable fields
source/dub/internal/dyaml/stdsumtype.d:2053:13: error: static assert:  "`handlers[0]` of type `template` never matches"
 2053 |             static assert(matches[].canFind(hid),
      |             ^
source/dub/internal/dyaml/stdsumtype.d:1717:52: note: instantiated from here: ‘matchImpl!(const(SumType!(NativePath, Version)))’
 1717 |         return matchImpl!(Yes.exhaustive, handlers)(args);
      |                                                    ^
source/dub/commandline.d:2319:43: note: instantiated from here: ‘match!(const(SumType!(NativePath, Version)))’
 2319 |                                 ovr.target.match!(
      |                                           ^
```

### How to reproduce?
Try to build dub using gdc

### Expected Behavior
Either the build succeeds as it did before, or dub emits a message that the compiler/version is unsupported for building dub.

ibuclaw:
Bad commit 8a25429f80cd7b6e1e34504b39f9219537fc5320

ximion:
If there was a (even temporary) fix available soonish, we could still get dub 1.30 into the next Debian release. Otherwise this will need a freeze exception, and even though it's a low popularity package, it technically also is a toolchain package which means it's not guaranteed to get that exception (switching to LDC would mean loosing architecture support, which comes with its own administrative issues).

WebFreak001:
building master seems to work on my machine with `gdc (GCC) 12.2.1 20230111`

ximion:
It fails here with the same error, using `gdc (Debian 12.2.0-14) 12.2.0` - did the .1 patch release change anything maybe?

ibuclaw:
> building master seems to work on my machine with `gdc (GCC) 12.2.1 20230111`

[...]

PR #2584, opened 2023-02-09
PackageManager: Don't use deprecated interface to PackageOverride

Geod24:
This should trigger a deprecation message with DMD, but doesn't. On the other hand, it triggers a linker error on GDC.

Related to #2577 
@ximion : Could you test this ?


ibuclaw:
> This should trigger a deprecation message with DMD

Because the deprecation is gagged.  Try building dub with `-de` and you'll get the same build error as gdc.

Geod24:
Should it be gagged ? If so, why ?

ibuclaw:
> Should it be gagged ? If so, why ?

Because it is template matching?  Haven't looked into the weeds of it though.

PR #2586, opened 2023-02-10
add pr_info GitHub actions run

WebFreak001:
Gives summary about some basic statistics, deprecations and warnings

### DIFF
--- a/.github/workflows/pr_info.yml
+++ b/.github/workflows/pr_info.yml
@@ -1,0 +1,67 @@
+name: PR Info
+
+# This workflow builds the whole project once and:
+# - comments build deprecations/warnings (highlighting new ones since last tested PR)
+
+on:
+  pull_request:
+    branches:
+      - master
+      - stable
+
+jobs:
+  pr_info:
+    name: PR Info
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_info.yml
```diff
@@ -0,0 +1,67 @@
...
+      - master
+      - stable
+
+jobs:
+  pr_info:
+    name: PR Info
+    runs-on: ubuntu-20.04
+    steps:
+    - name: 'Prepare sticky comment'
+      uses: marocchino/sticky-pull-request-comment@v2
```
WebFreak001, line 18: should we use an exact commit / version with this action or trust on them not introducing malicious content?

I think an exact commit would be more secure.

(2 occurrences of this)

rikkimax:
It looks like there is some sort of permission problem and hence isn't running right now, so not currently mergeable until someone with admin privileges fixes it (I think).

WebFreak001:
no, the problem is that I PRd from my repository, which has a different access token - so the comment action fails to run here because running modified CI scripts in the security context of the target repository would just be broken.

It will start working once merged, see the linked PR

rikkimax:
Okay looks good +1

Issue #2538, opened 2022-11-27
Build failure when linking DUB dependencies with long names


mbierlee:
### System information
- **dub version**: DUB version 1.30.0, built on Nov 14 2022
- **OS Platform and distribution**: Windows 10
- **compiler version** DMD32 D Compiler v2.101.0-dirty

### Bug Description
When adding a dub project with a very long name, a linker error occurs. In the following example it tries to build a dependency called `poodinis-mirage-config-injector`:
```
Build directory C:\Users\[..]\AppData\Local\dub\packages\poodinis-mirage-config-injector-1.0.1\poodinis-mirage-config-injector\.dub\build\library-debug-windows-x86_64-dmd_v2.101.0-dirty-AD9F674B70EEEA986AF59B29018C0DC50245C96E020D599F9A3615568E3B78BA\ is not writable. Falling back to direct build in the system's temp folder.
    Building poodinis-mirage-config-injector 1.0.1 [library]
    Building dublongpaths ~master: building configuration [application]
     Linking dublongpaths
lld-link: error: undefined symbol: _D8poodinis13valueinjector6mirage10loadConfigFOCQBu9container19DependencyContainerxAyaZv
>>> referenced by D:\Temp\dublongpaths\source\app.d:22
>>>               .dub\build\application-debug-windows-x86_64-dmd_v2.101.0-dirty-BEE0070E13F5C7E5928CD6E33EABF03EF415566C2C52A4FE23C21BB581FD23B2\dublongpaths.obj:(_Dmain)

lld-link: error: undefined symbol: _D8poodinis13valueinjector6mirage12__ModuleInfoZ
>>> referenced by .dub\build\application-debug-windows-x86_64-dmd_v2.101.0-dirty-BEE0070E13F5C7E5928CD6E33EABF03EF415566C2C52A4FE23C21BB581FD23B2\dublongpaths.obj:(_D3app12__ModuleInfoZ)
Error: linker exited with status 1
Error D:\DLang\dmd2\windows\bin\dmd.exe failed with exit code 1.
```

This is possibly related to issue #911. The dependency fails to compile so the linker fails to find the lib.

### How to reproduce?
See project: [dublongpaths.zip](https://github.com/dlang/dub/files/10099155/dublongpaths.zip)
Just `dub run` it.

mbierlee:
[...]

Issue #911, opened 2016-08-17
Long file paths crash compilation on windows


WebFreak001:
When I compile my dub package in a nested folder somewhere in like `C:\Users\MyUsername\Development\My_D_Projects\From2016\Old\my-super-awesome-project-with-some-long-filename` and I have some dependency in it like `experimental_allocator` it will crash with this error message:

```
Error: Error writing file '..\..\..\..\Roaming\dub\packages\experimental_allocator-2.70.0-b1\experimental_allocator\.dub\build\library-release-windows-x86-dmd_2071-09237FFF1A90ED11A7976B16E7AB1A89\experimental_allocator.lib'
dmd failed with exit code 1.
```

On Windows there is by default a file path length limit of 259 characters and either dub or dmd aren't using the extended windows paths using `\\?\` which increases the length limit to 32767 characters. So Windows doesnt allow you to write the file.

Reproducable project in my case:

``` bat
cd %appdata%
cd ..
cd Local\Temp
mkdir workspaced-install-636070322740907409
cd workspaced-install-636070322740907409
git clone https://github.com/Hackerpilot/DCD.git DCD
cd DCD
dub build --build=release --config=client --force
```

crash

WebFreak001:
Another issue which probably has the same cause:
```
libdparse ~master: building configuration "library"...
Error: failed to create path to file: .dub\obj\libdparse\.dub\build\library-debug-windows-x86_64-ldc_0-DCDD4FDB2EDFDAE1509F1E37E3AAEA55\dparse.lib
no such file or directory
ldc2 failed with exit code 1.
```

WebFreak001:
Workaround for Windows 10: in the registry set `HKLM\SYSTEM\CurrentControlSet\Control\FileSystem LongPathsEnabled` to 1 and then reboot or restart the command line and editor

helgaros:
I think it's useful to use the long path tool program

s-ludwig:
Optlink can surely be added to the list, too. I think the DUB fix for this should happen in Phobos. There is no reason to let this ugly piece of WinAPI design leak through to all application code in a cross platform environment.

s-ludwig:
Just mentioning that a tiny improvement to this has been merged: #1242

WebFreak001:
fixed by #2589

PR #2589, opened 2023-02-12
make build names much smaller

WebFreak001:
Should fix windows path name limitations in most cases (unless your package name is very long)

Should improve verbose log output / the linker line in regular output + debugging paths in broken builds and compiler invocations.

In the unittest we just pick a smaller name to fix the windows test runner.

Moved platforms (linux.posix) + architecture (x86_64) + compiler (dmd_v2.102.0) out of the filename into the hash. While they may be interesting, I think, especially architecture and such, are rather internal. So I decided to un-expose those in the filenames.

From the sha256 we only take half the hash now (should be plenty of bytes still) + instead of hex strings, which are 200% of byte size, we use base64 (url, with padding removed), which is only 133% of byte size.

Sample in naming changes:

```
/home/webfreak/.dub/cache/PACKAGE_NAME/~master/build/application-debug-linux.posix-x86_64-dmd_v2.102.0-D95E91464418B72BDBF518DAA90BF13D4018E81A07B52C64F0C5EB6A60977D58/TARGET_NAME
->
/home/webfreak/.dub/cache/PACKAGE_NAME/~master/build/application-debug-aCpueoVTBosJsJSHQ-aldQ/TARGET_NAME
```

@Geod24 do you think this is a breaking change? I think at least the checksum could be halfed in length without incompatibilities.

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -34,19 +34,21 @@ string getObjSuffix(const scope ref BuildPlatform platform)
 
 string computeBuildName(string config, in GeneratorSettings settings, const string[][] hashing...)
 {
-	import std.digest.sha : SHA256, toHexString;
+	import std.digest.sha : SHA256;
+	import std.base64 : Base64URL;
 
 	SHA256 hash;
 	hash.start();
 	void addHash(in string[] strings...) { foreach (s; strings) { hash.put(cast(ubyte[])s); hash.put(0); } hash.put(0); }
 	foreach(strings; hashing)
 		addHash(strings);
-	const hashstr = hash.finish().toHexString();
+	addHash(settings.platform.platform);
+	addHash(settings.platform.architecture);
+	addHash(settings.platform.compiler);
+	addHash(settings.platform.compilerVersion);
+	const hashstr = Base64URL.encode(hash.finish()[0 .. $ / 2]).stripRight("=");
 
-    return format("%s-%s-%s-%s-%s_v%s-%s", config, settings.buildType,
-			settings.platform.platform.join("."),
-			settings.platform.architecture.join("."),
-			settings.platform.compiler, settings.platform.compilerVersion, hashstr);
+	return format("%s-%s-%s", config, settings.buildType, hashstr);
 }
 
 class BuildGenerator : ProjectGenerator {
--- a/test/issue2051_running_unittests_from_dub_single_file_packages_fails.d
+++ b/test/issue2051_running_unittests_from_dub_single_file_packages_fails.d
@@ -1,5 +1,5 @@
 /+ dub.sdl:
-   name "issue2051_running_unittests_from_dub_single_file_packages_fails"
+   name "issue2051"
  +/
 
 import std.algorithm : any;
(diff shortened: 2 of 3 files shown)


WebFreak001:
@ibuclaw fixes the issue you commented about

ibuclaw (approved):

Geod24:
When I first did the change I considered reducing this as well. At the time I didn't envision the Windows path issue, so the tradeoff didn't seem worth it. How long do we have to get this in @ibuclaw ?

WebFreak001:
other advantages than windows path issues:
- verbose log is more readable
- compiler invocations are more readable

disadvantages of this change:
- can't look into internal build files to see which compiler was used, etc.

@atilaneves are you using these path based informations somewhere in your DUB wrappers perhaps?

ibuclaw:
@Geod24 if you want this in the next point release, until the 15th.  I've just tagged the beta now. :-)

Geod24:
From the top of my head, what I found most useful was to know which version of the compiler a project was compiled with.
It becomes less important with a GC for the cache (and I just realize that GCing the cache might need to use those informations).

ibuclaw:
FAOD, the 2.103 stable cycle begins on 1st March, so if you'd rather defer to next major release, that's fine as well - there's absolutely no pressure to rush things into 2.102 minor releases if they aren't necessary.

WebFreak001:
I would suggest we could wait with this for the next minor release, not a patch release. It's not a terribly important feature and it's a new feature, so it shouldn't actually really go into a patch release.

WebFreak001:
will fix #911, fix #2538

atilaneves:
No, reggae does its own thing.

WebFreak001:
ok let's merge this then

Geod24:
@WebFreak001 : Thanks for fixing this. We have plenty of evidence that it triggers too often for user, so glad to see it's going to be in the next release.

kinke:
Yeppa, thanks a lot!

source/dub/generators/build.d
```diff
@@ -34,19 +34,21 @@ string getObjSuffix(const scope ref BuildPlatform platform)
...
+	addHash(settings.platform.architecture);
+	addHash(settings.platform.compiler);
+	addHash(settings.platform.compilerVersion);
+	const hashstr = Base64URL.encode(hash.finish()[0 .. $ / 2]).stripRight("=");
 
-    return format("%s-%s-%s-%s-%s_v%s-%s", config, settings.buildType,
-			settings.platform.platform.join("."),
-			settings.platform.architecture.join("."),
-			settings.platform.compiler, settings.platform.compilerVersion, hashstr);
+	return format("%s-%s-%s", config, settings.buildType, hashstr);
```
kinke, line 51: As an aside - why is `buildType` there? I'm especially wondering about `unittest` - AFAIK, that only really applies to the main package. But all *dependencies* then get `unittest` too, although they aren't compiled with `-unittest` etc., so might likely have to be recompiled even if a suitable regular `debug` build might be in the build cache already.

PR #2592, opened 2023-02-13
Add OpenBSD package to README.md

ibara:
Includes installation instructions.


thewilsonator (approved):

WebFreak001:
linked upstream package tracker from dlang.org has shut down: https://openports.se/ https://openports.se/lang/dmd

ibara:
You shouldn't use openports.se -- it is not official and it is never worked correctly when it was working.
Learning that it has now shut down is the best news I've heard today.

Repology.org is a much better, cross-platform solution: https://repology.org/project/dub/versions

PR #2593, opened 2023-02-13
Fix #2574: Suggest alternative spelling when a command is mistyped

### LINKED ISSUE #2574: Mistyping commands
### Bug Description

I very often mistype:
- `dub ugprade`  instead of `dub upgrade`
- `dub clean-caches` instead of `dub clear-caches`

### Expected Behavior

Instead of displaying the full help text, suggest a more appropriate command name, like git does.

```
C:\Users\guill\Desktop\Dplug>git puke
git: 'puke' is not a git command. See 'git --help'.

The most similar command is
        prune
```

elvodqa:
Suggesting similar command names

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -465,6 +465,8 @@ int runDubCommandLine(string[] args)
 	}
 
 	if (cmd is null) {
+		logInfoNoTag("USAGE: dub [--version] [<command>] [<options...>] [-- [<application arguments...>]]");
+		writeln();
 		logError("Unknown command: %s", command_name_argument.value);
 		import std.algorithm.iteration : filter;
 		import std.uni : toUpper;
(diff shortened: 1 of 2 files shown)


WebFreak001:
problem: unknown commands should not print the whole usage, only a shortened version

e.g. right now it does:

```
       Error Unknown command: upfrade
             Did you mean 'upgrade'?

USAGE: dub [--version] [<command>] [<options...>] [-- [<application arguments...>]]

Manages the DUB project in the current directory. If the command is omitted,
DUB will default to "run". When running an application, "--" can be used to
separate DUB options from options passed to the application.
...
```

but this would be better:

```
USAGE: dub [--version] [<command>] [<options...>] [-- [<application arguments...>]]

       Error Unknown command: upfrade
             Did you mean 'upgrade'?

```

WebFreak001:
can you add `USAGE: dub [--version] [<command>] [<options...>] [-- [<application arguments...>]]` at the top of the output? (using `logInfoNoTag`)

Geod24 (approved):

PR #2604, opened 2023-02-27
Make getDUBExePath returns a NativePath

Geod24:
Fairly simple diff. Trying to reduce dependency on string being used for paths.
Note that no deprecation is necessary here, as this is an internal function (`dub.internal.utils : getDUBExePath`).


WebFreak001 (approved):

WebFreak001:
CI failure will be there for a while, since dmd master removed dip25, which breaks the dub tests

PR #2614, opened 2023-03-05
Make all static library exports contribute to dlls on Windows

rikkimax:
Fixes: https://github.com/ldc-developers/ldc/issues/4343

### DIFF
--- a/changelog/static_libraries_exported_dll.dd
+++ b/changelog/static_libraries_exported_dll.dd
@@ -1,0 +1,8 @@
+Static libraries now contribute towards DLL exports on Windows
+
+Previously if you did not explicitly pull in an object file within a static library, it was elided by the linker automatically.
+This pulls them in automatically for linkers compatible with Microsoft's linker via the ``/WHOLEARCHIVE:file`` flag. Supports LLD.
+
+It does not affect executables, although DLL's being built as dependencies by DUB will include it.
+
+If you have previously used a linker script (.def) or ``/WHOLEARCHIVE`` you may be able to remove them from your builds.
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -322,4 +322,10 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 				    || arg.startsWith("-mtriple=");
 		}
 	}
+
+	bool isWindowsCOFF(in BuildPlatform platform)
+	{
+		// What will happen on ARM Windows? Who knows. Once LDC ships for ARM, lets find out!
+		return platform.isWindows();
+	}
 }
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -217,6 +217,9 @@ interface Compiler {
 
 		return build_platform;
 	}
+
+	/// Given a platform specification, determine if a compiler is on Windows and PE-COFF with MSVC link compatible linker.
+	bool isWindowsCOFF(in BuildPlatform platform);
 }
 
 private {
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -434,4 +434,10 @@ config    /etc/dmd.conf
 				    || arg.startsWith("-defaultlib=");
 		}
 	}
+
+	bool isWindowsCOFF(in BuildPlatform platform)
+	{
+		// x86_omf and x86_mscoff shouldn't be something you have to worry about here, but just in case something leaks
+		return platform.isWindows && platform.architecture.canFind("x86", "x86_64", "x86_mscoff") && !platform.architecture.canFind("x86_omf");
+	}
 }
--- a/source/dub/compilers/gdc.d
+++ b/source/dub/compilers/gdc.d
@@ -259,6 +259,11 @@ class GDCCompiler : Compiler {
 
 		return  dflags;
 	}
+
+	bool isWindowsCOFF(in BuildPlatform platform)
+	{
+		return false;
+	}
 }
 
 private string extractTarget(const string[] args) { auto i = args.countUntil("-o"); return i >= 0 ? args[i+1] : null; }
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -97,7 +97,32 @@ class BuildGenerator : ProjectGenerator {
 		const copyDynamicLibDepsLinkerFiles = rootTT == TargetType.dynamicLibrary || rootTT == TargetType.none;
 		const copyDynamicLibDepsRuntimeFiles = copyDynamicLibDepsLinkerFiles || rootTT == TargetType.executable;
 
-		bool[string] visited;
+		// Check to see if given a compiler and platform target
+		//  are Windows and linking using a MSVC link compatible linker.
+		const isWindowsCOFF = settings.compiler.isWindowsCOFF(settings.platform);
+
+		bool[string] visited, visitedStaticInDll;
+
+		void visitStaticLibsInDll(ref BuildSettings bs, string target) {
+			if (target in visitedStaticInDll) return;
+			visitedStaticInDll[target] = true;
+
+			auto ti = targets[target];
+			if (ti.buildSettings.targetType != TargetType.staticLibrary)
+				return;
+
+			const ldepPath = target_paths[target].toNativeString();
+
+			// Add the MSVC link /WHOLEARCHIVE flag with static library path passed in
+			//  the purpose of this is to allow all exports from a static library to contribute
+			//  towards the dll's exports.
+			bs.addLFlags("/WHOLEARCHIVE:" ~ ldepPath);
+
+			foreach (ldep; ti.linkDependencies) {
+				visitStaticLibsInDll(bs, ldep);
+			}
+		}
+
 		void buildTargetRec(string target)
 		{
 			if (target in visited) return;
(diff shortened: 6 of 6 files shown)


rikkimax:
Looks like dmd-master failing isn't related to this PR.

source/dub/compilers/ldc.d
```diff
@@ -322,4 +322,9 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 				    || arg.startsWith("-mtriple=");
 		}
 	}
+
+	bool isWindowsCOFF(in BuildPlatform platform)
+	{
+		return platform.isWindows && platform.architecture.canFind("x86", "x86_64");
```
thewilsonator, line 328: What about ARM bed windows? This should just `return platform.isWindows;`, no?

changelog/static_libraries_exported_dll.dd
```diff
@@ -0,0 +1,8 @@
+Static libraries now contribute towards DLL exports on Windows
+
+Previously if you did not explicitly pull in an object file within a static library, it was elided by the linker automatically.
+This pulls them in automatically for linkers compatible with Microsoft's linker via the ``/WHOLEARCHIVE:file`` flag. Supports LLD.
+
+It does not affect executables, although DLL's being built as dependencies by DUB will include it.
+
+If you have previously used a linker script (.def) or ``/WHOLEARCHIVE`` you may be able to remove them from your builds.
```
thewilsonator, line 8: EOL before EOF

thewilsonator (commented):
Any way to test this?

source/dub/compilers/ldc.d
rikkimax: ARM idk about. I have no experience there, no way to test or any idea what the architecture string would be.

No its MSVC linker or compatible linker, PE-COFF only. So no Optlink.

rikkimax:
> Any way to test this?

The last time I looked at the test suite it didn't run on Windows and it wouldn't be the simplest of test to write, since you'd need a static lib + DLL + something to verify symbols were exported.

Scream test might be the best we get, and people who use it will certainly scream quickly if it's not in.

source/dub/compilers/ldc.d
rikkimax: LDC doesn't ship ARM builds, has dub even been tested on ARM Windows?

thewilsonator: > So no Optlink.

Correct, but LDC doesn't support that anyway IIRC, or is DUB still capable of passing that to LDC  and still need to be filtered out?

> LDC doesn't ship ARM builds,

Cross-compilation.

> has dub even been tested on ARM Windows?

No idea.

> No its MSVC linker or compatible linker, PE-COFF only.

More silly questions: what does LLD (e.g. `-link-internally`) fall into then? 
does it change its command line to be MSVC compatible on windows or is there a separate driver for it?

thewilsonator:
> Scream test might be the best we get, and people who use it will certainly scream quickly if it's not in.

Lol, fair enough.

source/dub/compilers/ldc.d
rikkimax: Gotcha, yeah it can probably be shrunk down a bit, will do.

LLD is MSVC link compatible (its in changelog that it is).

> $ lld-link /?
> OVERVIEW: LLVM Linker
> 
> USAGE: C:\Tools\D\dmd_2.102.0\windows\bin64\lld-link.exe [options] file...
> 
> OPTIONS:
>   /align:<value>          Section alignment
>   /aligncomm:<value>      Set common symbol alignment
>   /allowbind:no           Disable DLL binding
>   /allowbind              Enable DLL binding (default)
>   /allowisolation:no      Disable DLL isolation
>   /allowisolation         Enable DLL isolation (default)
>   /alternatename:<value>  Define weak alias
>   /appcontainer:no        Image can run outside an app container (default)
>   /appcontainer           Image can only be run in an app container
>   /base:<value>           Base address of the program
>   /Brepro                 Use a hash of the executable as the PE header timestamp
>   --color-diagnostics=<value>
>                           Use colors in diagnostics; one of 'always', 'never', 'auto'
>   --color-diagnostics     Use colors in diagnostics
>   /debug:<value>          Embed a symbol table in the image with option
>   /debugtype:<value>      Debug Info Options
>   /debug                  Embed a symbol table in the image
>   /def:<value>            Use module-definition file
>   /defaultlib:<value>     Add the library to the list of input files
>   /delayload:<value>      Delay loaded DLL name
>   /demangle:no            Do not demangle symbols in output
>   /demangle               Demangle symbols in output (default)
>   /dll                    Create a DLL
>   /driver:<value>         Generate a Windows NT Kernel Mode Driver
>   /dynamicbase:no         Disable ASLR (default when /fixed)
>   /dynamicbase            Enable ASLR (default unless /fixed)
>   /entry:<value>          Name of entry point symbol
>   /errorlimit:<value>     Maximum number of errors to emit before stopping (0 = no limit)
>   /export:<value>         Export a function
>   /failifmismatch:<value>
>   /filealign:<value>      Section alignment in the output file
>   /fixed:no               Enable base relocations (default)
>   /fixed                  Disable base relocations
>   /force:multipleres      Allow multiply defined resources when creating executables
>   /force:multiple         Allow multiply defined symbols when creating executables
>   /force:unresolved       Allow undefined symbols when creating executables
>   /force                  Allow undefined and multiply defined symbols when creating executables
>   /functionpadmin:<value> Prepares an image for hotpatching
>   /guard:<value>          Control flow guard
>   /heap:<value>           Size of the heap
>   /highentropyva:no       Disable 64-bit ASLR
>   /highentropyva          Enable 64-bit ASLR (default on 64-bit)
>   /ignore:<value>         Specify warning codes to ignore
>   /implib:<value>         Import library name
>   /include:<value>        Force symbol to be added to symbol table as undefined one
>   /includeoptional:<value>
>                           Add symbol as undefined, but allow it to remain undefined
>   /incremental:no         Overwrite import library even if contents are unchanged
>   /incremental            Keep original import library if contents are unchanged
>   /integritycheck:no      No effect (default)
>   /integritycheck         Set FORCE_INTEGRITY bit in PE header
>   /largeaddressaware:no   Disable large addresses (default on 32-bit)
>   /largeaddressaware      Enable large addresses (default on 64-bit)
>   /libpath:<value>        Additional library search path
>   /lib                    Act like lib.exe; must be first argument if present
>   /linkrepro:<value>      Dump linker invocation and input files for debugging
>   /lldltocache:<value>    Path to ThinLTO cached object file directory
>   /lldltocachepolicy:<value>
>                           Pruning policy for the ThinLTO cache
>   /lldsavetemps           Save temporary files instead of deleting them
>   /machine:<value>        Specify target platform
>   /manifest:<value>       NO disables manifest output; EMBED[,ID=#] embeds manifest as resource in the image
>   /manifestdependency:<value>
>                           Attributes for <dependency> element in manifest file; implies /manifest
>   /manifestfile:<value>   Manifest output path, with /manifest
>   /manifestinput:<value>  Additional manifest inputs; only valid with /manifest:embed
>   /manifestuac:<value>    User access control
>   /manifest               Create .manifest file
>   /merge:<value>          Combine sections
>   /mllvm:<value>          Options to pass to LLVM
>   /natvis:<value>         Path to natvis file to embed in the PDB
>   /no-color-diagnostics   Do not use colors in diagnostics
>   /nodefaultlib:<value>   Remove a default library
>   /nodefaultlib           Remove all default libraries
>   /noentry                Don't add reference to DllMainCRTStartup; only valid with /dll
>   /nxcompat:no            Disable data execution provention
>   /nxcompat               Enable data execution prevention (default)
>   /opt:<value>            Control optimizations
>   /order:<value>          Put functions in order
>   /out:<value>            Path to file to write output
>   /pdb:<value>            PDB file path
>   /pdbaltpath:<value>     PDB file path to embed in the image
>   /pdbsourcepath:<value>  Base path used to make relative source file path absolute in PDB
>   --rsp-quoting=<value>   Quoting style for response files, 'windows' (default) or 'posix'
>   /safeseh:no             Don't produce an image with Safe Exception Handler
>   /safeseh                Produce an image with Safe Exception Handler (only for x86)
>   /section:<value>        Specify section attributes
>   /stack:<value>          Size of the stack
>   /stub:<value>           Specify DOS stub file
>   /subsystem:<value>      Specify subsystem
>   /swaprun:cd             Make loader run output binary from swap instead of from CD
>   /swaprun:net            Make loader run output binary from swap instead of from network
>   /swaprun:<value>        Comma-separated list of 'cd' or 'net'
>   /thinlto-emit-imports-files
>                           Emit .imports files with -thinlto-index-only
>   /thinlto-index-only:<value>
>                           -thinlto-index-only and also write native module names to file
>   /thinlto-index-only     Instead of linking, emit ThinLTO index files
>   /thinlto-object-suffix-replace:<value>
>                           'old;new' replace old suffix with new suffix in ThinLTO index
>   /thinlto-prefix-replace:<value>
>                           'old;new' replace old prefix with new prefix in ThinLTO outputs
>   /threads:no             Do not run the linker multi-threaded
>   /threads                Run the linker multi-threaded (default)
>   /timestamp:<value>      Specify the PE header timestamp
>   /tsaware:no             Create non-Terminal Server aware executable
>   /tsaware                Create Terminal Server aware executable (default)
>   /version:<value>        Specify a version number in the PE header
>   --version               Print version information
>   /wholearchive:<value>   Include all object files from this archive
>   /WX:no                  Don't treat warnings as errors
>   /WX                     Treat warnings as errors

thewilsonator (approved):

kinke:
This was merged too soon; this can have undesired side effects like larger binaries and link-time regressions. We've had a similar regression just a few weeks back at work - some library containing a module depending on some 3rd-party lib only available on Linux, and that module (totally unused on Windows) forcefully being linked in suddenly causing lots of undefined symbol errors. The really bad thing here is that unlike default `-fvisibility=public -dllimport=all`, these linker flags cannot be overridden/prevented. 

And there are other ways to work around this problem; `-dllimport=all` was designed for all other D library deps being DLLs. When really wanting to link other libs statically into the DLL, a `sourceLibrary` would work best in terms of generated code (but suboptimal regarding build times). Compiling the static lib as single object via `-singleobj` would also work in almost all cases, but come with similar disadvantages as `/WHOLEARCHIVE`.

rikkimax:
What we could do is add a build requirement to turn this behavior off for a given shared library. So if you need control you have it, but the default will still be it is most likely going to work.

Geod24:
No test, no review, let's revert: https://github.com/dlang/dub/pull/2616

PR #2615, opened 2023-03-06
On Windows support disabling of forcing of all object files in a static library to be in a DLL

rikkimax:
Augments https://github.com/dlang/dub/pull/2614

I'll do dub-docs once given all clear.

@kinke is there anything else you want me to do?

### DIFF
--- a/changelog/static_libraries_exported_dll.dd
+++ b/changelog/static_libraries_exported_dll.dd
@@ -6,3 +6,10 @@ This pulls them in automatically for linkers compatible with Microsoft's linker 
 It does not affect executables, although DLL's being built as dependencies by DUB will include it.
 
 If you have previously used a linker script (.def) or ``/WHOLEARCHIVE`` you may be able to remove them from your builds.
+
+Warning: This may have the unfortunate side affect of preventing previously working builds of DLLs to no longer link.
+If this occurs use the build requirement flags of ``allowDLLObjectFileEliding`` or ``allowDLLObjectFileElidingAll`` via the ``buildRequirements`` directive.
+The difference between the two is that the latter can be applied anywhere in the dependency tree and the former must be placed on the DLL package recipe or any of its static library dependencies.
+
+Warning: DLL's may grow in size compared to previous versions of dub due to all the object files being included by default.
+If it was working for you previously, you may allow eliding.
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -355,17 +355,19 @@ enum TargetType {
 }
 
 enum BuildRequirement {
-	none = 0,                     /// No special requirements
-	allowWarnings        = 1<<0,  /// Warnings do not abort compilation
-	silenceWarnings      = 1<<1,  /// Don't show warnings
-	disallowDeprecations = 1<<2,  /// Using deprecated features aborts compilation
-	silenceDeprecations  = 1<<3,  /// Don't show deprecation warnings
-	disallowInlining     = 1<<4,  /// Avoid function inlining, even in release builds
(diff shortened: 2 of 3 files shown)


thewilsonator (approved):

Geod24:
I have a few issues with this:
- Having separate things for project and package has historically worked very poorly;
- There are no tests;
- It's blankly stating that the previous PR introduced a regression;
- I'm not quite sure what even te problem is here, and why this is a good solution;

WebFreak001 (changes requested):
needs a test to at least demonstrate what this is even doing

kinke:
I think what almost all users would want is something like https://github.com/symmetryinvestments/dub/commit/f841c6dfe4bfdfe700b86db42277ab337dcc5034 - being able to override default `-fvisibility=public -dllimport=all` (designed for a scenario where every dub dependency is a DLL) with e.g. `-fvisibility=hidden -dllimport=defaultLibsOnly` for an isolated DLL (potentially containing static lib deps, and only exporting explicitly exported symbols).

Geod24:
@kinke : We need to get the flag propagation fixed. I had the same problem with `-preview=in` and that's why we have the ugly dependency `dflags` hack.

kinke:
Yeah - including all those duplicates when a flag is passed down to all deps and then passed upwards again. I've seen *pages* (on a FullHD monitor) of identical flags in ugly-ass cmdlines.

rikkimax:
What I'm thinking is we need to have a directive to set the visibility. So that dub can actually understand it and add behavior like dllimport (which is kinda a hack we, unfortunately, need currently) and ``/WHOLEARCHIVE:``.

If we combine that directive with my work and add a way to disallow it, not just allow eliding, that should cover our basis here.

I think I'll leave it to you @kinke to get this to completion.

rikkimax:
Problem is, once it is needed you have people who don't know what a linker is asking why it doesn't work. When you give them an explanation they complain very loudly and are more likely to leave D even if it isn't our fault.

Ignoring it isn't a good solution.

kinke:
I probably shouldn't have changed the default compile-DLL flags with LDC on Windows with dub v1.30 yet, as it introduced such problems for existing DLL projects. It newly enables working with shared-lib dependencies, but broke at least some existing isolated DLLs.

But what's IMO important is taking the time to come up with a good, generic-as-possible solution. E.g., this specific problem could also be handled by making dub default to building all dependencies of a shared library project as shared libs themselves (recursively).

https://github.com/symmetryinvestments/dub/commit/f841c6dfe4bfdfe700b86db42277ab337dcc5034 then allows to customize exports and imports for a DLL or executable. I haven't tested that in production yet but will do, hopefully pretty soon. It would cover 2 problems we currently have at Symmetry:
* The main executable is linked against shared druntime/Phobos via `-link-defaultlib-shared`, so that we can load D plugins (shared libs) which also use that shared druntime/Phobos (incl. shared GC, shared thread lists etc. etc.). All static lib dependencies of the executable need to be compiled with `-dllimport=defaultLibsOnly` (or `-link-defaultlib-shared`, which defaults to that), and dub currently doesn't handle that.
* We have about 25 plugins - pretty isolated shared libs compiled with `-fvisibility=hidden -dllimport=defaultLibsOnly`, so that they only export a few select symbols and only share druntime/Phobos. All static lib deps of these plugins need to be compiled with the same flags. They all depend on one base project, so by adding `-fvisibility=hidden -dllimport=defaultLibsOnly` for that single project and making sure these propagate up and down the dub dependencies graph, that is handled in a single central place.

rikkimax:
Okay so we've got something else in the works, which we need to evaluate
which is much broader in scope.

Makes sense to wait.

PR #2616, opened 2023-03-06
Revert "Make all static library exports contribute to dlls on Windows"

Geod24:
This reverts commit 076022c1fe7fe15614959c0d50d2866f1149f78f / PR2614.

See https://github.com/dlang/dub/pull/2614#issuecomment-1455370459 for why.


Geod24:
@rikkimax : Note that I'm only reverting because it was merged too quickly. Please re-submit a complete diff (or update the pending PR) and we can get it properly reviewed and merged. Thanks.

Issue #2618, opened 2023-03-08
TargetType none is crashing dub on macOS


MrcSnm:
### System information

- **dub version**: 1.31.1
- **OS Platform and distribution**: MacOS 12.6 Monterey
- **compiler version**: 2.102.1

### Bug Description

I'm trying to pull my libraries using targetType none, but on macOS I get the following error:

`core.exception.AssertError@source/dub/compilers/dmd.d(303): No target name set.`

### How to reproduce?
```json
"name": "hipremeengine_ios",
	"authors": [
		"Marcelo Silva"
	],
	"targetType": "library",
	"targetPath": "libs",
	"copyright": "Copyright © 2023, Marcelo Silva",
	"dependencies": {
		"metal": {"path": "/Users/marcelosilva/Desktop/D/HipremeEngine/dependencies/metal/"}
	},

	"configurations": [
		{
			"name": "dependencies",
			"targetType": "none",
			"targetName": "metal",
			"postBuildCommands": [
				"cp /Users/marcelosilva/Desktop/D/HipremeEngine/dependencies/metal/libmetal.a libs"
			]
		}
	],
	"postBuildCommands": [
		"dub build -c dependencies"
	]
```

thewilsonator:
please run `dub` with `-v` and post the log of what it prints out. Also are you on ARM? is this a problem with Just DMD, or are LDC/GDC also affected?

MrcSnm:
I'm only using DMD since I'm also using Objective-C externs.
No, this is X86_64 for the while.

MrcSnm:
```
% dub -v 
Using dub registry url 'https://code.dlang.org/'
Note: Failed to determine version of package hipremeengine_ios at .. Assuming ~master.
  Version selection for dependency metal (metal) of hipremeengine_ios is missing.
  Adding local metal in /Users/marcelosilva/Desktop/D/HipremeEngine/dependencies/metal/
Note: Failed to determine version of package metal at .. Assuming ~master.
  Found dependency metal @/Users/marcelosilva/Desktop/D/HipremeEngine/dependencies/metal/
Generating using build
Configuring dependent hipremeengine_ios, deps:"metal"
  Configuring dependent metal, deps:
    Starting Performing "debug" build using /Library/D/dmd/bin/dmd for x86_64.
  Up-to-date metal ~master: target for configuration [library] is up to date.
[...]

PR #2619, opened 2023-03-09
add test case for #2618

WebFreak001:
only added a test case in this first commit to ensure that the tests fail when not changing any of dubs code, going to push the fix afterwards.

This only breaks in debug builds of dub / dub builds with asserts enabled. Luckily no functionality or logic needs to be changed, it's just a case of moving an assert down after an empty check and adding another empty check in a calling function, which makes the code logic more reasonable anyway.


Geod24:
Can you add a more descriptive commit / PR title ?

PR #2637, opened 2023-05-08
Exclude importPaths that start with ".."

CyberShadow:
> One use case for requiring such a declaration is when the module root is outside the Dub project root.  For example, the library "foo" containing a D module "foo.bar" may be in a directory called "foo" and have "bar.d" at the directory root (same directory as "dub.sdl").
>
> Currently, setting importPaths to ".." has the unpleasant side effect of scanning the entire parent directory (which may host other, unrelated projects).  Dub does this for reasons such as caching; the build process itself does not require a list of all files that may or may not be imported, as the compiler discovers them not by globbing, but by path construction and file existence checks, based on the names of imported modules and the list of import paths specified with -I.
>
> Ideally, Dub should NEVER recursively glob the importPaths list, and instead communicate with the compiler to discover the full list of files that were actually imported (e.g. from compilers' verbose output).  However, this change (which should not affect canonical use cases) should facilitate this particular directory structure.

What do you think?


WebFreak001:
can you add a test for this?

CyberShadow:
That's tricky, as the thing to test here is that Dub does not try to read this unrelated file / enter this unrelated directory. It does gracefully skip broken symlinks, so aside something like a logging / error-injecting FUSE filesystem I can't think of anything right now.

CyberShadow:
If I do `chmod 000` on a directory, that does cause current Dub to fail, but I don't think that would make for a very good test. For one, it's POSIX-only; for two, that should probably be fixed to skip unreadable directories anyway, in the same way that broken symlinks are.

CyberShadow:
Added it as I can't think of anything better right now, and I see the tests are POSIX-only anyway.

source/dub/recipe/packagerecipe.d
```diff
@@ -531,11 +531,24 @@ struct BuildSettingsTemplate {
...
+				.array
+			))
+			.assocArray;
+
  		// collect import files and remove sources
 		import std.algorithm : copy, setDifference;
 
 		auto importFiles =
-			chain(collectFiles(importPaths, "*.{d,di}"), collectFiles(cImportPaths, "*.h"))
+			chain(collectFiles(projectImportPath, "*.{d,di}"), collectFiles(cImportPaths, "*.h"))
```
WebFreak001, line 551: don't forget to also do this for cImportPaths

CyberShadow: Um... why? Well OK, maybe let's just do it for all paths. I don't see why not. Let's see what breaks.

WebFreak001: oh no, you definitely want to be able to specify source files with `..`, since for sub-packages they might want to share some kind of info file across all sub-packages that for example just contain some compile time helpers.

Also not all globs in parent directory scan recursively, but only specify single files, where this kind of exclusion would seem really arbitrary to the user.

WebFreak001:
here are some projects which rely on .. in globbing, to test if they still work properly:

for source paths:
- https://github.com/andrey-zherikov/argparse/blob/6fca01c219828b8223b64e1c7899e1cd7c2b2eb9/examples/completion/single_main/app/dub.json#L5

for source files:
- https://github.com/dokutoku/dxlib-d/blob/2e539d4dfda5804fed1a4a0c29957e7dd96f7819/examples/windows/example/dub.json#L20

for import paths:
- https://github.com/dlang-community/gitcompatibledubpackage/blob/master/dub.sdl#L22-L25

it seems that this is even intended documented behavior, although I don't get what it has to do with git submodules.

I think it would have made more sense that `importPaths ".."` would only scan its own directory, since it's used to make the DUB package name being used as module package name for all files within (and no source directory needed)

WebFreak001:
I think what we should do for import paths: (and C import paths, since they are the same thing for some compilers)
- make `".."` the only allowed import path that can escape the project root (means it will just pass that to the compiler, but only grep within the current project)
	- this requires a deprecation period if users used this for git submodules or things where it would find more files than what is in the package directory
	- this should ensure the folder name is the DUB package name
	- this should not affect regular DUB packages, since they are stored as `packagename/version/packagename/{project files}`, so the parent folder is known to only contain a single folder with D/C files, with a known and fixed name
- disallow all other relative globs that can escape the project root
- keep absolute globs allowed, for testing purposes, but probably deny them from being published to the registry
	- for users who want to actually use other relative escaping paths, one could use the $PACKAGE_DIR variable to build an absolute path like this

this way we also have proper semantics for importPaths ".." that users can utilize effectively and with solid guarantees.

WebFreak001:
> Perhaps this is the wrong approach entirely, and we need something like a moduleRoot declaration. So then, moduleRoot "foo" would tell Dub that a bar.d in the project root corresponds to the module foo.bar, and that the directory containing the project should be named foo, and that its parent directory should be added to the import (-I) paths; or even use the -mv compiler feature to make things align if the directory structure doesn't allow it.

I think `-mv` should be avoided by the user, I more see it as something that tools could use to do special purpose stuff, such as sharing the same dependency with multiple versions in a build. Mapping module structure to the file/folder structure makes it easier to reason about what happens and explain to people how things work.

moduleRoot could be one thing how this could be done, but I think keeping "name" for it already makes sense, although it breaks when users use `-` in their package name.

Do you have a requirement / use-case why you need to control the parent folder/current folder relationship?

CyberShadow:
> I think `-mv` should be avoided by the user, I more see it as something that tools could use to do special purpose stuff

Yeah sorry, I meant that Dub could use `-mv` under the hood to implement `moduleRoot`.

> moduleRoot could be one thing how this could be done, but I think keeping "name" for it already makes sense, although it breaks when users use `-` in their package name.

Well the Dub package name doesn't always correspond to the module name, and special symbols aren't the only case. Probably the most prominent example I'm aware of is `arsd_official`.

> Do you have a requirement / use-case why you need to control the parent folder/current folder relationship?

It's `ae` and projects structured like it (see that gist above). The structure allows using both Dub and Git submodules for dependency versioning.

Of course it would be nice to understand and fix the problem as generally as possible. For example, what about something like a package called `lib_foo` in a directory `foo` which holds modules in the `lib.foo` namespace, a package called `lib_bar` in a directory `bar` which holds modules in the `lib.bar` namespace... then, the equivalent `importPath` would be `"../.."`.

Issue #2567, opened 2023-01-17
How do I put comment in JSON now?


p0nce:
### System information
- **dub version**: DUB version 1.31.0-rc.1, built on Jan 17 2023
- **OS Platform and distribution**: Windows 11
- **compiler version** LDC 1.30.0

### Bug Description

DUB complains about non-existent keys in the schema, even those called `"comment"`.
But those JSON comments are there to avoid mistakes for the users of the project.

### How to reproduce?

Build this project => https://github.com/AuburnSounds/Dplug/tree/master/examples/ms-encode who has a "comment-stuff" key.

### Expected Behavior

A special way to write comment in dub.json, like we were advised to do. I don't want to go `.sdl`, it would disrupt users more.

### Logs

```
     Warning C:\Users\guill\Desktop\Dplug\examples\ms-encode\dub.json(15:4): comment: Key is not a valid member of this section. There are 53 valid keys: name, description, homepage, authors, copyright, license, toolchainRequirements, configurations, buildTypes, -ddoxFilterArgs, -ddoxTool, subPackages, version, dependencies, systemDependencies, targetType, targetPath, targetName, workingDirectory, mainSourceFile, subConfigurations, dflags, lflags, libs, sourceFiles, sourcePaths, excludedSourceFiles, injectSourceFiles, copyFiles, extraDependencyFiles, versions, debugVersions, versionFilters, debugVersionFilters, importPaths, stringImportPaths, preGenerateCommands, postGenerateCommands, preBuildCommands, postBuildCommands, preRunCommands, postRunCommands, environments, buildEnvironments, runEnvironments, preGenerateEnvironments, postGenerateEnvironments, preBuildEnvironments, postBuildEnvironments, preRunEnvironments, postRunEnvironments, buildRequirements, buildOptions
```

WebFreak001:
I think we could go with something like reserving underscore prefixed keys to be always comments - thoughts @Geod24 ?

On the other hand we could also start supporting regular JS-style comments. However right now there are most likely development apps reading dub.json files manually to look for certain keys, so we wouldn't want to break those. To properly read what the DUB recipe does you should rather use `dub describe` or at least `dub convert`, but both of those are still way too slow, taking up possibly several hundred milliseconds - so it's not feasible to expect users to use them in all cases right now.

[...]

PR #2639, opened 2023-05-15
Config keys starting with _ prefix do not trigger error or warnings.

p0nce:
This allow to use:
```json
    "_anything": "Here is a comment"
```

in JSON.

### DIFF
--- a/source/dub/internal/configy/Read.d
+++ b/source/dub/internal/configy/Read.d
@@ -490,6 +490,11 @@ private TLFR.Type parseMapping (alias TLFR)
                         if (k[p.length .. $].length > 1 && k[p.length] == '-')
                             continue FIELD;
 
+                // Ignore key starting with _ (escape hatch for comments in JSON).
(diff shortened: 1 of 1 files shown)


p0nce:
Is a solution for #2567

p0nce:
Note that all comments still need a different `_key`

p0nce:
Note that my PR seems incomplete, if multiple "_comment" are found the parser still complains:

> Warning Error was: dub.internal.dyaml.composer.ComposerException@source\dub\internal\dyaml\composer.d(370): Key '_comment' appears multiple times in mapping (first: file C:\Users\guill\Desktop\Dplug\examples\distort\dub.json,line 17,column 5)

p0nce:
Why is JSON parsed with a YAML parser? YAML doesn't allow duplicated keys, but JSON does.

p0nce:
Oh I see, the right fix is perhaps a parsing option in upstream "configy"

PR #2642, opened 2023-05-18
Cache json db

### LINKED ISSUE #2640: Dub support in Meson is broken
Change #2589 has broken Meson support for Dub packages.
Meson will not try to build Dub packages directly but rely instead on the Dub cache to find a package built by Dub.
The way it is done is to find the relevant info in the cache dir name.
For example `library-debug-linux.posix-x86_64-ldc_2081-EF934983A3319F8F8FF2F0E107A363BA`.
Now in the new shorter cache dir name, it is not possible anymore.

If short names are that important, can I suggest that the dir names is composed of two hashes:
 1. one hash that contains infos meson needs and that can be reliably computed by meson as well
    a. compiler
    b. compiler version
    c. arch
    d. platform
 2. a second hash for all info dub needs and meson has no clue about

E.g.
```
library-debug-bUZ3sLErP9iQlvgdxNNYQg
```
would become
```
library-debug-xXxXx-yYyYyYyYyYyYyYyYyYyYyY
              ~~~~~ ~~~~~~~~~~~~~~~~~~~~~~
                1            2
```
5 chars in base64 is more than a billion possibilities. Even if not perfectly distributed, it is more than enough to prevent accidental collisions.

rtbo:
As discussed in #2640, proposal to write Json files in the cache for 3rd party software like Meson to assess compatibility of artifacts in the Dub cache.
I've made the choice to write small Json files per package version rather than a gigantic one for the whole cache.
This comes at the price of stabilizing the directory structure within the cache (up to where the json file is written).

The PR add the following files in the cache
```
~/.dub/cache/vibe-d/0.9.6/+http/build-db.json
~/.dub/cache/vibe-d/0.9.6/+http/build-db.lock
```
To ensure interoperability, the following stabilization apply:
```
~/.dub/cache/vibe-d/0.9.6/+http/build-db.json
~~~~~~~~~~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      1             2
```
1. `$DUB_HOME/cache` _default_ location is stable (env-vars and settings customization still apply)
2. Json file path relative to `$DUB_HOME/cache` is stable

The final artifact location remains up to Dub developers.

fixes #2640

### DIFF
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -321,7 +321,7 @@ class BuildGenerator : ProjectGenerator {
 			db = Json.emptyArray;
 		}
 
-		foreach (entry; db) {
+		foreach_reverse (entry; db) {
 			if (entry["buildId"].get!string == build_id) {
 				// duplicate
 				return;
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -285,10 +285,57 @@ class BuildGenerator : ProjectGenerator {
 		buildWithCompiler(settings, cbuildsettings);
 		target_binary_path = getTargetPath(cbuildsettings, settings);
 
-		if (!settings.tempBuild)
+		if (!settings.tempBuild) {
 			copyTargetFile(target_path, buildsettings, settings);
+			updateCacheDatabase(settings, cbuildsettings, pack, config, build_id, target_binary_path.toNativeString());
+		}
 
 		return false;
+	}
+
+	private void updateCacheDatabase(GeneratorSettings settings, BuildSettings buildsettings, in Package pack, string config,
+		string build_id, string target_binary_path)
+	{
+		import dub.internal.vibecompat.data.json;
+		import core.time : seconds;
+
+		// Generate a `build-db.json` in the package version cache directory.
+		// This is read by 3rd party software (e.g. Meson) in order to find
+		// relevant build artifacts in Dub's cache.
+
+		enum jsonFileName = "build-db.json";
+		enum lockFileName = "build-db.lock";
+
+		const pkgCacheDir = packageCache(settings.cache, pack);
+		auto lock = lockFile((pkgCacheDir ~ lockFileName).toNativeString(), 3.seconds);
+
+		const dbPath = pkgCacheDir ~ jsonFileName;
+		const dbPathStr = dbPath.toNativeString();
+		Json db;
+		if (exists(dbPathStr)) {
+			const text = stripUTF8Bom(cast(string)readFile(dbPath));
+			db = parseJsonString(text, dbPathStr);
+			enforce(db.type == Json.Type.array, "Expected a JSON array in " ~ dbPathStr);
+		}
+		else {
+			db = Json.emptyArray;
+		}
+		Json entry = Json.emptyObject;
+
+		entry["package"] = pack.name;
+		entry["configuration"] = config;
+		entry["buildType"] = settings.buildType;
+		entry["compiler"] = settings.platform.compiler;
+		entry["compilerVersion"] = settings.platform.compilerVersion;
+		entry["compilerBinary"] = settings.platform.compilerBinary;
+		entry["architecture"] = serializeToJson(settings.platform.architecture);
+		entry["platform"] = serializeToJson(settings.platform.platform);
+		entry["buildId"] = build_id;
+		entry["targetBinaryPath"] = target_binary_path;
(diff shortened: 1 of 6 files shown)


WebFreak001:
since you want to ensure that the output doesn't change, do you want to add a test in the tests folder that checks the file to match exactly using `diff`?

source/dub/generators/build.d
```diff
@@ -285,12 +285,59 @@ class BuildGenerator : ProjectGenerator {
...
+		entry["buildType"] = settings.buildType;
+		entry["compiler"] = settings.platform.compiler;
+		entry["compilerVersion"] = settings.platform.compilerVersion;
+		entry["compilerBinary"] = settings.platform.compilerBinary;
+		entry["architecture"] = serializeToJson(settings.platform.architecture);
+		entry["platform"] = serializeToJson(settings.platform.platform);
+		entry["buildId"] = build_id;
+		entry["targetBinaryPath"] = target_binary_path;
+
+		db ~= entry;
```
WebFreak001, line 336: this seems to grow infinitely / doesn't check for duplicates. E.g. when you build with `--force`, it will add duplicate entries.

WebFreak001 (changes requested):

rtbo:
> since you want to ensure that the output doesn't change, do you want to add a test in the tests folder that checks the file to match exactly using `diff`?

The Json file contains the build ID, which varies with compiler version. So I can't predict in a test what the exact content is.
At best I can check if all the Json fields are there and the value of the immutable ones.

source/dub/generators/build.d
```diff
@@ -285,12 +285,68 @@ class BuildGenerator : ProjectGenerator {
...
+		if (exists(dbPathStr)) {
+			const text = stripUTF8Bom(cast(string)readFile(dbPath));
+			db = parseJsonString(text, dbPathStr);
+			enforce(db.type == Json.Type.array, "Expected a JSON array in " ~ dbPathStr);
+		}
+		else {
+			db = Json.emptyArray;
+		}
+
+		foreach (entry; db) {
```
WebFreak001, line 324: maybe this should be a
```suggestion
		foreach_reverse (entry; db) {
```
since latest entries are gonna be at the bottom and it's most likely cache hits are from recent builds

test/pr2642-cache-db/source/test_cache_db.d
```diff
@@ -0,0 +1,78 @@
...
+        rmdirRecurse(dubhome);
+    }
+
+    const string[string] env = [
+        "DUB_HOME": dubhome,
+    ];
+    const fetchProgram = [
+        environment["DUB"],
+        "fetch",
+        "vibe-d@0.9.6",
```
WebFreak001, line 23: `gitcompatibledubpackage` might be better for build time

WebFreak001 (approved):

WebFreak001:
thanks, seems to be a well designed and useful feature!

rtbo:
Thanks for the quick review !

Geod24:
I've missed the discussion here, but why are we adding JSON files for other tools to depend on ? `dub describe` was meant as the way to do this, why is it not enough ?

Geod24:
I'm not sure what "stabilization" means in this context, either. The commit just adds comments.

rtbo:
`dub describe` doesn't give any info about cached builds.
e.g. on my system
```
tree ~/.dub/cache/urld
/home/remi/.dub/cache/urld
└── 2.1.1
    ├── build
    │   ├── library-debug-cu9jbmcFOeHiBceMMo2VbA
    │   │   └── liburld.a
    │   └── library-debug-Z7qINYX4IxM8muBSlyNGrw
    │       └── liburld.a
    ├── db.json
    └── db.lock
```
Meson needs to know what these 2 archives are (compiler, compiler version etc.) in order to use them properly.
Meson uses `dub describe` for everything else (import paths, sub-dependencies etc.), but the cache info is not available.
I have to say that enriching `dub describe` output would have been a viable option as well (e.g. add the buildId).

rtbo:
> I'm not sure what "stabilization" means in this context, either. The commit just adds comments.

What is stable is what should not be changed in future Dub versions.

Geod24:
> What is stable is what should not be changed in future Dub versions.

The only thing that should not be changed in future dub version is whatever is part of the API. And the API to access dub's internal data is `dub describe`, hence my comment.

PR #2643, opened 2023-05-18
Fix flags Json serialization

rtbo:
These flags always appear empty in `dub describe`


WebFreak001 (approved):

WebFreak001:
build failures unrelated / pre-existing

WebFreak001 (commented):
first thought this may have been introduced with the yaml changes, but it looks like this issue was present before, so not a regression but a bug fix. Will merge now

PR #2644, opened 2023-05-21
add the build ID in TargetDescription

rtbo:
Allow 3rd party programs (e.g. Meson) to locate artifacts in the cache.
Following @Geod24 comment in #2642.

@WebFreak001 @Geod24 This would be enough for Meson to locate Dub's artifacts.
If you prefer to use this proposal, feel free to revert #2642.

### DIFF
--- a/source/dub/description.d
+++ b/source/dub/description.d
@@ -121,6 +121,7 @@ struct TargetDescription {
 	string[] packages; /// All packages contained in this target (e.g. for target type "sourceLibrary")
 	string rootConfiguration; /// Build configuration of the target's root package used for building
 	BuildSettings buildSettings; /// Final build settings to use when building the target
+	string cacheArtifactPath; /// The full path of the built target in the cache
 	string[] dependencies; /// List of all dependencies of this target (package names)
 	string[] linkDependencies; /// List of all link-dependencies of this target (target names)
 }
--- a/source/dub/generators/build.d
+++ b/source/dub/generators/build.d
@@ -263,7 +263,7 @@ class BuildGenerator : ProjectGenerator {
 			m_tempTargetExecutablePath = target_path = getTempDir() ~ format(".dub/build/%s-%s/%s/", packageName, pack.version_, build_id);
 		}
 		else
-			target_path = packageCache(settings.cache, pack) ~ "build/" ~ build_id;
+			target_path = targetCacheDir(settings.cache, pack, build_id);
 
 		if (!settings.force && isUpToDate(target_path, buildsettings, settings, pack, packages, additional_dep_files)) {
 			logInfo("Up-to-date", Color.green, "%s %s: target for configuration [%s] is up to date.",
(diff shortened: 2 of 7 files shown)


rtbo:
Instead of `cacheBuildId`, I could use `cacheArtifactPath` with the full path to the built artifact in the cache.
It would avoid Meson to take any assumption to the internal cache structure of Dub
i.e.
```
$dubhome/cache/$pkg/$version[/+$subpkg]/build/$buildId/$targetFileName
```
Even with the buildId available, if this structure change, Meson will be lost again.

PetarKirov:
A full path would be better, as Meson shouldn't depend on implementation details of Dub. As @Geod24 mentioned, `dub describe` is what other tools should use and if it doesn't contain all necessary information we should simply update it to fix that.

rtbo:
There seem to be a bug in `ProjectGenerator.configurePackages` that prevent this PR to work as intended.
For example, when I run `dub describe eventcore`, the target `taggedalgebraic` get versions from target `eventcore`, despite having no dependency towards `eventcore`:
```json
				"versions": [
					"EventcoreEpollDriver",
					"Have_taggedalgebraic"
				],
```
However when I run `dub build eventcore`, `taggedalgebraic` is not built (because static libs) and when I run `dub build taggedalgebraic`, it is built without the `EventcoreEpollDriver` version.

It means that `dub describe` is currently describing targets that can't actually be built.

This come down to this piece of code, which looks wrong to me. I don't understand why dependencies need versions from their parents.
```d
		// 1. downwards inherits versions, debugVersions, and inheritable build settings
		static void configureDependencies(const scope ref TargetInfo ti, TargetInfo[string] targets, size_t level = 0)
		{
			// do not use `visited` here as dependencies must inherit
			// configurations from *all* of their parents
			logDebug("%sConfigure dependencies of %s, deps:%(%s, %)", ' '.repeat(2 * level), ti.pack.name, ti.dependencies);
			foreach (depname; ti.dependencies)
			{
				auto pti = &targets[depname];
				mergeFromDependent(ti.buildSettings, pti.buildSettings);
				configureDependencies(*pti, targets, level + 1);
			}
		}

		configureDependencies(*roottarget, targets);
```

WebFreak001:
> This would be enough for Meson to locate Dub's artifacts.
> If you prefer to use this proposal, feel free to revert https://github.com/dlang/dub/pull/2642.

actually I think having both would be beneficial, as dub describe will only tell you about a single configuration, but you might want to know what everything in the folder is.

rtbo:
When I completely remove the "downwards inherits versions etc." section from Dub's code, I still make valid builds.
Do you know the purpose of this (dependencies receiving versions from their parents) ?

Right now `dub build` and `dub describe` are not fully compatible with each other.

rtbo:
> It is used quite heavily, see bindbc projects that use it to specify the version of the library it needs to load.

Ah yes, quite obvious.
Anyway this PR is ready. It will be fully useful if the `--deep` flag I've proposed is merged.

rtbo:
Hi, can this be merged before next Dub release?
The change seems to have consensus. 
I'd like to use it to fix Meson support for Dub dependencies.

WebFreak001 (approved):

PR #2646, opened 2023-05-24
support specifying --d-versions

WebFreak001:
Usually you should prefer using configurations and build types, but sometimes, especially when wrapped in other build tools or when supporting e.g. different target platforms on the same target architecture and configuration, you may want to manually manage certain versions.


WebFreak001:
CI fails unrelated (happening on master all the time as well)

PetarKirov (approved):
Looks good to me - straightforward addition!

Issue #2209, opened 2022-02-05
How to build every dependency of a project (given a dub.json / dub.selections.json)?


Panke:
I am trying to improve dub2nix and make Nix work better with D. To this end I want to build all dependencies of a package without building that package itself. I tried to fetch the dependencies from a dub.selections.json file and build them, but this fails: 

`[nix-shell:~/dlang/dlang-mix-meson-playground]$ HOME=. dub fetch stdx-allocator@2.77.5
Fetching stdx-allocator 2.77.5...

[nix-shell:~/dlang/dlang-mix-meson-playground]$ dub add-local .dub/packages/stdx-allocator-2.77.5/stdx-allocator 2.77.5
Registered package: stdx-allocator (version: 2.77.5)

[nix-shell:~/dlang/dlang-mix-meson-playground]$ HOME=. dub build --skip-registry=all stdx-allocator@2.77.5
No package stdx-allocator was found matching the dependency 2.77.5
`

Panke:
If I move the package to another place and `dub add-local` it there and allow the registries it will 1.) fetch the package again to `/.dub/packages`, but 2.) build the package in the place specified with `dub add-local`.

rikkimax:
Try skipping ``configured`` or ``standard`` instead of ``all``.

Panke:
That does not make a difference.

Panke:
I am going now a different route without `dub add-local`. 

1. I fetch all packages from a `dub.selections.json` file into a custom file-repository.

2. HOME=. dub build --skip-registry=all --registray=file://what/ever <pkg>@<version>
foreach package in `dub.selections.json` and keep the ./dub/packages/ as a build dependency

3.  For the actual build, restore the ./dub/packages folder and get the required flags via `dub describe`.

WebFreak001:
this might help: https://github.com/dlang/dub/issues/2224

https://code.dlang.org/packages/dub-build-deep

https://github.com/mesonbuild/meson/pull/10045

MrcSnm:
Is there some way to get a `staticLibraryBundle` target type?
This could be helpful for my engine. I have done a simple script to do something like that but the work is still very manual.

WebFreak001:
@MrcSnm consider opening a new issue for that

PR #2647, opened 2023-05-25
add --deep flag to `dub build`

### LINKED ISSUE #2224: Build a static library and all its sub-dependencies
When building a Dub static library (e.g. `dub build vibe-d:http`), Dub will build the static library itself, but not the dependencies of this library (simply because I think it is not needed).

Is it possible to get Dub build the complete chain with a single command?
This is useful when using Meson. In a Meson build, if one depends on say `vibe-d:http`, Meson expect to find all the needed Dub libraries already built. 
Running `dub build` for a dozen of packages is quite tedious and error prone as some packages will require a specific configuration.
I wrote a [script](https://gist.github.com/rtbo/52ca40f119e0b6c892cbfaebbab4f49b) to support that, but it would be better if it comes from Dub.

### System information

- DUB version 1.27.0
- Arch Linux
- DMD v2.098.1

rtbo:
Allows to build all dependencies of a static library.

This provides the same functionality as the [dub-build-deep](https://code.dlang.org/packages/dub-build-deep) package.
The difference is that all dependencies are built in the same DUB invocation, which imply that the build will be fully compatible with `dub describe`.

E.g.
`dub describe vibe-d:http` will describe the build settings of all dependencies of `vibe-d:http` and `dub build --deep vibe-d:http` will build all dependencies with the exact same build settings.
If we build dependencies one by one (as done by `dub-build-deep`), some settings will differ.

See also #2644

Fixes #2224 (and possibly #2209)

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1198,6 +1198,7 @@ class GenerateCommand : PackageBuildCommand {
 	protected {
 		string m_generator;
 		bool m_printPlatform, m_printBuilds, m_printConfigs;
+		bool m_deep; // only set in BuildCommand
 	}
 
 	this() @safe pure nothrow
@@ -1274,6 +1275,7 @@ class GenerateCommand : PackageBuildCommand {
 		gensettings.runArgs = app_args;
 		// legacy compatibility, default working directory is always CWD
 		gensettings.overrideToolWorkingDirectory = getWorkingDirectory();
+		gensettings.buildDeep = m_deep;
 
 		logDiagnostic("Generating using %s", m_generator);
 		dub.generateProject(m_generator, gensettings);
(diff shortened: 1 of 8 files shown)


MrcSnm (changes requested):
@rtbo So far, the only thing I see missing is a test. Add a test using the --deep flag
I'm also really looking for that since I have been using quite a hack for my engine

MrcSnm (approved):
LGTM @WebFreak001 @Geod24

MrcSnm:
@rtbo I have wrote this file here:

```
Added `--deep=` switch to dub build

By specifying this flag, you can now build all the dependencies
of a staticLibrary. The default behavior is to only build the
library located in the root directory of the dub configuration
file. This allows better integration with other build systems
which require the libraries to be built upfront.

```
dub build --deep
```

If a staticLibrary A depends on staticLibrary B, and the --deep
flag is specified, dub will output both the A and B libraries.
```

Feel free to modify as you see fit. This is a required file which goes inside the changelog folder.

MrcSnm:
Hey there @WebFreak001 @Geod24 can you take a look into this PR?

WebFreak001 (approved):

PR #2648, opened 2023-05-27
Fix failure of test suite on POSIX

rtbo:
Fixing annoying failure of test suite on POSIX systems with recent DMD versions

### DIFF
--- a/test/4-describe-data-2-dmd.sh
+++ b/test/4-describe-data-2-dmd.sh
@@ -34,8 +34,12 @@ fi
 
 # Create the expected output path file to compare against.
 expected_file="$CURR_DIR/expected-describe-data-2-dmd-output"
+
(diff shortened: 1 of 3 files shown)


CyberShadow:
What is the DMD change?

CyberShadow:
> This change I believe [dlang/phobos#8684](https://github.com/dlang/phobos/pull/8684)

Ah, I see. So, I think the problem is then that the Dub test suite depends on the *exact algorithm* that Phobos uses for escaping shell parameters.

If possible, I think it would better to turn this around, and parse the quoted arguments into an array using the shell and then comparing that.

WebFreak001 (approved):
this runs in CI and passes on mac and linux, let's just pull it in like this

a better fix can be made in the future

PR #2650, opened 2023-05-30
Allow building deprecated modules without deprecation warnings

CyberShadow:
The compiler issues a deprecation warning only on a "non-deprecated" -> "deprecated" transition.  As such, building or testing Dub packages which have deprecated modules when a dub_test_root.d file needs to be generated will cause deprecation warnings.

This in turn makes it impossible to use `buildRequirements "disallowDeprecations"` in this situation.

Fix this by simply making the auto-generated root module (dub_test_root) deprecated. As it is the compiler's entry point, the module itself will not cause a deprecation warning, and it importing deprecating modules will not cause deprecation warnings (as there is no longer a transition of deprecation state).

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -1911,6 +1911,7 @@ final class SelectedVersions {
 
 /// The template code from which the test runner is generated
 private immutable TestRunnerTemplate = q{
+deprecated // allow silently using deprecated symbols
 module dub_test_root;
 
 import std.typetuple;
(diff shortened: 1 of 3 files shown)


source/dub/project.d
```diff
@@ -1911,7 +1911,7 @@ final class SelectedVersions {
 
 /// The template code from which the test runner is generated
 private immutable TestRunnerTemplate = q{
-module dub_test_root;
+deprecated module dub_test_root;
```
Geod24, line 1914: Maybe add an explanation here so a person looking at it is not puzzled as to why we do this.

CyberShadow: Done

PR #2652, opened 2023-06-01
dub.generators.generator: Don't create targetPath dir with --temp-build

### LINKED ISSUE #2651: Don't try to write to the `--cache=system` cache as an unprivileged user
### System information
Reproducible on `docker.io/debian:latest` with Dub from D 2.103.1

### Bug Description
Some context here: https://gist.github.com/ximion/fe6264481319dd94c8308b1ea4e8207a

### How to reproduce?
1. As root, run `dub fetch --cache=system foo`
2. As non-root, try to build any package which depends on `foo` with `--cache=system`

Full reproducer here: https://gist.github.com/CyberShadow/3dca5becd83f3fef3b582b1a6299a0cc/c6ae623f4e8e9642be02ffaf8e2d0a67ddb4a80f

### Expected Behavior
Dub should not try to write to the `--cache=system` location as an unprivileged user (maybe even if it is a privileged user).

We want to roughly do the same thing Python does with its `/usr/lib/python3.10/site-packages`, i.e. use them if we can (putting any temporary files in the current directory or `$TMPDIR`), don't use them if we can't (version mismatch).

AFAIU, fixing this or having something like this in Dub would make packaging Dub packages in Debian much more realistic.

### Logs
The reproducer above fails with:
```
Error /var/lib/dub/packages/emsi_containers-0.9.0/emsi_containers/build: Permission denied
```

CyberShadow:
Fix #2651. 

Is this it?? It works for me with this change.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -963,7 +963,8 @@ private void finalizeGeneration(in Package pack, in Project proj, in GeneratorSe
 	}
 
 	if (generate_binary) {
-		ensureDirectory(NativePath(buildsettings.targetPath));
+		if (!settings.tempBuild)
+			ensureDirectory(NativePath(buildsettings.targetPath));
 
 		if (buildsettings.copyFiles.length) {
 			void copyFolderRec(NativePath folder, NativePath dstfolder)


CyberShadow:
I'm not sure how to write a test for this, because the test target is a directory created by `dub fetch`.

Geod24 (approved):

PR #2660, opened 2023-06-13
Draft: LDC: Apply dflags affecting symbol visibility to all deps

kinke:
On Windows.


kinke:
Example scenarios:
* If an .exe is linked against druntime/Phobos DLLs (e.g., to enable loading other D DLLs, all sharing central druntime/Phobos), all static-lib dub dependencies need to be compiled with `-dllimport=defaultLibsOnly` too [the default with `-link-defaultlib-shared`].
* To make large DLLs with few selective `export`ed symbols work after #2412, the implicit flags need to be overridden - for the DLL itself and all its static-lib dependencies. This can currently only be accomplished by setting the `DFLAGS` environment variable. This PR enables overriding these `dflags` either in the DLL root project/config directly, or in some dependency shared by multiple DLLs. E.g., for Symmetry, we have an .exe with lots of plugin DLLs. All plugins and the .exe have a shared dependency, so adding `dflags "-fvisibility=hidden" "-dllimport=defaultLibsOnly" platform="windows-ldc"` once in the `dub.sdl`of that shared dependency suffices to compile *everything* with those required flags.

TODO:
* Tests
* Decide whether this should be Windows only, or whether `-fvisibility` should be handled identically on Posix platforms, enabling to use hidden visibility for all static-lib deps of some .so/dylib too.

kinke:
> Does `-mscrtlib` need to be passed through too?

Nope, that flag really only affects linking, not compilation. And I think LDC defaults to static `libcmt` with `-link-defaultlib-shared=false`, for DLLs too [if an MSVC installation is found].

rikkimax:
My one concern (apart from code review), is that this would be better suited to be handled from a pretty significant refactor in the form of placing directives to have the relationship to other packages of ``just this``, ``goes up``, ``goes down``. But since this would be the default, its still a good PR in theory.

WebFreak001:
no longer draft? in that case also adjust title pls

kinke:
Rebased, changelog entry added, and not passing stuff down to *shared*-library deps.

While I fully agree that there would ideally be a generic way of controlling the flags propagation, this automatism solves a number of use cases which can currently only be handled by setting the DFLAGS env variable. As DMD seems to be adopting the relevant LDC switches (`-fvisibility` and `-dllimport`), the approach could eventually be generalized to DMD too. GDC has `-fvisibility` already, so we could include it already (but the main use cases are on Windows).

PR #2663, opened 2023-06-17
Skip path-specified optional dependencies if they don't exist

CyberShadow:
I think this is a useful and harmless alternative way to understand "optional" in this context.

The motivation is being able to do autoconf-like optional dependencies: D packages may install a Dub package file somewhere on the system, which we can pull in if it is present, but silently skip if it's absent.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -473,6 +473,11 @@ class Project {
 				Dependency vspec = dep.spec;
 				Package p;
 
+				if (vspec.optional && vspec.path() != NativePath.init && !existsDirectory(vspec.path())) {
+					logDiagnostic("%sOptional dependency %s path %s does not exist, not using.", indent, dep.name, vspec.path());
+					continue;
+				}
+
 				auto basename = getBasePackageName(dep.name);
 				auto subname = getSubPackageName(dep.name);
 
--- a/test/pr2663-optional-path/.gitignore
+++ b/test/pr2663-optional-path/.gitignore
@@ -1,0 +1,1 @@
+/pr2663-optional-path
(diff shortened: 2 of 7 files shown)


source/dub/project.d
```diff
@@ -473,6 +473,11 @@ class Project {
 				Dependency vspec = dep.spec;
 				Package p;
 
+				if (vspec.optional && vspec.path() != NativePath.init && !existsDirectory(vspec.path())) {
+					logDiagnostic("%sOptional dependency %s path %s does not exist, not using.", indent, dep.name, vspec.path());
```
WebFreak001, line 477: I think this should rather fall under a warning, since when typo-ing the path, the user won't understand why the modules they import aren't being found and probably first try to search in the dependency.

WebFreak001 (changes requested):

source/dub/project.d
CyberShadow: But on the other hand, do we really want Dub to print a warning every time you build a project even when nothing is wrong and everything is as intended and there is nothing to fix?

WebFreak001: well imo that's better than no warning for such issues

CyberShadow: Rule of silence???

WebFreak001: I would agree if this would be a warning that the user can't get rid of, however you can remove it by fixing the path (what people typoing usually want) or currently just commenting out such dependencies when they are optional. However I think we can have a better solution than this, like a new setting per dependency.

I'm not a fan of reusing optional, since users might just manually set a path in existing dependencies that are optional. While it's unlikely, I don't really see a major reason not to add a new key for this.

CyberShadow: Alright.

Issue #2666, opened 2023-06-23
preGenerateCommands should reparse dub.json/dub.sdl, if needed


WebFreak001:
When running preGenerateCommands, dub will already re-evaluate all the globs and variables from the dub recipe, however it will never actually re-parse the dub recipe. I think if the recipe has changed (modification time), we should also re-parse the recipe, so that more advanced use-cases that edit the recipe are possible.

Note that dub right now only considers itself, not dependencies or dependents when running preGenerateCommands, which is a limitation that keeps the implementation sane though.

Geod24:
> so that more advanced use-cases that edit the recipe are possible.

I don't see self-modifying code as an advanced use case, it's an error prone hack and I really don't think we should do that.

PR #2667, opened 2023-06-23
make package recipe const

WebFreak001:
So that only package_.d may modify it and it's easier to reload it from disk in the future.

Just some refactorings right now, should make #2666 a lot easier to implement.


Geod24:
Isn't this a breaking change ?

WebFreak001:
technically yes, for API users. I don't think we can offer any good migration path either. For the test failures, need to investigate.

PR #2689, opened 2023-08-18
CI: Enable test/run-unittest.sh on Windows

kinke:


test/run-unittest.sh
```diff
@@ -39,10 +39,10 @@ FRONTEND="${FRONTEND:-}"
 
 if [ -z ${FRONTEND:-} ]; then
     if [ "$DC_BIN" == "ldc2" ]; then
-        FRONTEND=$(ldc2 --version | grep -Po "based on DMD v\K(2\.\d+\.\d)")
+        FRONTEND=$(ldc2 --version | grep 'based on DMD v2.' | sed -E -n 's/^.*DMD v(2\.[0-9]+\.[0-9]).*$/\1/p')
```
kinke, line 42: The grep in my Windows VM complained about `grep -P` only working for some locales.

test/issue2452/.no_test
kinke, line 1: This file is needed because the path of the unittest runner executable on Windows exceeds the 260 allowed chars... but there aren't any unittests to run either and no extra dub config, so skip that seemingly superfluous test build altogether.

.github/workflows/main.yml
```diff
@@ -95,7 +95,17 @@ jobs:
...
+
+          # FIXME: DMD fails a few tests on Windows; remove them for now
+          if [[ '${{ matrix.dc }}' = dmd* ]]; then
+            # DLL support is lacking
+            rm -rf test/{1-dynLib-simple,2-dynLib-dep,2-dynLib-with-staticLib-dep}
+            # Unicode in paths too
+            rm -rf test/issue130-unicode-СНА*
+            # ImportC probably requires set-up MSVC environment variables
+            rm -rf test/use-c-sources
+          fi
```
kinke, line 108: Not sure if there's a more elegant way to handle this, or making DMD's limitations on Windows more prominent.

.github/workflows/main.yml
```diff
@@ -88,14 +88,22 @@ jobs:
...
+          test/run-unittest.sh
```
shoo, line 107: I have worked on this subject in the past. Check out my outcomes below:
https://github.com/dlang/dub/compare/master...shoo:dub:enh_test_windows

The difference from this PR's approach is that in my approach I changed `run-unittest.d` to be more like `run-unittest.sh` and not dependent on `.sh` .

In windows, there are several ways to run `.sh` . Cygwin(or MSYS2) / git-bash / WSL
GitHub Actions uses git-bash, which I think should be avoided for compatibility reasons.

kinke: AFAIK, the gnu tools bundled with git for windows originate from msys2. Why do you think it should be avoided for compatibility reasons? Since almost any Windows dev has git, he/she has bash + gnu tools too. I've been using bash on Windows for years, and it's absolutely okay, especially for such trivial stuff that is easier handled in bash than D.

kinke:
Ping, why does this have to wait for so long? It significantly increases test coverage on Windows.

WebFreak001 (approved):
sorry, LGTM

PR #2690, opened 2023-08-24
LDC: Don't keep around temporary object files for static libs

kinke:
This was a workaround for older LDC versions. Since LDC v1.25, `-cleanup-obj` implies that the temporary object files are placed into a unique temp directory (per compiler invocation), so that object file collisions across parallel dub invocations cannot happen anymore.

### DIFF
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -243,10 +243,9 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 				if (platform.frontendVersion >= 2071) {
 					settings.addDFlags("-cleanup-obj");
 				}
-				if (platform.frontendVersion >= 2095) {
-					// since LDC v1.25, -cleanup-obj additionally defaults to a unique temp -od directory
-				} else {
-					// resort to a unique-ish -od directory
+				if (platform.frontendVersion < 2095) {
+					// Since LDC v1.25, -cleanup-obj defaults to a unique temp -od directory
+					// We need to resort to a unique-ish -od directory before that
 					settings.addDFlags("-od=" ~ settings.targetPath ~ "/obj");
 				}
 				break;
--- a/source/dub/compilers/ldc.d
+++ b/source/dub/compilers/ldc.d
@@ -104,7 +104,6 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 
 	void prepareBuildSettings(ref BuildSettings settings, const scope ref BuildPlatform platform, BuildSetting fields = BuildSetting.all) const
 	{
-		import std.format : format;
 		enforceBuildRequirements(settings);
 
 		// Keep the current dflags at the end of the array so that they will overwrite other flags.
@@ -117,9 +116,6 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 				if (settings.options & t[0])
 					settings.addDFlags(t[1]);
 		}
-
-		// since LDC always outputs multiple object files, avoid conflicts by default
-		settings.addDFlags("--oq", format("-od=%s/obj", settings.targetPath));
 
 		if (!(fields & BuildSetting.versions)) {
 			settings.addDFlags(settings.versions.map!(s => "-d-version="~s)().array());
@@ -241,7 +237,18 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 			case TargetType.executable: break;
 			case TargetType.library:
 			case TargetType.staticLibrary:
-				settings.addDFlags("-lib");
+				// -oq: name object files uniquely (so the files don't collide)
+				settings.addDFlags("-lib", "-oq");
+				// -cleanup-obj (supported since LDC v1.1): remove object files after archiving to static lib
+				if (platform.frontendVersion >= 2071) {
+					settings.addDFlags("-cleanup-obj");
+				}
+				if (platform.frontendVersion >= 2095) {
+					// since LDC v1.25, -cleanup-obj additionally defaults to a unique temp -od directory
+				} else {
+					// resort to a unique-ish -od directory
+					settings.addDFlags("-od=" ~ settings.targetPath ~ "/obj");
+				}
 				break;
 			case TargetType.dynamicLibrary:
 				settings.addDFlags("-shared");
--- a/test/removed-dub-obj.sh
+++ b/test/removed-dub-obj.sh
@@ -3,19 +3,14 @@
 . $(dirname "${BASH_SOURCE[0]}")/common.sh
 cd ${CURR_DIR}/removed-dub-obj
 
-DUB_CACHE_PATH="$HOME/.dub/cache/removed-dub-obj/"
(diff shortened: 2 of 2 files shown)


source/dub/compilers/ldc.d
```diff
@@ -234,7 +231,9 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 			case TargetType.executable: break;
 			case TargetType.library:
 			case TargetType.staticLibrary:
-				settings.addDFlags("-lib");
+				// -oq: name object files uniquely (so the files don't collide)
+				// -cleanup-obj: remove object files after archiving to static lib, and put them in a unique temp directory
+				settings.addDFlags("-lib", "-oq", "-cleanup-obj");
```
WebFreak001, line 236: are the new flags not required for the dynamicLibrary case below?

test/removed-dub-obj.sh
```diff
@@ -3,19 +3,13 @@
...
-if [[ ${DC} == *"ldc"* ]]; then
-    if [ ! -f $DUB_CACHE_PATH/~master/build/library-*/obj/test.o* ]; then
-        ls -lR $DUB_CACHE_PATH
-        die $LINENO '$DUB_CACHE_PATH/~master/build/library-*/obj/test.o* was not found'
-    fi
-fi
-
-exit 0
+numObjectFiles=$(find "$DUB_CACHE_PATH" -type f -iname '*.o*' | wc -l)
+[ "$numObjectFiles" -eq 0 ] || die $LINENO "Found left-over object files in $DUB_CACHE_PATH"
```
WebFreak001, line 15: this change seems to revert what has previously been published in the changelogs: https://github.com/kinke/dub/commit/176ea9cc13c9c36f285dea628954d049a04aa0f2

> Enables co-existence and parallel compilation of the same project with different settings (e.g. cross compilation) by moving `.dub/obj` to `$DUB_TARGET_PATH/obj`.

Although that changelog entry seems faulty right now, since it didn't actually write to the target path (e.g. `"targetPath": "bin"`) but to the build cache path.

So I think this change is fine.

WebFreak001 (commented):
since the `-cleanup-obj` flag was introduced in 2021 and we don't really have any tool version / stability policy, can you restrict the changes into `if (platform.frontendVersion >= 2095)` to be on the safe side?

source/dub/compilers/ldc.d
kinke: Nope; the object files for executables and shared libs are persisted/kept around, by dub and the compilers alike. Static libs with DMD are a special case, as it writes the static lib directly itself, without creating separate object files and then archiving them, like LDC and GDC. `ldmd2 -lib` also injects these 2 extra flags for the `ldc2` cmdline.

kinke:
The flag was introduced in LDC v1.1 in 2016: https://github.com/ldc-developers/ldc/commit/141310c77cb433ca358bcc65e7117166b975061f

It's just that since v1.25 it additionally implies a concurrent-safe unique temp dir instead of the cwd (if you don't specify a `-od` objects dir explicitly).

test/removed-dub-obj.sh
kinke: With LDC v1.25+, this PR solves the concurrency problem more robustly - each compiler invocation uses a unique temp dir for its object files, no need to specify unique-ish paths explicitly.

kinke:
@WebFreak001: Okay, I've re-added support for all LDC versions.

source/dub/compilers/ldc.d
```diff
@@ -241,7 +237,18 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
...
+				// -cleanup-obj (supported since LDC v1.1): remove object files after archiving to static lib
+				if (platform.frontendVersion >= 2071) {
+					settings.addDFlags("-cleanup-obj");
+				}
+				if (platform.frontendVersion >= 2095) {
+					// since LDC v1.25, -cleanup-obj additionally defaults to a unique temp -od directory
+				} else {
+					// resort to a unique-ish -od directory
+					settings.addDFlags("-od=" ~ settings.targetPath ~ "/obj");
+				}
```
Geod24, line 251:
```suggestion
				if (platform.frontendVersion < 2095) {
					// Since LDC v1.25, -cleanup-obj defaults to a unique temp -od directory
					// We need to resort to a unique-ish -od directory before that
					settings.addDFlags("-od=" ~ settings.targetPath ~ "/obj");
				}
```

Geod24 (approved):

Geod24:
@kinke : This broke `dub build -b ddox` for one of my project:
```
Error: Output file '__dummy_docs/package.html' for module `barcode.qr` collides with previous module `barcode`. See the -oq option
```
While building a project that depends on https://github.com/deviator/barcode
Working:
```
ldc2 -o- -w -lowmem -Xf=docs.json -Dd=__dummy_docs --oq -od=submodules/barcode/obj -d-version=_GLIBCXX_USE_CXX98_ABI -d-version=Have_barcode -d-debug=AgoraDebugAPIs -Isubmodules/barcode/source/ submodules/barcode/source/barcode/code128.d submodules/barcode/source/barcode/code39.d submodules/barcode/source/barcode/ean13.d submodules/barcode/source/barcode/itf.d submodules/barcode/source/barcode/package.d submodules/barcode/source/barcode/qr/ecl.d submodules/barcode/source/barcode/qr/package.d submodules/barcode/source/barcode/qr/qrcode.d submodules/barcode/source/barcode/qr/qrsegment.d submodules/barcode/source/barcode/qr/util.d submodules/barcode/source/barcode/qrwrap.d submodules/barcode/source/barcode/svgdraw.d submodules/barcode/source/barcode/types.d submodules/barcode/source/barcode/util.d -preview=in -revert=dtorfields -preview=in -revert=dtorfields -vcolumns
```

Broken:
```
ldc2 -o- -w -lowmem -Xf=docs.json -Dd=__dummy_docs -d-version=_GLIBCXX_USE_CXX98_ABI -d-version=Have_barcode -d-debug=AgoraDebugAPIs -Isubmodules/barcode/source/ submodules/barcode/source/barcode/code128.d submodules/barcode/source/barcode/code39.d submodules/barcode/source/barcode/ean13.d submodules/barcode/source/barcode/itf.d submodules/barcode/source/barcode/package.d submodules/barcode/source/barcode/qr/ecl.d submodules/barcode/source/barcode/qr/package.d submodules/barcode/source/barcode/qr/qrcode.d submodules/barcode/source/barcode/qr/qrsegment.d submodules/barcode/source/barcode/qr/util.d submodules/barcode/source/barcode/qrwrap.d submodules/barcode/source/barcode/svgdraw.d submodules/barcode/source/barcode/types.d submodules/barcode/source/barcode/util.d -preview=in -revert=dtorfields -preview=in -revert=dtorfields -vcolumns
```

kinke:
I guess it needs `-oq` (what about DMD, which AFAIK doesn't have that switch? maybe just no colliding-output-files check, so a silent overwrite?). The command doesn't seem to go through `setTarget()`, where the `-oq` would be applied (for static libs).

kinke:
=> https://github.com/dlang/dub/pull/2994

PR #2694, opened 2023-09-04
improve github CI definitions

WebFreak001:
- skip CI on changes only to changelog/
- reduce permissions of GH token for PR comment


WebFreak001:
stuff copied from docs, but to see if it really works there is nothing really viable here other than just testing in production - so merging this (since it's not a major issue if the CI comments break for the next PR either)

PR #2699, opened 2023-09-23
Add betterC to probePlatform calls

LunaTheFoxgirl:
When building D projects without a runtime for new platforms, dub will fail when there's no druntime for said platform.

This PR makes all the probePlatform calls use betterC so that runtime-less platforms can be built for with dub.

I need this for building D projects for the SuperH CPU architecture (as used in the SEGA Dreamcast and SEGA Saturn)

NOTE: This PR could be improved by checking the compiler versions before calling probePlatform, to ensure that older compilers without betterC support for the platform probe still works.

### DIFF
--- a/source/dub/compilers/utils.d
+++ b/source/dub/compilers/utils.d
@@ -326,13 +326,22 @@ NativePath generatePlatformProbeFile()
 	enum probe = q{
 		module object;
 
-		template _d_arrayappendcTXImpl(Tarr : T[], T)
+		static if (__VERSION__ < 2104)
 		{
-			ref Tarr _d_arrayappendcTX(return ref scope Tarr px, ulong n) @trusted pure nothrow
+			// Work around bug that in older FE versions CTFE would still emit
+			// symbol calls to druntime, which we don't have with betterC
+
+			// Starting with DMD 2.104 this is no problem anymore however, so we
+			// don't need to use this hack there.
+
+			template _d_arrayappendcTXImpl(Tarr : T[], T)
 			{
-				// this should never be called, since we use CTFE only
-				// avoids linker errors when compiler attempts to emit a call to this
-				assert(false);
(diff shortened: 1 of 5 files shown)


LunaTheFoxgirl:
I've added a workaround for a GDC bug related to this, the platform check would fail because __ctfe is broken and will try to compile for runtime anyways.

The cursed multi-line pragma improves compatibility and has been tested across GDC, DMD and LDC. It also now fully does not depend on libphobos for output compilation which should speed it up too.

source/dub/compilers/utils.d
```diff
@@ -326,31 +326,37 @@ NativePath generatePlatformProbeFile()
 	enum probe = q{
 		module object;
 
+		template _d_arrayappendcTXImpl(Tarr : T[], T)
```
rikkimax, line 329: This can't go in.

It is a time bomb waiting to happen.

Get rid of the appends instead.

WebFreak001: why not? it's just for the platform probe and it will be _very_ apparent if it breaks

WebFreak001: did you add a test case for that to the CI? Otherwise it might get introduced again

rikkimax: > why not? it's just for the platform probe and it will be _very_ apparent if it breaks - technically could wrap this in a version LDC/GDC or static if, when LDC and GDC fix this issue

Exactly, we'll know about it, There won't be a workaround and it limits what frontend versions dub can work for.

It is far too easy for this to fail, especially when compiler hooks are in flux.

WebFreak001: fixed, solution will can no longer break with newer compilers anymore and works with several tested older compilers, as well as CI should cover it.

WebFreak001 (approved):

Geod24:
Superseded by #2753
Sorry this wasn't merged earlier, and thank you for your work! Let us know if #2753 does not work for you for some reason.

PR #2703, opened 2023-09-26
remove GeneratorSettings.cache

WebFreak001:
instead the cache is a property of PackageManager

This way dub lib users don't have to try to figure out where the real cache path is and can just continue using the generator and describe APIs like in previous versions.

Customizing cache path is no longer easily possible and now must be done through creating a custom PackageManager and use that on `new Project()` to pass into the generators. If needed, APIs to modify m_dirs on startup could be added to Dub subclasses in the future.

Also for how the other code was setup, this looks like a more reasonable place to put the code, as well as PackageManager being a good place to handle path modification / generation.

Sadly the GeneratorSettings public API has already been released in a tag and thus this is a breaking change, that for example reggae also uses: https://github.com/atilaneves/reggae/blob/2302be1809edde6e1ac3552ee595970601f960ad/src/reggae/dub/interop/dublib.d#L70

However, at least for reggae, the migration would be trivial, since it already instantiates a custom PackageManager: https://github.com/atilaneves/reggae/blob/2302be1809edde6e1ac3552ee595970601f960ad/src/reggae/dub/interop/dublib.d#L208 and I suppose other apps wanting to control how DUB builds are not using `new Dub` either and instead create custom PackageManager instances.

For the usual dub-as-a-library user this PR allows users to start a DUB build, reusing the default DUB cache directory, without hardcoding or copying the DUB cache paths or SpecialDirs implementation from DUB itself. So I think this is definitely a better API than burdening the users with figuring out where the cache should go for builds if they just want to call `dub build` through the API.

Note: there was `metadata_cache.json` that still used the old non-central cache path? I moved that into the central cache path now, but maybe @Geod24 knows why this was not yet in the central cache? That part might still need a change.

### DIFF
--- a/source/dub/generators/generator.d
+++ b/source/dub/generators/generator.d
@@ -121,11 +121,13 @@ class ProjectGenerator
 	protected {
 		Project m_project;
 		NativePath m_tempTargetExecutablePath;
+		PackageManager m_packageMan;
 	}
 
 	this(Project project)
 	{
 		m_project = project;
+		m_packageMan = project.packageManager;
 	}
 
 	/** Performs the full generator process.
@@ -331,7 +333,7 @@ class ProjectGenerator
 
 		if (genSettings.filterVersions)
 			foreach (ref ti; targets.byValue)
-				inferVersionFilters(ti);
+				inferVersionFilters(m_packageMan, ti);
 
 		// mark packages as visited (only used during upwards propagation)
 		void[0][Package] visited;
@@ -629,7 +631,7 @@ class ProjectGenerator
 	}
 
 	// infer applicable version identifiers
-	private static void inferVersionFilters(ref TargetInfo ti)
+	private static void inferVersionFilters(PackageManager packageManager, ref TargetInfo ti)
 	{
 		import std.algorithm.searching : any;
 		import std.file : timeLastModified;
@@ -656,7 +658,7 @@ class ProjectGenerator
 		auto srcs = chain(bs.sourceFiles, bs.importFiles, bs.stringImportFiles)
 			.filter!(f => dexts.canFind(f.extension)).filter!exists;
 		// try to load cached filters first
-		const cacheFilePath = packageCache(NativePath(ti.buildSettings.targetPath), ti.pack)
+		const cacheFilePath = packageManager.packageCache(ti.pack)
 			~ "metadata_cache.json";
 		enum silent_fail = true;
 		auto cache = jsonFromFile(cacheFilePath, silent_fail);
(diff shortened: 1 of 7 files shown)


rikkimax:
I take it that there is no possibility of adding a deprecated workaround to give existing users some time to upgrade?

WebFreak001:
I don't think there is any way really that wouldn't make the code very ugly, like keeping the old stuff in as second branch (with a lot of copy-pasted code)

The API has been released for over 6 months now already so I don't exactly like entirely removing it either.

I opted for implementing it this way, since it looked like pretty obviously @Geod24 wanted to get rid of the public access for the cache dir entirely, naming the getter "cachePathDontUse"

Nice thing with this PR is at least we are removing that `cachePathDontUse` method since it's no longer needed and I personally think like putting the path handling inside PackageManager is the better idea here than some global methods inside generator.d that need to pass in the user-specified cache directory. Although from a user perspective having an API to change the cache directory is nice as well.

I realized this was an issue because I didn't specify any `.cache` to the GeneratorSettings (just using code from the old API basically) and it would `assert(false)` inside TargetGenerator since it also exposed the cache file path since a few months ago and thus started making old code break.

However actually fixing the cache path to use what DUB uses isn't possible since `cachePathDontUse` is not public and also pretty obviously doesn't look like it should be made public. I don't know entirely what other plans Mathias has here with the cache code, so I don't want to just expose it either.

Alternatively to this PR I could also make it actually use the `cachePathDontUse` function as a default when no cache path is specified, which I think from a user's point of view would be just as valid of a fix. This would also move it from the command line, where private/package APIs really shouldn't be used, somewhere into the implementation so it might no longer need to be "don't use".

source/dub/generators/generator.d
```diff
@@ -121,11 +121,13 @@ class ProjectGenerator
 	protected {
 		Project m_project;
 		NativePath m_tempTargetExecutablePath;
+		PackageManager m_packageMan;
```
Geod24, line 124: I think that's the reason I didn't do it in the first place - Having the `PackageManager` in the generator is a bit icky.

Geod24 (commented):
I think I never manage to figure out a way to make it "nice". This approach is a bit of a slippery slope as other places that may need the cache may end up depending on `PackageManager`. Instinctively, I think that the cache part should be linked to the *Package*, but currently the class is not ready for this.

Geod24:
@WebFreak001 : See https://github.com/dlang/dub/pull/2764
Generator really shouldn't know about the PackageManager. How would you like to proceed ?

PR #2704, opened 2023-09-27
keep track of specified source paths in target generator

WebFreak001:
Makes target generator & dub describe able to print which source _paths_ are specified for each package, instead of only listing all the *.d files that have been globbed.

Someone at Funkwerk told me their packages are often using sourcePaths separately from importPaths for packages, so this should cover that use-case and make it possible to implement properly in serve-d.


WebFreak001:
bump

PR #2714, opened 2023-11-11
fix memory deprecations

WebFreak001:
makes dub build without deprecations again

the preview=in seems to taint other packages if they depend on dub as a library though


Geod24:
#2783

PR #2717, opened 2023-11-20
Initial work towards a parallel dub building by supporting deferred logging

rikkimax:
This introduces the deferred logging struct, unless enabled it should act identical to how it has in the past.

The idea is we would swap every function that ``ProjectGenerator.generate`` calls to use the ``DereferredLogState`` by ref instead of the global functions.

When parallelized we would turn on the deferment and once all the builds have completed for a given parallel set it would print them so that they will not have cross pollution between the builds.

### DIFF
--- a/source/dub/internal/logging.d
+++ b/source/dub/internal/logging.d
@@ -333,40 +333,7 @@ void log(T...)(
 	lazy T args
 ) nothrow
 {
-	if (level < _minLevel)
-		return;
-
-	auto hasTag = true;
-	if (level <= LogLevel.diagnostic)
-		hasTag = false;
-	if (disableTag)
-		hasTag = false;
-
-	auto boldTag = false;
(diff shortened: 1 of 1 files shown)


PetarKirov:
This design decision looks odd to me. I can't think of any build system (all of which build independent jobs in parallel) that defer logging until after a job is finished. Typically you'd have the output from different jobs running parallel prefixed with the name of the job and often also colorized differently to make the it easy to follow visually.

PetarKirov:
E.g. nix build with nix-output-monitor, which I use almost every day:
https://asciinema.org/a/7hJXH2iFLEkKxG1lL25lspqNn

rikkimax:
That sounds like it has good tradeoffs too.

I'm happy to see that implemented, although I feel I've done my part for the initial work on parallelization.

Issue #2190, opened 2021-11-10
windows: dub won't run if TEMP is unset or empty


huglovefan:
### System information
- **dub version**: dub 1.27.0
- **OS Platform and distribution**: Windows 7 64-bit
- **compiler version** dmd 2.098.0

### Bug Description

if the `TEMP` environment variable is missing or has an empty value, dub will error out with `Environment variable not found` on startup

### How to reproduce?

in command prompt:

1. `set TEMP=`
2. `dub --help`

### Expected Behavior

should work the same as with `TEMP=.` or with the proper value (`C:\Users\qemu\AppData\Local\Temp`)

### Logs

i'm getting this when running dub through cygwin sshd:

```
% ssh qemu_win7 'dub --help'

object.Exception@std\process.d(217): Environment variable not found: TEMP
----------------
0x0033C37B
0x0033C132
```

i think this is the place it's failing:

https://github.com/dlang/dub/blob/4906205/source/dub/commandline.d#L396 (added in 4609540ac1c8bb60f8a575fb7736451b0438c572)

is that workaround still needed? i have no problems running rdmd inside cygwin with `TEMP=/tmp` (or unset) and all the paths printed by `rdmd --dry-run` are backslash-separated already

PR #2719, opened 2023-11-20
Fix issue #2190 - windows: dub won't run if TEMP is unset or empty

rikkimax:
I haven't ran this fix, but it's the only place that has a ``"TEMP"`` string so its pretty safe that's the cause.

Fixes issue #2190, I may have messed up the auto close attribute so if the merge of this doesn't please close.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -371,6 +371,8 @@ unittest {
 */
 int runDubCommandLine(string[] args)
 {
+	import std.file : tempDir;
+
 	static string[] toSinglePackageArgs (string args0, string file, string[] trailing)
 	{
 		return [args0, "run", "-q", "--temp-build", "--single", file, "--"] ~ trailing;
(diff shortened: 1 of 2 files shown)


Geod24 (changes requested):
Needs a test

rikkimax:
Okay CI went green prior to that force push (forgot to rename the test module module name after copying it).

source/dub/commandline.d
```diff
@@ -382,11 +384,23 @@ int runDubCommandLine(string[] args)
...
-	version(Windows){
-		// rdmd uses $TEMP to compute a temporary path. since cygwin substitutes backslashes
-		// with slashes, this causes OPTLINK to fail (it thinks path segments are options)
-		// we substitute the other way around here to fix this.
-		environment["TEMP"] = environment["TEMP"].replace("/", "\\");
+	{
+		// Guarantee that this environment variable is set
+		//  this is specifically needed because of the Windows fix that follows this statement.
+		// While it probably isn't needed for all targets, it does simplify things a bit.
+		if ("TEMP" !in environment)
```
WebFreak001, line 391: why did you put this outside the `version(Windows)` block? Spawned processes might behave differently if they suddenly have this environment variable set on POSIX

rikkimax: Generally speaking, I always prefer consistent behavior between platforms, otherwise its too easy to introduce platform-specific behavior that will break when things change or during porting.

However I can see that in this case with ``TMP`` it could add additional complications, and should leave it for Windows only.

WebFreak001 (approved):

rikkimax:
> Needs a test

That was done like 2 weeks ago.

WebFreak001:
that message was done 2 weeks ago as well :p

rikkimax:
> that message was done 2 weeks ago as well :p

Yeah, it was showing as waiting on that being completed.

PR #2733, opened 2023-12-17
Deprecate unused packageInfoFilenames

Geod24:
This is of little value (it's just a map) and not used within dub.


thewilsonator (approved):

Geod24:
Found the user: https://github.com/dlang/dub-registry/blob/ce84086011862249d64b21fda8c31a14e0935ee7/source/dubregistry/registry.d#L714
Will look into the code to see how it can be changed, I still believe the deprecation is warranted.

PR #2743, opened 2023-12-19
Remove logging from dub.dependency

Geod24:
The logging was not really useful: One unittest one, and a spurious one with no action.
There is a lot of value in having dependency be data-only, not having side effect or much dependency to the rest of dub.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -13,7 +13,6 @@ import dub.internal.vibecompat.data.json;
 import dub.internal.vibecompat.inet.path;
 import dub.package_;
 import dub.semver;
-import dub.internal.logging;
 
 import dub.internal.dyaml.stdsumtype;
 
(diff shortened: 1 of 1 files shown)


source/dub/dependency.d
```diff
@@ -327,9 +326,6 @@ struct Dependency {
 		Dependency dep;
 		if( verspec.type == Json.Type.object ){
 			if( auto pp = "path" in verspec ) {
-				if (auto pv = "version" in verspec)
-					logDiagnostic("Ignoring version specification (%s) for path based dependency %s", pv.get!string, pp.get!string);
```
WebFreak001, line 331: move this log message to where it should be logged instead please, I think this one has value since it indicates that the recipe / selections isn't properly defined.

WebFreak001 (changes requested):

source/dub/dependency.d
Geod24: There is nowhere to move it without exposing the content of the JSON object. Note that if we want to be strict about the content of the selection files (which I want to be!) it's better to have a generic mean of doing that (hence my previous work on the parser) rather than ad-hoc here and there.

Also note that this message is unlikely to be helpful because it's a `logDiagnostic` and hence only shows with `-v`.

WebFreak001 (approved):

Geod24:
@WebFreak001 : Thanks!

PR #2744, opened 2023-12-19
Move data part of dub.platform to dub.data

Geod24:
We want to separate the data part, which is exposed to the PackageRecipe, from the actual platform check code, which is used when building.

Note that it's exposed to `PackageRecipe` indirectly, through its `compiler` import, which I aim to remove.

### DIFF
--- a/build-files.txt
+++ b/build-files.txt
@@ -6,6 +6,7 @@ source/dub/compilers/dmd.d
 source/dub/compilers/gdc.d
 source/dub/compilers/ldc.d
 source/dub/compilers/utils.d
+source/dub/data/platform.d
 source/dub/data/settings.d
 source/dub/dependency.d
 source/dub/dependencyresolver.d
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -9,7 +9,7 @@ module dub.compilers.compiler;
 
 public import dub.compilers.buildsettings;
 deprecated("Please `import dub.dependency : Dependency` instead") public import dub.dependency : Dependency;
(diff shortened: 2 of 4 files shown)


WebFreak001 (approved):

s-ludwig:
@Geod24  Can you explain that in more detail? I don't see how `BuildPlatform` has anything to do with "data". If anything, you might want to separate the `PackageRecipe` logic from the data part. What is the reasoning for getting rid of the `compiler` import?

Geod24:
@s-ludwig : Sure. I want to separate everything that has to do with the `PackageRecipe` type definition in its own module.
That means moving `dependency`, `buildsettings`, `packagerecipe` to `dub.data`, and making sure those are (almost) leaf modules.

`dub.data` is not the greatest package name, I will give you that. But having the separation between data types and logic is what I am after and this will yield much cleaner code because module dependencies will be lower, and code will be easier to understand.

I hope this makes more sense ?

s-ludwig:
In that case, I would suggest to look into moving `getPlatformSettings` out of `packagerecipe` and letting `BuildPlatform` where it is. It could certainly be argued that `getPlatformSettings` is on a level closer to `Project` anyway.

Geod24:
Looking into it more, I'm not so convinced anymore that this PR was right. Can I have until year-end to complete what I am working on, and if I cannot make a compelling argument, we revert this ? I don't think it's going to take that long but I want a bit of margin.

s-ludwig:
@Geod24 Just a reminder so that this doesn't get forgotten.

Geod24:
Good point: https://github.com/dlang/dub/pull/2788

PR #2753, opened 2023-12-24
Add betterC probing for performance and for applications without drun…

etcimon:
This improves performance and also allows me to compile a custom druntime with my application. (post-switches = [] in the ldc2.conf)


rikkimax:
Duplicate of: https://github.com/dlang/dub/pull/2699

etcimon:
My solution involves minor changes and reverts to old behaviour, easier to approve.

source/dub/compilers/compiler.d
```diff
@@ -183,18 +183,32 @@ interface Compiler {
 		import dub.compilers.utils : generatePlatformProbeFile, readPlatformJsonProbe;
 		import std.string : format, strip;
 
-		auto fil = generatePlatformProbeFile();
+		NativePath fil = generatePlatformProbeFile!true();
+		auto result_betterC = execute(compiler_binary ~ args ~ "-betterC" ~ fil.toNativeString());
```
thewilsonator, line 187: `-betterC` is called something else with GDC ,`-f-no-druntime` or something like that.

etcimon: Seems to be -fno-druntime

source/dub/compilers/utils.d
```diff
@@ -266,16 +266,28 @@ private enum probeEndMark = "__dub_probe_end__";
...
+		alias string = const(char)[];
+		alias size_t = uint;
+		template _d_arrayappendcTXImpl(Tarr : T[], T){
+			ref Tarr _d_arrayappendcTX(return ref scope Tarr px, size_t n) @trusted pure nothrow{return px;}
+		}
+		const(char)[][] _d_arrayappendcTX(return scope ref const(char)[][] px, size_t n){return px;}
+		Tarr _d_arrayappendT(Tarr : T[], T)(return ref scope Tarr x, scope Tarr y){return x;}
+	} : q{
+		module dub_platform_probe;
+	};
```
Geod24, line 286: I'm a bit concerned about this part, as it seems to be specific to the current state of BetterC. Future improvements might make this out of date. Am I right that the problem arises because the platform probe has a dependency on the runtime ? Couldn't we achieve something similar with a sufficiently sized static array ?

etcimon: It's the string[] appender running as CTFE that calls the runtime as a stub, if there's any improvements we would be removing some definitions and that's pretty much it. Either way if anything breaks it'll show a warning and fallback to using the runtime (through the bool retry in the catch)

etcimon: The CTFE doesn't support using static storage either, you have to use the GC to create an allocation and then you can reference it in scope

etcimon: I managed to remove the declarations by using only pragma(msg, ) and formatting the string at the end

Geod24:
Just to be sure I follow: Do we need the `-betterC` flag if we make it `-betterC`-compliant ? As in, if you get it to a point it does not pull anything from the runtime, the flag becomes superfluous, doesn't it ?

rikkimax:
> Just to be sure I follow: Do we need the `-betterC` flag if we make it `-betterC`-compliant ? As in, if you get it to a point it does not pull anything from the runtime, the flag becomes superfluous, doesn't it ?

No, it doesn't.

You will still be linking against druntime, and that has cost.

source/dub/compilers/compiler.d
```diff
@@ -177,27 +177,29 @@ interface Compiler {
 			args			=	arguments for the probe compilation
 			arch_override	=	special handler for x86_mscoff
 	*/
-	protected final BuildPlatform probePlatform(string compiler_binary, string[] args,
-		string arch_override)
+	protected final BuildPlatform probePlatform(string compiler_binary, string[] args, string arch_override)
 	{
-		import dub.compilers.utils : generatePlatformProbeFile, readPlatformJsonProbe;
+		import dub.compilers.utils : generatePlatformProbeFile, readPlatformSDLProbe;
```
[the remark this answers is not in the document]
etcimon: I wasn't able to append commas properly in the pragmas so I thought sdl syntax is easier. This probe is basically an internal hack anyways.

etcimon:
The `probePlatform` routine takes ~25ms (vs ~38ms previously) for me using a `StopWatch` when building with these improvements. It's now also compatible with projects that don't include druntime.

etcimon:
On dmd removing the `-betterC` flag improves it to 18ms, but with ldc it worsens it to 40ms.

Geod24:
> You will still be linking against druntime, and that has cost.

Why do we need to link ?

.gitignore
```diff
@@ -16,6 +16,7 @@ __dummy.html
 /bin/dub-test-library
 /bin/libdub.a
 /bin/dub-*
+/bin/dub.*
```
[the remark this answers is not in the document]
etcimon: It's a minor annoyance when using vscode source control on windows, I was thinking it didn't require mention, nor its own pull request.

etcimon:
It wouldn't link, it doesn't write an object file, compile flag `-o-`. It's mostly dmd.dinterpret in the compiler doing most of the work and returning after.

etcimon:
Originally it's because this project https://github.com/etcimon/libwasm implements its own druntime and for it you must remove the includes in ldc2.conf, but probing fails on that configuration if it depends on druntime.

Geod24:
> but probing fails on that configuration if it depends on druntime.

I assume that probing fails because it tries to import druntime, rather than link it ?

etcimon:
It complains that there is no `module object`

Geod24:
@kinke : Any chance you could have a look at this ? I think the use case is good, but I would like to make sure it keeps on working with newer versions of DMD/LDC with minimal maintenance.

source/dub/compilers/compiler.d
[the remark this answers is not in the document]
Geod24: Well just `pragma(msg, ",");` ? Or something like:
```
pragma(msg, `"key": "`, value, `",`);
```
?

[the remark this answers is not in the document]
etcimon: The trailing comma is invalid json and it's quite complicated to sanitize it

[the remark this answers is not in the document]
Geod24: Fair. I'd like to reduce the dependency on SDL and keep the JSON one, but I don't think you should be blocked on this.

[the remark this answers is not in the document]
etcimon: My solution would involve having invalid json spit out by the probe and then replacing `,]` with `]` afterwards, if that looks fine to you I rewrite it with json, but honestly even having a csv-like format and no json would make more sense.

etcimon:
> So why is `-betterC` relevant here ?

It works without `-betterC`, but it does help with both performance and guarantees.

source/dub/compilers/compiler.d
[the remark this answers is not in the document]
Geod24: Leave it as is, if I care enough I will make it use JSON5, but as you said, it's internal, so we can change it anytime.

kinke:
AFAICT after a quick glance, the main change here is that the generated probe module becomes the special `object` module, only defining the `string` alias (not even `size_t`), so that the (fat) druntime one isn't imported. `-betterC` then shouldn't really make a difference, as its effects should be enabled automatically without `TypeInfo` class etc. in `object`.

While such a minimalistic, fully self-contained D module is surely fast to analyze with `-o-` and working for ~all exotic targets too, the cost is the ugly duplication in `source/dub/platform.d`. I guess that could be improved though by post-processing the existing string, e.g., replacing `ret ~= <literal>;` by `pragma(msg, ´ <literal>´);` and stripping off `string[] ret;` and `return ret;`.

etcimon:
> I guess that could be improved though by post-processing the existing string

The post-processing using regex, matchAll and replace makes it impossible to evaluate the string with ctfe, but I doubt this would make it slower.

.gitignore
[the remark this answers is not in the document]
Geod24: What artifacts does vscode generate ? We can keep it in this PR, I just could not tell where it's coming from.

source/dub/compilers/compiler.d
```diff
@@ -177,27 +177,26 @@ interface Compiler {
...
+		NativePath fil = generatePlatformProbeFile();
 
 		auto result = execute(compiler_binary ~ args ~ fil.toNativeString());
 		enforce!CompilerInvocationException(result.status == 0,
 				format("Failed to invoke the compiler %s to determine the build platform: %s",
 				compiler_binary, result.output));
+		BuildPlatform build_platform;
 
-		auto build_platform = readPlatformJsonProbe(result.output);
+		build_platform = readPlatformSDLProbe(result.output);
```
Geod24, line 193: Merge those two lines ?

source/dub/compilers/compiler.d
```diff
@@ -177,27 +177,26 @@ interface Compiler {
...
+		string ver = determineVersion(compiler_binary, result.output).strip;
 		build_platform.compilerBinary = compiler_binary;
 
-		auto ver = determineVersion(compiler_binary, result.output)
-			.strip;
 		if (ver.empty) {
 			logWarn(`Could not probe the compiler version for "%s". ` ~
-				`Toolchain requirements might be ineffective`, build_platform.compiler);
+					`Toolchain requirements might be ineffective`, build_platform.compiler);
```
Geod24, line 199: Revert

source/dub/compilers/utils.d
```diff
@@ -268,78 +268,103 @@ private enum probeEndMark = "__dub_probe_end__";
 */
 NativePath generatePlatformProbeFile()
 {
+
```
Geod24, line 271: Revert

source/dub/compilers/utils.d
```diff
@@ -268,78 +268,103 @@ private enum probeEndMark = "__dub_probe_end__";
...
-	import dub.internal.vibecompat.data.json;
 	import dub.internal.utils;
 	import std.string : format;
 
+	enum moduleInfo = q{
+		module object;
+		alias string = const(char)[];
+	};
+
 	// try to not use phobos in the probe to avoid long import times
```
Geod24, line 281: I would also modify the comment so the next person knows the platform probe should not rely on the runtime to make your use case possible.

Geod24 (approved):
Looks pretty nice, just a few nits. I don't know if you want/can add a test to avoid a regression, I'm happy to merge as-is otherwise once the nits have been addressed.

.gitignore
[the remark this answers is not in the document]
etcimon: This is for the dub.exe and dub.pdb files that are generated when building under windows.

etcimon:
Should I squash it?

PR #2759, opened 2023-12-27
Add unittest framework and some dependency tests

Geod24:
```
Dub has always been notoriously hard to test due to the amount of IO it does.
Over the past few years, I have progressively refactored it to allow
dependency injection to take place, which this finally put into action.
By overriding the `PackageManager`, `PackageSupplier`, and a few strategic
functions, we can start to unittest Dub's behavior solely in unittests.
This should hopefully tremendously helps with adding regression tests
for the package manager side of Dub (the build side still need work
to have similar capabilities).
```

See each commit for details. In the long run we might want to go with a full filesystem abstraction, especially when it comes to testing the building side, but that should allow us to write some good tests for over half of Dub.

### DIFF
--- a/source/dub/test/base.d
+++ b/source/dub/test/base.d
@@ -13,10 +13,11 @@ public import std.algorithm;
 
 import dub.data.settings;
 public import dub.dependency;
-import dub.dub;
-import dub.package_;
+public import dub.dub;
+public import dub.package_;
 import dub.packagemanager;
 import dub.packagesuppliers.packagesupplier;
+import dub.project;
 
 // TODO: Remove and handle logging the same way we handle other IO
 import dub.internal.logging;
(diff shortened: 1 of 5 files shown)


Geod24:
> ⚠️ This PR introduces new deprecations:

Yeah that's a DMD bug. Even though we're in a deprecated context, `map`ping a deprecated function triggers a deprecation 🤷

thewilsonator (approved):

source/dub/test/base.d
```diff
@@ -0,0 +1,277 @@
...
+		import dub.recipe.io;
+		final switch (fmt) {
+			case PackageFormat.json:
+				auto recipe = parsePackageRecipe(str, "dub.json");
+                recipe.version_ = vers.toString();
+                return new Package(recipe);
+			case PackageFormat.sdl:
+				auto recipe = parsePackageRecipe(str, "dub.sdl");
+                recipe.version_ = vers.toString();
+                return new Package(recipe);
```
maxhaton, line 115: it github broken or is the indentation all screwed up here

PR #2760, opened 2023-12-28
Test framework: Add documentation and an example UT

Geod24:
The documentation is in the diff.


PetarKirov (approved):
Nice work (not just this, but the unit test framework and the refactoring PRs as whole) 👍

PR #2761, opened 2023-12-28
Test framework: Return fully typed TestPackageManager from TestDub

Geod24:
Using covariant return type, we can avoid the need to change the base class, and leave all the overloading in the child class.

### DIFF
--- a/source/dub/test/base.d
+++ b/source/dub/test/base.d
@@ -83,6 +83,17 @@ public class TestDub : Dub
 	public alias loadPackage = Dub.loadPackage;
 
 	/**
+	 * Returns a fully typed `TestPackageManager`
+	 *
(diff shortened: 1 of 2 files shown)


source/dub/test/base.d
```diff
@@ -212,6 +223,21 @@ package class TestPackageManager : PackageManager
 
 		return null;
     }
+
+    /**
+     * Adds a `Package` to this `PackageManager`
+     *
+     * This is currently only available in unittests as it is a convenience
```
WebFreak001, line 230: documentation is lying, this is publicly available now and not only in unittests.

Also, using spaces instead of tabs

Geod24: There is a `version (unittest)` at the top of the module.

WebFreak001 (approved):

PR #2763, opened 2023-12-28
Remove fallback to ~master on dub fetch, fix #2679

### LINKED ISSUE #2679: `dub fetch` for non-existent package name should have better error message
### System information

- **dub version**: 1.33.0
- **OS Platform and distribution**: Linux Mint 21
- **compiler version** dmd v2.104.0

### How to reproduce?

```
nick@nmt-pc:~/git/english$ dub fetch libssl-dev 
             Getting a release version failed: (1): Error: Got JSON of type null_, expected object.
             Retry with ~master...
Error (1): Error: Got JSON of type null_, expected object.
```
### Expected Behavior

The code should show an error like:

    Error: No package `libssl-dev` found

Geod24:
The retry does not make sense, as one cannot add a package to the registry without a release version.
It is also not in line with Dub strategy, as we are trying to reduce the usage of branch dependencies.

This also fixes issue https://github.com/dlang/dub/issues/2679 which pointed out the error message
was sub-par. The API of the registry seems to have changed and
we now get `null` instead of `undefined`, leading to an error
inside the JSON module instead of the enforce being triggered.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -2176,8 +2176,7 @@ class FetchCommand : FetchRemoveCommand {
 			}
 			catch(Exception e){
 				logInfo("Getting a release version failed: %s", e.msg);
-				logInfo("Retry with ~master...");
-				dub.fetch(name, VersionRange.fromString("~master"), location, fetchOpts);
+				return 1;
 			}
 		}
 		return 0;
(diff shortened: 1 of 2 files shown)


Geod24:
Edited to pass CI and fix an issue in the process

thewilsonator (approved):

PR #2769, opened 2024-01-02
Deprecate `Package.load`, use `PackageManager.load` instead

Geod24:
In order for the dependency injection approach to work, we need to be able to override every single instance of `Package.load`. We could not override `Package.load` itself as it is a static function.
However, in the process of adding support for path-based dependencies to the test suite, it became obvious that the functionality offered by `Package.load` is actually better suited for the `PackageManager`. Deprecating `Package.load` confirms this: Not a single call was outside of `PackageManager`, and no extra parameter had to be provided anywhere to make this work.
This also nicely simplifies the `TestPackageManager` implementation, as we can now overload `load` and let `getPackage` alone.

### DIFF
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -80,9 +80,10 @@ static immutable string[] builtinBuildTypes = [
 /**	Represents a package, including its sub packages.
 */
 class Package {
+	// `package` visibility as it is set from the PackageManager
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

PetarKirov (commented):
👍

PR #2785, opened 2024-01-09
Provide a consistent experience when interacting with dub fetch

Geod24:
'dub fetch doesNotExists' and 'dub fetch doesNotExists@1.0.0' were behaving differently, both when they succeeded and when they failed. Making them consistent was just a matter of deleting a few lines of code, allowing us to also remove a branch.

PR #2786, opened 2024-01-09
fetch: Support multiple arguments, projects, recursive fetch

Geod24:
```
Dub fetch was overly limited for no reason: Since we migrated from
`--version` to `name@version`, there is no reason to only allow one
argument, multiple packages can be fetched. Additionally, it could
be useful to fetch a package's dependency without building it,
which the command now supports.
```


Geod24:
Based on #2785

thewilsonator (approved):

PR #2790, opened 2024-01-10
Add overrideCachePaths alongside customCachePaths

nordlow:
This is needed in a scenario where all third-party dependencies of project are vendored under an in-repo directory tree mimicing the structure of `~/.dub/packages`.

### DIFF
--- a/source/dub/data/settings.d
+++ b/source/dub/data/settings.d
@@ -38,6 +38,7 @@ public enum SkipPackageSuppliers {
 package(dub) struct Settings {
     @Optional string[] registryUrls;
     @Optional NativePath[] customCachePaths;
+    @Optional NativePath[] overrideCachePaths;
 
     SetInfo!(SkipPackageSuppliers) skipRegistry;
     SetInfo!(string) defaultCompiler;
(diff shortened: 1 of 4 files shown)


source/dub/data/settings.d
```diff
@@ -171,3 +172,90 @@ unittest {
      auto m3 = Settings.init.merge(c1);
      assert(m3 == c1);
 }
+
```
nordlow, line 175: Shall I instead parameterize the unittest on settings key being either `customCachePaths` or `overrideCachePaths`?

source/dub/packagemanager.d
```diff
@@ -167,6 +167,25 @@ class PackageManager {
...
 
+	/** Override (read-only) package cache paths to search for packages.
+
+		Cache paths have the same structure as the default cache paths, such as
+		".dub/packages/".
+
+		Note that previously set custom paths will be removed when setting this
+		property.
+	*/
+	@property void overrideCachePaths(NativePath[] override_cache_paths)
```
nordlow, line 178: Alternatively, we could add an enum flag to `customCachePaths that describes how the lookup path is customized; typically as appended, prepended or replaced.

Geod24:
I don't think we should do this. I had a long discussion with @kinke about what is needed for the monorepo.

We want to have a repository with multiple, independent packages within it, *except* for the fact they share the same set of dependencies. We want to be able to ensure this, and for foreign (non-internally developed) packages to have their dependencies overriden.

To me, it sounds like we want to cripple Dub's abilities as a package manager to its bare minimum and just use it as a build system. We also want to do so in a way that is *local*, e.g. without having our developers need to change their user or system-wide settings.

This doesn't really help with this. Today we could just store packages in `.dub/packages/` and we would not need this diff.

nordlow:
> This doesn't really help with this. Today we could just store packages in `.dub/packages/` and we would not need this diff.

Closing this then.

PR #2796, opened 2024-01-13
Let uesr $DFLAGS override build settings as much as possible

the-horo:
This change is mostly about being able to build programs without `-Werror` or `-w`.

`dub` setting up so that warnings are errors by default can be nice for developers making their code more correct but it is useless for consumers. Currently, there is no way to disable that without modifying the project configuration to add `allowWarnings` to `buildRequirements`. Users should, at least, be able to add `-wi` to `DFLAGS` and have the preference respected.

This isn't so much a problem with `dmd` or `ldc` in my experience as they mostly report deprecations, the problem is with `gdc` where every deprecation is a warning and the `-Werror` makes builds fail where they otherwise should have succeeded with only some warnings.

### DIFF
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -230,6 +230,11 @@ config    /etc/dmd.conf
 	{
 		enforceBuildRequirements(settings);
 
+		// Keep the current dflags at the end of the array so that they will overwrite other flags.
+		// This allows user $DFLAGS to modify flags added by us.
+		const dflagsTail = settings.dflags;
+		settings.dflags = [];
+
 		if (!(fields & BuildSetting.options)) {
 			foreach (t; s_options)
 				if (settings.options & t[0])
@@ -281,6 +286,8 @@ config    /etc/dmd.conf
 
 		if (platform.platform.canFind("posix") && (settings.options & BuildOption.pic))
 			settings.addDFlags("-fPIC");
+
+		settings.addDFlags(dflagsTail);
 
 		assert(fields & BuildSetting.dflags);
 		assert(fields & BuildSetting.copyFiles);
--- a/source/dub/compilers/gdc.d
+++ b/source/dub/compilers/gdc.d
@@ -89,6 +89,11 @@ class GDCCompiler : Compiler {
 	{
 		enforceBuildRequirements(settings);
 
+		// Keep the current dflags at the end of the array so that they will overwrite other flags.
+		// This allows user $DFLAGS to modify flags added by us.
+		const dflagsTail = settings.dflags;
+		settings.dflags = [];
+
 		if (!(fields & BuildSetting.options)) {
 			foreach (t; s_options)
 				if (settings.options & t[0])
@@ -137,6 +142,8 @@ class GDCCompiler : Compiler {
 
 		if (settings.options & BuildOption.pic)
 			settings.addDFlags("-fPIC");
+
+		settings.addDFlags(dflagsTail);
 
 		assert(fields & BuildSetting.dflags);
 		assert(fields & BuildSetting.copyFiles);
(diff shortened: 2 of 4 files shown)


Geod24:
Currently you can work around this by using `dflags` on dependencies. It isn't pretty but it works.
GDC should really treat deprecations and warnings differently though.

the-horo:
> Currently you can work around this by using `dflags` on dependencies. It isn't pretty but it works.

Can you explain me how this solution would go as I didn't know you could do that?

Geod24:
```json
 "dependencies": {
        "vibe-d":           { "version": "~>0.9", "dflags" : [ "-preview=in", "-revert=dtorfields" ] }
    }
```

It will apply transitively, so `eventcore`, `vibe-core`, etc... will have those DFLAGS defined.
Not ideal but until we take the time to untangle how we want the build system to look like, this is the "best" way to do it.

the-horo:
> ```json
>  "dependencies": {
>         "vibe-d":           { "version": "~>0.9", "dflags" : [ "-preview=in", "-revert=dtorfields" ] }
>     }
> ```
> 
> It will apply transitively, so `eventcore`, `vibe-core`, etc... will have those DFLAGS defined. Not ideal but until we take the time to untangle how we want the build system to look like, this is the "best" way to do it.

My issue is more of the flags being put in the wrong place and overwriten rather then not applying transitively:
```
$ DFLAGS='-Wno-error' dub -v build --compiler=gdc
...
    Building dub 1.36.0-beta.1+commit.37.gf86ed3e9: building configuration [application]
gdc -Wno-error -o <cache>/dub -Werror -Wall -fversion=DubUseCurl -fversion=DubApplication ...
```
The same thing happens if I put `dflags "-Wno-error"` in `dub.sdl`.

There should be a way to have flags be passed later on the command line so that their values are respected, that's what I'm trying to fix.

Geod24:
Have you tried `"buildRequirements": [ "allowWarnings" ]` ?

PetarKirov:
@Geod24 I think there is a common use case which Dub doesn't support very well. While developers working on a given project may find that adding `"buildRequirements": [ "allowWarnings" ]` to their dub.sdl/json may be the ideal approach, the same is not true for consumers of projects which ideally should not need to modify the source code of any of their dependencies. E.g. imagine a user of Linux distro that wants to rebuild their entire system for their exact CPU. Many of them would like the simplicity exporting `CFLAGS=-march=native` and hitting "rebuild all".

the-horo:
> Have you tried `"buildRequirements": [ "allowWarnings" ]` ?

Yes, that fixes it, but my issue is not about having a way to disable it, it's having an easy-access way to do it. Editing the `dub.*` file I don't consider enough. You can also consider the second issue of having user flags be respected more, which I think is always nice.

> @Geod24 I think there is a common use case which Dub doesn't support very well. While developers working on a given project may find that adding `"buildRequirements": [ "allowWarnings" ]` to their dub.sdl/json may be the ideal approach, the same is not true for consumers of projects which ideally should not need to modify the source code of any of their dependencies. E.g. imagine a user of Linux distro that wants to rebuild their entire system for their exact CPU. Many of them would like the simplicity exporting `CFLAGS=-march=native` and hitting "rebuild all".

I agree with this. Gentoo, for which I maintain D packages, does require disabling `-Werror`: https://devmanual.gentoo.org/ebuild-writing/common-mistakes/index.html#-werror-compiler-flag-not-removed.

Geod24:
@the-horo : I can see the use case being useful, I myself had my share of similar issues.
I'm starting to think the way we currently do DFLAGS does not make much sense, as it's its own build type. Anyway, looking at the code with a fresh pair of eyes, I think we should go ahead with it.

Geod24 (approved):

PR #2801, opened 2024-01-15
Introduce PackageName, to fully type a package name

Geod24:
```
Package names being passed around in the Dub codebase are string, which is problematic as it suffers from validation / semantic issues. Validation because we cannot impose restrictions on characters or length easily, and semantic because some functions expect a main package while some would accept a subpackage. Introducing a type will remedy both those issues.
```

Note that this just replaces use of the two functions that deal with package names as string. Deprecating / replacing the API is another undertaking that I felt was best done separately.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1557,7 +1557,7 @@ class BuildCommand : GenerateCommand {
 		if (packageParts.name.startsWith(":"))
 			return 0;
 
-		const baseName = getBasePackageName(packageParts.name);
+		const baseName = PackageName(packageParts.name).main;
 		// Found locally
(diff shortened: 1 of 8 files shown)


Geod24:
This is a much simpler approach than #2360 and can be built on.
CC @nordlow

PetarKirov (approved):

PetarKirov:
@Geod24 does it make sense to rebase #2360 now that this PR is merged? Are there useful stuff that we can salvage?

Geod24:
Not that I can think of. I started to adapt the APIs locally, mostly for the benefit of the `PackageManager`. Already found a bug!

PR #2806, opened 2024-01-17
Deprecate Package.getInternalSubPackage, simplify DepedencyResolver

Geod24:
The PackageManager will always eagerly load subpackages, whether they are path based or embedded (see addPackages). So calling getInternalSubPackage if getSubPackage returned null will never yield a non-Nullable object.
In short, we can simplify that code greatly, and as Package.getInternalSubPackage does not handle all use case and require using the PackageManager anyway, just deprecate it and direct people to the right function.


Geod24:
@thewilsonator : This one too ? ;)

thewilsonator (approved):

PR #2808, opened 2024-01-17
Improve Dub.fetch API

Geod24:
The API was not ideal, as it meant the caller has
to be a bit repetitive in what it was doing:
- defaultPlacementLocation should be used when possible;
- A library user would most likely want FetchOptions.none;
- The most simple use case would be 'dub.fetch(pkgName)', meaning get me the latest version of pkgName, as is used by the tools fetching code; All of this is now supported, in addition to a transition to using the PackageName API.


Geod24:
CC @atilaneves as I think you use this API.

thewilsonator (approved):

PR #2820, opened 2024-01-19
Improve unittest framework to have a virtual FS

Geod24:
This gets the unittest framework closer to the actual behavior, and allows us to mock basic FS operations such as mkdir and writeFile. Further improvements can be made to support more operations, and to have package setup done before `TestDub` instantiation.


Geod24:
There's a mismatch between what our vibecompat `path` offers and what vibe-core has. Fix in: https://github.com/dlang/dub/pull/2821

Geod24:
I used a different approach since #2821 introduces its own set of issues.

thewilsonator (approved):

PR #2851, opened 2024-02-12
Test: Re-enable 'scan' behavior in TestPackageManager

Geod24:
Now that 'scan' is implemented in the TestPackageManager, there is no reason to deviate from what the base class and Dub binary does.


Geod24:
Actually I should do it the other way around

PR #2856, opened 2024-02-14
Tests: Use TestDub FSEntry constructor for more accurate tests

Geod24:
The previous approach added Packages in the PackageManager directly, completely sidestepping the scanning the PackageManager does, which is a behavior we absolutely want to test.

By setting up the filesystem before we instantiate our `TestDub` instance, we can remove a bunch of exceptions / overrides, and get a much closer behavior to the final product.


Geod24:
Wonderful, this uncovered a bug, which is now fixed.

PR #2867, opened 2024-02-17
add some missing PackageName overloads

WebFreak001:


source/dub/project.d
```diff
@@ -1946,9 +1952,13 @@ public class SelectedVersions {
...
 		valid and it is a version selection.
 	*/
-	deprecated("Use the overload that accepts a `PackageName`")
 	Dependency getSelectedVersion(string packageId) const
 	{
+		// TODO: we want to deprecate this overload, however we can't really do
+		// that until the selectedPackages exposes PackageName, otherwise this
+		// doesn't work without warning, but obviously should:
+		// foreach (key; project.selections.selectedPackages)
+		//     ... = project.selections.getSelectedVersion(key);
```
Geod24, line 1961: The following will work:
```
project.selections.getSelectedVersion(PackageName(key));
```

Not ideal but `PackageName.this` is fairly light.

WebFreak001: that really doesn't look like its intended API for how to iterate over selected packages though, selectedPackages should obviously just return PackageName[] instead - getSelectedVersion should only take as input what selectedPackages returns (string) - we don't want to encourage the library users to put PackageName(...) everywhere, it should just be constructed where explicitly needed, e.g. when converting from user input (string) to it. When it comes from the library it should be PackageName everywhere where applicable.

Also, while I toyed around with this tried to use PackageName everywhere, it quickly looked like an issue that PackageName.main also returns PackageName - we should probably have an extra type like RootPackageName that is not allowed to have any sub-package in it. (it can implicitly convert into PackageName though) Otherwise it can quickly become ambiguous and will quickly suffer from the same problems as using string everywhere. We should definitely add such a type before PackageName is rolled out as library part.

Geod24: > that really doesn't look like its intended API for how to iterate over selected packages though, selectedPackages should obviously just return PackageName[] instead - getSelectedVersion should only take as input what selectedPackages returns (string) - we don't want to encourage the library users to put PackageName(...) everywhere, it should just be constructed where explicitly needed, e.g. when converting from user input (string) to it. When it comes from the library it should be PackageName everywhere where applicable.

Yes. Unfortunately we can't do overloads based on return type. Same reason why `Package.name` is still `string`. This will be a v2 change. So the current state is IMO the best compromise. We could also just introduce an overload, e.g. `byKey`, that returns `PackageName`.

> Also, while I toyed around with this tried to use PackageName everywhere, it quickly looked like an issue that PackageName.main also returns PackageName - we should probably have an extra type like RootPackageName that is not allowed to have any sub-package in it. (it can implicitly convert into PackageName though) Otherwise it can quickly become ambiguous and will quickly suffer from the same problems as using string everywhere. We should definitely add such a type before PackageName is rolled out as library part.

I hit a similar problem. In practice, dependency resolution barely cares about non-base/root package names because the version  for subpackages is always the same. Likewise for `dub fetch`. But at least this is now somewhat visible / documented in the code. Note that the original version of `PackageName` returned `string` for `main`. It didn't work well.

WebFreak001: there have already been backwards incompatible changes in this regard, so I'd suggest we just edit some things such as the selections and other places where it's obvious immediately and break this usage all at once instead of breaking parts of it every release.

Geod24: Where do we have backward incompatible changes ? I never intended to make breaking changes, that's why there are so many deprecated overloads in Dub ATM.

WebFreak001: ok sorry turned out it's just a single thing that I had to deal with, so it's not so bad, however it did need changes:
- PackageDependency.name from `Package.getAllDependencies()` changed to PackageName without deprecation period

I really like the PackageName change and would like to use it in more places. My suggestion would be for more rarely used APIs to already change them to PackageName early.

Should package names inside the recipe struct be of type string or PackageName? PackageName implies semantics that it would have been checked already so I don't think it'd be the right thing that early on. Also we need a parse method that actually verifies the contents and some kind of fromTrustedString method that asserts on error, similar to the vibe.d APIs for stuff like paths.

Geod24: > PackageDependency.name from Package.getAllDependencies() changed to PackageName without deprecation period

Let me have a look at it. It was a bit annoying to deal with `PackageDependency` when doing the migration, but I found a way to make it work. However, it might have been relying on the `alias this` which I then removed...

> Should package names inside the recipe struct be of type string or PackageName? PackageName implies semantics that it would have been checked already so I don't think it'd be the right thing that early on.

In general we should check user input as early as possible. Having it checked when parsing makes sense to me because we can then point the user to the right place. Configy has various hooks (`fromString`, constructor overloads) that it can call, and if they throw, it will display  the error message with all the required information.
For example:
```D
File Edit Options Buffers Tools D Help
/++ dub.json:
 { "name": "test", "dependencies": { "configy": "*" } }
+/
module c;

import std.format;
import std.stdio;
import configy.Read;

struct PackageName {
    this (string value)
    {
        if (value != "42")
            throw new Exception("Value %s is not right".format(value)\
);
    }

    private string val;
}

struct Nested { PackageName name; }
struct Recipe { Nested nested; }

void main ()
{
    try
        parseConfigString!Recipe(`{ "nested": {
 "name": "notQuiteRight"
}}`, "dub.json");
    catch (ConfigException exc)
        writefln("%S", exc);
}
```

This displays (with colors):
```
% dub c.d
dub.json(1:9): nested.name: Value notQuiteRight is not right
```

So it would be fairly trivial to do this change, however it's a breaking change. We are getting closer to a point where Dub 2.0 is inevitable and I'd prefer to pack the breaking changes for that moment.

Geod24:
@WebFreak001 : Can we move forward here ?

WebFreak001:
yeah does anything need to be changed?

Geod24:
I would say limit it to adding the overload, and keep the deprecation / comment out ?

Geod24:
Ping @WebFreak001

PR #2872, opened 2024-02-21
Allow tests to call customCachePaths

Geod24:
This is now possible, as refresh has been mostly implemented.


WebFreak001:
should we add a test using this in that case?

PetarKirov (approved):

PR #2876, opened 2024-02-22
Remove the need for intermediate `.zip` file in PackageSupplier

Geod24:
```
By using a different overload of `retryDownload`,
we return the data instead of writing to disk,
ensuring that we don't have a lingering file
in memory and reducing IO.

This follow the same approach for deprecations as previous `PackageSupplier` deprecations: it's an inevitable, but easily fixable, breaking change for classes that implement `PackageSupplier`, but it's a proper deprecation for client code, based on the assumption that all the classes implementing `PackageSupplier` are in Dub.
```

### DIFF
--- a/source/dub/dub.d
+++ b/source/dub/dub.d
@@ -1021,13 +1021,11 @@ class Dub {
 		{
 			import std.zip : ZipException;
 
-			auto path = getTempFile(name.main.toString(), ".zip");
-			supplier.fetchPackage(path, name.main, range, (options & FetchOptions.usePrerelease) != 0); // Q: continue on fail?
-			scope(exit) removeFile(path);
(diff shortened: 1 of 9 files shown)


thewilsonator (approved):

thewilsonator:
This fails CI

Geod24:
Yeah, because the registry actually issue a redirection to Github and `retryDownload` doesn't follow redirection. Will fix.

Geod24:
Yep that did it

Geod24:
@s-ludwig : Your input here would be appreciated. Do you see a reason why this is the wrong direction ?

In the long run I think having the `PackageSupplier` do the extraction makes more sense, so we're not tied to `.zip` files, but I wanted to remove the intermediate file in the meantime. It seems wasteful considering the data will be loaded in memory multiple times (first via download, then via readFile).

PR #2881, opened 2024-02-24
Simplify mocking IO in PackageManager, add a test for `add-path`

Geod24:
@s-ludwig : I would like to remove the scanning behavior in `PackageManager`. This is a source of pain for us, as it slows down builds, including with @atilaneves Reggae.

Last time that introduced a regression which you fixed in https://github.com/dlang/dub/pull/2481 .
This simplifies mocking quite a bit and introduces a test for it (see last commit / `source/dub/test/other.d` in the diff).
If you could have a quick look at the test and confirm that it covers your use case ?

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -896,7 +896,7 @@ class Command {
 
 	private bool loadCwdPackage(Dub dub, bool warn_missing_package)
 	{
-		auto filePath = Package.findPackageFile(dub.rootPath);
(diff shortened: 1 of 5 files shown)


Geod24:
This has quite a lot of improvements I need to fix other bugs (write more tests), so I'm going to merge it in ~10 hours unless anyone objects.

thewilsonator (approved):

PetarKirov (commented):
I reviewed the commits one-by-one and everything looks moving in the right direction. Awesome work @Geod24 👍

PR #2883, opened 2024-02-27
Draft: Make PackageManager.lookup() not call refresh()

nordlow:
Adding MR to trigger CI to see if anything breaks.

At least

```sh
dub build --compiler=dmd && DC=dmd DUB=bin/dub test/run-unittest.sh 
```

passes locally.


Geod24:
Done in #2923

PR #2897, opened 2024-04-11
Speed up module parsing with hand-written parser

atilaneves:
In one project at work it reduced the time to create the test configuration from 80ms to 40ms. This new vesion also passes a unit test that the old version did not.

### DIFF
--- a/source/dub/internal/utils.d
+++ b/source/dub/internal/utils.d
@@ -609,29 +609,81 @@ string determineModuleName(BuildSettings settings, NativePath file, NativePath b
 
 /**
  * Search for module keyword in D Code
+ * A primitive parser to skip comments and whitespace to get
+ * the module's name from the module declaration.
  */
 string getModuleNameFromContent(string content) {
-	import std.regex;
-	import std.string;
+	import std.ascii: isAlpha, isAlphaNum;
+	import core.exception: RangeError;
 
-	content = content.strip;
-	if (!content.length) return null;
+	enum keyword = "module";
 
-	static bool regex_initialized = false;
-	static Regex!char comments_pattern, module_pattern;
+	size_t i = 0;
+	auto foundKeyword = false;
 
-	if (!regex_initialized) {
-		comments_pattern = regex(`//[^\r\n]*\r?\n?|/\*.*?\*/|/\+.*\+/`, "gs");
-		module_pattern = regex(`module\s+([\w\.]+)\s*;`, "g");
-		regex_initialized = true;
+	auto ch() {
+		return content[i];
 	}
 
-	content = replaceAll(content, comments_pattern, " ");
-	auto result = matchFirst(content, module_pattern);
+	try {
+		while(i < content.length) {
+			if(ch == keyword[0] && content[i .. i + keyword.length] == keyword) {
+				// -1 because the end of the loop will advance by 1
+				i += keyword.length - 1;
+				foundKeyword = true;
+			}
+			else if(ch == '/') {
+				++i;
+				// line comment?
+				if(ch == '/') {
+					while(ch != '\n')
+						++i;
+				}
+				// block comment?
+				else if(ch == '*') {
+					++i;
+					while(ch != '*' && content[i + 1] != '/')
+						++i;
+					++i; // skip over closing '/'
+				}
+				// nested comment?
+				else if(ch == '+') {
+					++i;
 
-	if (!result.empty) return result[1];
+					size_t level = 1;
 
-	return null;
+					while(level > 0) {
+						if(ch == '/') {
+							++i;
+							if(ch == '+') {
+								++i;
+								++level;
+							}
+						}
+						if(ch == '+') {
+							++i;
+							if(ch == '/') {
+								--level;
+							}
+						}
+						++i;
+					}
+				}
+			}
+			else if(isAlpha(ch) && foundKeyword) {
+				const start = i;
+				while(isAlphaNum(ch) || ch == '.') {
+					++i;
+				}
+				return content[start .. i];
+			}
+			++i;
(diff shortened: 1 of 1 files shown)


source/dub/internal/utils.d
```diff
@@ -609,29 +609,81 @@ string determineModuleName(BuildSettings settings, NativePath file, NativePath b
...
+							++i;
+							if(ch == '/') {
+								--level;
+							}
+						}
+						++i;
+					}
+				}
+			}
+			else if(isAlpha(ch) && foundKeyword) {
```
rikkimax, line 673: The module name is an identifier list.

An identifier is UAX31 and is not limited to ASCII.

Instead ignore white space and look for semicolon removing the need to care about what identifier tables are in use.

rikkimax (changes requested):
As is this won't work due to identifier character sets. See comment for potential solution.

source/dub/internal/utils.d
atilaneves: The old version had `\w` in the regex, which is the same thing as here but worse because of numbers.

rikkimax: The old regex was unlikely to match what D had with the C99 tables.

Identifier tables can be interesting.

rikkimax: Appears ``_`` is missing causing at least one CI failure.

So yes, this approach most definitely is missing things.

Going with a positive approach will cause problems for us long term. Going negative with what is not valid will be a lot easier as the set is rather small (whitespace and terminator ``;``).

source/dub/internal/utils.d
```diff
@@ -609,29 +609,81 @@ string determineModuleName(BuildSettings settings, NativePath file, NativePath b
...
+							if(ch == '+') {
+								++i;
+								++level;
+							}
+						}
+						if(ch == '+') {
+							++i;
+							if(ch == '/') {
+								--level;
+							}
```
WebFreak001, line 667: add `else continue;`, otherwise `++/` will break with this (also test this, e.g. `/++ ++/`)

source/dub/internal/utils.d
WebFreak001: yeah `_` is missing, otherwise it's better than the code we had before and is

WebFreak001 (changes requested):

source/dub/internal/utils.d
rikkimax: No, it's worse.

std.regex will decode as required and compare against the alpha/numeric character tables.

This is ASCII only.

s-ludwig: > Going with a positive approach will cause problems for us long term. Going negative with what is not valid will be a lot easier as the set is rather small

Agreed, this should be the way to go here, code complexity and performance wise. The parser only has to work correctly under the assumption that the code is correct anyway. A sanity check for the identifier format can be done afterwards to avoid any shenanigans (e.g. injecting relative paths), but it wouldn't even matter if we had a digit in front, the compiler will already error out.

atilaneves:
TIL that `\w` in D does things it doesn't in Perl.

source/dub/internal/utils.d
```diff
@@ -609,34 +609,102 @@ string determineModuleName(BuildSettings settings, NativePath file, NativePath b
...
+				++i;
+				// line comment?
+				if(ch == '/') {
+					while(ch != '\n')
+						++i;
+				}
+				// block comment?
+				else if(ch == '*') {
+					++i;
+					while(ch != '*' && content[i + 1] != '/')
```
s-ludwig, line 651:
```suggestion
					while(ch != '*' || content[i + 1] != '/')
```

Right now, this matches "*?" and "?/", "?" standing for any character.

Test cases that should currently fail: `"/** module foo*/ module bar;"` `"/* / module foo*/ module bar;"`

source/dub/internal/utils.d
```diff
@@ -609,34 +609,102 @@ string determineModuleName(BuildSettings settings, NativePath file, NativePath b
...
+				if(!isIdentChar(ch))
+					return content[startIndex .. endIndex];
+				else continue;
+			} else if(!isIdentChar(ch) && foundKeyword && startIndex != 0) {
+				return content[startIndex .. endIndex];
+			}
+			++i;
+		}
+		return "";
+	} catch(RangeError) {
```
s-ludwig, line 700: Although this will probably work fine unless the code get compiled with `-release` or `-noboundscheck`, I'd consider this a serious code smell, as there are several things that can go wrong when catching `Error`s in general, and this also hides possible bugs in the implementation.

In case you think this good enough anyway, this should at least be marked `@safe` so that bounds checks remain in place for `-release`.

s-ludwig (changes requested):

source/dub/internal/utils.d
atilaneves: The reason I did that was that I thought this was a lot easier than checking the bounds over and over again.

s-ludwig: The undefined behavior that results from throwing `Error` is due to the fact that scoped destruction may be skipped on the way up to the exception handler - I also had a quick look and couldn't find this in the spec, but this is what Walter has stated in the past. In this case, there is nothing like this, but IMO with general a risk like this in production code, an `Error` should always be treated as a fatal condition (except that there should be a chance to preempt a `OutOfMemoryError`, but that's a different topic).

Anyway, I can still sleep with this approach, the more important issue is the parsing logic bug.

s-ludwig (approved):

kinke:
This fails/regressed:
```
assert(getModuleNameFromContent("module modules.foo;") == "modules.foo");
```

PR #2902, opened 2024-04-19
Speed up Package.getPackageConfigs

atilaneves:
`Package.getPackageConfigs` was taking 3 *seconds* for a particular dub package, which was made worse by the fact it was called multiple times with different arguments. The changes in this PR bring it down to just under 200ms, which IMHO is still slow.

Investigating led me to realise that `determineDependencyConfigs` was being called over and over again for the exact same packages, especially common dependencies like `mir-algorithm` and `unit-threaded`. This PR does two things:

First, it makes the array of package dependencies smaller via `.sort.uniq`.  This cut down the time by half to 1.5s.

Secondly, after taking a while to understand the inner workings of the function, it returns early for packages that have already been analysed. This brought it further down to the 100ms range.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -99,6 +99,12 @@ struct PackageDependency {
 		this.spec = s;
 	}
 
+	int opCmp(in typeof(this) other) @safe const {
+		return name == other.name
+			? spec.opCmp(other.spec)
+			: name.opCmp(other.name);
+	}
+
 	/// Name of the referenced package.
 	PackageName name;
 
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -573,13 +573,18 @@ class Package {
 	// Left as package until the final API for this has been found
 	package auto getAllDependenciesRange()
 	const {
+		import std.algorithm: sort, uniq;
+		import std.array: array;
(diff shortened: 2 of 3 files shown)


thewilsonator (approved):

s-ludwig:
Looks good as far as I can see. Two additional things caught my eye:
1. `getAllDependencies` now performs a redundant allocation of the returned list - since `getAllDependenciesRange` is `package`, we could now just scrap it and put the implementation directly into `getAllDependencies`
3. Hard to tell without profiling this in some way, but it looks like it may make sense to pre-compute the result for each `Package` instance, as `getAllDependencies` is called in some places that look like they might be called often during a run

s-ludwig:
Turns out there is actually code that mutates the dependencies of a `Package` through its `recipe` property, so the pre-computation won't work just like that, unfortunately. But on the other hand, for a certain project, this code got called more than 10 times per package on average when performing a build, so if this is something that turns up in a profiler, that should still be worthwhile to pursue.

atilaneves:
I considered speeding up `getAllDependencies` but it wasn't where the time was being spent. I even cached it at one point and it made no difference.

PR #2903, opened 2024-04-27
Respect skipRegistry value from configuration files

the-horo:
Currently, the skipRegistry value in configuration files is ignored, the final value is either the one the user passed on the command line, which is the intended behavior, or the value `none` if the user didn't pass a --skip-registry argument.

This commit fixes this by introducing the new value `default_` to the SkippackageSuppliers enum to know when the user didn't pass the skip-registry argument and the value should be taken from the configuration files.

Be aware that the output of an invalid `skipRegistry` value in `dub.settings.json` has changed:
Before this PR:
```
     Warning SkipPackageSuppliers does not have a member named 'blah'
```
After this PR:
```
     Warning /tmp/test/dub.settings.json(1:17): skipRegistry: SkipPackageSuppliers does not have a member named 'blah'
```

### DIFF
--- a/source/dub/data/settings.d
+++ b/source/dub/data/settings.d
@@ -17,6 +17,7 @@ public enum SkipPackageSuppliers {
     standard,   /// Does not use the default package suppliers (`defaultPackageSuppliers`).
     configured, /// Does not use default suppliers or suppliers configured in DUB's configuration file
     all,        /// Uses only manually specified package suppliers.
+    default_,   /// The value wasn't specified. It is provided in order to know when it is safe to ignore it
 }
 
 /**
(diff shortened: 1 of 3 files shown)


thewilsonator (approved):

source/dub/data/settings.d
```diff
@@ -17,6 +17,7 @@ public enum SkipPackageSuppliers {
     standard,   /// Does not use the default package suppliers (`defaultPackageSuppliers`).
     configured, /// Does not use default suppliers or suppliers configured in DUB's configuration file
     all,        /// Uses only manually specified package suppliers.
+    default_,   /// The value wasn't specified. It is provided in order to know when it is safe to ignore it
```
Geod24, line 20: Instead of modifying the enum, passing a `Nullable` would be a better option, wouldn't it ?

the-horo: When I wrote this I considered both passing a Nullable in the constructor or modifying the enum as possible solutions. My final conclusion was that I would rather not change the constructor semantics and, instead, modify the enum.

I wouldn't mind doing as you say and changing the approach to use a Nullable.

PR #2905, opened 2024-05-03
Speed up configuration resolution

s-ludwig:
This speeds up the `getPackageConfigs` call issued by `getDefaultConfiguration` from 14.8 ms to 3.6 ms for the particular project that I'm benchmarking. There is still a lot more to gain, but this is a good start.

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -643,6 +643,8 @@ class Project {
 	/// Returns a map with the configuration for all packages in the dependency tree.
 	string[string] getPackageConfigs(in BuildPlatform platform, string config, bool allow_non_library = true)
 	const {
+		import std.typecons : Rebindable, rebindable;
+
 		struct Vertex { string pack, config; }
 		struct Edge { size_t from, to; }
 
@@ -650,9 +652,24 @@ class Project {
 		Edge[] edges;
 		string[][string] parents;
 		parents[m_rootPackage.name] = null;
-		foreach (p; getTopologicalPackageList())
-			foreach (d; p.getAllDependencies())
-				parents[d.name.toString()] ~= p.name;
+		// cached package information to avoid continuous re-computation
+		// during the resolution process
+		const(Package)[] package_list;
+		Rebindable!(const(Package))[string] package_map;
+		string[const(Package)] package_names;
+		PackageDependency[][const(Package)] package_dependencies;
+
+		package_list.reserve(m_dependencies.length);
+		foreach (p; getTopologicalPackageList()) {
+			auto pname = p.name;
+			auto pdeps = p.getAllDependencies();
+			package_list ~= p;
+			package_map[pname] = p;
+			package_names[p] = pname;
+			package_dependencies[p] = pdeps;
+			foreach (d; pdeps)
+				parents[d.name.toString()] ~= pname;
+		}
 
 		size_t createConfig(string pack, string config) {
 			foreach (i, v; configs)
@@ -724,16 +741,18 @@ class Project {
 
 		void determineDependencyConfigs(in Package p, string c)
 		{
+			auto pname = package_names[p];
+			auto pdeps = package_dependencies[p];
 
 			// below we call createConfig for the main package if
 			// config.length is not zero.  Carry on for that case,
 			// otherwise we've handle the pair (p, c) already
-			if(haveConfig(p.name, c) && !(config.length && p.name == m_rootPackage.name && config == c))
+			if(haveConfig(pname, c) && !(config.length && pname == m_rootPackage.name && config == c))
 				return;
 
 			string[][string] depconfigs;
-			foreach (d; p.getAllDependencies()) {
-				auto dp = getDependency(d.name.toString(), true);
+			foreach (d; pdeps) {
+				auto dp = package_map.get(d.name.toString(), rebindable(const(Package).init));
 				if (!dp) continue;
 
 				string[] cfgs;
@@ -748,15 +767,15 @@ class Project {
 				// if no valid configuration was found for a dependency, don't include the
 				// current configuration
 				if (!cfgs.length) {
-					logDebug("Skip %s %s (missing configuration for %s)", p.name, c, dp.name);
+					logDebug("Skip %s %s (missing configuration for %s)", pname, c, dp.name);
 					return;
 				}
 				depconfigs[d.name.toString()] = cfgs;
 			}
 
 			// add this configuration to the graph
-			size_t cidx = createConfig(p.name, c);
-			foreach (d; p.getAllDependencies())
+			size_t cidx = createConfig(pname, c);
+			foreach (d; pdeps)
 				foreach (sc; depconfigs.get(d.name.toString(), null))
 					createEdge(cidx, createConfig(d.name.toString(), sc));
 		}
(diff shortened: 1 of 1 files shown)


s-ludwig:
Added two more optimizations to get down to 3.1 ms.

s-ludwig:
Now down to 2.6 ms.

source/dub/project.d
```diff
@@ -643,64 +645,91 @@ class Project {
...
-						return false;
-					} else if (configs[e.to].pack == configs[i].pack) {
-						still_has_dep_to_pack[e.from] = true;
-					}
-					if (e.from == i) return false;
-					return true;
-				}).array;
+			// eliminate all edges that connect to config 'i'
+			size_t eti = 0;
+			foreach (ei, e; edges) {
```
atilaneves, line 707: I don't think this for loop is small enough to justify variable names such as `eti`, `ei`, and `e`.

source/dub/project.d
```diff
@@ -643,16 +643,33 @@ class Project {
...
 		string[][string] parents;
 		parents[m_rootPackage.name] = null;
-		foreach (p; getTopologicalPackageList())
-			foreach (d; p.getAllDependencies())
-				parents[d.name.toString()] ~= p.name;
+		// cached package information to avoid continuous re-computation
+		// during the resolution process
+		const(Package)[] package_list;
+		Rebindable!(const(Package))[string] package_map;
+		string[const(Package)] package_names;
```
atilaneves: How is this faster/better than `package.name`?

source/dub/project.d
```diff
@@ -724,16 +741,18 @@ class Project {
 
 		void determineDependencyConfigs(in Package p, string c)
 		{
+			auto pname = package_names[p];
```
atilaneves: I don't think this is superior to `p.name`.

source/dub/project.d
```diff
@@ -645,44 +645,50 @@ class Project {
 	const {
```
atilaneves, line 647: AA lookups are costly?

atilaneves:
I have a concern about code complexity. I found the original function hard to understand; it took me a couple of hours of summarising it in English before I understood what it did well enough to be able to optimise it. This PR makes even more complicated.

Having said that, I tried this PR on the pathological case that motivated my own PR and the runtime for `getPackageConfigs` went from ~290ms to ~90ms. The "problem" is that it's still death by a thousand cuts. Packages with fewer dependencies don't take that long to begin with so our total runtime went from 15s to 12s.

source/dub/project.d
s-ludwig: The loop contains 8 lines of code. What's your personal threshold?

source/dub/project.d
s-ludwig: As the commit message states, `package.name` needs to allocate for every call. In particular, it needs to construct the full name from the parent package name concatenated with '":"' and the shallow name of the sub package. Also note that the final version of the code uses a linear array instead of an associative array.

source/dub/project.d
s-ludwig: The profiler thinks otherwise.

source/dub/project.d
s-ludwig: Compared to what?

s-ludwig:
> I have a concern about code complexity. I found the original function hard to understand; it took me a couple of hours of summarising it in English before I understood what it did well enough to be able to optimise it. This PR makes even more complicated.

I agree with that. Part of it is based on the unfortunate design choice of making `Package.recipe` and `Package.rawRecipe` mutable references, which is something that would be nice to remedy at some point - and then ultimately reducing the complexity here again. Other parts could be improved by introducing new container types (`Set!T`, `ListSet!T`), but in the end this is an algorithmic complexity issue and not a code style issue. Also, the code can be refactored for more clarity for sure, but that is a different topic, as that would target the high-level structure (which is not changed by this PR).

> Having said that, I tried this PR on the pathological case that motivated my own PR and the runtime for `getPackageConfigs` went from ~290ms to ~90ms. The "problem" is that it's still death by a thousand cuts. Packages with fewer dependencies don't take that long to begin with so our total runtime went from 15s to 12s.

Without being able to reproduce your pathological case or more precise information, I cannot help you there. You identified this particular piece of code to be responsible for a considerable slowdown, so I went head and optimized it. But apart from that, a 20% total improvement is not what I would associate with "death by a thousand cuts".

atilaneves:
> You identified this particular piece of code to be responsible for a considerable slowdown, so I went head and optimized it

Yes, and thanks!

> a 20% total improvement is not what I would associate with "death by a thousand cuts".

Normally, no, but the majority of the runtime is still in `getPackageConfigs`. It gets called twice to make things worse, once to add the test runner configuration and another in the generator when getting the build information (so with different arguments). Our build has 80 dub packages, which means death by 160 cuts. If each "cut" is ~50ms on average it means an added runtime of 8s.

source/dub/project.d
atilaneves: > The loop contains 8 lines of code. What's your personal threshold?

For one letter variables? 1 line.

s-ludwig: For me, a handful of lines is okay, as long variable names are not only making the code easier to understand, but also harder to read - although non-standard variables like the ones here are more critical than the standard ones, if used correctly (i, j, k , x, y, z). However, the parameter `i` to `removeConfig` actually violates my own style guide and should be a full word instead, which would free up the `i` for the `foreach` loop.

But IMO, what would really be the right way to go here would be to generalize the array filter algorithm, so that this becomes `edges = edges.filterInPlace!((e) => ...)` (analogous to `std.algorithm.mutation.remove`). Unfortunately, Phobos doesn't have that AFAIK and I didn't want to mix non-local refactoring (or any substantial refactoring for that matter) into the same PR that does optimization.

s-ludwig:
> Normally, no, but the majority of the runtime is still in getPackageConfigs. It gets called twice to make things worse, once to add the test runner configuration and another in the generator when getting the build information (so with different arguments). Our build has 80 dub packages, which means death by 160 cuts. If each "cut" is ~50ms on average it means an added runtime of 8s.

But no matter how often `getPackageConfigs` gets called, a 2x improvement stays a 2x improvement. The death by a thousand cuts would IMO apply more to a situation where a thousand different places each contribute a thousandth of the runtime.

Just to quickly get the numbers straight here:
- The overall number for you improved by roughly 25% - assuming that `getPackageConfigs` takes up 50% of the total time after the optimization,  that would mean a speedup for `getPackageConfigs` of only 1.5x
- Assuming the 160 calls per total runtime is true, that means with the optimized version a single call takes about 38 ms on average (15x more than for my benchmark project)
- Your pathological test case improved by roughly 3.2x
- In my test case, I'm getting an improvement of around 5.7x
- Applying the 3.2x factor above to estimate the overall runtime taken by `getPackageConfigs` would result in just 29% of the total runtime of 15 s (11% afterwards for the 12 s)

What seems odd here is that, speaking of the pathological test case, you both seem to have a larger payload (assuming the single-thread performance of the machine is not much lower) and a considerably worse speedup at the same time. I would usually expect the opposite to be true and it would be interesting to find out why that happens.

On the other hand, if I should bet, I would say that the reality is that the majority of the runtime is actually spent somewhere else. Taking the estimation from the last bullet point, this would mean that optimizing `getPackageConfigs` alone will never result in less than maybe 11 s. Did you measure absolute time, or is this a statistical measurement from a sampling profiler (which can be heavily skewed)?

atilaneves (approved):

PR #2910, opened 2024-05-31
[stable] Fix regression in module-declarations parser

kinke:
Introduced in https://github.com/dlang/dub/pull/2897, which hasn't landed in any stable release yet. So please merge quickly before v1.38.0 final!


kinke:
Yay, looks like a CI runner image bump broke dub CI (bash doesn't like unescaped backslashes anymore?).

WebFreak001 (approved):

PR #2912, opened 2024-05-31
ci: Bump setup-dlang

the-horo:
Update the setup-dlang action.

It's probably best to do this after the new release so that `release.yml` doesn't break as I've only tested it once.


thewilsonator (approved):

the-horo:
There are still changed being made to setup-dlang@v2 so let's wait until that release has its details ironed out.

PR #2916, opened 2024-06-03
Remove support for OMF, make MsCOFF be x86

Geod24:
Now that OMF support is being removed, dmd-latest is broken in the CI. Since DMD no longer supports it, and we are mostly concerned about supporting the latest 10 releases, we're keeping x86_omf and x86_mscoff available but making them aliases to x86.


ibuclaw:
I think this should target stable?

Geod24:
Changed to stable.

thewilsonator (approved):

PR #2917, opened 2024-06-04
Enhancement of integration test for windows

shoo:
Modified `run-unittest.d` to run integration tests for Windows.
Also, since OMF was removed in the latest release (2.109.0), the OMF build test was removed.

With this PR, it is likely that `run-unittest.sh` will become unnecessary, but it has been left in place. This is to ensure that all tests can still be executed as before. As a result, some tests are currently being run twice. If there are no issues after some time, it would be preferable to remove it.

### DIFF
--- a/test/run-unittest.d
+++ b/test/run-unittest.d
@@ -10,12 +10,13 @@ import common;
 
 int main(string[] args)
 {
-	import std.algorithm : among, endsWith;
+	import std.algorithm : among, endsWith, startsWith, count;
 	import std.file : dirEntries, DirEntry, exists, getcwd, readText, SpanMode;
 	import std.format : format;
 	import std.stdio : File, writeln;
-	import std.path : absolutePath, buildNormalizedPath, baseName, dirName;
-	import std.process : environment, spawnProcess, wait;
+	import std.path : absolutePath, buildPath, baseName, dirName, stripExtension, globMatch;
+	import std.process : environment, spawnProcess, spawnShell, wait, ProcessConfig = Config;
+	import std.string: cmp, splitLines;
 
 	//** if [ -z ${DUB:-} ]; then
 	//**     die $LINENO 'Variable $DUB must be defined to run the tests.'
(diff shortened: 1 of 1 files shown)


Geod24:
Regarding the OMF change, we need to merge stable into master, there is a fix for it.
Regarding general direction, I would really like to be able to use the [test framework](https://github.com/dlang/dub/blob/master/source/dub/test/base.d) more, so if you feel like trying to convert some tests, it will be welcome! Regardless, removing the `.sh` is a good goal.

test/run-unittest.d
```diff
@@ -10,12 +10,13 @@ import common;
...
-	import std.algorithm : among, endsWith;
+	import std.algorithm : among, endsWith, startsWith, count;
 	import std.file : dirEntries, DirEntry, exists, getcwd, readText, SpanMode;
 	import std.format : format;
 	import std.stdio : File, writeln;
-	import std.path : absolutePath, buildNormalizedPath, baseName, dirName;
-	import std.process : environment, spawnProcess, wait;
+	import std.path : absolutePath, buildPath, baseName, dirName, stripExtension, globMatch;
+	import std.process : environment, spawnProcess, spawnShell, wait, ProcessConfig = Config;
+	import std.string: cmp, splitLines;
```
Geod24, line 19: At this point I would just get rid of selective imports :)

test/run-unittest.d
```diff
@@ -46,12 +47,15 @@ int main(string[] args)
 	//** DC_BIN=$(basename "$DC")
 	//** CURR_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
 	//** FRONTEND="${FRONTEND:-}"
-	const dc_bin = baseName(dc);
+	const dc_bin = baseName(dc).stripExtension;
```
Geod24, line 50: Why stripExtension ? IIRC Windows supports it just fine ?

test/run-unittest.d
```diff
@@ -46,12 +47,15 @@ int main(string[] args)
...
 	const curr_dir = __FILE_FULL_PATH__.dirName();
-	const frontend = environment.get("FRONTEND", "");
+	const frontend = environment.get("FRONTEND", __VERSION__.format!"%04d");
```
Geod24, line 52: I assume that is to make it easier to test with the host compiler ?

test/run-unittest.d
```diff
@@ -62,31 +66,131 @@ int main(string[] args)
...
+
+		//#First we build the packages
+		//if [ ! -e $pack/.no_build ] && [ ! -e $pack/.no_build_$DC_BIN ]; then # For sourceLibrary
+		bool build;
+		if (!pack.name.buildPath(".no_build").exists
+			&& !pack.name.buildPath(".no_build_" ~ dc_bin).exists
+			&& !pack.name.buildPath(".no_build_" ~ os).exists)
+		{
+			//build=1
+			build = true;
```
Geod24, line 117:
```suggestion
		bool build =(!pack.name.buildPath(".no_build").exists
			&& !pack.name.buildPath(".no_build_" ~ dc_bin).exists
			&& !pack.name.buildPath(".no_build_" ~ os).exists);
		if (build)
		{
```

Slightly more readable.

Geod24 (approved):
Thanks! Just rebase on master and address the comment and we're G2G.

test/run-unittest.d
shoo: Yes. That is correct. This is useful for manually redoing failed tests. Only DUB variables are required to run.
`dub --single run-unittest.d -- xxxx`

PR #2919, opened 2024-06-04
Clear an unprinted deprecation during dub build

Geod24:
This deprecation is currently not printed by DMD,
but shows up if '-de' is used while compiling.
This solution is a workaround, as a proper fix
would mean not using the function in the first place, but that requires a more in-depth rework.


Geod24:
Ping @thewilsonator

thewilsonator (approved):
Sorry, I missed this one somehow.

PR #2928, opened 2024-06-11
VersionRange: Always output minor version for ~> and ^

### LINKED ISSUE #2830: `dub add` breaks `~>x.0` dependency specification
### System information

- **dub version**: 1.32.0
- **OS Platform and distribution**: ArchLinux `6.6.8-zen1-1-zen`
- **compiler version** v2.106.0

### Bug Description

DUB documentation specifies the two version specifications `~>a.b` and `~>a.b.c`, where the former accepts any minor and patch release since `a.b.0`, and the latter accepts any patch release since `a.b.c`. The documentation does not define any behavior for specification `~>a` like `~>5`.

DUB appears to incorrectly reformat any version specification following the `~>a.b` format, if the minor version is specified to be `0`, to the `~>a` format. Since behavior of this format is undocumented, it is unclear how it is to be understood — would `~>5` also include version `6.0.0` or not?

### How to reproduce?

Add a dependency and set the version specification. For example:

```json
{
	"authors": [
		"artha"
	],
	"dependencies": {
		"raylib-d": "~>5.0"
	},
	"description": "A minimal D application.",
	"license": "proprietary",
	"name": "dubtest"
}
```

Add any other dependency with `dub add`, for example `dub add fluid`. The package will be reformatted, and the previously added dependency, which in this example is `raylib-d`, will have the "dot zero" part removed.

### Expected Behavior

The version specifier should stay as is, with minor version preserved. DUB should reject the major-only specification `~>a` as invalid.

### Logs

```
Using dub registry url 'https://code.dlang.org/'
Note: Failed to determine version of package dubtest at .. Assuming ~master.
  Version selection for dependency raylib-d (raylib-d) of dubtest is missing.
  Missing dependency raylib-d ~>5 of dubtest
             Adding dependency fluid >=0.6.0-alpha <0.7.0-0
```

Geod24:
```
Dub was correct in outputting only the major version, however this can lead to some confusion for users
as pointed out in 2830. Accepting ~>5 but only exposing two states (~>5.0 and ~>5.0.0) will hopefully make things a bit more obvious.
```

@s-ludwig : I was about to close #2830 as WONTFIX but thinking about it more reducing possible state is probably best. Thoughts ?

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -729,7 +729,10 @@ unittest {
 unittest {
 	assert(VersionRange.fromString("~>1.0.4").toString() == "~>1.0.4");
 	assert(VersionRange.fromString("~>1.4").toString() == "~>1.4");
-	assert(VersionRange.fromString("~>2").toString() == "~>2");
+	// https://github.com/dlang/dub/issues/2830
+	assert(VersionRange.fromString("~>2").toString() == "~>2.0");
+	assert(VersionRange.fromString("~>5.0").toString() == "~>5.0");
+
 	assert(VersionRange.fromString("~>1.0.4+1.2.3").toString() == "~>1.0.4");
 	assert(VersionRange.fromString("^0.1.2").toString() == "^0.1.2");
 	assert(VersionRange.fromString("^1.2.3").toString() == "^1.2.3");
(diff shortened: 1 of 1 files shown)


s-ludwig:
I'd say the way you have it in the unit tests makes most sense. Following the general scheme, `~>1` would be the same as `>=1.0.0`, which doesn't really mean anything useful in SemVer semantics, so IMO, we shouldn't advertise it as a supported syntax, and neither output it. Disallowing it completely might be subject to a deprecation phase, though, considering that people might have picked up on it.

Geod24:
@s-ludwig : Anything else to do for now ? We probably want to deprecate it later but I don't think it's urgent.

s-ludwig (approved):
Should be good to go for now!

PR #2935, opened 2024-06-18
CI: Upgrade some more actions to their latest versions

Geod24:


Geod24:
The CI is lying, this doesn't work: https://github.com/dlang/dub/actions/runs/9566236746

PR #2941, opened 2024-06-22
Tests: Mock IO in the `Dub` class too

Geod24:
This will allow to write more tests for the Dub class. In particular I am interested in being able to write unittests around the config file, which require `loadConfig` to be callable. We still need to call `thisExePath` at the moment but that can be solved later as it has no visible side effect.


thewilsonator (approved):

Geod24:
This exposed a bug in the underlying Mockfs. Fixed in its own commit, and green at last!

PR #2951, opened 2024-07-12
chore(path): Make a commonly triggered assert more verbose

Geod24:
We see this assert triggered frequently, even sometimes in production, so at least make it easier to track down the data involved.


Geod24:
Alpine fix: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/69170

thewilsonator (approved):

PR #2952, opened 2024-07-21
Add runtests subcommand

shoo:
As with many official tools, projects often require integration testing for CI/CD. In such cases, it is useful to run integration tests in a unified format.
The runtests subcommand I added is similar to `test/run-unittest.sh` or `test/run-unittest.d` so integration tests of dub projects can be run dub itself.
To test dub, run the following command:
```
dub run -- runtests --skip-build
```


rikkimax:
Ideally these two would be merged together by looking for the json/sdl comment at start of file.

```
- `*.d` : The D language source code is executed by rdmd.
- `*.script.d` : Runs the D language source code as a file as a single-file DUB project.
```

shoo:
> Ideally these two would be merged together by looking for the json/sdl comment at start of file.
> 
> ```
> - `*.d` : The D language source code is executed by rdmd.
> - `*.script.d` : Runs the D language source code as a file as a single-file DUB project.
> ```
Although not impossible, only the minimum necessary functions have been implemented in this PR.
This change would be able to be enhanced later without breaking changes.

Here is a list of others that I can think of that would be nice to have but have not yet been implemented:

- `.fail_build` / `.no_build` / `.no_test` / `.no_run` / `.min_frontend` / `.testignore` / `--testdir` These could be better configured, for example integrated into the project settings.
- When running tests, it would be useful to have an environment variable that contains the path to the binary file under test.
- A way to output the test results in CSV format, etc., for easier handling by other tools would be helpful.
- It would be useful to have an option to specify the directory where the coverage results are stored.
- An option to run each test in parallel may be useful to speed up the process.

If you can think of any very important enhancements, glitches, or future enhancements that might require breaking changes, please let me know.

rikkimax:
Support directory based dub projects.

A directive to pick the directory by default on where the tests are instead of ``test`` or provided each time by the CLI argument.

shoo:
>Support directory based dub projects.
If you are talking about test cases, they are already implemented.
The added test `test/runtests/test/test01` is a directory based dub project.

>A directive to pick the directory by default on where the tests are instead of test or provided each time by the CLI argument.
It is difficult to detect with certainty which directory contains test cases.
For example, dub includes three directories (test, examples, scripts) with similar file characteristics, but it is difficult to mechanically identify which directory is for integration testing. Furthermore, if this were to be automatically deduced, it would be confusing to the user, unless it were deduced with fairly simple logic.

rikkimax:
> > Support directory based dub projects.
> 
> If you are talking about test cases, they are already implemented. The added test `test/runtests/test/test01` is a directory based dub project.

No, I'm talking about ``test/runtests/dub.sdl``.

> > A directive to pick the directory by default on where the tests are instead of test or provided each time by the CLI argument.
> 
> It is difficult to detect with certainty which directory contains test cases. For example, dub includes three directories (test, examples, scripts) with similar file characteristics, but it is difficult to mechanically identify which directory is for integration testing. Furthermore, if this were to be automatically deduced, it would be confusing to the user, unless it were deduced with fairly simple logic.

I didn't suggest automatic detection. I suggested a directive in your dub file to specify the directory to use.

shoo:
> No, I'm talking about `test/runtests/dub.sdl` .
More information, please. Are you saying that you want to recognize the directory as being for the dub project without the dub.sdl?

> I didn't suggest automatic detection. I suggested a directive in your dub file to specify the directory to use.
I see. However, I considered this but decided not to include it in this PR. I believe that adding dub project configuration directives with a higher priority than the default but lower than command line arguments can be added in a future update, not this time.

rikkimax:
> > No, I'm talking about `test/runtests/dub.sdl` .
> 
> More information, please. Are you saying that you want to recognize the directory as being for the dub project without the dub.sdl?

Below the tests directory, below that is either files that you're running. I would like it to also support dub projects at that level.

It could be a misunderstanding on my part, but that changelog doesn't imply it.

shoo:
>Below the tests directory, below that is either files that you're running. I would like it to also support dub projects at that level.
>
>It could be a misunderstanding on my part, but that changelog doesn't imply it.

When you run `dub runtests` at the root of a dub project, it will execute directory based dub projects contained within the `test` directory. This behavior should align with your expectations. However, in the case of the `test/runtests` directory, it is excluded from the tests because it contains a `.no_build` file.
It might be confusing, but the test case I created runs `test/runtests.script.d` with `test/runtests` as the current directory. This test executes the test cases in `test/runtests/test/*` by running the dub runtests command.

shoo:
>First thing that comes to mind: We already have a special configuration for unittest, why not do that for integration tests ? Why a special command ?

The idea of having a special configuration selected when running runtests subcommand might be a good one. By having a directive to replace `--testdir`, such as allowing `--config` to switch test directories, we could achieve good synergy.
On the other hand, I still think a special subcommand is necessary. Just as there is a `dub test` subcommand instead of `dub run --config unittest -- --DRT-testmode=test-only` .

PR #2960, opened 2024-08-18
Do dependency resolution after checking valid configs

### LINKED ISSUE #2695: nonsensical unresolvable dependencies message
in my project:

```
   Upgrading project in /home/webfreak/dev/serve-d/
Error Unresolvable dependencies to package automem:
  cachetools 0.4.1 depends on automem >=0.0.0
  concurrency 5.0.3 depends on automem 0.6.4
```

while obviously `automem 0.6.4` would satisfy the printed constraints.

When trying to `dub upgrade` a package with the following recipe contents:

```sdl
dependency "cachetools" version="0.4.1"
dependency "concurrency" version="5.0.3"
```

the error messages get even worse:

```
   Upgrading project in /tmp/a/
Error Unresolvable dependencies to package unit-threaded:
  automem 0.6.4 depends on unit-threaded >=0.0.0
  automem 0.6.4 depends on unit-threaded >=0.0.0
  automem 0.6.4 depends on unit-threaded >=0.0.0
  cachetools 0.4.1 depends on unit-threaded ~>0
  concurrency 5.0.3 depends on unit-threaded >=0.0.0
  concurrency 5.0.3 depends on unit-threaded >=0.0.0
  concurrency 5.0.3 depends on unit-threaded >=0.0.0
  ikod-containers 0.0.22 depends on unit-threaded >=0.0.0
```

grillo-delmal:
This is my attempt to fix #2695 

This PR changes the dependency resolver algorithm so that it marks which resulting config is going to be used after all the incompatible configs have been discarded.

### DIFF
--- a/source/dub/dependencyresolver.d
+++ b/source/dub/dependencyresolver.d
@@ -110,6 +110,14 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		ulong loop_counter = this.loop_limit;
 		constrain(root, context, loop_counter);
 
+		// Get best available results
+		foreach (base; context.configs.keys)
+			foreach (j, ref sc; context.configs[base])
+				if (sc.included){
+					context.result[base] = sc.config;
+					break;
+				}
+
 		// remove any non-default optional dependencies
 		purgeOptionalDependencies(root, context.result);
 
@@ -176,9 +184,6 @@ class DependencyResolver(CONFIGS, CONFIG) {
 		assert(base in context.configs);
 		if (context.isVisited(n.pack)) return;
 		context.setVisited(n.pack);
-		context.result[base] = n.config;
-		foreach (j, ref sc; context.configs[base])
-			sc.included = sc.config == n.config;
 
 		auto dependencies = getChildren(n);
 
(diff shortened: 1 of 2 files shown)


Geod24:
Could you use the new testing framework ? If you need some help setting it up let me know. I can also try to write a test myself but I won't be able to get to it until end of next week.

grillo-delmal:
> Could you use the new testing framework ? If you need some help setting it up let me know. I can also try to write a test myself but I won't be able to get to it until end of next week.

Sure, happy to help ^^ I'm just kinda out of the loop, so IDK what the new testing framework is :sweat_smile: 

Looking in the latest commits I'm assuming that it means to add it as a unit test on the source/test folder?

I made a branch with that workflow here https://github.com/dlang/dub/compare/master...grillo-delmal:dub:unittest_issue_2695 by adding a unittest to the dependencies.d file

Tell me if I got it right or if it needs some changes (like using a new file or something) and we can continue from there ;).

Geod24:
> Looking in the latest commits I'm assuming that it means to add it as a unit test on the source/dub/test folder?

Correct.

> I made a branch with that workflow here https://github.com/dlang/dub/compare/master...grillo-delmal:dub:unittest_issue_2695 by adding a couple of unittest to the dependencies.d file

That looks good! In both cases, you might want to enforce that the right dependency is selected (`0.0.0` over `0.0.1` IIUC?).

> Tell me if I got it right or if it needs some changes (like using a new file or something) and we can continue from there ;).

I try to group the tests by category but it's a bit arbitrary and categories will likely evolve with time. Anything related to dependency resolution was added to `dependencies.d` so far so that's the exact place I would have put it.

grillo-delmal:
Ok, updated the tests and rebased the PR over it ^^.

Geod24 (approved):
Thanks!

PR #2962, opened 2024-08-22
Ignore --arch switch if it doesn't affect the build

the-horo:
This change has two consequences, one it can improve caching as specifying `--arch=<default_arch>` will lead to the same build id as not specifying it at all but, more importantly, it improves the usability of external tools that rely on dub generated files.

Take as an example the meson build system. It supports for a project to depend on a dub build library. The internal code calls `dub describe --build=... --compiler=... --arch=...` and gathers the cache artifact file from that. Because of always specifying `--arch` the package will always need to be rebuilt when used by meson. This is not intuitive to someone who doesn't know how the --arch switch is used internally in dub.

This has come up as an issue in meson's CI system. The CI system is setup to build test images to be used in a container, periodically, which include all the needed dependencies and common files that are shared across test runs to improve test times. For dub this implies that dub packages that are needed during the tests are fetched + built during the building of the CI images, not when running tests.

However, the packages were built with `dub build --compiler=... <pkg>` which would log to stdout that it built `<pkg>` debug configuration with `<comp>` on x86_64. During the tests meson would fail to find the built packages because of the `--arch` switch it used internally and it recommended that the package should be built with `--compiler=<comp> --build=debug --arch=x86_64` witch is exactly what dub reported the package was built with. This has lead to frustrating debugging and the install scripts calling dub like `dub run --yes dub-build-deep -- <pkg> --arch=x86_64 --compiler=<comp> --build=debug` because nobody can figure out what flags are actually needed.

### DIFF
--- a/source/dub/compilers/compiler.d
+++ b/source/dub/compilers/compiler.d
@@ -171,20 +171,24 @@ interface Compiler {
 			format("%s failed with exit code %s.", args[0], status));
 	}
 
+	/** Default compiler arguments for performing a probe. They should
+		be the D compiler equivalent of "don't output executables"
+	*/
+	protected string[] defaultProbeArgs() const;
+
 	/** Compiles platform probe file with the specified compiler and parses its output.
 		Params:
 			compiler_binary =	binary to invoke compiler with
-			args			=	arguments for the probe compilation
-			arch_override	=	special handler for x86_mscoff
+			arch_flags      =   compiler specific flags derived from the user's arch override
 	*/
-	protected final BuildPlatform probePlatform(string compiler_binary, string[] args, string arch_override)
+	protected final BuildPlatform probePlatform(string compiler_binary, string[] arch_flags)
 	{
 		import dub.compilers.utils : generatePlatformProbeFile, readPlatformSDLProbe;
 		import std.string : format, strip;
 
-		NativePath fil = generatePlatformProbeFile();
+		immutable fileArg = generatePlatformProbeFile().toNativeString;
 
-		auto result = execute(compiler_binary ~ args ~ fil.toNativeString());
+		auto result = execute(compiler_binary ~ defaultProbeArgs ~ arch_flags ~ fileArg);
 		enforce!CompilerInvocationException(result.status == 0,
 				format("Failed to invoke the compiler %s to determine the build platform: %s",
 				compiler_binary, result.output));
@@ -200,22 +204,20 @@ interface Compiler {
 			build_platform.compilerVersion = ver;
 		}
 
-		// Skip the following check for LDC, emitting a warning if the specified `-arch`
-		// cmdline option does not lead to the same string being found among
-		// `build_platform.architecture`, as it's brittle and doesn't work with triples.
-		if (build_platform.compiler != "ldc") {
-			if (arch_override.length && !build_platform.architecture.canFind(arch_override) &&
-				!(build_platform.compiler == "dmd" && arch_override.among("x86_omf", "x86_mscoff")) // Will be fixed in determinePlatform
-			) {
-				logWarn(`Failed to apply the selected architecture %s. Got %s.`,
-					arch_override, build_platform.architecture);
-			}
-		}
-
 		return build_platform;
 	}
+
 }
 
 private {
 	Compiler[] s_compilers;
+}
+
+/// Adds the given flags to the build settings if desired, otherwise informs the user
+package void maybeAddArchFlags(ref BuildSettings settings, bool keep_arch, string[] arch_flags, string arch_override) {
+	if (keep_arch)
+		settings.addDFlags(arch_flags);
+	else if (arch_override.length) {
+		logDebug("Ignoring arch_override '%s' for better caching because it doesn't affect the build", arch_override);
+	}
 }
--- a/source/dub/compilers/dmd.d
+++ b/source/dub/compilers/dmd.d
@@ -110,34 +110,37 @@ config    /etc/dmd.conf
 	{
 		// Set basic arch flags for the probe - might be revised based on the exact value + compiler version
 		string[] arch_flags;
-		if (arch_override.length)
-			arch_flags = [ arch_override != "x86_64" ? "-m32" : "-m64" ];
-		else
-		{
-			// Don't use Optlink by default on Windows
-			version (Windows) {
-				const is64bit = isWow64();
-				if (!is64bit.isNull)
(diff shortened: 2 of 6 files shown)


the-horo:
Example of meson changes caused by this issue: https://github.com/mesonbuild/meson/pull/13550/files#diff-a0eb8ae9cfbf214701a150d556ecba59c3776cff6c5ba9c1f712a2d6a6616443

source/dub/compilers/compiler.d
```diff
@@ -212,6 +214,23 @@ interface Compiler {
 			}
 		}
 
+		if (arch_flags.length) {
+			auto result_without_arch = execute(compiler_binary ~ args ~ fileArg);
+			with (result_without_arch) {
```
Geod24, line 219: Please don't use `with`. We can just shorten the name here (`res_noarch`?)

source/dub/compilers/compiler.d
```diff
@@ -174,17 +174,19 @@ interface Compiler {
...
 		Params:
 			compiler_binary =	binary to invoke compiler with
-			args			=	arguments for the probe compilation
+			args			=	arguments for the probe compilation, typically the compiler specific "don't output executables"
+			arch_flags      =   compiler specific flags derived from the user's arch override
 			arch_override	=	special handler for x86_mscoff
+			arch_override_does_something = will be set to true if arch_flags modify the build environment compared to not specifying them.
 	*/
-	protected final BuildPlatform probePlatform(string compiler_binary, string[] args, string arch_override)
+	protected final BuildPlatform probePlatform(string compiler_binary, string[] args, string[] arch_flags, string arch_override, out bool arch_override_does_something)
```
Geod24, line 182: I think we should look into simplifying this. Having 3 parameters around arch is a red flag.

Geod24 (commented):
Thanks for doing this. It is indeed a problem if `dub` behave differently when an `--arch` that matches the default is passed as opposed to not provided.  I made a few comments about the approach but the underlying goal is 💯

source/dub/compilers/compiler.d
the-horo: Let me think about this for a little bit, enough time has passed since I first wrote this and I changed my mind on some implementations details. I'll make the changes that I want and see if I can reduce the number of flags here in order not to have a garbage API.

Geod24: Sorry for the delay - I can also try to take a look. At the end of the day, if this solves a problem you have, I would be happy to go ahead with it to avoid another great delay.

the-horo: The only thing that I care for changing is modifying the algorithm to preserve `-arch x86_64-unknown-linux-gnu` like flags, so fully specifying the target, but try to strip flags like `-arch x86_64` since those just map to `-m64`. I'm just a little worried that stripping the triple will lead to some obscure errors on weird setups but stripping `-m32` and whatnot should always be safe.

I'm not sure I'll get it done by the end of the day and I'm not the one affected by this issue so I also don't care that much for getting this merged ASAP.

the-horo: So I went with another approach. To determine if the platform changes I just call `probePlatform` twice in each compiler's `determinePlatform`.

I've also taken the warning about the architecture not applying from `probePlatform` and put it into `dmd.d` since it seems to be mostly relevant for `x86_mscoff` vs `x86_omf`, which gdc doesn't suffer from. This, along side moving the default probe compiler flags (`-o- -v -c`) into a separate function, reduced the number of parameters to `probePlatform` to only two: the compiler and the arch flags.

Like I've said that I wanted now full triples like `x86_64-unknown-linux-gnu` are not stripped.

Geod24: The new code looks better, my only concern is calling probePlatform twice, do you know how much time that would add to the build time ? I assume the platform file is fairly easily parsed / analyzed but I never actually measured how much it takes on my system.

Geod24: Did a quick check on my system, got min/max/med of 7/11/10

the-horo: > The new code looks better, my only concern is calling probePlatform twice, do you know how much time that would add to the build time ? I assume the platform file is fairly easily parsed / analyzed but I never actually measured how much it takes on my system.

Using a stopwatch to time `determinePlatform` on my system gives, roughly:
| compiler | Without --arch  | With --arch |
| -------- | --------------- | ----------- |
| dmd      | 3.5ms           | 6.5ms       |
| gdc      | 9ms             | 17ms        |
| ldc      | 15ms            | 28ms        |

The old code also invoked the compiler twice to determine if the arch changed, it just did that in the `probePlatform` function in order to reuse the probe file. Timing that probe file generation gives me 0.038ms so, since it keeps the code cleaner I think it's better to simply recall `probePlatform`. I don't think we have a way to check if the architecture changes without reinvoking the compiler, maybe there's a way to do it simply by looking at `compiler -v` and guessing the default architecture from that but it sounds very brittle.

source/dub/compilers/compiler.d
```diff
@@ -200,20 +204,17 @@ interface Compiler {
...
-				logWarn(`Failed to apply the selected architecture %s. Got %s.`,
-					arch_override, build_platform.architecture);
-			}
-		}
-
 		return build_platform;
 	}
+
+	/// Adds the given flags to the build settings if desired, otherwise informs the user
+	final void maybeAddArchFlags(bool keep_arch, ref BuildSettings settings, string[] arch_flags, string arch_override) {
```
the-horo, line 211: Maybe this can become `protected static` instead of `public final`?

Geod24: Or move it outside the `class`:
```D
package void maybeAddArchFlags(ref BuildSettings settings, bool keep_arch, string[] arch_flags, string arch_override)
```

Reordering args to make it UFCS-friendly.

Geod24:
@the-horo : This is ready to be merged. LMK if you want to act on https://github.com/dlang/dub/pull/2962#discussion_r1771148885 and maybe rebase on master.

Geod24 (approved):
Thanks!

Geod24:
BTW @the-horo any way I can reach you ? Are you on the Dlang Slack / an email I can use ?

the-horo:
I'm not on, what I assume is, https://dlang.slack.com but I wouldn't mind joining. Email's fine otherwise, it's the same as the one I sign-off my commits with: a.horodniceanu@proton.me. Pings on github also work.

PR #2972, opened 2024-10-22
Throw instead of assertion and better message

MrcSnm:
The assertion crashed code-d and gave no information on what was wrong.

### DIFF
--- a/source/dub/dependency.d
+++ b/source/dub/dependency.d
@@ -33,8 +33,11 @@ public struct PackageName
 		if (auto idx = fn.indexOf(':'))
 			this.separator = idx > 0 ? idx : fn.length;
 		else // We were given `:foo`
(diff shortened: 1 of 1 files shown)


Geod24 (changes requested):
What was the stack trace ? The error is in the calling code.

MrcSnm:
I didn't get any stack trace somehow (I had to run on debugger), but this is a minimal reproducible example:

[duberr.zip](https://github.com/user-attachments/files/17481422/duberr.zip)

Geod24:
It's not handling the following propery:
```json
			"dependencies": {
				":version-a": "*"
			}
```
I think that's a limitation of the parser, will see if I can fix it properly.

Geod24:
Closing in favor of the fix: #3028

PR #2974, opened 2024-10-29
[stable] Fix #2973: [Reg] error when depending on a subpackages that is already resolved

### LINKED ISSUE #2973: Serious regression in stable/v1.39 wrt. 'Non-optional dependency not found in dependency tree!?'
This trivial project fails the dependency resolution if the single dependency is on disk:
```
--- dub.sdl
name "foo"
dependency "arsd-official:cgi" version="~>10.9.0"

--- dub.selections.json
{
    "fileVersion": 1,
    "versions": {
        "arsd-official": "10.9.10"
    }
}
```

dub v1.38.0 works fine. New behavior with v1.39.0-beta.1 (latest stable) and latest master:
```bash
$ rm -rf ~/.dub/packages
$ dub build # fetches arsd-official and works fine
$ dub build --vverbose
Note: Failed to determine version of package foo at .. Assuming ~master.
Collecting dependencies for foo
Lazily loading package arsd-official:10.9.10 from /home/mkinkelin/.dub/packages/arsd-official/10.9.10/arsd-official/
  Found dependency arsd-official:cgi 10.9.10
  Collecting dependencies for arsd-official
core.exception.AssertError@source/dub/project.d(197): Non-optional dependency 'arsd-official:cgi' of 'foo' not found in dependency tree!?.
----------------
source/dub/project.d:197 [0x5af4b63cbc0a]
source/dub/project.d:208 [0x5af4b63cb85b]
source/dub/project.d:890 [0x5af4b63d790f]
source/dub/project.d:909 [0x5af4b63d40be]
source/dub/project.d:647 [0x5af4b63cc3f4]
source/dub/project.d:241 [0x5af4b63cc266]
source/dub/commandline.d:1490 [0x5af4b68236bc]
source/dub/commandline.d:1559 [0x5af4b6823d31]
source/dub/commandline.d:535 [0x5af4b66ec592]
source/app.d:39 [0x5af4b6833ccc]
```

kinke:


source/dub/project.d
```diff
@@ -563,7 +563,8 @@ class Project {
 					(VersionRange range) {
 						// See `dub.recipe.selection : SelectedDependency.fromYAML`
 						assert(range.isExactVersion());
-						return m_packageManager.getPackage(dep.name, vspec.version_);
+						auto tmp = m_packageManager.getPackage(basename, vspec.version_);
+						return resolveSubPackage(tmp, subname, true);
```
kinke, line 567: This fixes the testcase in #2973. Not sure if this is the right fix though; `PackageManager.getPackage()` returning the parent package when given a subpackage name doesn't make a lot of sense IMO.

source/dub/test/dependencies.d
```diff
@@ -98,7 +98,7 @@ version "1.0.0"`, PackageFormat.sdl);
     assert(dub.project.hasAllDependencies(), "project has missing dependencies");
     assert(dub.project.getDependency("b", true), "Missing 'b' dependency");
     assert(dub.project.getDependency("c", true), "Missing 'c' dependency");
-    assert(dub.project.getDependency("c", true), "Missing 'd' dependency");
+    assert(dub.project.getDependency("d", true), "Missing 'd' dependency");
```
kinke, line 101: Just happened to notice this while browsing the tests.

source/dub/project.d
Geod24: Yeah it doesn't make a lot of sense but it's always been this way. It was one of the things I wanted to fix when I introduced `PackageName` but never got around to it.
For dependency resolution we always store references to parent package, not subpackage, because all subpackages should resolve to the same parent. I would need to inspect the code to see if that is really the right fix.

Geod24 (commented):
Thanks! By when do you need this reviewed ?

source/dub/project.d
kinke: Okay fine if it's always been like that.

kinke:
> By when do you need this reviewed ?

I assume you've already reviewed this, and if you haven't, just expand the diff context - all other branches do the subpackage resolution (after getting the base package). :) - This is IMO a major release blocker, so v1.39 final should definitely have this.

source/dub/test/subpackages.d
```diff
@@ -38,3 +38,20 @@ unittest
...
+unittest
+{
+    scope dub = new TestDub((scope Filesystem root) {
+        root.writeFile(TestDub.ProjectPath ~ "dub.json",
+            `{ "name": "a", "dependencies": { "b:a": "~>1.0" } }`);
+        root.writeFile(TestDub.ProjectPath ~ "dub.selections.json",
+            `{ "fileVersion": 1, "versions": { "b": "1.0.0" } }`);
+        root.writePackageFile("b", "1.0.0",
+            `{ "name": "b", "version": "1.0.0", "subPackages": [ { "name": "a" } ] }`);
+    });
```
kinke, line 52: Btw thx again for the test framework, sooo much better than the old tests...

Geod24: I wish we had a similar filesystem abstract in Phobos to make this kind of unittests the norm. It is so much more expressive.
I really need to finish it though, there's an issue with Windows at the moment which prevents mimicking the real world properly.

Geod24:
> I assume you've already reviewed this, and if you haven't, just expand the diff context - all other branches do the subpackage resolution (after getting the base package). :) - This is IMO a major release blocker, so v1.39 final should definitely have this.

Do you know what introduced the regression ? I am fairly sure it used to work ?

kinke:
> Do you know what introduced the regression ? I am fairly sure it used to work ?

Oh, I thought this was all revamped after v1.38, but no, the 'problematic' line is the same as was in v1.38.0, which works for the original test case. Hmm...

kinke:
Hmm, this looks suspicious: https://github.com/dlang/dub/blob/7a50f4c9605433116747ff79d6b3910a7e41f2d8/source/dub/packagemanager.d#L1740-L1747

It returns the package from the `path`, which is always going to be a parent package.

source/dub/packagemanager.d
```diff
@@ -1323,6 +1324,29 @@ symlink_exit:
...
+			return pack;
+
+		// available sub-packages have been appended
+		foreach_reverse (sp; dst_repos) {
+			if (sp.parentPackage is pack && sp.name == nameToResolve.toString())
+				return sp;
+		}
+
+		return null;
+	}
```
kinke, line 1349: I've totally reworked the fix, tackling the `PackageManager` API instead and resolving the subpackage in the lazy-loading `getPackage()` case (that was the regression AFAICT), as well as newly in `loadSCMPackage()`.

With more context now, I see that `resolveSubPackage()` is actually to be avoided, as it results in a full packages scan, killing the lazy-loading advantages. It's still used if the `dub.selections.json` contains a path-based selection, so currently we still get a full scan whenever depending on a subpackage of a path-selected parent package.

Geod24: Right, anything that calls `getPackageIterator` needs to go / be reworked.

Geod24 (approved):

PR #2977, opened 2024-11-05
PackageManager: Support directly loading a sub-package via path-based `getOrLoadPackage()`

kinke:
To avoid a costly full packages scan to resolve the sub-package later.

### DIFF
--- a/source/dub/packagemanager.d
+++ b/source/dub/packagemanager.d
@@ -271,8 +271,7 @@ class PackageManager {
 				foreach (ovr; repo.overrides)
 					if (ovr.package_ == name.toString() && ovr.source.matches(ver)) {
 						Package pack = ovr.target.match!(
-							(NativePath path) => getOrLoadPackage(path, NativePath.init, false,
-									StrictMode.Ignore, name),
+							(NativePath path) => getOrLoadPackage(path, NativePath.init, name),
 							(Version	vers) => getPackage(name, vers, false),
 						);
 						if (pack) return pack;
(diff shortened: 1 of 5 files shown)


source/dub/packagemanager.d
```diff
@@ -271,7 +271,8 @@ class PackageManager {
 				foreach (ovr; repo.overrides)
 					if (ovr.package_ == name.toString() && ovr.source.matches(ver)) {
```
kinke, line 272: I guess the overrides should only apply to base packages as a whole, not on a sub-package basis. I.e., `ovr.package_ == name.main.toString()`.

source/dub/packagemanager.d
```diff
@@ -377,21 +378,41 @@ class PackageManager {
...
 	{
 		path.endsWithSlash = true;
-		foreach (p; this.m_internal.fromPath)
-			if (p.path == path && (!p.parentPackage || (allow_sub_packages && p.parentPackage.path != p.path)))
+		const nameString = name.toString();
+
+		foreach (p; this.m_internal.fromPath) {
+			if (p.path != path) continue;
+			if (!nameString.empty && p.name != nameString) continue;
+			if (!p.parentPackage || (allow_sub_packages && p.parentPackage.path != p.path))
```
kinke, line 398: Hmm wait, this would reject an already loaded subpackage if it's defined in the parent recipe.

source/dub/packagemanager.d
```diff
@@ -377,21 +377,42 @@ class PackageManager {
...
-			allow_sub_packages = Also return a sub package if it resides in the given folder
+			name = Optional (sub-)package name if known in advance. Required if a sub-package is to be returned.
 			mode = Whether to issue errors, warning, or ignore unknown keys in dub.json
 
 		Returns: The packages loaded from the given path
 		Throws: Throws an exception if no package can be loaded
 	*/
 	Package getOrLoadPackage(NativePath path, NativePath recipe_path = NativePath.init,
-		bool allow_sub_packages = false, StrictMode mode = StrictMode.Ignore)
+		PackageName name = PackageName.init, StrictMode mode = StrictMode.Ignore)
```
kinke, line 387: The breaking API change is in a separate 2nd commit, in case we cannot/don't want to break it.

Geod24 (approved):

Geod24:
LGTM. Feel free to rebase + merge. I've been wanting to have a much more in-depth change to the PackageManager since forever (storing packages hierarchically in memory) but never finished it. I think that's a pretty good improvement, sorry for forgetting to review it for so long.

PR #2982, opened 2024-12-30
Fix Github CI following runner image changes

Geod24:
This is only needed for Dub's testsuite.


rikkimax:
DMD was upgraded to OSX-13 ages ago, except for nightlies. I upgraded that to 13 last month.

The rest does indeed look specific to dub.

Geod24:
It was `diff` being different. I hate those shell scripts with a burning passion.

PR #2983, opened 2024-12-31
fix: Do not install pkg-config via brew as it's already installed

Geod24:
Installing pkg-config led to the following warning for each macOS pipeline:
pkgconf 2.3.0_1 is already installed and up-to-date. To reinstall 2.3.0_1, run: brew reinstall pkgconf


thewilsonator (approved):

Geod24:
Reduced scope to just clearing the warning as the openssl thing is still required (although I need to look into why, I'm sure we can clear it up).

PR #2987, opened 2025-01-03
fix: Rename source/dub/test/other.d to plural to match module declaration

Geod24:


Geod24:
Merging quickly as this is preventing me from testing on Windows.

PR #2989, opened 2025-01-03
fix: Ensure nameless FSEntry cannot be constructed

Geod24:
We currently do not expect to have nameless FSEntry save for the root, so add an assertiong that would catch such a case. If a use case arise, this can always be removed.


Geod24:
Depends on #2990

thewilsonator (approved):

Issue #2664, opened 2023-06-20
Cross-compiling should use host platform for pre/post commands


jacob-carlborg:
### System information
```
 $ dub --version
DUB version 1.31.1, built on Mar 12 2023
$ ldc2 --version
LDC - the LLVM D compiler (1.32.0):
  based on DMD v2.102.2 and LLVM 15.0.7
  built with LDC - the LLVM D compiler (1.32.0)
  Default target: arm64-apple-darwin22.5.0
  Host CPU: apple-m1
$ uname -a
Darwin  22.5.0 Darwin Kernel Version 22.5.0: Mon Apr 24 20:53:44 PDT 2023; root:xnu-8796.121.2~5/RELEASE_ARM64_T8103 arm64
```

### Bug Description

When cross-compiling, I think the pre and post commands, i.e. `preGenerateCommands`, should use the host platform instead of the target platform.

### How to reproduce?

```d
$ cat main.d 
/+ dub.sdl:
    name "foo"
    preGenerateCommands "echo aarch64" platform="aarch64"
    preGenerateCommands "echo x86_64" platform="x86_64"
+/

import std;

void main()
{
    writeln("foo");
}
```

```
$ dub --arch x86_64 --single main.d                                                                                                   09:47:53
     Pre-gen Running commands for foo
x86_64
    Starting Performing "debug" build using /Users/jacobcarlborg/.dvm/compilers/ldc-1.32.0/bin/ldc2 for x86_64.
    Building foo ~master: building configuration [application]
     Linking foo
     Running foo
foo
```

### Expected Behavior

The output for the Pre-gen Running commands prints `x86_64`. I think it makes more sense to print `aarch64`, since the commands will be running on the host machine. One can use the environment variables `$DUB_ARCH ` and `$DUB_PLATFORM ` to get the target platform inside the commands.

I guess this is a breaking change and I'm sure someone will rely on the current behavior. If that's not acceptable perhaps a new syntax is required for this.

rikkimax:
The current behavior is pretty consistent and makes the most sense for the target.

However, having a constraint to support the host sounds like a good idea for those that need it.

WebFreak001:
I don't think this makes sense with what we have at the moment and would break packages depending on this / calling commands based on the target.

You should instead check for the platform in the command, e.g. in a shell file or when running D files just checking for `version (X86_64)`, etc.

[...]

PR #3004, opened 2025-03-08
Improve dub interop with webassembly

LunaTheFoxgirl:
Adds missing wasm platform and arch specifiers for wasm32, wasm64 and emscripten.


kassane:
Wouldn't it be enough to use ldc2 triple target for now?

With this new target being added, would it affect `pre/pos` commands build?

I even commented on the need to run commands through dub for the wasm32/64 or emscripten target.
- https://github.com/dlang/dub/issues/2664#issuecomment-2661096178

LunaTheFoxgirl:
> Wouldn't it be enough to use ldc2 triple target for now?
> 
> With this new target being added, would it affect `pre/pos` commands build?
> 
> I even commented on the need to run commands through dub for the wasm32/64 or emscripten target.
> 
>     * [Cross-compiling should use host platform for pre/post commands #2664 (comment)](https://github.com/dlang/dub/issues/2664#issuecomment-2661096178)

It'll only affect platform specifications, making them actually work. Since before you couldn't make a configuration exclusive to eg. wasm32 or wasm64, or exclusive to emscripten.

LunaTheFoxgirl:
Okay, this PR is growing a bit in size, wasm32 and wasm64 only works with relocatable objects, so added a case to ensure this when compiling with the `object` target type.

LunaTheFoxgirl:
> Wouldn't it be enough to use ldc2 triple target for now?
> 
> With this new target being added, would it affect `pre/pos` commands build?
> 
> I even commented on the need to run commands through dub for the wasm32/64 or emscripten target.
> 
>     * [Cross-compiling should use host platform for pre/post commands #2664 (comment)](https://github.com/dlang/dub/issues/2664#issuecomment-2661096178)

So despite the extra updates; this does *not* add any new arch types, all it does is handle the result from the triple better by conforming to wasm requirements (such as objects being relocatable, wasm32 and wasm64 having slightly different linker flags, etc.)

source/dub/platform.d
```diff
@@ -102,6 +103,10 @@ enum string archCheck = q{
 	version(LoongArch64) ret ~= "loongarch64";
 	version(LoongArch_SoftFloat) ret ~= "loongarch_softfloat";
 	version(LoongArch_HardFloat) ret ~= "loongarch_hardfloat";
+	version(WebAssembly) {
+		version(D_LP64) ret ~= "wasm64";
+		else ret ~= "wasm32";
```
WebFreak001, line 108: making everything non-64-bit be wasm32 doesn't seem right, should this maybe instead be:
```
version (WebAssembly) ret ~= "webassembly";
version (WebAssembly) version (D_LP64) ret ~= "wasm64";
```

so we are also closer to the D Version identifiers and make it easier for us to add new wasm variants in the future?

LunaTheFoxgirl: There is only wasm32 and wasm64 in terms of architecture; and they are sufficiently different that they need to be seperated. (Otherwise the linker will be upset)

source/dub/compilers/ldc.d
```diff
@@ -259,6 +260,14 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
 				break;
 			case TargetType.object:
 				settings.addDFlags("-c");
+
+				// When using wasm-ld on output objects, we need to explicitly
+				// not strip dead symbols, otherwise we'll get a linker error.
+				// as wasm-ld only works on relocatable objects.
+				if (p.canFind("wasm")) {
+					settings.addDFlags("--disable-linker-strip-dead");
+					settings.addLFlags("-r");
```
WebFreak001, line 269: this doesn't quite seem like the right place to add the dflags per platform, since there is also no other such check here, but I haven't been able to check the code in other compiler wrappers here yet, perhaps consider just disregarding this comment

source/dub/platform.d
WebFreak001: oh I see, also read some more here and I think this should be fine

source/dub/compilers/ldc.d
LunaTheFoxgirl: This requirement only applies to the object target type. Couldn't find anywhere better to put it given that .o files have to be reloctable and can't be section gc'd. If these flags aren't there for the object type then you will be unable to use wasm-ld and/or emscripten to link your module with eg. Emscripten ports or other external libraries.

WebFreak001 (approved):

WebFreak001:
I think we should probably add some tests for webassembly as well since it's quite easy to test, would you like to add that in this PR?

Perhaps @Geod24 also has things to say here otherwise LGTM

LunaTheFoxgirl:
Any updates on getting this merged? @WebFreak001 @Geod24

source/dub/compilers/ldc.d
```diff
@@ -287,10 +296,17 @@ config    /etc/ldc2.conf (x86_64-pc-linux-gnu)
...
 		args ~= settings.sourceFiles;
-		if (platform.platform.canFind("linux"))
-			args ~= "-L--no-as-needed"; // avoids linker errors due to libraries being specified in the wrong order
+
+		// Avoids linker errors due to libraries being specified in the wrong order.
+		// However, the wasm-ld linker does not have --no-as-needed and emscripten is
+		// implicitly treated as a "linux" platform.
+		if (p.canFind("linux") && !p.canFind("emscripten"))
+			args ~= "-L--no-as-needed"; 
+
```
WebFreak001, line 309:
```suggestion
			args ~= "-L--no-as-needed";

```

just CI complaining about this is remaining

Geod24 (approved):

Geod24:
@LunaTheFoxgirl Sorry for the delay. Can you rebase on master and we'll get this in ?

LunaTheFoxgirl:
Sorry for taking a while, I'll get the rebase done later today

Geod24:
Sure. Ping me when done as I don't get notifications on push.

PR #3008, opened 2025-04-14
chore(configy): Upgrade Configy to the latest HEAD

Geod24:
This version is not yet released but brings in a lot of changes, some of which are beneficial to Dub. Non-exhaustive list of changes:

- The YAML dependency has been abstracted behind three interfaces, opening up the possibility to use a JSON parser to parse dub.json, and to write an SDL parser;
- Line and column numbers for errors in `dub.json` were off-by-one, this has been corrected;
- Converters have been removed: this slightly ugligy the code for ToolchainRequirements as the struct now needs to be wrapped;
- `fromYAML` has been renamed to `fromConfig` and the template argument to `ConfigParser` has been removed;


thewilsonator (approved):

Geod24:
Well it actually caught a bug so that's neat.

PR #3015, opened 2025-04-16
draft: Remove normalization wrapper

Geod24:
As quite a lot has changed since the original implementation.


thewilsonator (approved):

Geod24:
I probably need to fix this upstream.

PR #3017, opened 2025-04-18
fix(convert): Validate the package format passed to 'dub convert'

Geod24:
```
Currently calling 'dub convert --format=nope' will trigger an assertion failure.
```

Also include a small future-proof refactoring for the test framework.


thewilsonator (approved):
Do we care enough about this to target stable?

Geod24:
> Do we care enough about this to target stable?

No it's just something I caught while doing some refactoring, don't think we even have an issue about it.

PR #3023, opened 2025-04-28
feat: Implement support for Github-based index, bypassing the registry

Geod24:
```
This implement a package 'index' similar to that found for Homebrew, Nix, Cargo, etc... It allows us to remove a SPOF in our critical infrastructure, as a Github outage would always cause a registry being unusable anyway.

There are multiple steps to having a useful index:
- For transition purpose, we add a hidden command to Dub that export an `index.yaml`;
- In the future, users should register their packages by adding an entry to `index.yaml`, the index definition file of the registry. This is used as the source of all packages;
- `dub` now has a hidden `index-build` command to allow it to build the index based on an index definition file (`index.yaml`). Using this, it queries the various APIs to generate JSON index files that are stored under a pre-defined hierarchy.
- Finally, a `PackageSupplier` is added to make use of this new feature;

In the future, the registration process needs to be moved from the registry to Github to make this migration complete. This *can* be done by exposing a user-friendly interface on `code.dlang.org`, if making an MR to the index is deemed too complicated.
```

This is still a WIP, albeit quite complete now. Things that still need to be done:
1) ~Description is not handled properly (needs to be extracted from the recipe file);~
2) Consider ways to limit / reduce impact on user's disk over a long period (currently uses 32 Mb of data);
3) We need to have the index in production for a while before enabling it by default for users;
4) ~Need to switch configy to a real JSON backend as the YAML one doesn't handle strings well.~
5) Consider scenario where the workspace is empty (e.g. in CI), do we always download a full cache ?
6) Fetching packages from GitLab and Bitbucket is not yet implemented;
7) Support for non-global instances of GitLab and Github could be trivially implemented;

FYI @s-ludwig

Issue #2575, opened 2023-01-31
`dub add` should not remove existing formatting / documentation comments in .sdl file


AndrejMitrovic:
Maybe this is in limbo since dub might be moving away from SDL, but currently calling `dub add <dependency>` will overwrite the existing package file while destroying any existing comments and formatting.

It would be nice if it only appended to the file.

PR #3027, opened 2025-05-14
feat(recipe): Add support YAML recipe

Geod24:
This makes dub recognize `dub.yaml` as a configuration file format. The code is very similar to the JSON one, as Dub has been using the YAML parser for multiple years. It allows us to use a configuration format that is in widespread use and is more human-oriented than JSON.


Geod24:
Note that I did mention it at a DLF meeting a while back.

s-ludwig:
I remember calling you out on this possible plan when you first integrated Configy. Now, the main argument that has been made in the big community outcry back when the JSON/SDL topic came up - namely fragmenting the package format landscape and adding maintenance burden - is still standing. What has lead people to change their mind on this? Back then there were very strong feelings associated with this.

Now, admittedly, I don't quite agree with the gravity of the fragmentation argument. It's more so that this whole discussion, which has been brought up time and time again, has cost me a *lot* of motivation and, apart from the objective aspects, it still bothers me a bit to now see someones favorite color of the shed just being sneaked in.

But regarding YAML itself, I really think the language design is flawed in many ways and not a *technically* good choice. The main argument is that it's frequently used for CI configuration files and in some other more or less related areas. But in terms of usability and simplicity, SDL is just superior for this use case.

The fact that something is technically easy to implement alone is a very weak argument when it comes to designing a system - no matter whether Confiy just happens to be able to parse YAML or DMD conveniently has C bitfield semantics already implemented. 

Anyway, apart from all of this, this is a change with a strong long-term impact that IMO warrants a proper community discussion. And if we go ahead with this, there is also quite a bit of documentation to be amended (and maintained, in case of new recipe features).

Geod24:
> I remember calling you out on this possible plan when you first integrated Configy. Now, the main argument that has been made in the big community outcry back when the JSON/SDL topic came up - namely fragmenting the package format landscape and adding maintenance burden - is still standing. What has lead people to change their mind on this? Back then there were very strong feelings associated with this.

Interestingly, things went the other way around for Configy. I ended up abstracting away the YAML dependency, and you can now use it to parse SDL, JSON, TOML, as long as you have a backend for it. I actually tried to remove the YAML dependency in Dub as a result (see the first commit of #3023 for the JSON backend), but I was not getting column information (only lines) so it was a bit of a regression.

For further reference, Configy was introduced in #2280 . My stance at the time was *against* adding YAML support: https://github.com/dlang/dub/pull/2280#issuecomment-1161558740
I think most of our discussion about YAML happened a bit later, in #2546 . I went through the thread on the forum again, where I request community feedback, and there was a lot of discussions, but little conclusion. The one thing I want to highlight is [Walter's post](https://forum.dlang.org/post/tu00ij$nkd$1@digitalmars.com):

> I don't know much technically about the various formats, but on a non-technical basis:
>
> 1. We win if we go with a well-known, popular format. We are not in the business of promoting an obscure format or inventing our own. We picked github and git long ago simply because it was the most popular, and that decision paid off handsomely for us. As you wrote, it minimizes user friction.
>
> 2. We need a quality import/export mechanism for it written in D with an easy license (preferably Boost).
>
> I'm on board with whatever you choose.

Additionally, @atilaneves ' post was [suggesting we migrate everyone away from one of the format](https://forum.dlang.org/post/bkznrfstadjuikghtkgw@forum.dlang.org).

The argument I remember at the time of the SDL discussion was that, had SDL not been made the default immediately, things would have gone over a lot smoother. It also didn't help that the discussion was started by Andrei. As you can see from the diff, YAML will not be the default, and it will only take precedence of `package.json`, because no one should be using `package.json` nowadays.

> Now, admittedly, I don't quite agree with the gravity of the fragmentation argument. It's more so that this whole discussion, which has been brought up time and time again, has cost me a lot of motivation and, apart from the objective aspects, it still bothers me a bit to now see someones favorite color of the shed just being sneaked in.

Given the amount of discussions we had on this, the feedback from the two BDFL, and the time that has passed, I hope that PR doesn't count as sneaking it in!

> But regarding YAML itself, I really think the language design is flawed in many ways and not a technically good choice. The main argument is that it's frequently used for CI configuration files and in some other more or less related areas. But in terms of usability and simplicity, SDL is just superior for this use case.

I'm not going to argue here. But when I open `dub.sdl` in VSCode, I also don't get syntax highlighting (which include highlighting syntax errors). Same thing when I open it in Github. We don't have the capacity / people to build the tooling required to make everyone's life easier, and using YAML is a way to sidestep that problem. I'd rather spend time making DMD-as-a-library a reality than implementing SDL support for Github / GitLab / VSCode.

> The fact that something is technically easy to implement alone is a very weak argument when it comes to designing a system - no matter whether Confiy just happens to be able to parse YAML or DMD conveniently has C bitfield semantics already implemented.

I strongly agree with this, and it became obvious to me, as soon as saw your message, I should have expanded on the rationale. Ease of implementation is not the reason we should do it. At most, ease of implementation helps build confidence that the feature works, and simplify code review, but it should never be a justification for the feature itself.
The reason we should add support for YAML is because, at the moment, we have the choice between two formats:
1) JSON, which I think we all agree is terrible for human-written configuration;
2) SDL, which, while superior, has not gained traction, lacks support across the board;

YAML has its flaws, but is not a *bad* language, and has definitely gained a massive amount of traction. As discussed, everyone on Github is exposed to YAML. It's hard to expose how important the lack of tooling is, because every single example can be discussed, cherry picked, but it's really a death by a thousand cut. No out-of-the box support for editors / online code viewer, few parsers / emitter available (especially in other languages), little references, very poor support for LLM generation as the language is just now that common, etc...

> Anyway, apart from all of this, this is a change with a strong long-term impact that IMO warrants a proper community discussion. And if we go ahead with this, there is also quite a bit of documentation to be amended (and maintained, in case of new recipe features).

I'm okay to revive the discussion, but I don't think the forum is particularly productive there. I know Walter and Atila are on board, or at least don't oppose it, and last time I mentioned it at the DLF meeting, I don't recall any objection.

s-ludwig:
I don't want to get into the SDL/JSON discussion again, I'm tired of it, as the same arguments are repeated over and over again, everyone focusing on their own specific favorite point. As far as I see it, the real mistake happened earlier and was keeping the JSON support in the first release version of DUB, the transition plan to SDL had already been there for a while at that point and JSON was never really a good format for the purpose, it just happened to be convenient due to the existing implementation(s).

> My stance at the time was against adding YAML support

I remember you pushing for YAML in multiple places and occasions, but I don't think it really matters in the end - I just think the now technically convenient possibility should receive the same scrutiny as the earlier package format topic, because a fundamental change like this has a strong long-term effect and we should carefully weigh the costs and benefits.

> I hope that PR doesn't count as sneaking it in!

Yes, I should have put that in quotation marks. Back when the SDL/JSON discussions happened there simply was a lot of community activity around this, which is not really happening in this case (maybe because the forums were just more active in general, maybe because the topic just isn't as interesting anymore). It would have been unthinkable to just open a PR back then to deprecate JSON support or to add another format without causing huge backlash.

> But when I open dub.sdl in VSCode, I also don't get syntax highlighting (which include highlighting syntax errors). Same thing when I open it in Github.

I actually implemented a syntax highlighting package for Sublime Text 10 years ago and I don't think it took me more than an evening and it should be easily adaptable to VSCode: https://github.com/s-ludwig/sublime-sdlang

GitHub appears to use third-party syntax highlighting modules, so that might then also be solvable with a simple support ticket: https://github.com/github-linguist/linguist/blob/main/vendor/README.md

> SDL, which, while superior, has not gained traction, lacks support across the board;

Despite JSON being the default when creating new packages, SDL makes up about one third of the packages on code.dlang.org. My question would be though, why is anything other than syntax highlighting support even relevant for our use case more in-depth SDL reference, parsers in more foreign languages)?

Just for fun, querying Gemma3 (12b) to write a package recipe with some platform specific parts, it did hallucinate in the dependency and platform specific parts, but it nailed the SDL syntax. JSON was very similar in that regard. This might differ for larger models, of course.

> I'm okay to revive the discussion, but I don't think the forum is particularly productive there.

Definitely, the forum is awful for anything that offers the opportunity for bike-shedding. However, I think the discussion should follow a similar structure and should not merely rely on a more ephemeral medium like a video conference or IRC/Slack.

The main points of concern that I see are long term vision (e.g. keep all three, narrow down on one/two?) and playing through different scenarios with an extended package format syntax (e.g. [DEP3](https://github.com/dlang/dub/wiki/DEP3)/[DEP4](https://github.com/dlang/dub/wiki/DEP4)). Also, do you have a complex example of how a complex YAML recipe would look like with this PR?

s-ludwig:
@Herringway: SDL support has been there from the beginning (before DUB was first released as such), the discussion broke out when SDL was to be made the default format. Apart from that, as already said, the plan was to make it the only format that is human authored, not an additional one. Lastly, it's a fallacy to try to justify a supposed mistake with an analogous earlier supposed mistake.

Herringway:
> @Herringway: SDL support has been there from the beginning (before DUB was first released as such), the discussion broke out when SDL was to be made the default format. Apart from that, as already said, the plan was to make it the only format that is human authored, not an additional one. Lastly, it's a fallacy to try to justify a supposed mistake with an analogous earlier supposed mistake.

My earliest packages (https://github.com/Herringway/natcmp, for example) predate the initial work to support `dub.sdl` packages in 4393dad533e2ad8faa2924677b81c86524a10db3. I was around for this.

Precedence is not a fallacy. It isn't necessarily proof, either, but I am seeing no reason to disagree with it at this time.

s-ludwig:
> My earliest packages (https://github.com/Herringway/natcmp, for example) predate the initial work to support dub.sdl packages in https://github.com/dlang/dub/commit/4393dad533e2ad8faa2924677b81c86524a10db3. I was around for this.

This was way before it became the official package manager and its first official release in 2016. Anyway, the plan at that time was to *replace* JSON, being a permanent alternative is what came out of that infamous discussion.

> Precedence is not a fallacy. It isn't necessarily proof, either, but I am seeing no reason to disagree with it at this time.

So if you shoot yourself in your left foot, it's okay to also shoot in the right one, because of precedence?

Herringway:
> > My earliest packages (https://github.com/Herringway/natcmp, for example) predate the initial work to support dub.sdl packages in [4393dad](https://github.com/dlang/dub/commit/4393dad533e2ad8faa2924677b81c86524a10db3). I was around for this.
> 
> This was way before it became the official package manager and its first official release in 2016. Anyway, the plan at that time was to _replace_ JSON, being a permanent alternative is what came out of that infamous discussion.
> 
> > Precedence is not a fallacy. It isn't necessarily proof, either, but I am seeing no reason to disagree with it at this time.
> 
> So if you shoot yourself in your left foot, it's okay to also shoot in the right one, because of precedence?

False equivalence is a fallacy. Shooting yourself in the foot has obvious negative consequences. Supporting a new format does not.

s-ludwig:
They may not be as obvious, but there are all kinds of potentially negative impacts that need to be taken into account (maintenance and documentation costs, system complexity and implementation size, user confusion and communication overhead, possibly making it more difficult to make format changes or adding constraints on how format extensions can be represented are a few that come to mind). The fact that removing an existing format has a very high cost, means that those costs and the benefits should be weighted very carefully against each other.

Herringway:
> They may not be as obvious, but there are all kinds of potentially negative impacts that need to be taken into account (maintenance and documentation costs, system complexity and implementation size, user confusion and communication overhead, possibly making it more difficult to make format changes or adding constraints on how format extensions can be represented are a few that come to mind). The fact that removing an existing format has a very high cost, means that those costs and the benefits should be weighted very carefully against each other.

It's been over ten years since SDL support was added. We have ten years of data to measure these "potential" impacts with.

Geod24:
> I actually implemented a syntax highlighting package for Sublime Text 10 years ago and I don't think it took me more than an evening and it should be easily adaptable to VSCode: https://github.com/s-ludwig/sublime-sdlang
>
> GitHub appears to use third-party syntax highlighting modules, so that might then also be solvable with a simple support ticket: https://github.com/github-linguist/linguist/blob/main/vendor/README.md

Note that the statement after the one you quote answers this: Why don't we have out of the box support for SDL in VSCode and Github ? Because:
> We don't have the capacity / people to build the tooling required to make everyone's life easier, and using YAML is a way to sidestep that problem. I'd rather spend time making DMD-as-a-library a reality than implementing SDL support for Github / GitLab / VSCode.

I know about linguist, and when things broke for syntax highlighting on Github the issue was handled quickly. But it's been 10 years and Github still doesn't support SDL syntax highlighting despite the ease to do so, as you pointed out. Perhaps the problem is not with the difficulty of the task, then.

> Despite JSON being the default when creating new packages, SDL makes up about one third of the packages on code.dlang.org. My question would be though, why is anything other than syntax highlighting support even relevant for our use case more in-depth SDL reference, parsers in more foreign languages)?

I'm not sure I follow the question. As pointed out earlier, this is death by a thousands (paper) cuts: unfamiliarity, lack of tooling, lack of libraries.

> Definitely, the forum is awful for anything that offers the opportunity for bike-shedding. However, I think the discussion should follow a similar structure and should not merely rely on a more ephemeral medium like a video conference or IRC/Slack.

I think, before said discussion happens, we should define the criterias for rejecting or approving it. We've always followed a BDFL approach, so I think if both Walter and Atila are in agreement, we should go ahead with it, unless you have another suggestion.
The discussion can happen in the next foundation / industry meeting if you'd like and are available. If there are requirements to this happening, such as deprecating JSON, they can also be voiced and agreed upon there. What do you think ?

> The main points of concern that I see are long term vision (e.g. keep all three, narrow down on one/two?) and playing through different scenarios with an extended package format syntax (e.g. [DEP3](https://github.com/dlang/dub/wiki/DEP3)/[DEP4](https://github.com/dlang/dub/wiki/DEP4)). Also, do you have a complex example of how a complex YAML recipe would look like with this PR?

I think we should add support for YAML and gradually deprecate JSON and SDL going forward, first by removing it from `dub init`, then by warning on building projects using them, then finally by erroring on projects that use it. When it comes to packages being used from the registry, I would like to keep supporting both package recipe for an extended period of time (we can discuss in terms of releases, or years), as that would be the most disruptive part.

When it comes to extend the package format, that was not yet in scope. I would not like to bundle too many changes at once, but perhaps it's an area of discussion as well.

Herringway:
> I suspect getting SDL into Github/Gitlab is more of a "Time and will power" problem than anything else. I very much doubt that they've specifically rejected it. IMO, GitHub and Gitlab not supporting SDL is more of an argument for doing the work to get it supported, rather than an argument against retiring/replacing it.

You may be waiting a very long time for someone to volunteer for this.

Herringway:
I would also like to point out that since dub has been using a YAML parser to parse dub.json files without JSON validation for some time now, it's quite possible that some misnamed dub.yaml files already exist in the wild.

LightBender:
So what I gather from this thread is that the only compelling reason to make this highly disruptive change is that GitHub/Gitlab support YAML syntax highlighting?

I'm struggling to see why that merits such a disruptive change.

schveiguy:
I vehemently disagree with yaml support (my position has not changed). It's a terrible format, and adding more format support isn't going to fix any of dub's real problems. It will just add to the problems.

Will bring this page up again:

https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

My recommendation has been if we want to support a better format to go with json5. But better still, we should focus on the existing problems in dub before adding more formats. Maybe start with making it so dub can edit its known file formats without destroying the formatting? #2575

schveiguy:
> What is SDL's main use case ? Dub recipe files. Where else is it used prominently ?

SDL is never going away. Anything we add here is adding, SDL will always be there, because nobody is going to go back through all released revisions of all projects and change `dub.sdl` to `dub.yaml`.

We will continue to have the problems of SDL, and add all the problems of YAML.

This goes for any new format we add. We should only add a format if it's worth adding on its own.

If you want a prominently used format, use JSON. There is plenty of tooling for JSON.

> All of those problems are fixed by quoting strings.

So we will have a format where it's easy to put in bizarre corner cases, and our response to users who have those issues is going to be that it's YAML, so you can write bad formats, nothing we can do? This reeks of "you're holding it wrong" scolding.

The beauty of JSON is that it is never ambiguous. It's definitive, and there are no "options" when it comes to how to specify things.

> last time I mentioned it at the DLF meeting, I don't recall any objection.

I missed this before. I have objected every time I'm at the DLF meeting. I shouldn't have to attend them all to bring this up every time.

Geod24:
> The beauty of JSON is that it is never ambiguous. It's definitive, and there are no "options" when it comes to how to specify things.

There are options, one of them you've been recommending (JSON5).
There's also JSONC (https://en.wikipedia.org/wiki/JSON#Supersets), and obviously YAML.

> SDL is never going away. Anything we add here is adding, SDL will always be there, because nobody is going to go back through all released revisions of all projects and change dub.sdl to dub.yaml.
>
> We will continue to have the problems of SDL, and add all the problems of YAML.

We need to continue supporting *reading* the *current* SDL format. That's it. We can get rid of SDL formatting code (support as target in `dub convert`), we don't need to support it in `dub add` (the aforementioned bug), and we don't need to support it going forward (can get rid of documentation, don't have to think about it if we upgrade the recipe format...).

Because YAML is a superset of JSON, we get it for (almost) free. SDL has a cost. Going with JSON5 would be less disturbing than YAML, and the upside is IMO pretty limited.

schveiguy:
> we don't need to support it going forward (can get rid of documentation, don't have to think about it if we upgrade the recipe format...).

So when someone looks at a package, and reads the dub.sdl, what are they going to do? How do they look up how to edit that file, or what things mean?

The documentation is already there, why would we ever get rid of it?

I don't see the point of removing SDL, even if we don't want to encourage it. We literally cannot get rid of it, because of existing packages. And if we have to support it, we should keep documenting the support.

s-ludwig:
> "Building the tooling" is not solely about writing code. It's also integrating it and maintaining it. We're all experienced programmers here and we all experienced how much maintainance actually cost compared to the initial development time.

There can, however, be extreme differences in the ratio here. In the concrete example of the syntax highlighting module, maintenance cost has been more or less zero.

>> I'm asking why you think that lack of libraries is relevant to DUB and I'm asking why you think that "little references" are relevant. I'm asking that in a concrete way, as I cannot see where you see the "thousand cuts" theory playing out in practice - to me this is just an empty phrase without at least some kind of evidence.
>
> D's ecosystem weakest point is tooling, I hope we can agree with this. To me this is a consequence of being spread to thin, something exacerbed by our NIH tendencies. Today if I want to build tooling for D that somehow uses the recipe file, not only do I have to build the tooling, but most likely I have to select among a short set of SDL parsing libraries, or build one myself. But then I also have to make sure I provide good error handling, and god forbid if we ever want to update the SDL format.

If you want to make a tool that interacts with the recipe file, you should use dub as a library and not reinvent the whole package parsing and higher level logic - that just begs for problems as the package format evolves and multiplies maintenance costs. But the more important point is, having a third format that needs to be supported doesn't make for a good argument in terms of facilitating easier writing of tooling. Even if everything but YAML would be dropped 10 years from now, you would still have made it harder to write tooling until then. And just not supporting the existing formats will not improve the tooling landscape, but will just make it appear more broken.

schveiguy:
My objections to YAML aren't personal. It's objectively a bad format. Yes, I understand that it has "won" when it comes to people being forced to use it. However, the plan here shouldn't be to force people to use YAML. This would be a bad move IMO.

The existence of SDL does not discourage people from using D. Removing SDL would definitely alienate existing users. Forcing YAML on people would definitely alienate users. Adding YAML as an option will not gain users (nobody is going to use D now because dub does YAML). The discussion here is literally only about us here, do we want YAML or not? All of us will continue to use D whether YAML is there or not. There is no checkboxes on anyones list that this helps with.

It costs almost nothing to support SDL. Why is SDL such a bad format that we should remove it? Nobody has explained this. The same goes for JSON. Note that neither SDL nor JSON formats have changed. Not including new dub features in JSON or SDL would be mostly spiteful. You would probably have to reject PRs that fix any issues.

I have multiple projects that I inherited (and some that I started) that use SDL as the dub format. I don't use it for new projects, because JSON is easier to use for me. But I certainly don't want to switch to JSON on those or YAML, SDL is fine to use for maintenance.

If YAML gets added, I will be sad, but move on. If SDL or JSON gets deprecated, I will be mad. Deprecating is a bad move here that gains us nothing, and just adds friction to using D.

Geod24:
> If YAML gets added, I will be sad, but move on. If SDL or JSON gets deprecated, I will be mad. Deprecating is a bad move here that gains us nothing, and just adds friction to using D.

Note that the time horizon for deprecation is *long*. Honestly long. The steps would be:
1) Add support for it, and ensure we have good documentation / examples;
2) Wait a X releases because you should never introduce a deprecation at the same time as you introduce the alternative;
3) During (2), make YAML the default;
4) Start to issue deprecation warnings for **projects** using `dub.json` / `dub.sdl`;
5) After Y releases, start to issue deprecation warnings for **packages** using `dub.json` / `dub.sdl`;
6) After Z releases, remove support for `dub.json` / `dub.sdl`;

If X = 4, Y = 10, Z = 10, you're looking at 6 years before we issue an error because your dependency uses `dub.sdl`. We could make it longer, or even not do (6), if the cost of converting old packages will be kept low-ish. Sometimes between (4) and (5), or maybe coupled with (5), we would also stop recognizing `dub.sdl` / `dub.json`. But that would only affect actively developed packages which are likely to transition easily.

Herringway:
> It is clear this is going to happen whether I like it or not.

I was getting the opposite impression. Nearly everything that interests me gets rejected in the end.

finnurthorisson:
> @finnurthorisson The people in this conversation so far (on both sides) have been using D for decades _plural_, I therefore suggest not making such judgements. If you have further comments which don't add constructively to the discussion, please use Discord. Thanks!

I am aware. I am not literally telling people to go use C++. I am making the point that "everyone else does it" is not a reason to do something, otherwise we should be using the most popular systems language instead of D. Read between the lines a little

The people that have been using D for decades _plural_ are already familiar with JSON/SDL. The main argument people have been giving is that YAML is familiar to newbies (or more directly saying "everyone else does it"), but the difficulty of learning SDL is not worth mentioning when compared with the difficulty of learning D and the exact keys/values DUB expects. Most newbies will glance at it and know how it basically works then copy chunks from the docs; it will not break if their editor is configured in a 'problematic' way either

There is the argument for tooling, like generating/parsing DUB recipes, but if you get to choose your format JSON has the best tooling, and if you can't fragmenting the ecosystem makes this worse

There is nothing to be gained from adding YAML and so deprecating SDL/JSON can do nothing but harm

LunaTheFoxgirl:
> > The beauty of JSON is that it is never ambiguous. It's definitive, and there are no "options" when it comes to how to specify things.
> 
> 
> 
> There are options, one of them you've been recommending (JSON5).
> 
> There's also JSONC (https://en.wikipedia.org/wiki/JSON#Supersets), and obviously YAML.
> 
> 
> 
> > SDL is never going away. Anything we add here is adding, SDL will always be there, because nobody is going to go back through all released revisions of all projects and change dub.sdl to dub.yaml.
> 
> >
> 
> > We will continue to have the problems of SDL, and add all the problems of YAML.
> 
> 
> 
> We need to continue supporting *reading* the *current* SDL format. That's it. We can get rid of SDL formatting code (support as target in `dub convert`), we don't need to support it in `dub add` (the aforementioned bug), and we don't need to support it going forward (can get rid of documentation, don't have to think about it if we upgrade the recipe format...).
> 
> 
> 
> Because YAML is a superset of JSON, we get it for (almost) free. SDL has a cost. Going with JSON5 would be less disturbing than YAML, and the upside is IMO pretty limited.

Adding yet another format and letting SDL languish would be a poor choice imho. A large majority of my packages use the SDL format because it's easy to read and write; and it doesn't use indentation for scoping which is a big pain point for yaml.

I have no desire or intention of switching away from SDL; so if that languishes that would be the point I look into switching away from hosting my D code on dub.

LunaTheFoxgirl:
An additional note; yaml has shown no tangible benefits to the Inochi2D Project. I already use it to write workflows for github CI and yaml makes that process very error prone. New contributors are easily onboarded to SDL, wouldn't be able to say the same for yaml given its footguns.

SDL works fine; while I can't decide whether or not it should be a default format: it is the format which is used throughout all of the Inochi2D Project and my business and the format I instruct any contributor working on new components/libraries to use. In my own experience SDL has not been a hindrance for adoption of DLang.

cyrusmsk:
Such PRs should be prepared only after some joint agreement of the majority of community.
Which is not the case (even for core devs)

PR #3033, opened 2025-05-22
Replace SkipPackageSuppliers.default_ usage with Nullable

the-horo:
Instead of having a different enum value for stating that the value is not specified and it should be taken from the configuration files, use a Nullable. This has been suggested in https://github.com/dlang/dub/pull/2903#discussion_r1582262457

The main reason for this change is going through the other places where the configuration files were not honored (e.g. dub lint) (which was the intent of the original PR). It would be a shorter change to modify the current code to use the .default_ enum value when it was appropriate but if I have to change the code I might as well go and implement the less error-prone solution with Nullable.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -35,7 +35,7 @@ import std.process : environment, spawnProcess, wait;
 import std.stdio;
 import std.string;
 import std.traits : EnumMembers;
-import std.typecons : Tuple, tuple;
+import std.typecons : Nullable, nullable, Tuple, tuple;
 
 /** Retrieves a list of all available commands.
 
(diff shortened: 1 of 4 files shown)


Geod24:
The diff looks good, but I can't help but think the real issue here is that argument parsing and configuration file loading are so far appart. If the `Dub` class was just provided with a "proper" configuration struct that whole exercise would not be required, right ?

the-horo:
Maybe, `dub`'s code is very intertwined in a bunch of places making it harder to follow. I don't think this instance is the cause of that. The main issue here was that the constructor had a default value for the `skipRegistry` argument which made the fallback value (the one from the configuration files) never be used.

PR #3035, opened 2025-05-25
refactor: Move IO-performing getPlatformSettings to buildsettings

Geod24:
It is better to have this function closer to buildsettings where it is used, than from buildSettingsTemplate which otherwise has almost no IO dependencies.

### DIFF
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -7,16 +7,22 @@
 */
 module dub.compilers.buildsettings;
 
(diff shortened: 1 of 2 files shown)


Geod24:
> source/mir/optim/least_squares.d(242,6): Error: `@safe` function `mir.optim.least_squares.__unittest_L219_C7` cannot call `@system` function `mir.optim.least_squares.__unittest_L219_C7.optimize!((x, y)
  
  That's the Buildkite failure, so not triggered by this PR.

thewilsonator (approved):

PR #3050, opened 2025-06-20
Fix #595 Error when using an executable subConfiguration.

### LINKED ISSUE #595: Error when using an executable subConfiguration
```json
{
	"name": "foobar",
	"dependencies": {
		"ddox": "~>0.10.8"
	},
	"subConfigurations": {
		"ddox": "application"
	}
}
```

```bash
$ dub build
Error executing command build:
Could not resolve configuration for package foobar
```

veelo:


veelo:
Thank you @Geod24. I am struggling a bit with `test\issue1651-custom-dub-init-type`, which is failing on macOS and ubuntu. I am running Windows locally. That particular test only contains a .zip file, and `run-unittest.d` seems to skip it on Windows. Any idea how I can reproduce the failure locally on Windows?

veelo:
> Though it looks like `run-unittest.d` is skipping `.sh` tests on windows unconditionally, maybe making it run them if `bash` is found in `PATH` is a better idea

Even better would be if there were no shell scripts in the test suite ;-) I prefer D scripts.

the-horo:
```diff
diff --git a/source/dub/package_.d b/source/dub/package_.d
index 18c5d485..dcc9cdc2 100644
--- a/source/dub/package_.d
+++ b/source/dub/package_.d
@@ -714,7 +714,7 @@ class Package {
 		}
 
 		// generate default configurations for root package if none are defined
-		if (m_info.configurations.length == 0 && m_parentPackage is null) {
+		if (m_info.configurations.length == 0) {
 			// check for default app_main
 			string app_main_file;
 			auto pkg_name = m_info.name.length ? m_info.name : "unknown";
diff --git a/source/dub/project.d b/source/dub/project.d
index e64cccad..199249a5 100644
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -749,7 +749,13 @@ class Project {
 				} else {
 					auto subconf = pack.package_.getSubConfiguration(c, packages[dp].package_, platform);
 					if (!subconf.empty) setConfigs(only(subconf));
-					else setConfigs(packages[dp].package_.getPlatformConfigurations(platform));
+					else {
+						setConfigs(packages[dp].package_.getPlatformConfigurations(platform));
+						if (depconfigs[dp].length == 0) {
+							// Try with the executables too
+							setConfigs(packages[dp].package_.getPlatformConfigurations(platform, true));
+						}
+					}
 				}
 
 				// if no valid configuration was found for a dependency, don't include the
```
got the testsuite to pass for me though maybe you could make it a little bit more readable

veelo:
I'm happy, this is ready for review.

thewilsonator (approved):

the-horo:
> @the-horo I have tried making it more readable, but I'm not sure. Is this better?

I'm not entirely sure as I don't even fully understand what that code is supposed to do. Conceptually, it picks the configurations for a package that has been pulled in as a dependency when a configuration override is not present. But is the behavior of this documented anywhere or are we just guessing what the code should do.?

My idea was that, when looking for a dependency, you want to try any of the *library*-ish configurations and fallback to the executable configurations if you couldn't find a match. I don't know if it would also work if you didn't go as far as to use a fallback, but simply iterate the dependency configs in the right order, something like `library_configs ~ executable_configs` and then leave the rest of the code to clean up unused edges in the dependency graph.

One edge case that came to mind would be:
- pkgA depends on pkgDep => should default to pkgDep[library]
- pkgB depends on pkgDep[application]

I think that this should lead to a failure as changing what `pkgA` gets as a result of some other package's config is hard to keep track of. I think that the current code allows this so maybe the solution would be:
> When dub looks for a dependency without a configuration being specified it tries to pick
> 1. The first non-executable-target config
> 2. In the absence of the former, the first executable-target config

But that sounds a little bit restrictive and it would interfere with:
1. pkgA depends on pkgDep => pick pkgDep[library]
2. pkgB depends on pkgDep[fancy-library] => pick pkgDep[fancy-library]
which would result in multiple configurations being pulled in the graph which would fail so, maybe it's better for dub to pick:
> 1. Any non-executable-target config
> 2. In the absence of the former, any executable-target config

I'm curios what @Geod24 thinks about this or if he has any suggestions.

Geod24:
> One edge case that came to mind would be:
> * pkgA depends on pkgDep => should default to pkgDep[library]
> * pkgB depends on pkgDep[application]
>
> I think that this should lead to a failure as changing what pkgA gets as a result of some other package's config is hard to keep track of.

I don't see why you would have a dependency to an application configuration unless you're using it for building / generating code / etc... In this case, the code is not included in the final binary, so the dependency is "local" - I don't see why two packages, depending on the same package, would conflict, as long as at least one of the configuration is application.

I haven't fully reviewed the code yet though.

veelo:
> I don't even fully understand what that code is supposed to do.

You are not alone...

veelo:
> I haven't fully reviewed the code yet though.

Friendly ping.

PR #3053, opened 2025-06-27
Update generated `.gitignore` during `init`

KaruroChori:
I added a filter for the file with pattern `libname.a`, as that is what I get on my platform by default and it was missing.  
For context, debian 13 and latest ldc2 from github releases.


thewilsonator (approved):

Geod24:
Thanks! I partly think that we should just exclude all executables (`*.so`, `.[oa]`, etc...) but this is a good enough fix for now.

PR #3059, opened 2025-07-05
Add --dest option

LunaTheFoxgirl:
This pull-request adds a `--dest=<path>` option for specifying a target directory to stage into.
This value is prefixed onto the targetPath and workingDirectory in the `configurePackages` step.


source/dub/generators/generator.d
```diff
@@ -259,6 +259,20 @@ class ProjectGenerator
 
 		auto roottarget = &targets[rootPackage.name];
 
+		// Handle the destination directory being overwritten.
```
schveiguy, line 262:
```suggestion
		// Handle the destination directory being overridden.
```
Slight confusion I had when first looking at this. I thought you meant that the output directory was going to be overwritten with new files.

schveiguy (approved):
LGTM, just a comment change.

Is there a plan to update the online docs? Not sure where those are kept...

LunaTheFoxgirl:
> LGTM, just a comment change.
> 
> Is there a plan to update the online docs? Not sure where those are kept...

Possibly? I'm not entirely sure either; to me it seems the docs are auto generated; but if not I could look into making a PR into the online docs.

LunaTheFoxgirl:
> apart from: [documentation](https://github.com/dlang/dub-docs/tree/master) (with most probably an entry in [dub-reference](https://github.com/dlang/dub-docs/tree/master/docs/dub-reference), I presume this is a build setting) test case and a [changelog entry](https://github.com/dlang/dub/tree/master/changelog), this looks good.

This is a command line option; as such from what I can tell the documentation should be auto-generated from the man pages.

PR #3060, opened 2025-07-06
Remove $DFLAGS build type (but keep supporting the $DFLAGS variable)

the-horo:
While it is useful and expected that dub would honor the $DFLAGS variable, it should not drop *all* flags that it adds on its own.

The most egregious way this is manifesting is that:
- `dub test` - runs unittests (duh)
- `DFLAGS= dub test` - does not run a single test

This is because dub would treat the presence (even if empty) of $DFLAGS as flags to override the current buildType. This has been changed and now $DFLAGS simply are appended to the flags derived from the buildType. If users actually want to prevent dub from passing any special flags they can do so with `dub test --build=plain`

### DIFF
--- a/changelog/remove-dflags-config.md
+++ b/changelog/remove-dflags-config.md
@@ -1,0 +1,16 @@
+Specifying `$DFLAGS` in the environment no longer overwrites dub-added flags
+
+When the `$DFLAGS` variable was present in the environment (even if
+empty) Dub would stop itself from adding any flags implied by the
+build type. This can lead to serious issues with `dub test` not
+running unittests, like in the below example:
+
+```
+$ DFLAGS= dub test # this does not run any unittests
+```
+
+This has been changes and now, the `$DFLAGS` variable is simply
+appended to the dub generated build flags.
+
+If you rely on the old behavior you can pass `--build=plain` to `dub`,
+in addition to specifying your custom flags in `$DFLAGS`.
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1289,7 +1289,7 @@ abstract class PackageBuildCommand : Command {
 	{
 		args.getopt("dest", &m_destPath, ["Base directory in which output atifacts will be placed"]);
 		args.getopt("b|build", &this.baseSettings.buildType, [
-			"Specifies the type of build to perform. Note that setting the DFLAGS environment variable will override the build type with custom flags.",
+			"Specifies the type of build to perform.",
 			"Possible names:",
 			"  "~builtinBuildTypes.join(", ")~" and custom types"
 		]);
@@ -1376,8 +1376,7 @@ abstract class PackageBuildCommand : Command {
 		}
 
 		if (this.baseSettings.buildType.length == 0) {
-			if (environment.get("DFLAGS") !is null) this.baseSettings.buildType = "$DFLAGS";
-			else this.baseSettings.buildType = default_build_type;
+			this.baseSettings.buildType = default_build_type;
 		}
 
 		if (!m_nodeps) {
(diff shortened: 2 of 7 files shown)


Geod24:
It's something I've been thinking about for a while now. I agree we should do it. I'm currently in vacations abroad so have limited time to review this week, but feel free to ping me next week if you see no action.

s-ludwig:
While I do agree that the usability of the current approach is lacking and doesn't really make sense in the case of `dub test`, I honestly don't think we can just remove it, because it is completely unclear how that affects existing build scripts, also considering that DUB itself sets the `DFLAGS` variable when executing (pre/post) build commands.

Maybe an alternative approach could work here, for example a command line argument (e.g. `dub -dflags="-foo -bar"`)?

the-horo:
>  I honestly don't think we can just remove it, because it is completely unclear how that affects existing build scripts

scripts:
- without `DFLAGS` stay the same
- with `DFLAGS` and `--build=...` stay the same
- with `DFLAGS` and without an explicit `--build=...`
  - `dub test` - was broken, is now fixed
  - `dub build` and `dub run` - now add `-g -debug`

This really don't seem like that big consequence compared to the fact that there is no way to safely pass flags to the compiler when scripting, which is a very common occurrence with `gdc` and `gdmd` that almost always need `-Wno-error` to get around deprecations and, half the time, need `-allinst`. I currently can only:
1) Set `DFLAGS` and modify all `dub test` invocations with `dub test --build=unittest`
2) (only for `gdmd`) write to `/etc/dmd.conf` and set `DFLAGS` there.

> also considering that DUB itself sets the DFLAGS variable when executing (pre/post) build commands.

As in it lets it leak into the environment, but doesn't actually set `$DFLAGS` to be used in build commands. The behavior stays the same in this sense.

Note that this is another issue that should be fixed/specified. Currently `dub` lets `DFLAGS` be inherited when spawning other processes which has 2 weird interactions
1. During compilation, `ldmd2` and `gdmd` use that variable and append it to their CLI, leading to flags being added twice: once by dub when generating the build and once when invoking the compiler
2. Like you said, pre/post build commands are also affected, which is weird because `$DC -run` which is used by a couple of projects would compile with `DFLAGS` in the case of `ldmd2` and `gdmd`

> Maybe an alternative approach could work here, for example a command line argument (e.g. dub -dflags="-foo -bar"

Nothing easy comes to mind but `--dflags` I would read as a switch that overwrites the `DFLAGS` environment variable, but it would still behave the same semantically. So:
```
DFLAGS=-foo dub --dflags=-bar
dub --dflags=-bar
DFLAGS=-bar dub
```
I would imagine do the same thing: compile with `-bar`

the-horo:
> The default mode should always be to treat "deprecations" as non-fatal warnings and "warnings" as errors. Code that doesn't compile without actual "warnings" should specify "allowWarnings" as a [build requirement](https://dub.pm/dub-reference/build_settings/#buildrequirements) instead of relying on DFLAGS, though, but I guess you are in a situation where you are compiling foreign code with a newer compiler?

This is the issue, `gcc` (and, by extension `gdc`) does not have the concept of deprecations. It only knows about warnings. if `-Werror` is passed to `gdc` or `-w` to `gdmd` then any warning, including depractions diagnostics, will cause the build to fail.

I understand that this was the initial intention, to treat warnings as errors but, given the current landscape where:
1. `dmd` and `ldc2` basically never output warnings, only deprecations
2. `gdc` and `gdmd` fail to build by default with deprecations

It sounds reasonable to abandon the original idea and to allow warnings by default, for `gcc` and `gdmd` at least. Given the current dub architecture where `gdmd` is treated as `dmd` it would mean that `dmd` would also be exempt from `-w` and if that's how it's going to go it makes sense to drop `-w` from `ldc2` as well.

All of this is unrelated to this PR though, so maybe the discussion should be continued in a separate issue.

> but I guess you are in a situation where you are compiling foreign code with a newer compiler?

yes, I usually hit it when trying to build somebody else's dub project, either in CI or as a package for Gentoo, and the build almost always prints deprecations, leading to ugly hacks to coerce dub into building with `gdc`. Gentoo also has a policy for disabling `-Werror` by default so that is a factor as well

PR #3061, opened 2025-07-07
Add ability to specify frameworks to link against

LunaTheFoxgirl:
This pull-request adds a new configuration variable called `frameworks` as an analogue to `libs`; which allows specifying which frameworks to link against.

Frameworks are semantically different from dylibs/shared objects, as such `libs` is not really able to provide this functionality.

The `frameworks` directive is ignored on non-darwin systems.

### DIFF
--- a/changelog/add-frameworks.dd
+++ b/changelog/add-frameworks.dd
@@ -1,0 +1,14 @@
+Add `frameworks` dubfile key.
+
+Adds support for specifying macOS frameworks to link against,
+this replaces the need to manually specify frameworks via lflags or dflags.
+
+Before:
+-------
+lflags "-framework" "Cocoa"
+-------
+
+After:
+-------
+frameworks "Cocoa" "OpenGL"
+-------
--- a/source/dub/compilers/buildsettings.d
+++ b/source/dub/compilers/buildsettings.d
@@ -37,6 +37,7 @@ struct BuildSettings {
 	string[] dflags;
 	string[] lflags;
 	string[] libs;
+	string[] frameworks;
 	string[] linkerFiles;
 	string[] sourceFiles;
 	string[] injectSourceFiles;
@@ -87,6 +88,7 @@ struct BuildSettings {
 		addDFlags(bs.dflags);
 		addLFlags(bs.lflags);
 		addLibs(bs.libs);
+		addFrameworks(bs.frameworks);
 		addLinkerFiles(bs.linkerFiles);
 		addSourceFiles(bs.sourceFiles);
 		addInjectSourceFiles(bs.injectSourceFiles);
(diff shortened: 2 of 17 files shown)


rikkimax:
I would suggest changing this to ``appleFrameworks`` instead.

Given that this is limited to Apple platforms anyway, it prevents us from using that directive for things that may be D-specific in the future. Which may be necessary for PhobosV3.

LunaTheFoxgirl:
> I would suggest changing this to `appleFrameworks` instead.
> 
> Given that this is limited to Apple platforms anyway, it prevents us from using that directive for things that may be D-specific in the future. Which may be necessary for PhobosV3.

Not limited to apple platforms; limited to ones that use mach-o, or which support mach-o.
Given changes happening in linux-land with stuff like darling, just calling it frameworks is more appropriate, imho.

LunaTheFoxgirl:
Failed test is unrelated to the PR, fetchzip apparently had a stroke in that run

LunaTheFoxgirl:
@thewilsonator added unit test passes; linking to Foundation.framework successfully.

the-horo:
@LunaTheFoxgirl is the `frameworks` test invoking the linker on osx? On my linux `dub build` simply builds a static library so there's no linking done if you specify a configuration without a target type.

LunaTheFoxgirl:
> @LunaTheFoxgirl is the `frameworks` test invoking the linker on osx? On my linux `dub build` simply builds a static library so there's no linking done if you specify a configuration without a target type.

Yes, it is linking given that it's using a symbol that needs to be resolved whether or not it's an executable. That being said the config should ensure it's compiled as an executable; but only on Darwin targets.

Is no-op if targeting anything else.

the-horo:
You can "use" a symbol at compile time and have it be unresolved. It will fail at link time. I am asking specifically because I can `dub build` that file on linux without it failing (because a static library is not linked). Is it the same on osx? Would the test pass without the `frameworks` in `dub.sdl` because there's no linking being done? If so, the test needs to be fixed to include the executable targetType.

LunaTheFoxgirl:
That being said; being more clear by adding targetType would be a good idea.

rikkimax:
> > You can "use" a symbol at compile time and have it be unresolved. It will fail at link time. I am asking specifically because I can `dub build` that file on linux without it failing (because a static library is not linked). Is it the same on osx? Would the test pass without the `frameworks` in `dub.sdl` because there's no linking being done? If so, the test needs to be fixed to include the executable targetType.
> 
> In my experience that is behavior exclusive to Linux. Has never happened to me on macOS or Windows.
> 
> I think the object file structure of macOS and Windows more or less requires linking to be done or weak be applied before such behavior presents itself for those. As mach-o objects themselves store library reference data. And same with PE/COFF objects, if I understand that spec correctly.
> 
> Also in my local testing an executable mach-o file was created.

Something here appears to be miscommunicated.

You can use any symbol that you want to as far as the D compiler is concerned.
Static libraries are produced by archivers, not linkers. They do no symbol resolution.

Linkers do most of symbol resolution, but not all. Usually you must communicate to the linker which binary includes a given symbol if its not found within the object files you passed to it.

Loaders will do the final symbol resolution.

The only way to know for certain that a symbol has been resolved is to use it. Anything other than this can have it be ignored.

PR #3071, opened 2025-09-02
refactor integration tests (part 1)

the-horo:
A smaller extract of #3064. I've tried to keep the old tests running so that means that the refactor would only be adding tests, and we will be left to remove the old ones at a later time when the refactor finishes. I've also tried to keep CI usable, running all of the old tests alongside the newer ones.

@Geod24 is this in the range of smaller but no so small PR, I don't expect problems if you want me to increase or decrease the number of commits per PR.


Geod24:
@the-horo : Thanks for the ping. Why are the tests under `new-tests` ?

PR #3084, opened 2025-12-27
Rename PackageName.main to .base to making naming consistent

s-ludwig:
This is and has always been referred to "base package name" throughout the code base and should be named like this here, too, to avoid confusion.

Also reduces the amount of explicit attribute annotations a bit.

### DIFF
--- a/source/dub/commandline.d
+++ b/source/dub/commandline.d
@@ -1604,7 +1604,7 @@ class BuildCommand : GenerateCommand {
 		if (packageParts.name.startsWith(":"))
 			return 0;
 
-		const baseName = PackageName(packageParts.name).main;
+		const baseName = PackageName(packageParts.name).base;
 		// Found locally
 		if (dub.packageManager.getBestPackage(baseName, packageParts.range))
 			return 0;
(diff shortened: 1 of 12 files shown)


thewilsonator (approved):

WebFreak001 (approved):

the-horo:
Some tests failure because `setup-dlang@v2` changed the way `DC` is set:
```
[INFO] Running /home/runner/work/dub/dub/test/dc-env.sh...
core.exception.AssertError@app.d(17): Expected 'dmd' but got '/opt/hostedtoolcache/dmd/2.112.0/x64/dmd2/linux/bin64/dmd'
----------------
??:? _d_assert_msg [0x55936729e140]
./app.d:17 _Dmain [0x55936727c875]
Error Program exited with code 1
Error:  /opt/hostedtoolcache/dmd/2.112.0/x64/dmd2/linux/bin64/dub:6 command failed
[ERROR] Script failure.
```

I guess I should merge https://github.com/dlang-community/setup-dlang/pull/94 since it's been staying there all this time, then you could do:
```yml
- name: Install latest DMD
   uses: dlang-community/setup-dlang@v2
   with:
     dc_format: basename
```
and the test should pass. Alternatively, making it less flasky is also a good idea: https://github.com/the-horo/dub/blob/ba66cf37fd8597fa830805872d169be1a957150d/test/new_tests/issue2012-dc-env/sample/app.d

PR #3086, opened 2026-01-03
Improve test runner output

CyberShadow:
Because I could not figure out why https://github.com/dlang/dub/actions/runs/20679582481/job/59371914064?pr=3085 failed without looking at the runner source code.


the-horo:
Just noting that I've done a lot of work in https://github.com/dlang/dub/pull/3064 about improving the unittests. Maybe it's worth looking into that one as well

PR #3088, opened 2026-01-03
ci(workflows): update macOS runner versions

CyberShadow:
- Update main workflow to use macos-14 instead of macOS-13
- Update release workflow to use macos-15-intel instead of macOS-13
- Normalize runner name casing to lowercase (macOS -> macos)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,8 @@ jobs:
       fail-fast: false
       matrix:
         # Latest stable version, update at will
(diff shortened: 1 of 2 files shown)


rikkimax:
Dmd does not support macos-15 btw.

Both @WalterBright and @LunaTheFoxgirl are aware of it.

CyberShadow:
I'll disable testing DMD on macOS for now, since it looks like there isn't a working combination with what we have today.

PR #3107, opened 2026-03-25
Fix #2598: don't propagate build commands from root buildType to deps

### LINKED ISSUE #2598: preBuildCommands/postBuildCommands tainting _all_ dependencies from root package
Like since always... this can't have ever been intended, it certainly wasn't documented. what...?

```sdl
name "dubrec"
dependency "semver" version="0.3.4"
buildType "broken" {
    # causes dependencies to be recompiled each time as well?
    buildOptions "syntaxOnly"

    # this is run inside every dependency?!?!
    preBuildCommands "ls"
}
```

output:

```diff
 $ dub build --build=broken
     Starting Performing "broken" build using /usr/bin/dmd for x86_64.
     Building semver 0.3.4 [library]
     Pre-build Running commands
-#  v  THIS SHOULDN'T BE HERE!  v
 dub.json  LICENSE  README.md  source
     Building dubrec ~master [application]
    Pre-build Running commands
 docs  docs.json  dubrec  dub.sdl  dub.selections.json  __dummy.html  source
```

tried and reproduces with dub 1.31.0, 1.30.0, 1.29.0, 1.6.0

this is a fatal issue in preBuildCommands/postBuildCommands. Luckily this only happens when you define the pre/post build commands directly in your root package you are trying to build, but it's still a fatal flaw.

queelius:
## Summary

- Fixes #2598: `preBuildCommands`/`postBuildCommands` (and all other command hooks) defined in a custom `buildType` in the root package were incorrectly executed for every dependency during the build.
- When `addBuildTypeSettings` is called for a non-root package, command fields and their associated environments are now cleared from the buildType settings before propagation, matching the existing pattern that already strips unittest options for dependencies.
- Added a unit test that verifies commands from a custom buildType are present for the root package but absent for dependencies.

## Root cause

In `Project.addBuildTypeSettings()` (`source/dub/project.d`), the root package's buildType settings (including all command hooks) were applied to every target via `processVars()`. The `for_root_package` flag was only used to strip unittest options, but commands were never filtered.

## Test plan

- [x] New unit test in `source/dub/test/others.d` that creates a project with a custom buildType containing all 6 command types, a dependency, and verifies commands appear only for the root package
- [x] Confirmed the test fails without the fix (1/56 modules FAILED) and passes with it (56 modules passed)
- [x] All existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/dub/project.d
+++ b/source/dub/project.d
@@ -987,6 +987,22 @@ class Project {
 				// unit tests aren't run anyway and the additional code may
 				// cause linking to fail on Windows (issue #640)
 				btsettings.removeOptions(BuildOption.unittests);
+
+				// don't propagate commands from the root package's buildType
+				// to dependencies - they should only run for the root package
+				// (issue #2598)
+				btsettings.preBuildCommands = null;
+				btsettings.postBuildCommands = null;
+				btsettings.preGenerateCommands = null;
+				btsettings.postGenerateCommands = null;
+				btsettings.preRunCommands = null;
+				btsettings.postRunCommands = null;
+				btsettings.preBuildEnvironments = null;
+				btsettings.postBuildEnvironments = null;
+				btsettings.preGenerateEnvironments = null;
+				btsettings.postGenerateEnvironments = null;
+				btsettings.preRunEnvironments = null;
+				btsettings.postRunEnvironments = null;
 			}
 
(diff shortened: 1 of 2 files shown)


queelius:
Closing this to keep your queue clean. Happy to reopen if useful, or feel free to pick up any of it. Thanks!

PR #3125, opened 2026-06-12
Fix -conf flag not being passed to linker

LunaTheFoxgirl:
One of my libraries uses a custom ldc config file that overrides the original configuration of ldc, this also involves the linker step.
This PR simply makes sure the -conf option is passed along to ldc.


thewilsonator (approved):

thewilsonator (commented):
you could also add a test

PR #3129, opened 2026-07-08
fix: add -makedeps tracking and -J flag support to ninja generator

ghost:
Adds -makedeps dependency tracking to the ninja dc rule so ninja 
can track implicit dependencies (string imports, headers) and 
trigger incremental rebuilds correctly.

Also fixes a bug where -J flags for stringImportPaths were not 
being emitted, which would silently break any project using 
string imports.

Test: builds a project with a string import, runs it to verify 
output matches the imported file content, modifies the file, 
rebuilds, and asserts the output changed — proving dependency 
tracking works end-to-end.


source/dub/generators/ninja.d
```diff
@@ -32,7 +32,9 @@ class NinjaGenerator : ProjectGenerator
         f.writeln("# Generated by DUB ninja generator");
         f.writeln();
         f.writeln("rule dc");
-        f.writeln("  command = ", compiler, " $flags -c $in -of=$out");
+        f.writeln("  command = ", compiler, " $flags -c $in -of=$out -makedeps=$out.dep");
+        f.writeln("  depfile = $out.dep");
+        f.writeln("  deps = gcc");
```
thewilsonator, line 37: what is this line for?

ghost: deps = gcc tells Ninja to parse the depfile in gcc/Makefile format (vs MSVC's /showIncludes). Combined with depfile = $out.dep, it lets Ninja pick up implicit dependencies from the -makedeps output — e.g. string imports like views/data.txt — so those trigger rebuilds too, not just the direct source file. This is validated across Windows too, not just Linux — the behavioral test passes on 8 different Windows compiler configs (dmd 2.100–latest, ldc 1.30–master); the only Windows failure is an unrelated LNK1318 linker flake in a different test.

PR #3130, opened 2026-07-08
fix: emit -J flags for stringImportPaths and add behavioral makedeps …

atilaneves:
…test

PR #3131, opened 2026-07-11
feat: add ar rule and static library support to ninja generator

ghost:
Adds an `ar` rule and routes `staticLibrary`/`library` targets through
it instead of `link`, so subpackages build into real `.a` archives.

Does not depend on #3130, cherry-picked cleanly onto current
upstream/master, verified independently via fresh clone.

Test builds the archive, compiles a separate consumer against it via
extern(C), links, runs the binary, and asserts the actual computed
output, not just that the file exists.


atilaneves (approved):

ghost:
Closing to resubmit from my primary account after GitHub restored it and flagged this second account per their one-account policy. Same commit, will reference this PR. Also investigating the current CI failures before resubmitting.

PR #3134, opened 2026-07-16
feat: regenerate build.ninja when dub.json or dub.selections.json change

hariprakazz:
Adds a regen rule and build.ninja edge so it rebuilds itself
automatically whenever dub.json changes (dub.selections.json too,
if present).

Independent of #3130 (makedeps-v3, still unmerged) - builds cleanly
on current upstream/master with subpkg (#3133) already merged.

Test verifies the actual behavior, not just config text: touches
dub.json, runs ninja, and asserts "Regenerating build.ninja" genuinely
appears in the output. Confirmed the test fails when the regen rule
is removed and passes when restored. Verified via fresh clone build.

### DIFF
--- a/source/dub/generators/ninja.d
+++ b/source/dub/generators/ninja.d
@@ -9,11 +9,13 @@ module dub.generators.ninja;
 import dub.compilers.compiler;
 import dub.generators.generator;
 import dub.project;
+import dub.internal.vibecompat.inet.path;
 
 import std.algorithm : map, startsWith;
 import std.array : join, replace;
 import std.path : stripExtension;
 import std.stdio : File;
+import std.file : exists;
 
 class NinjaGenerator : ProjectGenerator
 {
@@ -45,6 +47,22 @@ class NinjaGenerator : ProjectGenerator
         f.writeln("rule ar");
         f.writeln("  command = $ar rcs $out $in");
         f.writeln("  description = Archiving $out");
+        f.writeln();
+
+        auto recipePath     = m_project.rootPackage.recipePath.toNativeString();
+        auto selectionsPath = (m_project.rootPackage.path ~ "dub.selections.json").toNativeString();
+
+        f.writeln("rule regen");
+        f.writeln("  command = dub generate ninja");
+        f.writeln("  generator = 1");
+        f.writeln("  description = Regenerating build.ninja");
+        f.writeln();
+
+        string[] regenInputs = [recipePath];
+        if (exists(selectionsPath))
+            regenInputs ~= selectionsPath;
+
+        f.writeln("build build.ninja: regen ", regenInputs.join(" "));
         f.writeln();
 
         foreach (name, info; targets)
--- a/source/dub/generators/ninja.d
+++ b/source/dub/generators/ninja.d
@@ -87,7 +87,7 @@ class NinjaGenerator : ProjectGenerator
             foreach (src; bs.sourceFiles)
             {
                 auto obj = objName(src);
-                f.writeln("build ", obj, ": dc ", src);
+                f.writeln("build ", obj, ": dc ", escapeNinjaPath(src));
                 if (flags.length)
                     f.writeln("  flags = ", flags);
                 objs ~= obj;
--- a/source/dub/generators/ninja.d
+++ b/source/dub/generators/ninja.d
@@ -15,7 +15,7 @@ import std.algorithm : map, startsWith;
 import std.array : join, replace;
 import std.path : stripExtension;
 import std.stdio : File;
-import std.file : exists;
+import std.file : exists, thisExePath;
 
 class NinjaGenerator : ProjectGenerator
 {
@@ -49,11 +49,11 @@ class NinjaGenerator : ProjectGenerator
         f.writeln("  description = Archiving $out");
         f.writeln();
 
-        auto recipePath     = m_project.rootPackage.recipePath.toNativeString();
+        const recipePath     = m_project.rootPackage.recipePath.toNativeString();
         auto selectionsPath = (m_project.rootPackage.path ~ "dub.selections.json").toNativeString();
 
         f.writeln("rule regen");
-        f.writeln("  command = dub generate ninja");
+        f.writeln("  command = ", thisExePath(), " generate ninja");
         f.writeln("  generator = 1");
         f.writeln("  description = Regenerating build.ninja");
         f.writeln();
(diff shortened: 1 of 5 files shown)


test/ninja-generator.sh
```diff
@@ -19,4 +19,33 @@ if ! grep -q "rule link" build.ninja; then
     die $LINENO 'build.ninja missing link rule!'
 fi
 
+if ! grep -q "rule regen" build.ninja; then
```
atilaneves: We shouldn't be looking for textual inclusion and instead test the *behaviour*. For instance, in this case, by testing that there's a regen when the selections file is changed.

test/ninja-generator.sh
```diff
@@ -19,4 +19,33 @@ if ! grep -q "rule link" build.ninja; then
...
+if ! grep -q "generator = 1" build.ninja; then
+    die $LINENO 'build.ninja missing generator attribute on regen rule!'
+fi
+
+if ! grep -q "^build build.ninja: regen" build.ninja; then
+    die $LINENO 'build.ninja missing self-regeneration build edge!'
+fi
+
+if ! grep "^build build.ninja: regen" build.ninja | grep -q "dub.json"; then
+    die $LINENO 'build.ninja self-regeneration edge missing dub.json dependency!'
```
atilaneves: We don't fire if the recipe file is `dub.sdl`?

atilaneves (changes requested):

test/ninja-generator.sh
```diff
@@ -19,33 +19,36 @@ if ! grep -q "rule link" build.ninja; then
...
-    die $LINENO 'build.ninja missing regen rule!'
+# Behavioral: verify build.ninja regenerates when dub.json changes
+ninja -t clean
+if ! ninja 2>&1; then
+    die $LINENO 'initial ninja build failed for regen test!'
 fi
 
-if ! grep -q "generator = 1" build.ninja; then
-    die $LINENO 'build.ninja missing generator attribute on regen rule!'
+touch dub.json
```
atilaneves: dub.sdl?

source/dub/generators/ninja.d
```diff
@@ -47,6 +49,22 @@ class NinjaGenerator : ProjectGenerator
         f.writeln("  description = Archiving $out");
         f.writeln();
 
+        auto recipePath     = m_project.rootPackage.recipePath.toNativeString();
```
atilaneves: `const` instead of `auto`.

test/ninja-generator.sh
```diff
@@ -19,4 +19,39 @@ if ! grep -q "rule link" build.ninja; then
     die $LINENO 'build.ninja missing link rule!'
 fi
 
-rm -f build.ninja
+# Behavioral: verify build.ninja regenerates when dub.json changes
```
atilaneves: I'm still not sure why the testing is written in bash.

atilaneves (commented):
This still doesn't handle dub.sdl

source/dub/generators/ninja.d
hariprakazz: Rewrote the regen test as test/ninja-generator.script.d, matching the existing .script.d convention instead of bash.

Added a dedicated dub.sdl fixture with a dependency check, not just existence, verified it fails when broken and passes when fixed.

Also caught a real bug: the regen rule used a bare dub in its command, which could resolve to the wrong binary. Switched to thisExePath().

test/ninja-generator.script.d
```diff
@@ -0,0 +1,71 @@
...
+import core.time : seconds;
+
+int main()
+{
+	const dub = environment.get("DUB", buildPath(__FILE_FULL_PATH__.dirName.dirName, "bin", "dub"));
+	const dc = environment.get("DC", "dmd");
+	const curr_dir = environment.get("CURR_DIR", buildPath(__FILE_FULL_PATH__.dirName));
+	const projDir = buildPath(curr_dir, "ninja-generator");
+
+	int fail(string msg)
```
atilaneves: Why not exceptions?

test/ninja-generator.script.d
```diff
@@ -1,71 +1,64 @@
...
 import core.time : seconds;
 
+import common;
+
+bool regenerated(string projDir, string touchedFile)
+{
+	Thread.sleep(1.seconds);
+	setTimes(buildPath(projDir, touchedFile), Clock.currTime, Clock.currTime);
+	const result = execute(["ninja"], null, Config.none, size_t.max, projDir);
+	return result.status == 0 && result.output.canFind("Regenerating build.ninja");
```
atilaneves: This could be passed by printing the message but not actually regenerating anything. The way to check this is to write a test that would fail unless the build.ninja file was actually regenerated. Imagine adding an import path to the dub recipe, for instance.

hariprakazz: Fixed. The test now checks the actual regenerated build.ninja content instead of just the log message. It adds a real import path to the recipe (dub.json/dub.sdl), runs ninja, and asserts the path actually shows up in build.ninja afterward.

Verified it catches the exact failure you described: broke importFlags so import paths get silently dropped, test failed correctly. Fixed it, test passed again.

Also fixed the same colon escaping bug in lflags, missed it on the first pass, same issue as src/regenInputs.

atilaneves (approved):

hariprakazz:
Hi Atila, thanks for approving this one. Quick follow up on #3130 (makedeps) too. I asked on the 22nd whether it landed some other way or needs resubmitting, haven't heard back yet. I've got the commit ready to reapply on current master if you'd like a fresh PR, just want your go ahead first since #3129/#3130 already caused mixup once. No rush, just flagging since final GSoC submissions open Aug 17.

atilaneves:
> Hi Atila, thanks for approving this one. Quick follow up on #3130 (makedeps) too. I asked on the 22nd whether it landed some other way or needs resubmitting, haven't heard back yet. I've got the commit ready to reapply on current master if you'd like a fresh PR, just want your go ahead first since #3129/#3130 already caused mixup once. No rush, just flagging since final GSoC submissions open Aug 17.

I closed that one in favour of yours?

hariprakazz:
checking the data on this: #3130 is under your fork (atilaneves/dub, branch hariprakash-makedeps), which looks like the version you set up after the duplicate account mess in July. It's currently closed and never merged, and I don't see -makedeps or depfile logic anywhere in current master.

Would you like me to resubmit that work as a fresh PR from my account, rebased on current master? Happy to do it however works best for you.

atilaneves:
> checking the data on this: #3130 is under your fork (atilaneves/dub, branch hariprakash-makedeps), which looks like the version you set up after the duplicate account mess in July. It's currently closed and never merged, and I don't see -makedeps or depfile logic anywhere in current master.
> 
> Would you like me to resubmit that work as a fresh PR from my account, rebased on current master? Happy to do it however works best for you.

Yes please, resubmit under your normal account.

PR #3151, opened 2026-09-04
Pin commit hashes of third-party actions on GHA

0xEAB:
This prevents potential supply chain attacks where an attacker swaps out the commit that a tag points to.


0xEAB:
Implementation detail:
GitHub’s own actions as well as `codecov/codecov-action@v4` are given the benefit of the doubt.

0xEAB:
It also assumes that the commits that the tags pointed to at the time of writing are fine.
I didn’t review them, and that’s only fair: The person who used the tag instead of the commit hash didn’t care either.

thewilsonator (approved):

PR #3156, opened 2026-09-09
Quote DC variable and use v2 of *setup-dlang*

0xEAB:


.github/workflows/main.yml
the-horo:
```suggestion
```

I think you want to do `setup-dlang` on all OSes

the-horo (changes requested):

.github/workflows/main.yml
0xEAB: yeah, just had the same brainwave. thanks ^^

the-horo:
If you can wait some more, I'll merge https://github.com/dlang-community/setup-dlang/pull/94 by the end of the week in the worst case and tomorrow in the best case and it should simplify things for you.

0xEAB:
@the-horo 
No hurry. If you want us to wait with this, we’ll do so ^^

0xEAB:
The current failures look spurious, anyway.

0xEAB:
What mad kind of tinkering is this testsuite? :roll_eyes:

0xEAB:
macOS 15: The `fetchzip.sh` heisenbug.
Windows failure: `Error: unrecognized switch '--d-version=must_be_defined'` — not sure why we suddenly pass LDC args to DMD.

(Funfact: Open D’s DMD would have happily accepted `--d-version`.)

